PR #17, opened 2016-12-20
Itemgetter

jacg:
SiPMNoise used a number of lambda expressions along the lines of `lambda x:x[0]`, which is just an ugly and CPU-costly way of writing `itemgetter(0)`.


jjgomezcadenas:
Branch gives conflict on Database/loadDB.py

jacg:
The conflict should be resolved now.

PR #29, opened 2016-12-23
Added OS X to Travis

jacg:
This commit runs four builds on Travis:

   linux & Python 2.7
   linux & Python 3.5
   OS X  & Python 2.7
   OS X  & Python 3.5

The Travis OS X servers do not provide Python installations, but this is
not a big problem for us, given that we use a conda installation of
Python rather than the Travis one anyway. I have switched the Travis
language from Python to generic, to make the OS X and linux procedures
as similar as possible. This should also bring the slight advantage of
making the Linux builds slightly faster, as we do not waste time on
installing the Travis Python, which we then do not use.

There are two small downsides to this, entirely related to Travis:

1. There aren't many Travis OS X servers, and there is sometimes quite a
   backlog of jobs for them. Expect the OS X builds to spend some time
   waiting in a queue before starting to run. In getting this set up I
   observed queue times from around 2 minutes, to 15 minutes.

2. The build takes about twice as long to execute on the OS X servers as
   compared to the Linux servers.

This takes the total execution time up to about 10 minutes (from about 4
minutes), and I think that there is a limit of 50 minutes. So there's
no danger of our jobs being cut short any time soon.

In short, this commit results in needing to wait longer (sometimes as
much as 20 minutes as opposed to 2 minutes) before getting official
Travis confirmation that a push is OK.

Of course, unofficial feedback from the two linux builds is available
just as quickly as it was before (about 2 minutes).

The upside is that we get automatic feedback of breakages on OS X. We
also get confirmation that it *DOES* work on OS X, so JJ doesn't have an
excuse for switching off all sorts of useful features and breaking all
sorts of code, just because his machine is misconfigured :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,21 @@
-sudo: false
-language: python
+language: generic
+sudo:     false
 
-python:
-- 2.7
-- 3.5
+env:
+  - IC_PYTHON_VERSION=2.7
+  - IC_PYTHON_VERSION=3.5
+
+os:
+  - linux
+  - osx
+
+
+before_install:
+    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export CONDA_OS=MacOSX; else export CONDA_OS=Linux; fi
 
 install:
   # Install conda
-  - wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
+  - wget http://repo.continuum.io/miniconda/Miniconda-latest-${CONDA_OS}-x86_64.sh -O miniconda.sh
   - bash miniconda.sh -b -p $HOME/miniconda
   - export PATH="$HOME/miniconda/bin:$PATH"
   - conda config --set always_yes yes --set changeps1 no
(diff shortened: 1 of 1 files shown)


jjgomezcadenas:
Hola,
Not very clear to me why we want to do build in OS X. It looks like the OS X builds are “machine dependent” (e.g, they can work in one machine and fail in another). Also, the last time I tried, ANACONDA3 did not work for the mac (problem with matplotlib). Is really worth the overhead?

I am thinking also in the possibility to develop from a LINUX virtual machine, so that all development is LINUX?

Please comment.
J

jacg:
> Not very clear to me why we want to do build in OS X.

Because you are developing on OS X.

>  It looks like the OS X builds are “machine dependent” (e.g, they can work in one machine and fail in another)

The same is true for any other operating system: I could mess with the configuration on my machine until builds fail, even though they succeed on the server!

> Also, the last time I tried, ANACONDA3 did not work for the mac (problem with matplotlib).

If I were a betting man, I would put money on the problem being with your machine's configuration, rather than Anaconda itself.

> Is really worth the overhead?

There is almost no overhead from the Travis and our point of view. The only question is the overhead of you battling with **your** personal machine.

> I am thinking also in the possibility to develop from a LINUX virtual machine, so that all development is LINUX?

If you are comfortable with that, the go ahead. But having Travis build on OS X is causing us no trouble at all, as far as I am aware. As long as it doesn't get in our way, I would leave it. And we might benefit from it.

jjgomezcadenas:
OK.

PR #32, opened 2016-12-23
Moving towards orthodox setup

jacg:
Replaced icompile.py with a setup.py that should eventually take on an
orthodox form. It should already make development and testing on
multiple python versions much easier (see below).

Created an overall `invisible_cities` package, which should make for
easier distribution, in comparison with the separate packages that we
had so far.

The recommended way to compile the Cython modules is now

    python setup.py develop

You should then immediately be able to run the tests with `pytest`.

To test the code on a different Python version

1. Open another shell,

2. (create, if you haven't got one yet, and) activate the conda env for
   that Python version,

3. source env.sh

4. python setup.py develop

5. pytest

Like this, the cython-compiled modules for different python versions can
happily co-exist and the correct ones should be picked up automatically
by the Python interpreter.


jjgomezcadenas (approved):
Convinced by divine inspiration that this is the way to go. If problems blame catholic church.

PR #39, opened 2017-01-06
Marking the OS X builds al allowed failures in Travis

jacg:
The OS X builds are sometimes very slow. This can significantly delay approval of pull requests. So they have been marked as allowed failures and the fast_build option has been switched on. This allowes pull requests to be merged as soon as the Linux builds pass

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: generic
 sudo:     false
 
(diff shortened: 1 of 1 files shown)


jjgomezcadenas (approved):
A terrific idea!

PR #43, opened 2017-01-09
Keeping an eye on Python 3.6

jacg:
Added Python 3.6 as an allowed failure to the Travis build
matrix. Like this we will know when all our dependencies are available
on conda in Python 3.6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,12 @@ matrix:
   fast_finish: true
   allow_failures:
   - os: osx
+  - env: IC_PYTHON_VERSION=3.6
+
 env:
   - IC_PYTHON_VERSION=2.7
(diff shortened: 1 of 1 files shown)


jjgomezcadenas (approved):
Agree. On the other hand, I wonder wether we should keep 2.7 as "required" in Travis. I believe that the right condition is that 3.5 runs, but we should not be required to make it run in 2.7 too. Soon enough (now...) is going to be too much work to maintain 2.7...

jacg:
At the moment, maintaining 2.7 compatibility is very cheap. As soon as it becomes more than we can be bothered with, it's just a question of adding

      - env: IC_PYTHON_VERSION=2.7

to the allowed_failures section it `.travis.yml`.

PR #48, opened 2017-01-09
Swapped meaning of ICDIR and ICTDIR

jacg:
The T in ICTDIR is supposed to stand for **TOP** directory.

Fixing this should reduce the scope for confusion.

### DIFF
--- a/invisible_cities/cities/diomira_test.py
+++ b/invisible_cities/cities/diomira_test.py
@@ -27,13 +27,13 @@ from invisible_cities.core.random_sampling\
 
 def test_diomira_run():
     """ Tests that DIOMIRA runs on default config parameters """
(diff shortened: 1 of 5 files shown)


jjgomezcadenas (approved):
In addition to clarify ICDIR and ICTDIR I see a new definition, DATABASE_LOCATION, also a environment variable?

jacg:
@jjgomezcadenas I would like to understand the purpose of your proposed DATABASE_LOCATION a bit better. I am reviving my localize branch as two separate branches:

- localize-tests
- localize-cython

Since I wrote that branch, a lot of `.h5` files have appeared in the test directory; a directory which isn't supposed to exist any more!

I guess that your DATABASE_LOCATION supposed to refer to the location of `localdb.sqlite3` only. But will we need another variable for all the `.h5`s?

PR #50, opened 2017-01-10
Move tests next to the implementations

jacg:
The test directory has been removed and any tests that used to be
there were placed in the same directory as the thing that they
test. Any data files used by the tests were moved into a subdirectory
called `test_data` of the directory containing the tests.

The idea is to keep test and implementations together, to encourage
people to read and write the tests, aswell as save them time by not
having to wonder where they are.


jjgomezcadenas (approved):
OK, with only a relevant question. Do ww want 

/database/test_data
or directly

/test_data

I am OK with both options,  just bringing the point.

jacg:
@jjgomezcadenas I chose `wherever/test_data` because, if the whole point of the change is to keep the tests close to the implementations, then it seems to make sense to keep the test data local too.

PR #53, opened 2017-01-10
Move Cython implementations next to pure python implementations

jacg:
In the same vein as #50, the motivation is to localize related code.

The `cython` directory has been removed. Its subdirectories' residents
have been moved to the corresponding subdirectories of
`invisible_cities`.

Where both a pure Python and a Cython implementation exist, the Cython
implementation gets a `_c` suffix.

Thus the import style changes from

    import invisible_cities.cython.wherever.whatever as namespace

to

    import invisible_cities.wherever.whatever_c as namespace


jjgomezcadenas (approved):
After thinking about this I agree that the proposed structure, with all related files in the same directory makes sense. The commit, however, results in conflicts, who should be solved by hand (probably pretty easy). I am approving the commit, decide who solves the conflicts.

jacg:
@jjgomezcadenas I was resolving the conflicts but #54 got in the way, and understanding what was going on wasted a lot of my time. I'll pick it up tomorrow.

PR #54, opened 2017-01-10
Remove file which should never have been added

jacg:
It turns out that #50 accidentally added `electrons_40keV_z250_RWF.h5`
to the repository. It should not be there as it is an output of one of
the tests.

A later PR will ensure that the tests in question do not leave such rubbish lying around.


jjgomezcadenas (commented):
Indeed, this file should not be in the repository. I can modify the test suite to ensure deletion after tests are done (it turns out that this is easier than leaving the file around). For the time being, we remove from the directory I will submit a PR to remove it automatically after tests.

jjgomezcadenas (approved):

PR #55, opened 2017-01-10
RWF file removed from working area

jjgomezcadenas:
diomira_tests.py produce a RWF file as a byproduct. The RWF file may
be needed for further tests (e.g, related with irene.py) but since it
can be re-generated fast (5 seconds) when needed. Consequently this
commit adds code to delete the RWF file at the end of the diomira
tests.

Furthermore, the obsolete script diomira_run.py has been eliminated.


jacg:
Did you really mean to remove `ic_Kr_tests_pmaps.h5` from the repository in this commit? And leave `electrons_40keV_z250_MCRD.h5`?

I'm having trouble understanding which of these files are outputs, which are inputs. And if some of them are outputs, then why were they added to the repo in the first place?

We clearly need to work on discipline when writing tests that use these types of files!

jjgomezcadenas:
To be clear.

The file MCRD is input and needs to be in the repo. RWF WAS NOT in the repo, managed to sneak in due to a bug and now it has been removed. Also ic_Kr_tests_pmpas (output) was left in the repo as an fossil and has now been removed.

I believe that for now the only files neeed in the repo are MCRD. The tests can produce RWF and PMPAS files, but erase them once they are done (this is what diomira_tests.py is doing now). On the other hand, I will need a few more MCRD files to test code with electrons of different energies (it is very relevant to understand that the code behaves as expected independently of energy of electrons).

PR #59, opened 2017-01-11
Move Cython implementations next to pure python implementations

jacg:
In the same vein as #50, the motivation is to localize related code.

The `cython` directory has been removed. Its subdirectories' residents
have been moved to the corresponding subdirectories of
`invisible_cities`.

Where both a pure Python and a Cython implementation exist, the Cython
implementation gets a `_c` suffix.

Thus the import style changes from

    import invisible_cities.cython.wherever.whatever as namespace

to

    import invisible_cities.wherever.whatever_c as namespace


jjgomezcadenas (approved):
Let's do it as the master says!

PR #60, opened 2017-01-11
Make one more module name PEP8 compatible

jacg:
BLR -> blr


jjgomezcadenas (approved):
yes, long life to PEP8!

PR #63, opened 2017-01-11
First draft of CONTRIBUTING.md

jjgomezcadenas:
Aims to document our preferred workflow and the git instructions
needed to achieve it.


jacg (approved):
Normally speaking this should be rejected as it is not yet ready and full of rough edges and gaps. But as we want to open the repo to other collaborators, and this document will help them getting used to THE WAY THINGS SHOULD BE DONE ... I'll accept it

PR #65, opened 2017-01-11
Make `manage.sh clean` give some feedback

jacg:
`manage.sh clean` used to work in complete silence. It now announces
that it is working and writes `rm somefilename` on separate lines for
each line that it removes. If it doesn't find anything to clean up, it
says so.


jjgomezcadenas (approved):
a clean chatting box? I love it!

PR #68, opened 2017-01-12
Ease making a number of temporary, shared test files

jacg:
The pytest tempfile-generating fixture has been:

- Moved out of `cities/diomira_test.py` into `cities/conftest.py`. In
  this way it can be used by any test in `cities`. If it needs to be
  shared by tests in distinct directories, this can be achieved by
  moving the `conftest.py` file higher up the directory structure into
  a common parent.

- Changed to produce a temporary directory rather than a single
  file. Multiple test output files can now be placed in that directory
  and shared by the tests.

This is so much better than writing these files in the project
directory because

- Tests for different versions of Python can now run concurrently on
  the same machine

- If the tests crash, there won't be any rubbish left lying around in
  the source directory

- There is no danger of anyone accidentally adding these files to the
  repo.

### DIFF
--- a/invisible_cities/cities/conftest.py
+++ b/invisible_cities/cities/conftest.py
@@ -1,0 +1,5 @@
+import pytest
+
+@pytest.fixture(scope='session')
+def irene_diomira_chain_tmpdir(tmpdir_factory):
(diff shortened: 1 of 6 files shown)


jacg:
Sorry, this was not meant for the nextic repo!

PR #70, opened 2017-01-12
Irene dev

jjgomezcadenas:
irene_test.py now includes a test that uses the fixture defined in
conftest.py. This permits to handle the files created by the test
chain through a temporary directory that vanishes when the test is
completed leaving no unwanted files in the source tree.

### DIFF
--- a/invisible_cities/cities/conftest.py
+++ b/invisible_cities/cities/conftest.py
@@ -1,0 +1,5 @@
+import pytest
+
+@pytest.fixture(scope='session')
+def irene_diomira_chain_tmpdir(tmpdir_factory):
+    return tmpdir_factory.mktemp('irene_diomira_tests')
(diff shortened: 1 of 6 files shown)


jacg (approved):
On my local machine I performed a test which Travis cannot really do
for us: run the tests concurrently for different Python versions *on
the same machine*. It worked. It used to fail.

It would be good to start using configuration files written
specifically for the tests and placed in the same temporary
directories, rather than the single hard-wired conf file that is in
the source repo. The tests should not be limited by that one conf file.

Also, [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) compliant dates would be nice.

jjgomezcadenas:
I believe that with the current solution and the freedom that configure gives us to change input, output and number of events, we should do fine, but if we need to change more parameters, I agree!

Issue #71, opened 2017-01-14
script manage.sh behaves in an unexpected way in Mac OS X


jjgomezcadenas:
source manage.sh c

Unrecognized command: c

Usage:

source -bash install_and_check X.Y
soruce -bash install X.Y
source -bash work_in_python_version X.Y
bash -bash make_environment X.Y
bash -bash run_tests
bash -bash compile_and_test
bash -bash download_test_db
bash -bash clean
Saving session...
...copying shared history...
...saving history...truncating history files...
...completed.
Deleting expired sessions...none found.

[Process completed]

jjgomezcadenas:
The error reported is no longer there, I was looking at the wrong fork. But there is still something that does not work completely well in the Mac

IC$ source -bash install_and_check 3.5
-bash: source: -b: invalid option
source: usage: source filename [arguments]

jjgomezcadenas:
Sorry again, wrong paste:
source -bash manage.sh install_and_check 3.5
-bash: source: -b: invalid option
source: usage: source filename [arguments]

jjgomezcadenas:
In other words, it appears that source --bash manage.sh does not work in OS. source manage.sh works

jacg:
Well, yes, the only 2 valid options are

1. `source manage.sh <command> [arg]`
2. `bash    manage.sh <command> [arg]`

The problem is that the usage message suggested nonsense (replacing `manage.sh` with `-bash`), and you believed and copied the nonsense.

I do not have access to OS X here, so please check that the upcoming PR really solves your problem before accepting it.

PR #72, opened 2017-01-14
Fix usage manage.sh message on OX X

jacg:
**I have not tested this on OS X (and neither will Travis, for the feature in question) so please test it manually on your OS X box before accepting.**

Correctly identifying whether a shell script is sourced or executed
and extracting the name of the script in a portable and reliable way
across different versions of bash and related shells such as zsh, is
not worth the trouble. This script is not meant to be relocatable; its
whole purpose is to sit in the top directory of IC, so I have taken
the simple and pragmatic approach: hard-wire its name into the usage
message.

### DIFF
--- a/manage.sh
+++ b/manage.sh
@@ -150,6 +150,8 @@ function clean {
 
 ## Main command dispatcher
 
+THIS=manage.sh
+
 case $COMMAND in
     install_and_check)      install_and_check ;;
     install)                install ;;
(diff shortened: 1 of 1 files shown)


jjgomezcadenas (changes requested):
The script works in OS-X, I have tested all the functions. In fact, in OS-X everything works with source manage.sh stuff (but also with bash). 

Since we are at it, there is a small typo in printout, line 2  (you can fix is using C-T in emacs) :)!

source manage.sh install_and_check X.Y
soruce manage.sh install X.Y
source manage.sh work_in_python_version X.Y
bash   manage.sh make_environment X.Y
bash   manage.sh run_tests
bash   manage.sh compile_and_test
bash   manage.sh download_test_db
bash   manage.sh clean

jjgomezcadenas:
I suggest that we could add jupyter notebook to the environments defined by manage.sh. Most users will eventually work with notebooks and adding the environment in manage.sh avoids that we need to add it later, when they want to run the notebook.  In this way, when they work in the NB repository they will only need to define an extra environment variable (ICNBTDIR) (to make the references to their code and files independent of user details).

If you agree, this can be also added to the next pull request on manage.sh

jacg:
>  there is a small typo in printout, line 2

Amended.

jjgomezcadenas:
Sorry to be a pain in the neck. Since we are at it, I would propose to add the Jupiter notebook to the environments. Also, a suggestion: in the ic_env function you print a number of messages such as
echo setting PYTHONPATH
But perhaps it would be more efficient to print something like 
PYTHONPATH set to $PYTHONPATH

in this way, the user is not only warned that we are setting PYTHONPAT (or ICDIR, ICTDIR) but also reads the value that takes for her.

Of course, this is just a ridiculous minor suggestion whose main aim is to show that I have read all the text of the functions...

jacg:
I am happy to consider all sorts of changes, but they do not belong in this PR which is a self-contained fix to #71.

I would need more information and time to consider and understand the Jupyter notebook issue. Please open an issue.

The `PYTHONPATH` suggestion seems reasonable. Please open an issue, or submit a PR.

jjgomezcadenas (approved):

PR #77, opened 2017-01-15
Jupyter

jjgomezcadenas:
Additions to manage.sh

Ths commit adds the package notebook to the conda environment,
introduces small changes in ic_env and adds a new function show_ic_env
that lists the conda environment and the environemnt variables ICDIR,
ICTDIR and PYTHONPATH

### DIFF
--- a/manage.sh
+++ b/manage.sh
@@ -75,6 +75,7 @@ dependencies:
 - scipy
 - matplotlib
 - jupyter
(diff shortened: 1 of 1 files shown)


jacg (approved):
With the increasing number of arguments to this script, we should consider teaching bash how to complete them. See #78.

PR #82, opened 2017-01-16
Enhance the code-symmetry emphasis in `irene.py`

jacg:


jjgomezcadenas (approved):
After this revision, Irene.py can be used as an excellent example to illustrate JCK.

Issue #46, opened 2017-01-09
investigate documentation tools, choices and structure


jjgomezcadenas:
We need to add a /docs directory to the repository. Proposal is to add it at the top level (next to invisible_cities), possibly with some extra sub-structure (such as /cities, /analysis, /core, etc.)

Open questions:
1) JJ argues that adding clean, well structured jupyter nb as docs is OK and good.  They need to be reviewed anyway, before entering repository.
2) Are we using Sphinx for docs? Any other tool?
3) What would be the structure of the /docs recommended by Sphinx?
4) Any other issues, suggestions?

jjgomezcadenas:
Further thinking leads to the proposal that we should keep an independent NB repository (already available in nextic) for NB docs. Documentation with Sphinx or similar tools can go into the IC repository

jmbenlloch:
I have tested sphinx in my fork. The idea is to have a doc/ folder containing one conf.py file for sphinx and as many rst files as we need for our documentation. This can also be linked to readthedocs.io, so we have a web like this http://ic-1.readthedocs.io/en/latest/ The documentation is recompiled for each commit automatically (we can also choose to have some "stable" versions).

I have made a pull request with the required changes in #84

jmbenlloch:
This is closed by #84.

PR #84, opened 2017-01-18
Sphinx example setup

jmbenlloch:
Test setup to solve issue #46


jacg:
I think we agreed that

- This is still work in progress
- The branch name sould be changed to `docs.sphinx`
- The docs directory should be changed from `docs` to `doc`

jmbenlloch:
Several things:

* `docs` folder is renamed to `doc`
* A Makefile is added. 

With the Makefile, to compile html one can just run:

`make html`

Without the Makefile sphinx-build is needed. It can be executed like this:

`sphinx-build -b html doc doc/build`

I think it is convenient to have the Makefile (specially if we want to add a more general one in the future #88)

And for last, I don't know how to rename the branch in github, I have tried doing it first locally and the pushing to the repository, but it didn't work. After a search on google, people suggest removing the old branch in github and pushing again the branch with the new name. I don't know what would happend with the already existing pull requests. Therefore, I have not updated the branch name.

jacg:
I think that the PR is coupled to the branch name, so changing the name essentially means closing the PR.

If we are still considering this work in progress, then it's

- no problem to close the PR,
- a good idea to rename the branch (to something starting with `docs.`), in orded to avoid Travis builds on subsequent pushes that are sure to come.

jjgomezcadenas (commented):
I have run the makefile and the html pages are correctly produced (once you install sphinx). Please add sphinx to  the environment

jmbenlloch:
I have added sphinx to the environment in manage.sh and also `doc/build` in `.gitignore`. It has been tested on a fresh installation and it works well.

jjgomezcadenas (approved):
The system works in two independent computers (my *fast* Mac and the *slow* linux new laptop).

Issue #78, opened 2017-01-15
Create a bash command argument completion file for `manage.sh`


jacg:
Our `manage.sh` script is acquiring (#77)  a large number of long and inconvenient to type command arguments. It would benefit immensely from completion help from the shell. I get the impression that most of us are using `bash` or some derivative. Let's provide a file which teaches `bash` how to complete `manage.sh`'s arguments for us.

How this can be done seems to be described [here](http://www.tldp.org/LDP/abs/html/tabexpansion.html).

Once made, this file can be sourced to enable the completion. You could do this in your `.bashrc` or similar.

jjgomezcadenas:
Earth talking to Galactica. I hear that Capitan JJ won't do it. I guess admiral JCK is busy next couple of weeks. Shall we assign democratically to lieutenant JMB?

jacg:
Get in the habit of @-mentioning people who you want to read the message. In this case mentioning @jmbenlloch would have been a good idea.

I suggest that he assign himself, if he feels like taking it on. Otherwise, it can sit there in the list of issues until somebody takes pity on it, or someone thinks it needs to be given a higher priority. It's possibly not of the highest priority at the moment.

jjgomezcadenas:
Ay, Ay, sir!

As for JM he is sitting by my side, so he has taken notice! And he is assigning himself to the task. He is indeed pretty busy, but also pretty fast, so no worries.

PR #86, opened 2017-01-18
First version of bash completion script for manage.sh

jmbenlloch:
This is a first proposal for bash completion in manage.sh (issue #78). First you need to execute the script:

`. bash_manage.sh`

Then you it will autocomplete `bash manage.sh` or `source manage.sh` in this way:

```
$ source manage.sh [TAB][TAB]
clean                   install                 run_tests
compile_and_test        install_and_check       work_in_python_version
download_test_db        make_environment        
```

If you chose an option requiring python version it will offer you 2.7 and 3.5:

```
$ source manage.sh work_in_python_version [TAB][TAB]
2.7  3.5  
```

The only problem is that some bash options are overwrite. For instance, this will happen before executing the script:

```
$ bash --[TAB][TAB]
--debug            --help             --noprofile        --restricted
--debugger         --init-file        --norc             --verbose
--dump-po-strings  --login            --posix            --version
--dump-strings     --noediting        --rcfile     
```

While after executing the script no option will be shown in that case. It is not clear to me how to improve this behavior, comments are welcomed.

### DIFF
--- a/bash_manage.sh
+++ b/bash_manage.sh
@@ -1,0 +1,74 @@
+manage.sh()
+{
+    local cur prev opts
+    COMPREPLY=()
+    cur="${COMP_WORDS[COMP_CWORD]}"
+    prev="${COMP_WORDS[COMP_CWORD-1]}"
+    prev2="${COMP_WORDS[COMP_CWORD-2]}"
+
+
+    case "${prev}" in
+        manage.sh)
+                        local opts="make_environment run_tests compile_and_test\
+                                download_test_db clean"
+            COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
+            return 0
+            ;;
+        *)
+        ;;
+    esac
+
+    if [[ ${prev2} == "manage.sh" ]] ; then
+            case "${prev}" in
+                        install_and_check|install|work_in_python_version|make_environment)
+                        local versions="2.7 3.5"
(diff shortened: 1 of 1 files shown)


jacg:
I see that the commit message is a bit sparse. If you could amend it to include the pull request message, that would be great.

I don't use `bash` myself, and it seems that it doesn't work out of the box with `zsh`, so I'll need a bit more time to have a look at it in detail. Basic experiments within `bash` suggest that it seems to be working. It already seems useful as it is, so I'm happy to accept, but perhaps you could amend the commit message as suggested first.

jacg:
Here's a recipe for getting it to work with `zsh`

    autoload bashcompinit
    bashcompinit
    source bash_manage.sh

Perhaps you could include this info in the commit message.

jacg:
I'm also wondering whether it is neccessary to put this in a separate file: Could this be squeezed into `manage.sh`? Would that be a good idea?

jacg:
It seems that the `manage.sh` commands which *must* be sourced appear as suggestions for `bash`. Would it be much work to get them only to appear for `source`?

It also seems that completion works for `source manage.sh` but not for `. manage.sh`. Would that be hard to do?

jacg (approved):
Already very useful, so let's include it, even if we may want to polish things later.

PR #87, opened 2017-01-19
Move from miniconda2 to miniconda3

jacg:
We were using miniconda2, which is based on Python 2. Given that our
primary version of IC is the Python 3 one, we might as well use
miniconda3 and save ourselves the need to drag down two Python
installations (one for conda and one for IC) in the primary version.


jjgomezcadenas (approved):
Makes sense.

Issue #81, opened 2017-01-16
Error in Irene


jmbenlloch:
I have run Irene on MC Kr at CERN, but 41 of the jobs had errors during processing. All of them are like this:

```
Opening ./ic_dst_NEXT_v0_08_06_Kr_ACTIVE_7_0_7bar_MCRD_10000.root.h5... Events in file = 10000
# PMT                  => 12
# SiPM                 => 1792
PMT WL                 => 48000
SIPM WL                => 1200
event in file = 99, total = 100
error
Traceback (most recent call last):
  File "/afs/cern.ch/work/j/jobenllo/software/IC/invisible_cities/cities/irene.py", line 548, in <module>
    IRENE(sys.argv)
  File "/afs/cern.ch/work/j/jobenllo/software/IC/invisible_cities/cities/irene.py", line 540, in IRENE
    nevt = fpp.run(nmax=nevts, store_pmaps=True)
  File "/afs/cern.ch/work/j/jobenllo/software/IC/invisible_cities/cities/irene.py", line 429, in run
    rebin  = False)
  File "invisible_cities/core/peak_functions_c.pyx", line 105, in invisible_cities.core.peak_functions_c.find_S12 (invisible_cities/core/peak_functions_c.c:5121)
  File "invisible_cities/core/peak_functions_c.pyx", line 134, in invisible_cities.core.peak_functions_c.find_S12 (invisible_cities/core/peak_functions_c.c:4232)
IndexError: Out of bounds on buffer access (axis 0)
```

I have checked the logs to find a file in which the error appears soon, in particular with:

`lxplus.cern.ch:/eos/experiment/next/dataic/DiomiraMS/Kr/ic_dst_NEXT_v0_08_06_Kr_ACTIVE_7_0_7bar_MCRD_10000.root.h5`

The error occurs before event 200. The config file I am using is this one:

```
PATH_IN .
PATH_OUT .
FILE_IN ic_dst_NEXT_v0_08_06_Kr_ACTIVE_7_0_7bar_MCRD_10000.root.h5
FILE_OUT pmaps_dst_NEXT_v0_08_06_Kr_ACTIVE_7_0_7bar_MCRD_10000.root.h5
COMPRESSION ZLIB4

RUN_NUMBER 0
NPRINT 100

NBASELINE 38000
THR_TRIGGER 5

NMAU 100
THR_MAU 3
THR_CSUM 0.5
S1_TMIN 10
S1_TMAX 590
S1_STRIDE 4
[...]

PR #98, opened 2017-01-23
Irene bug test

jjgomezcadenas:
This PR consists of two commits (not squashed since they may be more informative as separate commits).

First commit:

In a few cases (less than 0.1 % of the events) NEXT simulation (NEXUS)
produces events with no energy in the PMTs of the energy plane. This
is not strictly a bug (in the sense that it simply reflects a
pathological type of events that occur rarely), but it can produce a
crash in the processing of the data.

While a protection could be set alreay at DIOMIRA level (not accpeting
events with empty energy plane) this would require a rather expensive
test from the computational point of view. A better place to set the
protection is down the processing line, in Irene, asking that the
calibrated sum of the PMTs is not zero.

Function test_diomira_identify_bug() does what its name says. It reads
a file with a single-event of type "empty energy plane", and asserts
that all PMTs are empy. Notice that the test would fail (as intended
for a normal file).

Two files are needed for the suite. irene_bug_Kr_ACTIVE_7bar_MCRD.h5
is use by test_diomira_identify_bug(), while
irene_bug_Kr_ACTIVE_7bar_RWF.h5 will be used by a test (next commit)
which reads the buggy file and would fail if the protection in irene
fails.

Second commit:

Following previous commit, a protection has been added in Irene to
skip events where the calibrates (ZS) sum is zero. A test has been
added checking that irene does not fail when reading a file with such
event and that the event is correctly looped out and counted.


jacg:
I am not qualified to comment on the value and correctness of this test and bugfix from the perspective of the phisics and the detector design (and I don't have time to gain the relevant knowledge, at present), so I shall restrict my observations to the code, and let someone who understands the dectector better than I do, to approve the PR.

jacg:
I'm glad to see that you had the presence of mind to mark the test with `@mark.serial` ... but in this case, it seems unnecessary: At first glance it looks like any files written by this test are consumed by this test and this test alone, and so do not interfere with any other tests. A quick check fails to falsify this hypothesis: removing the mark does not cause any tests to fail, but it *does* reduce the total test run time from about 23 seconds to roughly 17 seconds.

So I think that the mark can be removed, unless there is a sensible way of splitting this test into two separate tests. (In which case I need to raise the priority of my TODO on grouping of serial tests).

jjgomezcadenas:
What is marked as serial is the test already existing

@mark.serial
def test_diomira_and_irene_run(irene_diomira_chain_tmpdir):

While the new test is not marked as serial

def test_empty_events(irene_diomira_chain_tmpdir):

def test_diomira_identify_bug()

So, we can decide to remove the mark_serial in test_diomira_and_irene (indeed it only consumes its own files, I have no problem with that).

jacg:
The fundamental question is 'does `test_diomira_and_irene_run` need to be marked as serial to function properly?'. As I said, my practical investigation suggests that it does not, but maybe I was just lucky. (And I don't have time today to read the code carefully enough to reach a confident theoretical conclusion.)

If the answer is 'no' then the mark should be removed.

The secondary question is 'Is `test_diomira_and_irene_run` logically a single test?'. If it can sensibly be split into multiple tests, then it sould ... but that is orthogonal to *this* PR.

In *this* PR, the serial mark should be removed (or rather, should not be added), it seems.

jacg:
Ah, I see: you have added another commit which reverses the additon of the mark. But the addition of `import pytest` still remains.

These should both be *edited* out of the second commit.

I might have the time to do it this evening. Or maybe you want me to go through it with you. Or maybe you can try it yourself. As it's the top of the two (real) commits, it's pretty easy.

jjgomezcadenas:
After the iteration, a new PR with *a single commit* has been sent. I believe that the commit text is sufficiently clear to explain everything done, and that the @mark issues is well documented in the conversation. Thus, A single commit seems to be the cleanest option.

---

In a few cases (less than 0.1 % of the events) NEXT simulation (NEXUS)
produces events with no energy in the PMTs of the energy plane. This
is not strictly a bug (in the sense that it simply reflects a
pathological type of events that occur rarely), but it can produce a
crash in the processing of the data.

While a protection could be set alreay at DIOMIRA level (not accpeting
events with empty energy plane) this would require a rather expensive
test from the computational point of view. A better place to set the
protection is down the processing line, in Irene, asking that the
calibrated sum of the PMTs is not zero.

Function test_diomira_identify_bug() does what its name says. It reads
a file with a single-event of type "empty energy plane", and asserts
that all PMTs are empy. Notice that the test is intended to fail if
it runs in a normal file. The purpose of this test is to demonstrate
that empty "empty energy plane events" exist.

To solve the problem, a protecction is added in irene.py which skip
events where the calibrates (ZS) sum is zero. A second test,
(test_empty_events()) has been
added checking that irene does not fail when reading a one-event file
containing an "empty energy plane" event. The test also checks that
the event is counted.

Two files are needed for the suite. irene_bug_Kr_ACTIVE_7bar_MCRD.h5
is use by test_diomira_identify_bug(), while
irene_bug_Kr_ACTIVE_7bar_RWF.h5 is used by test_empty_events().
Notice that test_empty_events() would fail if the "empty event" slips
through the protection.

jmbenlloch:
I have tested this solution with a file different from the one in #81 (in particular `lxplus:/eos/experiment/next/dataic/DiomiraMS/Kr/ic_dst_NEXT_v0_08_06_Kr_ACTIVE_0_0_7bar_MCRD_10000.root.h5 ic_dst_NEXT_v0_08_06_Kr_ACTIVE_0_0_7bar_MCRD_10000.root.h5`) and it works well. At the end it reports how many empty events have been analyzed:

```
reached max nof of events (=2000)
Energy plane empty events (skipped) = 1
run 2000 evts in 762.133505821228 s, time/event = 0.38106675291061404
``` 

The problem and the issues mentioned by @jacg seem to be solved, so I will approved this PR and close #81.

jmbenlloch (approved):

PR #100, opened 2017-01-23
Reinstate serial mark on `test_diomira_cwf_blr`

jacg:
This test cannot be run in parallel with some of the other, as it
needs the temporatry file `electrons_40keV_z250_RWF.h5`.

The mark seems to have been removed mistakenly in #98.


jacg:
When adding or removing serial test marks (especially when removing) please check with `bash manage.sh run_tests_par` to see what effect it has or the tests when run in parallel!

jjgomezcadenas (approved):
Sorry about this accidental remove!

PR #102, opened 2017-01-24
Bug.test core dump

### LINKED ISSUE #95: Investigate why test_define_window crashes on Travis Linux
It works on Travis OS X, it works on JG's linux machine, it works on JJ's Mac ... but core dumps on Travis Linux.

The test can be found in `invisible_cities/core/core_functions_test.py`.

It has been marked as `skipif` on Travis Linux.

jacg:
Use `np.histogram` instead of `plt.hist` and fix #95

`test_define_window` was core dumping in headless environments (for
us, this meant Linux-on-Travis). This was caused by the use of
`matplotlib.pyplot.hist` in order to *calculate* a histogram inside
the test. Because `plt.hist` tries to *display* the histogram, it's
not surprising that it crashed on systems that did not have displays.

As the test had no real need to actually display the histogram, merely
to calculate it, this problem was easily solved by using
`numpy.histogram` which performs the same calculation, without the
(pointless and damaging) displaying side-effect.

Moral of the story: don't use drawing functions to perform pure
calculations!

(I am leaving the first commit unsquashed, in order to leave a record of the `skipif` condition for headless environments, in case anyone needs it in the future.)

### DIFF
--- a/invisible_cities/core/core_functions_test.py
+++ b/invisible_cities/core/core_functions_test.py
@@ -6,8 +6,6 @@ import sys, os
 from . import core_functions as core
 import pandas as pd
 import numpy as np
-import matplotlib.pyplot as plt
-
 
 def test_lrange():
     assert core.lrange(10) == [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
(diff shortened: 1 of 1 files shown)


jjgomezcadenas (approved):
Yes, it is clear and obvious! (or, in good Spanis... obvio a toro pasado)!

Lesson learned!

PR #103, opened 2017-01-24
Clean up `.pyx` files

jacg:
Mostly JCK-1 stuff.

However, most of these files seem to have no tests! (Remember the
rule: if a change to existing code doesn't make an existing test fail,
then that change must be ok.)

There seems to be a lot of copy-n-paste code duplication in these
files, but that's something that will require a bit of thought to
consider (unlike these changes, which were mostly mechanical and
exactly the sort of thing I can do to relax after a long day at the
office.)

### DIFF
--- a/invisible_cities/core/peak_functions_c.pyx
+++ b/invisible_cities/core/peak_functions_c.pyx
@@ -5,13 +5,13 @@ JJGC December, 2016
 from __future__ import division
 
 cimport numpy as np
-import numpy as np
+import  numpy as np
 from scipy import signal
 
 cpdef calibrated_pmt_sum(double [:, :] CWF,
                          double [:] adc_to_pes,
-                         int n_MAU=200,
-                         double thr_MAU=5):
+                         int      n_MAU = 200,
+                         double thr_MAU =   5):
     """
     Computes the ZS calibrated sum of the PMTs
     after correcting the baseline with a MAU to suppress low frequency noise.
@@ -25,15 +25,15 @@ cpdef calibrated_pmt_sum(double [:, :] CWF,
 
     cdef int j, k
     cdef int NPMT = CWF.shape[0]
-    cdef int NWF = CWF.shape[1]
+    cdef int NWF  = CWF.shape[1]
     cdef double [:] MAU = np.array(np.ones(n_MAU),
-                                   dtype=np.double)*(1 / n_MAU)
+                                   dtype = np.double) * (1 / n_MAU)
 
 
     # CWF if above MAU threshold
     cdef double [:, :] pmt_thr = np.zeros((NPMT,NWF), dtype=np.double)
-    cdef double [:] csum = np.zeros(NWF, dtype=np.double)
-    cdef double [:] MAU_pmt = np.zeros(NWF, dtype=np.double)
+    cdef double [:]       csum = np.zeros(      NWF , dtype=np.double)
+    cdef double [:]    MAU_pmt = np.zeros(      NWF , dtype=np.double)
 
     for j in range(NPMT):
         # MAU for each of the PMTs, following the waveform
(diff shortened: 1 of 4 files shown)


jjgomezcadenas (approved):
Actually, the duplication and apparent cut-and-paste occurs only in cblr.pyx and it's not accidental. This module was kept in sierpe to track the historical development of the deconvolution algorithm and "document" the different options (some of them quite equivalent). The final choice of algorithm was moved to blr.pyx and this algorithm is tested extensively in the test suite and documented in the notebooks. 

I guess that the right thing to do is to eliminate cblr.pyx which is doing nothing, but I would like to tag the repository first.

The functions in core are weakly tested in the test-suite (although they are excercised in the notebooks and work well). I will add more tests shortly.

As for the amends to the code, I like most but not all of them. One example of choice that I don't like is to write something of the type
np.zeros(.        NWF,       dtype)

I don't like at all the blank spaces after the parenthesis (even if it is clear that they are there for alignment purposes). I would prefer a different type of alignment, which, even if not perfect does not leave blank spaces after opening a parenthesis (for some reason this bothers me and is certainly very unPEPISH). 

In any case,  since most changes improve the readability of the code I take them and if I decide I cannot live with the blank spaces after the parenthesis I will submit a patch.

jacg:
Yes, it seems that `cblr.pyx` *should* be removed in that case. Remember, it's still there in the history of the repo, and thus can still serve as documentation. But it should not be one of the files in the *current* version.

> I don't like at all the blank spaces after the parenthesis (even if it is clear that they are there for alignment purposes). I would prefer a different type of alignment, which, even if not perfect does not leave blank spaces after opening a parenthesis (for some reason this bothers me and is certainly very unPEPISH).

:-)

I sympathize. But do note that you are effectively saying "I acknowledge that it's better but the indoctrination I suffered in my youth makes me dislike it.

If whitespace after open parenthesis bothers you and you find it hard to resist the temptation to remove it, then you have various options:

1. Remove the whitespace
1. Move the parenthesis (leaving the whitespace *before* it)
1. Live with it
1. Learn to love it

Option 2 is beter than 1, but still not as good as 3. 4 is even better.

What matters is not which one you like most: What matters is which one makes mistakes and patterns most obvious to the human eye.

jjgomezcadenas:
> On 25 Jan 2017, at 08:57, jacg <notifications@github.com> wrote:
> 
> Yes, it seems that cblr.pyx should be removed in that case. Remember, it's still there in the history of the repo, and thus can still serve as documentation. But it should not be one of the files in the current version.
> 
> 

Agreed. If no complains I will put a tag first, though (v0.05?), then remove the file. I think we are not far to be able to tag a first release (v0.1). Please advise. 
> I don't like at all the blank spaces after the parenthesis (even if it is clear that they are there for alignment purposes). I would prefer a different type of alignment, which, even if not perfect does not leave blank spaces after opening a parenthesis (for some reason this bothers me and is certainly very unPEPISH).
> 
> :-)
> 
> I sympathize. But do note that you are effectively saying "I acknowledge that it's better but the indoctrination I suffered in my youth makes me dislike it.
> 
> If whitespace after open parenthesis bothers you and you find it hard to resist the temptation to remove it, then you have various options:
> 
> Remove the whitespace
> Move the parenthesis (leaving the whitespace before it)
> Live with it
> Learn to love it
> Option 2 is beter than 1, but still not as good as 3. 4 is even better.
> 
> What matters is not which one you like most: What matters is which one makes mistakes and patterns most obvious to the human eye.
> 
> 
Agreed, oh Jedi. Let my should be elevated and LOVE blacks after parenthesis.

jacg:
>  I will put a tag first, though (v0.05?)

No! Read about semantic versioning, or wait until I have the time to talk to you about it.

PR #104, opened 2017-01-25
Add two utility functions to core.

jjgomezcadenas:
Two utility functions have been added to core.
1) A function to compare CWF and BLR waveforms (wfm_functions)
2) A function to plot S2Si objects as a SiPM map (to sensor_functions)

The first function is actually a rendering of one of the tests in the
test-suite that ensures that CWF and BLR are equal to high
precission. The second function is a drawing function, exercized in
the irene notebook.


jacg:
I feel I must attract your attention to point 3 of the [workflow summary](https://github.com/nextic/IC/blob/master/CONTRIBUTING.md#workflow-summary) of our [contribution guide](https://github.com/nextic/IC/blob/master/CONTRIBUTING.md).

And also to the [Testing](https://github.com/nextic/IC/blob/master/CONTRIBUTING.md#testing) section of the same, specifically the paragraph which states

> Write tests for any new code that you write. Before your code can be merged into the main repository it must be reviewed by someone else. Expect reviewers to reject your code if it does not come with adequate tests.

It's interesting that you say that one of these functions is 'a rendering of one of the tests in the
test-suite'. I do not currently have the time to look into what this means exactly, and its consequences.

In any case, the second function appears to be completely untested, and we are trying to establish a discipline of only adding new functionality if it has accompanying tests.

jjgomezcadenas:
Function 

def compare_cwf_blr(pmtrwf, pmtblr, event_list, plot=False, window_size=500):
    """ Compute CWF wfm deconvoluting input RWF and compare with BLR."""

returns the differences between the CWF and the BLR waveforms (which are expected to be very small).

Function

def test_diomira_cwf_blr(irene_diomira_chain_tmpdir):

in diomira_test, in facts ASSERTS that those differences are very small. 

The second function is a plotting function. I think we agreed that plotting functions are exercised in the notebooks but do not need a test. I propose that we agree that anything starting with *plot* may (should) be exercised in a notebook but does not need a test.

jacg:
> I think we agreed that plotting functions are exercised in the notebooks but do not need a test.

Here we are saying, testing this is hard so let's not bother.

That might be a perfectly valid choice. However, from my perspective, I *have* to investigate how these could be tested, and how much it would cost us as a team in terms of invested effort, and what we get back for that investment.

> I propose that we agree that anything starting with plot may (should) be exercised in a notebook but does not need a test.

An interesting suggestion. I wouldn't go around changing function names to fit this proposal just yet. Let's think about it a bit more. (I don't have time this week.)

jacg:
> Function `compare_cwf_blr` [...] returns the differences between the CWF and the BLR waveforms (which are expected to be very small).
> [...]
> Function `test_diomira_cwf_blr` [...] in diomira_test, in facts ASSERTS that those differences are very small.

So if I fiddle with the definition of `compare_cwf_blr` and break it, `test_diomira_cwf_blr` will fail? If so, I'll be extatically happy.

jacg:
> So if I fiddle with the definition of `compare_cwf_blr` and break it, `test_diomira_cwf_blr` will fail? If so, I'll be extatically happy.

Hmm, it looks like I'll have to delay my extatic happiness.

It seems that `compare_cwf_blr` mixes the orthogonal concerns of

1. performing a calculation 
1. plotting something

That's a pretty bad idea. Unfortunately, that's all I have time for right now.

jjgomezcadenas:
The problems spotted by @jacg are corrected in the last commit.

jacg:
I have pushed a few minor commits on top of this, for your consideration. (They should be squashed in to the main commit, if accepted.) This just arose from a superficial look at the code: I haven't had the time for a closer look, I'm afraid.

jjgomezcadenas:
Commits accepted and squashed. @jacg

jmbenlloch (approved):

PR #110, opened 2017-01-31
Core devel

jjgomezcadenas:
Add new test suite

A new test suite (peak_functions_test.py) has been added. The module
runs several exahustive tests over peak_functions, including:

1. Comparison of BLR and CWF waveforms.
2. Comparison of python and cython functions (for this purpose the
python version of several existing cython functions have been added
to peak_functions.py)
3. Analysis of a toy model.

In addition, a few utility functions (needed for the test toy model) have
been added to core_functions (together with their corresponding test
in core_functions_test).

Some minor refactoring has been carried out along with the addition of this
module (moving an obsolete function out of peak_functions.py).

Also, some minor modifications (mostly cosmetic) have been added to several core modules.

The main ideas on the test suite are illustrated here:

https://anaconda.org/nexticw/peak_functions_test

### DIFF
--- a/invisible_cities/cities/irene.py
+++ b/invisible_cities/cities/irene.py
@@ -393,7 +393,7 @@ class Irene:
                           n_baseline  = self.n_baseline,
                           thr_trigger = self.thr_trigger)
                         # calibrated PMT sum
-                        csum = cpf.calibrated_pmt_sum(
+                        csum, _ = cpf.calibrated_pmt_sum(
                           CWF,
                           self.adc_to_pes,
                             n_MAU = self.  n_MAU,
@@ -519,7 +519,7 @@ class Irene:
                 row.append()
                 self.runInfot.flush()
             self.pmapFile.close()
-        
+
         if self.print_empty:
             print('Energy plane empty events (skipped) = {}'.format(
                    self.empty_events))
--- a/invisible_cities/core/mpl_functions.py
+++ b/invisible_cities/core/mpl_functions.py
@@ -63,10 +63,10 @@ def plot_signal(signal_t, signal, title="signal",
 
 
 def plot_signal_vs_time_mus(signal,
-                            t_min=0,
-                            t_max=1200,
-                            signal_min=0,
-                            signal_max=200):
+                            t_min      =    0,
+                            t_max      = 1200,
+                            signal_min =    0,
+                            signal_max =  200):
     """Plot signal versus time in mus (tmin, tmax in mus). """
     tstep = 25 # in ns
     PMTWL = signal.shape[0]
@@ -74,16 +74,16 @@ def plot_signal_vs_time_mus(signal,
     ax1 = plt.subplot(1, 1, 1)
     ax1.set_xlim([t_min, t_max])
     ax1.set_ylim([signal_min, signal_max])
-    set_plot_labels(xlabel="t (mus)",
-                    ylabel="signal (pes/adc)")
+    set_plot_labels(xlabel = "t (mus)",
+                    ylabel = "signal (pes/adc)")
     plt.plot(signal_t, signal)
 
 
 def plot_pmt_signals_vs_time_mus(pmt_signals,
-                                 t_min=0,
-                                 t_max=1200,
-                                 signal_min=0,
-                                 signal_max=200):
+                                 t_min      =    0,
+                                 t_max      = 1200,
+                                 signal_min =    0,
+                                 signal_max =  200):
     """Plot all the PMT signals versus time in mus (tmin, tmax in mus)."""
 
     tstep = 25
@@ -94,8 +94,8 @@ def plot_pmt_signals_vs_time_mus(pmt_signals,
         ax1 = plt.subplot(3, 4, i+1)
         ax1.set_xlim([t_min, t_max])
         ax1.set_ylim([signal_min, signal_max])
-        set_plot_labels(xlabel="t (mus)",
-                        ylabel="signal (pes/adc)")
+        set_plot_labels(xlabel = "t (mus)",
+                        ylabel = "signal (pes/adc)")
 
         plt.plot(signal_t, pmt_signals[i])
 
--- a/invisible_cities/core/mpl_functions.py
+++ b/invisible_cities/core/mpl_functions.py
@@ -62,6 +62,44 @@ def plot_signal(signal_t, signal, title="signal",
     # plt.show()
 
 
+def plot_signal_vs_time_mus(signal,
+                            t_min=0,
+                            t_max=1200,
+                            signal_min=0,
+                            signal_max=200):
+    """Plot signal versus time in mus (tmin, tmax in mus). """
+    tstep = 25 # in ns
+    PMTWL = signal.shape[0]
+    signal_t = np.arange(0., PMTWL * tstep, tstep)/units.mus
+    ax1 = plt.subplot(1, 1, 1)
+    ax1.set_xlim([t_min, t_max])
+    ax1.set_ylim([signal_min, signal_max])
+    set_plot_labels(xlabel="t (mus)",
+                    ylabel="signal (pes/adc)")
+    plt.plot(signal_t, signal)
+
+
+def plot_pmt_signals_vs_time_mus(pmt_signals,
+                                 t_min=0,
+                                 t_max=1200,
+                                 signal_min=0,
+                                 signal_max=200):
+    """Plot all the PMT signals versus time in mus (tmin, tmax in mus)."""
+
+    tstep = 25
+    PMTWL = pmt_signals[0].shape[0]
+    signal_t = np.arange(0., PMTWL * tstep, tstep)/units.mus
+    plt.figure(figsize=(12, 12))
+    for i in range(len(pmt_signals)):
+        ax1 = plt.subplot(3, 4, i+1)
+        ax1.set_xlim([t_min, t_max])
+        ax1.set_ylim([signal_min, signal_max])
+        set_plot_labels(xlabel="t (mus)",
+                        ylabel="signal (pes/adc)")
+
+        plt.plot(signal_t, pmt_signals[i])
+
+
 def set_plot_labels(xlabel="", ylabel="", grid=True):
     """Short cut to set labels in plots."""
     plt.xlabel(xlabel)
--- a/invisible_cities/core/peak_functions.py
+++ b/invisible_cities/core/peak_functions.py
@@ -39,20 +39,20 @@ def calibrated_pmt_sum(CWF, adc_to_pes, n_MAU=200, thr_MAU=5):
 
     pmt_thr = np.zeros((NPMT, NWF), dtype=np.double)
     csum    = np.zeros(       NWF,  dtype=np.double)
+    csum_mau    = np.zeros(   NWF,  dtype=np.double)
     MAU_pmt = np.zeros(       NWF,  dtype=np.double)
 
+    MAUL=[]
     for j in range(NPMT):
         # MAU for each of the PMTs, following the waveform
         MAU_pmt = signal.lfilter(MAU, 1, CWF[j,:])
-
+        MAUL.append(MAU_pmt)
+        csum += CWF[j] * 1 / adc_to_pes[j]
         for k in range(NWF):
             if CWF[j,k] >= MAU_pmt[k] + thr_MAU: # >= not >. Found testing
                 pmt_thr[j,k] = CWF[j,k]
-
-    for j in range(NPMT):
-        for k in range(NWF):
-            csum[k] += pmt_thr[j, k] * 1 / adc_to_pes[j]
-    return csum
+        csum_mau += pmt_thr[j] * 1 / adc_to_pes[j]
+    return csum, csum_mau, np.array(MAUL)
 
 
 def wfzs(wf, threshold=0):
--- a/invisible_cities/core/peak_functions.py
+++ b/invisible_cities/core/peak_functions.py
@@ -10,35 +10,198 @@ from time import time
 import tables as tb
 import matplotlib.pyplot as plt
 
+from scipy import signal
+
 import invisible_cities.core.system_of_units as units
 import invisible_cities.sierpe.blr as blr
 import invisible_cities.core.peak_functions_c as pf
 from invisible_cities.database import load_db
 
+def calibrated_pmt_sum(CWF, adc_to_pes, n_MAU=200, thr_MAU=5):
+    """Compute the ZS calibrated sum of the PMTs
+    after correcting the baseline with a MAU to suppress low frequency noise.
+    input:
+    CWF         : Corrected waveform (passed by BLR)
+    adc_to_pes  : a vector with calibration constants
+    n_MAU       : length of the MAU window
+    thr_MAU     : treshold above MAU to select sample
+
+    NB: This function is used mainly for testing purposes. It is
+    programmed "c-style", which is not necesarily optimal in python,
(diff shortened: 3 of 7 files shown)


jmbenlloch:
I think everything is ok, but is there a reason to have this changes in two separate commits? Maybe we can squash them.

jjgomezcadenas:
Is just a question of clarity, the first commit introduces minor changes which are of cosmetic nature, the second commit is where all the meat is. I have no problem with squashing them in any case.

jmbenlloch:
Precisaly because the changes included in the first commit are minor I think it would be better to squash them, it would be easier to read the history of the main repository. But if you think it is more clear in this way, we can approve this now. Tell me and I'll give you the review.

jjgomezcadenas:
Does the Jedi have an opinion?

jacg:
When deciding whether/what to squash, you should keep two things in
mind:

1. A single commit should be easy to understand.

2. Removing, reversing or otherwise undoing the commit, in order to
   (temporarily or permanently) remove a piece of work, should not
   result in the removal of some other work.

From the description I see above, it seems that there are two
distinct pieces of work being proposed here

1. Some cosmetics.
2. A new test suite.

If we decide that the test suite is a load of rubbish, we should be
able to remove it without losing the cosmetics. If we decide that the
cosmetics are horrible and violate all our coding conventions, we
should be able to throw them away without losing the test suite.

This suggests that we need two separate commits.

The reviewer, or someone trying to understand the test suite addition
in the future, should be able to look at the diff and see changes
pertaining to the test suite only, without being distracted by
cosmetic changes. Similarly, a reviewer checking the cosmetic changes,
should not be distracted by the test suite.

This suggests that need two separate commits.

I appreciate that lots of tiny commits pollute the history to some
extent, but it is a price worth paying for maintaining orthogonality
of commits. In some sense it would be nice if we could squash cosmetic
commits into larger commits containing only cosmetic changes, and this
*can* be done, but it's almost certainly not worth the effort, as it
would require either regular rewriting of published history, or the
delay of cosmetic commits.

One might argue that these commits should have come as separate PRs,
but I think that would be a waste of effort.

I think that these being separate commits is the right thing, in
principle ... now I have to look at the actual code!

jjgomezcadenas:
OK, the reasons you argue are the reasons why I made two separated commits. 

J

jacg:
I have extracted unrelated work from this PR and re-submitted under separate PRs (#113, #114). Additionally I have removed `diomira.ipynb` (as suggested by @jjgomezcadenas elsewhere) in yet another PR (#115). I have merged all three of these into `nextic/master` and rebased this PR on top of it.

As a result we, hopefully, have a clean pull request addressing the peak function tests only.

jjgomezcadenas:
Thanks master. 
Concerning multiple (smallish) PR, one question that arises is as follows. Let us follow the scenario in this PR.

1) small cosmetic changes.
2) small tests (using hypothesis).
3) Remove an unwanted file.
4) A larger test suite.

Indeed, there is an advantage of submitting separated PR, but I am hesitant about how to branch this. All the 4 actions could be branches sprouting from master, or could be branches sprouting sequentially.

e.g master —>cosmetic-branch (PR) —> small-tests-branch (PR) —> remove-unwanted (PR) —>test-suite (PR).

What is best here? And also, if PR 2 needs say, PR1, then both should go in the same PR, right?

Just trying to learn to do it better next time…

jacg:
> Concerning multiple (smallish) PR, one question that arises is as follows.

Yes, there's something I dislike about the workflow when it comes to
multiple PRs in rapid succession, where 'rapid' means 'a subseqent PR
is opened before a previous one is closed'.

> All the 4 actions could be branches sprouting from master, or could
> be branches sprouting sequentially.  e.g master -> cosmetic-branch
> (PR) -> small-tests-branch (PR) -> remove-unwanted (PR) ->
> test-suite (PR). What is best here? And also, if PR 2 needs say,
> PR1, then both should go in the same PR, right? Just trying to learn
> to do it better next time…

Yes, it's these sorts of considerations I mean.

If the later branches don't depend on any of the changes in the
previous ones (and don't conflict with them), then it's probably
simplest and best to branch them all off master.

If one branch depends on another, then it's probably better to base
the later one on the earler one. The problem with this is that the
later PR will contain commits that belong to the earlier one. This is
not major problem, as long as the earlier one is merged into master
first. Once that is done, the later one can be rebased on top of
master, and the first PR's commits will disappear from the second
PR. In such a case the later PR should specifically mention the earlier one, to point out that the earlier one appears within it, and should, be dealt with first.

But I still have a feeling of dissatisfaction about this. I'll look
into it a bit more.

jacg:
There is something very fishy about this function:

    def toy_pmt_sum(v, npmt = 10):
        """Return the sum of npmt waveforms"""
        vsum = np.zeros(v.shape[0])
        for i in range(npmt):
            vsum = np.add(vsum, v)
        return vsum

The docstring claims that the function adds something, but what the function really does is multiply all the elements of `v` by `npmt`.

The looping variable `i` is not used anywhere. (If you are syntactically required to specify a name which you never intend to use, then you should signal this by using the name `_` (underscore): that way, people reading your code are made aware that you are *deliberately* ignoring it rather than having forgotten). But given that the function doesn't do what its docstring advertises, I suspect it's an oversight rather than being deliberate. Which leads me on to my second point.

Why are you writing a summing routine by hand in a loop? Python and numpy come with a plethora of loop abstractions, which you should use in preference to writing your own loops. There are two main reasons for doing this

1. You are less likely to make the sort of mistake that you appear to have made.
1. Readers of your code get a high-level message about what the code does, rather than having to figure it out for themselves by carefully reading your low-level code in order to extract the big picture from it?

Is there any reason why you couldn't have used `np.sum` or one of its friends for this?

jjgomezcadenas:
> On 2 Feb 2017, at 10:03, jacg <notifications@github.com> wrote:
> 
> There is something very fishy about this function:
> 
> def toy_pmt_sum(v, npmt = 10):
>     """Return the sum of npmt waveforms"""
>     vsum = np.zeros(v.shape[0])
>     for i in range(npmt):
>         vsum = np.add(vsum, v)
>     return vsum
> The docstring claims that the function adds something, but what the function really does is multiply all the elements of v by npmt.
> 
> 
Yes, this is correct. The (twisted) logic behind the argument is that I am adding 10 identical vectors (which is the same than multiplying each element of the vector by 10). This toy-sum is then compared with the CSUM being tested (which is adding all the elements of the vectors in input and does not know whether the vectors are the same or not, of course).

Thus, yes, **sorry** the real code is

> def toy_pmt_scale(v, npmt = 10):
>     “””Scale v by pmt"""
>     Return pmt*v
To repeat, what the test does is:

Compute V = toy_pmt_scale(v, npmt=10) (scale v by 10)
Compute CSUM adding an array of vectors that contains 10 identical copies of v. 

The result must be the same within tolerance.

I hope this is clearer now.

jacg:
S12[0] = []
    S12[0].append([T[0],P[0]])

is just a very convoluted way of saying 

    S12[0] = [[T[0],P[0]]]

Also, `[T[0],P[0]]` seems to be used as an immutable pair, so it's
better to use a tuple rather than a list.

jacg:
Please learn that while other languages might force you to write the equivalent of

    lo < value and value < hi

Python allows you to write the much more convenient and legible

    lo < value < hi

Please get used to doing this.

jacg:
>  The (twisted) logic behind the argument is

I would recommend avoiding twisted logic, if at all possible. Twisted logic invaiably leads to technical debt. Technical debt has to be repaid with interest, unless the project goes technically bankrupt.

It looks like this is best reviewed face-to-face. Unfortunately I am about to be otherwise occupied for the next 24 hours, or so.

jacg:
>Thus, yes, **sorry** the real code is
>
>     def toy_pmt_scale(v, npmt = 10):
>         """Scale v by pmt"""
>         return pmt*v

So, we have a function with a misleading name, misleading docstring, misleading implementation, implementing in 4 lines something that is built-in to the language, and being used in only one place?

Why can't we just say `vsum = v*10` (preferably without the magic number) in `test_csum_zs_s12`? Probably with a comment explaining what you just explained above.

I'm all for introducing abstractions, so I understand why one might want to write a named function for something that is used only once, but in this particular case it just added 4 layers of misdirection and confusion!

jjgomezcadenas:
Agreed.
Do you want me to implements this and submit a new version of the PR? Notice that I am leaving for Valencia tomorrow and will have little time to work in the upcoming week(s).

jacg:
> Do you want me to implements this and submit a new version of the PR?

No!

I'm about to push changes to the PR. Most of the stuff (largely cosmetic) I'll just squash into your original. (You will have to pull and rebase, if you want to add any changes yourself.) I'll put two new commits on top for you to consider

1. `toy_pmt_sum` -> `toy_pmt_scale`
1. Remove `toy_pmt_scale` altogeteher.

That's all I will have time for today. (I may have a bit of time towards the end of the day.)

The preferred one will have to be squashed into the main commit in the PR.

jacg:
I have extracted the orthogonal first commit and submitted it as a separate PR #116.

Please pull and rebase if you want to contribute to this branch.

jjgomezcadenas:
Sorry for being thick.

1) PR 116 needs to be approved. Can I approve it?
2) Then I can pull and rebase 
3) Then I understand you are adding stuff to PR?

jacg:
> PR 116 needs to be approved. Can I approve it?

No! (Github will let you, but you shouldn't)

You shouldn't review this, because *you wrote it* and *you are the only person who has seen it so far*!

I haven't had the time to look at this, I merely extracted it from #110 and submitted it under separate cover. @jmbenlloch could review it, but not you, as that would amount to self-review, which defeats the point of the exercise.

> Then I understand you are adding stuff to PR?

I have already added stuff to #110. Pull and rebase that. Add whatever you like on top of what is there. I would like you to decide whether `test_csum_zs_s12` is better understood with or without `toy_pmt_scale`, and push a commit that clarifies it as much as you can manage. Don't bother squashing, I'll deal with that later.

jjgomezcadenas:
The changes are very minor, so I think @jmbenlloch <https://github.com/jmbenlloch> can review it

jacg:
@jmbenlloch I've lost track on where we are with Sphinx. Have we got something set up that automatically builds the docs? Can I see how the rst file added in this PR is rendered somewhere, other than by building the docs it in my local environment?

> The changes are very minor, so I think @jmbenlloch can review it

Yes. The coding changes seem OK, but the rst looks a bit irregular.

jmbenlloch:
The docs are build here: http://ic.readthedocs.io/en/latest/ but as far as I know it is not done for pull requests, I'll check if there is a way to do it.

jacg:
`find_S12` was added to `core/wfm_to_df_functions.py`. In the process of reviewing this function, I was making some minor changes and running the full test suite to check that I hadn't broken anything. Much later, I realized that one of my changes introduced a syntax error, which got checked in! because none of the tests in our test suite checks `find_S12`.

It turns out that if you *remove the whole `wfm_to_df_functions.py` file* from the IC codebase, *all the tests pass*!

Remember, if there are no tests for it, then anyone fixing or refactoring it has no way of telling whether the changes break something. This effectively means that you don't mind if the code gets broken in the future.

Let me reiterate this point: **commiting code without accomparying tests, is essentially equivalent to saying that you don't mind if the code gets broken**!

Before approving this PR, I will go through any modules that this PR touches, and introduce random bugs: if running the full IC test suite does not alert me to the presence of such bugs, I will not approve the PR.

jjgomezcadenas:
I am afraid that I have introduced again a misunderstanding.

Actually, find_S12 was *removed* from peak_functions and sent to wmf_to_df_functions which contains a collection of currently not used (and thus not tested functions). Those functions are legacy from the quick exponential development of December. They manipulate pandas data frames. The current reconstruction (diomira & Irene) *do not* use data frames (we will presumably use them in later stages) and do not use any of the functions that I have put in wmf_to_df_functions. I should probably remove this module (together with cblr) from the repository, but before I wanted to complete the current PR. 

The test test_csum_zs_s12() in the suite ***checks*** find_S12
S12L1 = pf.find_S12(wfzs_ene, wfzs_indx,
             tmin = 0, tmax = 1e+6,
             lmin = 0, lmax = 1000000,
             stride=4, rebin=False, rebin_stride=40)

    S12L2 = cpf.find_S12(wfzs_ene, wfzs_indx,
             tmin = 0, tmax = 1e+6,
             lmin = 0, lmax = 1000000,
             stride=4, rebin=False, rebin_stride=40)

    for i in S12L1:
        t1 = S12L1[i][0]
        e1 = S12L1[i][1]
        t2 = S12L2[i][0]
        e2 = S12L2[i][1]
        npt.assert_allclose(t1, t2)
        npt.assert_allclose(e1, e2)

Checking that the python and python function work the same…
And checking that one gets the expected result

E = np.array([ 11,  12,  13,  14,  15,  16,  17,  18,  19,  20,
                   20,  20,  20,  20,  20,  20,  20,  20,  20,  20,
                   20,  19,  18,  17,  16,  15,  14,  13,  12,  11])

    for i in S12L2.keys():
        e = S12L2[i][1]
        npt.assert_allclose(e,E)

    # rebin
    S12L2 = cpf.find_S12(wfzs_ene, wfzs_indx,
             tmin = 0, tmax = 1e+6,
             lmin = 0, lmax = 1000000,
             stride=10, rebin=True, rebin_stride=10)


    E = np.array([155,  200,  155])

But of course, this is the find_S12 found in peak functions (both in python and python version), while the obsolete find_S12 version had been moved to wf_to_df. I believe that I wrote in the commit message that this was an obsolete function and was being moved out from the peak_functions module. 

I hope this explains the misunderstanding. As for introducing bugs in order to approve or not the PR. I think the current tests are meaningful and if you find them correct, you could approve them. If more tests are needed we can write them.

jacg:
Bleargh! What a mess. I recommend that you forget about this one, until we get a chance to sit down together and sort it out. The only thing we'll achieve remotely is to make an even bigger mess.

jjgomezcadenas:
As you wish. But what is a mess? The test themselves are nots that stupid, I think...

jacg:
> But what is a mess?

 Everything together. Lots of technical debt has been accumulated. We need to repay this debt: the sooner, the better. If we work on this remotely, I fear that we will accumulate more technical debt, rather than reducing it.

> Those functions are legacy from the quick exponential development of December.

Right there you have one obvious source of technical debt. The fact that I spent quite some time and effort trying to understand and improve a function which, according to the diff I had in front of my eyes, was *added* in this PR, only to find (eventually) that

1. the test suite I was running wasn't testing my changes *at all*, so I was adding crap to the code without noticing;
1. the function is legacy code and is about to be removed, so it was all pointless anyway;

is a clear example of the project *paying lots of interest on technical debt*: my time could have been spent on something useful instead; your time could have been spent on something more productive than addressing my queries. The technical debt accumulated in the code base diverted our precious resources away from things that matter.

Less haste, more speed. Avoid accumulating tecnical debt.

jjgomezcadenas:
OK, got it. I will be back next Friday. Let’s clear the mess the very same day, if possible.

jacg:
On Friday I will be *completely* unavailable and unreachable all day, I'm afraid.

jjgomezcadenas:
Saturday?

jacg:
Adding to my comments about technical debt: the interest and further accumulation of technical debt would have been much reduced, had we been pair programming.

Many people think that they cannot afford to do pair programming in their projects. In reality, they probably can't afford *not* to do it!

jacg:
> Saturday?

I don't know. Quite possibly.

PR #111, opened 2017-02-01
Add test suite for configure module.

jjgomezcadenas:
There was a subtle and hard-to-catch "bug" in configure.py. The bug
itself appears to be in argparse which in python 3 will not take an
int as an argument (as was the case in python2.7). The fix is simply
to modify the argument to str and later cast to int.

A test suite has been added, and a notebook describing the issue is
avaialble at
https://anaconda.org/nexticw/configure_test/notebook

Notice that diomira.ipynb is deleted in this commit. This goes in line
with the policy to hold all notebooks (including those of IC) in ICARO.

PR #112, opened 2017-02-01
Bug in configure module identified. Test added.

jjgomezcadenas:
configure.py fails in MacOS Xwhen passing an integer argument through argparse.
This is demonstrated by a test (test_configure), marked as xfail. The
test will fail if run normally but will not interrumpt execution of
tests. Run  pytest --runxfail configure_test.py to force a failure and
report.

Notice that diomira.ipynb is deleted in this commit. This goes in line
with the policy to hold all notebooks (including those of IC) in ICARO.


jacg:
Deleting `diomira.ipynb` has nothing to do with the argparse issue: These should be two separate commits.

Do we agree that the test fails in Python 2 aswell as Python 3?

jjgomezcadenas:
It just got in my way, and I wanted to get rid of the darned thing. 
I am not sure whether the test fails in python2 I have not tried it.

jacg:
As I pointed out in the other PR: the arguments that you write on the command line are just text. These arrive in `sys.argv` as python strings, and it is python strings which are supposed to be passed to `parser.parse_known_args`. Instead of passing strings, your test does this:

    n = 10
    s = 0
    p = 1
    CFP = conf.configure(['program_name',
                     '-c', CONF_file_def,
                     '-i', 'rwf_file',
                     '-o', 'pmp_file',
                     '-n', n,
                     '-s', s,
                     '-p', p])

Notice that `n`, `s`, and `p` are Python integers, which you pass in a list as the only positional argument to `conf.configure`. The latter binds this list to its `input_options` parameter, splits it into `program` and `args`, so `args` now contains some integers (along with some strings). Then you pass `args` (which contains some integers) to `parse_known_args` which chokes on the integers.

The problem is: you are passing integers to `parse_known_args`.

When you tell `argparse` that some argument must be an integer, you are telling it that the string it receives from the command line which corresponds to that argument, should be parsed as an integer, but it shouldn't *be* an integer *before* it is parsed. Before parsing, it should be a string! Yet you are giving it an integer.

The problem is that your test feeds rubbish to `parse_known_args`.

jjgomezcadenas:
Incidentally. I didn’t write that code.

jacg:
> I am not sure whether the test fails in python2 I have not tried it.

1. (If you haven't already got an open shell with the IC2.7 environment enabled (which, as an IC developer, you should!) Open new shell
1. `cd <wherever you have your IC source>`
1. source manage.sh work_in_python_version 2.7
1. pytest -k config --runxfail

This takes less than a minute, less than 30 seconds, unless you type veeeery slowly.

As an IC developer, working on this issue "I am not sure whether the test fails in python2 I have not tried it." is not really an acceptable comment!

jjgomezcadenas:
The problem had not been detected until now, I think,  because we were using configuration files (in which the arguments are parsed as strings) and overwriting only the names of the files (which are strings). So, how do we fix it?

jjgomezcadenas:
Apologies. Too many things at the same time.

jacg:
My point is, running all our tests in *all* our supported python versions should be a matter of habit. If you're developing something new and doing lots of iterations, by all means test it just in 3.6. But, every now and then you should give it a spin in the other versions, and **definitely** before pushing anything that will get Travis' attention, and **MOST definitely** before submitting a PR.

jacg:
> So, how do we fix it?

I still do not see that there is anything to fix other than a badly written test. The test can be fixed by replacing 

    n = 10
    s = 0
    p = 1
with

    n = '10'
    s = '0'
    p = '1'

jacg:
> It just got in my way, and I wanted to get rid of the darned thing.

I assume that this refers to `diomira.ipynb`. That's fine. Get rid of it. Commit getting rid of it in a separate commit. Get on with your life. Don't include its removal in a commit that has nothing to do with it!

jacg:
You are too quick on the trigger with closing PRs.

As I mentioned elsewhere, #111 should have been kept open and improved, instead of opening this one (#112). Rather than closing this one, it should have been kept open and

1. the notebook removal should have been removed
1. the test should have been fixed
1. the fix to the 

In summary, don't open and close multiple PRs for a single issue: improve the contents of a single PR until it is done.

Similarly, you don't *need* to have separate issues and PRs. Sometimes it's better to open a PR (even if it is not ready yet) rather than opening an issue and then a separate PR which addresses that issue. Issues and PRs are very similar. If you have some code which can act as the beginning of a solution to an issue, you might as well start a PR rather than an issue.

jacg:
Just to emphasize: that test, even though there's a bug in it, is valuable. The solution is not to kill the test, the solution is to fix the bug in the test.

PR #119, opened 2017-02-03
Show how to write matplotlib tests with image comparison (close #106)

### LINKED ISSUE #106: Investigate how plot functions might be tested
Our code base contains functions whose purpose is to create plots. These are currently untested. Options include

1. not test them at all
1. ad-hoc test them, perhaps in notebooks
1. try to write proper automated tests

@jacg is to investigate the feasibility of option 3

@jjgomezcadenas please make a list of all such functions. (I would do it myself, but given that matplotlib is sometimes used in our code to perform calculations rather than to make plots, I'm likely to get it wrong :-)

jacg:
- [ ] Eliminate inconsistencies between Python 3.6 and other versions
- [ ] Ensure that setting of Agg backend works, and that there are no
      crashes on Travis

Added nose to the conda environment, because matplotlib uses it in its
testing infrastructure. [This should be unnecessary in the long run, because of matplotlib's [migration to pytest](https://github.com/matplotlib/matplotlib/issues/6731)]

Added pytest-mpl to the conda environment, because it facilitates
using matplotlib's testing infrastructure in pytest.

Note that the matplotlib backend must be set to Agg *before* impolting
`matplotlib.pyplot` in order to avoid crashes in headless
environments.

It seems that the labels on the y-axis are rendered slightly higher in
Python 3.6 than in Pythons 3.5 and 2.7. Consequently the example test
fails in Python 3.6.

**NOTE: agg backend setting still doesn't seem to work correctly** in
spite of having tried a couple of solutions.


jjgomezcadenas (commented):
Travis is still failing... is this what we expect?

jacg:
> Travis is still failing... is this what we expect?

Please inspect the checkboxes at the top of this PR

jjgomezcadenas:
If I understand well Agg backend is not working on travis and the creates the failure, right?

jacg:
Not quite.

Travis is headless: it does not have a `$DISPLAY` on which any graphical output can be drawn. `Agg` is the only matplotlib backend which does not require a `$DISPLAY`. So we need to run this test with `Agg` as the backend, but this turns out to be difficult to achieve: importing `matplotib.pyplot` automatically sets some other default backend. It is probably this, possibly combined with stuff that happens when pytest does discovery, that makes all my attempts at choosing `Agg` fail. Until we manage to get the test to run under `Agg` it will keep failing on Travis.

Apart from that, *you* sould run this test on your local machine to see if you observe the same inconsistency between Python 3.6 and the other versions, as I do.

jmbenlloch:
I've managed to "hack" a little bit our travis machines to get a reverse shell (this job https://travis-ci.org/jmbenlloch/IC-1/jobs/204364293)

Along several runs I have checked that despite using Agg backend it doesn't work. A displacement in the labels of y-axis can be seen. In particular I have looked at the run with python 2.7 in travis, you can find the images generated here: http://cern.ch/jm.benlloch/mpl_travis.zip

jacg:
There are two orthogonal problems here:

1. Get the jobs not to crash on Travis because of its headless environment.

1. Discrepancy between Python 3.6 & the rest WRT y-axis label position.

I think that the first step is to get the tests to run on Travis *without hacking Travis*, rather by persuading pytest to run the tests with an appropriate backend. Once we have that, if we want to use these kinds of tests, then the cheap-and-pragmatic solution is probably to `@skipif` them depending on Python version.

jjgomezcadenas:
What do we do about this? It looks to me that checking plots via PYTEST may not be worth the effort. Opinions, everyone, but in particular @jacg?

jacg:
I would leave this open, quietly waiting in the background until we need it. Certainly not a priority for now, but the knowledge about how to do it should be available.

I've rebased it, so that it easy to apply whenever the need arises.

jacg:
It seems that matplotlib is [migrating from nose to pytest](https://github.com/matplotlib/matplotlib/issues/6731), which should make things slightly easier for us, if and when we pick this on up.

PR #120, opened 2017-02-04
Add comprehensive test for configuration parser.

jacg:
This is an expansion of the test that appeared in JJ's (withdrawn) PRs
#111 and #112.

Rather than using the real `irene.conf` file (and commenting in the
test that the test will fail if anyone ever changes that file), we
create temporary config files from within the test itself. This allows
us to inject a variety of values into the config file used by the
test, aswell as ensuring that changes to `irene.conf` will not break
the test.

We take care to use different values for all the settings wherever
possible, to minimize the chances of accidentally getting the right
answer.

The test is parametrized, to allow easy orthogonal testing of
independent arguments, and easy addition of new test cases.

### DIFF
--- a/invisible_cities/conftest.py
+++ b/invisible_cities/conftest.py
@@ -3,3 +3,7 @@ import pytest
 @pytest.fixture(scope='session')
 def irene_diomira_chain_tmpdir(tmpdir_factory):
     return tmpdir_factory.mktemp('irene_diomira_tests')
+
+@pytest.fixture(scope='session')
+def config_tmpdir(tmpdir_factory):
+    return tmpdir_factory.mktemp('configure_tests')
--- a/invisible_cities/core/configure.py
+++ b/invisible_cities/core/configure.py
@@ -37,41 +37,28 @@ def configure(input_options=sys.argv):
     """
     program, args = input_options[0], input_options[1:]
     parser = argparse.ArgumentParser(program)
-    parser.add_argument("-c", metavar="cfile", type=str,
-                        help="configuration file", required=True)
-    parser.add_argument("-i", metavar="ifile", type=str,
-                        help="input file")
-    parser.add_argument("-o", metavar="ofile", type=str,
-                        help="output file")
-    parser.add_argument("-n", metavar="nevt", type=int,
-                        help="number of events to be processed")
-    parser.add_argument("-s", metavar="skip", type=int, default=0,
-                        help="number of events to be skipped")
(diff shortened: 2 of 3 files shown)


jjgomezcadenas (commented):
This commit seems to do two different things. One is declared in the comment (the comprehensive parser) the other is not referred to anywhere (the modification in configure.py file). 

I have not understood the need to modify this file (it the code not working properly? are those cosmetic changes only?) but it seems to me that the teaching of the Jedi would probably recommend to different commits (and maybe to different PRs). If the changes to configure.py are cosmetic, two commits should do. If the changes are more substantial, then I think we are talking about two different PRs, since you are proposing to modify code that works (as far as I know) correctly. So there must be a good reason for the changes and probably you want to explain the whys.

jacg:
The changes to `config.py` are mostly cosmetic. There are one or two which use standard higher-level utilities in preference to error-prone hand-coding (`path.join`). There may also have been a small bugfix, but I don't remember exactly.

I was in two minds as to whether make two separate commits. In the end, given that 100% of the new code is contained in one file, while 100% of the other changes are elsewhere, I put it all together. But if you would prefer to separate this into two commits, then that's absolutely fine with me.

I'd be less inclined to put it in separate PRs: it doesn't seem worth it; but I wouldn't object strongly to it either.

> code that works (as far as I know) correctly.

A bold claim, for code that had no tests!

jjgomezcadenas:
Just trying to understand… the commit message didn’t say anything about the changes in configure.py thus I was not sure what to expect. On the other hand, I did write tests for configure.py (remember I made a PR, later withdraw, but the test did work in my machine and in Travis)…

Anyhow, no need to make a new commit or PR.

jjgomezcadenas (approved):
Although I am not able to follow completely the code (perhaps the hour or perhaps my limitations as programmer) I believe I understand the logic, and it is certainly much more powerful than the original version.

jacg:
>  I did write tests for configure.py (remember I made a PR, later withdraw, but the test did work in my machine and in Travis)…

Which is the very first thing I said in the description of this PR.

You may also recall that the test you wrote was feeding integers into a function that was expecting `sys.argv` (which only ever contains strings) and that the test was therfore failing. I told you how to fix the test, but you withdrew it instead.

You may also recall that your test led you to conclude that `argparse` couldn't handle integers in Python3, even though it could in Python2 ... a conclusion that is not consistent with reality.

In short, I don't think your test told you much about the correctness of the functionality it was supposed to test!

PR #122, opened 2017-02-04
Test and fix cast

jacg:
A hypothesis-based test was added, to catch the obvious (and maybe
some not so obvious) failures in `configure.cast`.

The original implementation tried to recognize ints and floats with an
extremely naive hand-written parser. The new implementation uses
Python's build-in parsers for floats and ints: their constructors.


jjgomezcadenas (approved):
A text-book improvement of the source code and very illustrative example of hypothesis-based test.

PR #123, opened 2017-02-09
Add new city Isidora

jmbenlloch:
City to compute corrected waveforms (CWF) and store them in an HDF5 file. There is one test included which check that Isidora runs and that the corrected waveforms is equal to the BLR one within 1%. One sample config file is also included.

### DIFF
--- a/invisible_cities/cities/isidora.py
+++ b/invisible_cities/cities/isidora.py
@@ -1,0 +1,221 @@
+from __future__ import print_function
+import sys
+
+from glob import glob
+from time import time
+import numpy as np
+import tables as tb
+
+from   invisible_cities.core.nh5 import RunInfo, EventInfo
+from   invisible_cities.core.nh5 import S12, S2Si
+import invisible_cities.core.mpl_functions as mpl
+import invisible_cities.core.pmaps_functions as pmp
+import invisible_cities.core.tbl_functions as tbl
+from   invisible_cities.core.configure \
+       import configure, define_event_loop, print_configuration
+
+import invisible_cities.sierpe.blr as blr
+import invisible_cities.core.peak_functions_c as cpf
+from   invisible_cities.core.system_of_units_c import SystemOfUnits
+from   invisible_cities.database import load_db
+
+units = SystemOfUnits()
+
+
+class Isidora:
+    """
+    The city of ISIDORA performs a fast processing from raw data
+    (pmtrwf and sipmrwf) to BLR wavefunctions.
+    It is optimized for speed (use of CYTHON functions) and intended
+    for fast processing of data.
+    """
+    def __init__(self, run_number=0):
+        """
+        Init the machine with the run number.
+        Load the data base to access calibration and geometry.
+        Sets all switches to default value.
+        """
+        # NB: following JCK-1 convention
+        self.run_number = run_number
+
+        DataPMT  = load_db.DataPMT (run_number)
+        DataSiPM = load_db.DataSiPM(run_number)
+
+        # This is JCK-1: text reveals symmetry!
+        self.xs              = DataSiPM.X.values
+        self.ys              = DataSiPM.Y.values
+        self.adc_to_pes      = abs(DataPMT.adc_to_pes.values).astype(np.double)
+        self.sipm_adc_to_pes = DataSiPM.adc_to_pes.values    .astype(np.double)
+        self.coeff_c         = DataPMT.coeff_c.values        .astype(np.double)
+        self.coeff_blr       = DataPMT.coeff_blr.values      .astype(np.double)
+
+        # BLR default values (override with set_BLR)
+        self.n_baseline   = 28000
+        self.thr_trigger  = 5 * units.adc
+
+        # MAU default values (override with set_MAU)
+        self.  n_MAU      = 100
+        self.thr_MAU      = thr_MAU = 3 * units.adc
+
+        # set switches
+        self.setFiles     = False
+        self.setSiPM      = False
+        self.setCwfStore  = False
+
+        # defaul parameters
+        self.nprint       = 1000000
+        self.signal_start =       0
+        self.signal_end   =    1200
+
+    def set_print(self, nprint=10):
+        """Print frequency."""
+        self.nprint = nprint
+
+    def set_input_files(self, input_files):
+        """Set the input files."""
+        self.input_files = input_files
+        self.setFiles    = True
+
+    def set_BLR(self, n_baseline=38000, thr_trigger=5 * units.adc):
+        """Parameters of the BLR."""
+        self.n_baseline  = n_baseline
+        self.thr_trigger = thr_trigger
+
+    def set_MAU(self, n_MAU=100, thr_MAU=3 * units.adc):
+        """Parameters of the MAU used to remove low frequency noise."""
+        self.  n_MAU =   n_MAU
+        self.thr_MAU = thr_MAU
+
+    def set_cwf_store(self, cwf_file, compression='ZLIB4'):
+        """Set the input files."""
+        # open cwf store
+        self.cwfFile = tb.open_file(
+            cwf_file, "w", filters=tbl.filters(compression))
+
+        # create a group
+        self.cwf_group = self.cwfFile.create_group(self.cwfFile.root, "BLR")
+
+        self.setCwfStore = True
+        self.compression = compression
+
+    def store_cwf(self, cwf):
+        """Store CWF."""
+        NPMT, PMTWL = cwf.shape
+        if "pmtcwf" not in self.cwf_group:
+            # create earray to store cwf
+            self.pmtcwf = self.cwfFile.create_earray(self.cwf_group, "pmtcwf",
+                                                atom    = tb.Float32Atom(),
+                                                shape   = (0, NPMT, PMTWL),
+                                                filters = tbl.filters(self.compression))
+        self.pmtcwf.append(cwf.reshape(1, NPMT, PMTWL))
+
+
+    def run(self, nmax):
+        """
+        Run the machine
+        nmax is the max number of events to run
+        """
+        n_events_tot = 0
+        # check the state of the machine
+        if not self.input_files:
+            raise IOError('input file list is empty')
+
+        if not self.setFiles:
+            raise IOError('must set files before running')
+
+        print("""
+                 ISIDORA will run a max of {} events
+                 Input Files ={}
+                          """.format(nmax, self.input_files))
+
+        # loop over input files
+        first = False
+        for ffile in self.input_files:
+            print("Opening", ffile, end="... ")
+            filename = ffile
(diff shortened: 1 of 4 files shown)


jacg:
Something fishy is going on: when I run the test suite simultaneously for our 3 supported python versions, `test_isidora_run` usually fails on one or two of them. If I run it one python version at a time, it (after a small number of attempts so far) seems to pass every time.

At first glance, you seem to be writing things in per-test-run temp directories, so there should be no interference.

Any ideas?

invisible_cities/cities/isidora_test.py
```diff
@@ -0,0 +1,124 @@
...
+
+    eps = 1
+    with tb.open_file(RWF_file, 'r') as h5in, tb.open_file(CWF_file, 'r') as h5out:
+        pmtblr = h5in.root.RD.pmtblr
+        pmtcwf = h5out.root.BLR.pmtcwf
+        for event in range(nevt):
+            BLR = pmtblr[event]
+            CWF = pmtcwf[event]
+            for pmt in range(len(BLR)):
+                diff = abs(np.sum(BLR[pmt][4950:5150]) \
```
jacg: Perhaps it would be helpful to replace the 4950 and 5150 *magic numbers* with meaningfully named variables?

invisible_cities/cities/isidora.py
```diff
@@ -0,0 +1,231 @@
...
+
+                        n_events_tot +=1
+                        if n_events_tot%self.nprint == 0:
+                            print('event in file = {}, total = {}'
+                                  .format(evt, n_events_tot))
+
+                        if n_events_tot >= nmax and nmax > -1:
+                            print('reached max nof of events (={})'
+                                  .format(nmax))
+                            break
```
jacg: What is the point of the `nmax > -1` check? Can `nmax` ever be less than 0? I imagine that `nmax == 0` (or even `nmax >= 0` could be taken to mean 'unlimited number of events' in which case this meaning  needs to be mentioned in the docstring ... and in which case the check should be

    nmax > 0 and n_events_tot >= nmax

which is expressed more pythonically (but perhaps less clearly from the motivation perspective) as 

    n_events_tot >= nmax > 0

invisible_cities/cities/isidora.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+            except Exception:
+                print('error')
+                raise
```
jacg: This pattern (printing 'error' whenever an exception comes out) is repeated throughout the cities. Is it really necessary? Does it really add anything? I have a suspicion that sometime, somewhere, someone who doesn't understand Python very well, put this in somewhere, and now it is being copied all over the place.

Can we identify the reason for this? If no good reason exists, could we please put a stop to this practice and excise all occurances from our code? (I'll do it in a separate PR

jacg: Please see my [remove-error-print branch](https://github.com/jacg/IC/tree/remove-error-print). It seems that the only other place where this is done in the new IC is `Irene.run()`. @jjgomezcadenas Please review the changes in the aforementioned branch to see whether there is any point in keeping this noise (best discussed face-to-face) and whether the finally-based version is preferable.

jacg:
I notice that `isidora_test.py:config_file_spec` contains
    PATH_OUT = '$ICDIR/database/test_data/'

and that `cwf_electrons_40keV_z250_RWF.h5` is therefore written by the test into our source tree. Tut-tut, very naughty!

This is

1. Probably the cause of the failing tests when running the tests for different python versions simultaneously (two tests trying to write to the same file at the same time, and getting in eachother's way).
1. Likely to cause headaches like the ones we had last time a test was writing output files into the source tree.

This test should write its output into a temp directory created by pytest.

jmbenlloch:
Thank you for your detailed review. Several comments:

- I have removed the magic numbers and put two variables for them (peak_start and peak_end).
- I have changed the dictionary with the configuration to use the temp dir.
- Regarding the nmax = -1 condition, you are right, that is a convention meaning no limit, run over all events in the input files. If we want to change that we need to change also the configuration of all cities, where we have:

`nevts = CFP['NEVENTS'] if not CFP['RUN_ALL'] else -1`

invisible_cities/cities/isidora_test.py
```diff
@@ -107,16 +103,19 @@ def test_isidora_run(config_tmpdir):
...
         for event in range(nevt):
             BLR = pmtblr[event]
             CWF = pmtcwf[event]
             for pmt in range(len(BLR)):
-                diff = abs(np.sum(BLR[pmt][4950:5150]) -
-                           np.sum(CWF[pmt][4950:5150]))
+                diff = abs(np.sum(BLR[pmt][peak_start:peak_end]) -
+                           np.sum(CWF[pmt][peak_start:peak_end]))
                 diff = 100 * diff / np.sum(pmtblr[event][pmt])
                 assert diff < eps
```
jacg: Do we really need a loop here? Can't this be expressed as an array operation and/or some numpy array assertion?

jjgomezcadenas:
What is wrong with the loop?
The test asserts that the difference between the CWF and the BLR in the given range 
 *for all events* and *for all pmts* is smaller than a small quantity. This is a legitimate and in my opinion clear way to express the assertion. While it probably can be expressed some other way, nothing fundamental is won, I think, by investing time in re-stating something that is clearly stated already.

invisible_cities/cities/isidora.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def run(self, nmax):
+        """
+        Run the machine
+        nmax is the max number of events to run
+        """
+        n_events_tot = 0
+        # check the state of the machine
+        if not self.input_files:
+            raise IOError('input file list is empty')
```
jacg: I see that this habit of using `IOError` in these situations is spreading throughout IC. I don't think that that is appropriate in such caes. Consider that from Python 3.3 onwards, `IOError` is just a synonym for `OSError`, and I hope that it's pretty clear that this is not an OS error! I think it would be better if we had specific IC exceptions with the exact meanings that we want to convey, and that we use them in these situations.

Let's nip this in the bud before too many people copy this bad practice!

jjgomezcadenas:
My poor understanding is that this was Input Output Error, which seemed appropriated. Please, suggest an specific alternative.

Also, I worry that we are spreading the review far and wide… many of the comments and issues to isidora (and I suspect to Maurilia) are in reality revisions to irene. I hope that we are not moving in circles…

jacg:
> What is wrong with the loop?

You should not write loops by hand, when you have higher-level ways of expressing the same idea. Thus, you should use `map`, `filter` and `reduce` rather than loops (any blog posts which claim the contrary are simply *wrong*) ... and you certainly shouldn't write loops over numpy arrays if you can broadcast over the array or use some ufunc or builtin loop operation.

> The test asserts that the difference between the CWF and the BLR in the given range *for all events* and *for all pmts* is smaller than a small quantity.

Which sounds like something that can be expressed much more cleanly with numpy array operations and numpy asserts. Please remember that operations on numpy arrays are designed to do things *for all* things in the array by default, *far more efficiently* and *far more concisely* than hand-coded loops. Hand-coded loops on numpy arrays are pretty much the number one numpy anti-pattern.

> While it probably can be expressed some other way, nothing fundamental is won, I think, by investing time in re-stating something that is clearly stated already.

If it were already stated as clearly as it can be stated, then I would agree with you, but I doubt that it is the case. If it is not stated as clearly as as possible, then it most certainly is worth reducing the technical debt of the code.

jjgomezcadenas:
Well, I think it is clearly stated, using legit python code. I wouldn’t call the use of a loop technical debt. And most of all, the test works, and we have too much work to do to stop and fix something  that is not broken. 

I have written other tests using numpy asserts but this one was clear enough, in my opinion in the way it was written and I still think you have not offer any convincing argument why is not the case. 
It states that for all events (first loop) and for all pmts (second loop) something minus something is small. It may not be the most elegant way to express it, but certainly is clear enough.

jjgomezcadenas:
Agree, the try/except block here seems pointless, I believe finally is not needed, except perhaps to announce happy ending. We are opening already the files using a with context, so no need to release them in case of trouble

jmbenlloch:
In my last commit I  have removed the for loops. Now the test is done using np.testing.all_close. We sum the region of interest for each pmt and for all events. This is compared with:

`np.testing.assert_allclose(pmtcwf_sum, pmtblr_sum, rtol=0.01, atol=0)`

The operation done is:

`absolute(pmtcwf - pmtblr) <= (0.01 * absolute(pmtblr))`

So we check the are the same within 1%.

jmbenlloch:
I have removed the print error, since it didn't seem to be useful.

I have also created a new exception called NoInputFiles, which should be raised if no input files are defined. This is defined in a new file `exceptions.py` in `core` module. Suggestions on where to defined IC exceptions are welcomed.

invisible_cities/core/exceptions.py
```diff
@@ -0,0 +1,7 @@
+"""
+Define IC-specific exceptions
+"""
+
+class NoInputFiles(Exception):
+    """ Input files list is not defined """
+    pass
```
jacg: Nice.

Could we please make this a subclass of `InvisibleCitiesException` or `ICException`.

Our exception hierarchy will require a but more thought, but this is fine for now.

jmbenlloch: I have already updated it in the last commit.

jacg:
I propose to squash this PR into three commits:

1. Add isidora (with everything not covered by the next two squashed into it)
1. Vectorize loops (to serve as a guiding example)
1. Add exceptions (to emphasize that we have started our own collection of exceptions)

I have not had the opportunity to look at the logic or the adequacy of the tests: I'll defer that until after the PR has been merged, as I'll have to do something similar for #110 and #124, and that is best done after talking to @jjgomezcadenas in person.

I have not given the slightest thought to the physical significance or desirability of this city: I trust that someone more qualified in that department than me, is reviewing that side of things; If I don't hear anything on that side, I'll approve and merge this PR in the next 24h.

I'll prepare the squashed version and push it to my fork so that you can have a look at exactly what I'm proposing.

jacg:
Oh, before I forget, I don't like the "-1 maxevents means unlimited" convention. It's very C-like. Using `None` to mean 'no limit' would be much more Pythonic. And `0` would be better than `-1` (but not as good as `None`).

But I don't want to delay this PR with that issue, so this is mostly a reminder to myself to look at it later.

jacg:
The squashed version is available for viewing [here](https://github.com/jacg/IC/tree/isidora)

jjgomezcadenas:
My five cents:
What this city does is simply to write down the deconvoluted waveforms in an hd5 file. This corresponds to a request of Pau et al, who want to compare the deconvoluted functions (offline, done by IC) with the deconvoluted functions online in the FPGA. The test checks that deconvoluted function (in the MC) agrees with the “true” function. 

From my side, the city and the tests appear OK.

PR #124, opened 2017-02-09
Maurilia

jerenner:
City extracting MC truth information (MCTracks PyTables) from HDF5 file which may contain other PyTables


jacg:
This history will have to be cleaned up before merging into nextic/master. I can do it for you, but perhaps you should practise doing it yourself.

jacg:
The test writes two files (`NEW_se_mc_1evt.h5` and `NEW_se_mc_reco_pmaps_1evt.h5`) into the source directory. Any files produced by the tests must be written into per-test temporary directories created by pytest. See `diomira_test.py` for inspiration.

jacg:
You can also get inspiration from [this commit](https://github.com/nextic/IC/pull/123/commits/03acc6e1c2532190b56605f32eb55462cdf36a6f) which solves exactly the same problem.

invisible_cities/cities/maurilia.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+    The city of MAURILIA extracts MC truth PyTable from HDF5 files.
+    """
+    def __init__(self):
+        """
+        Sets all switches to default value
+        """
+        # Set the machine default state
+        self.set_input_files_    = False
+        self.set_output_file_    = False
```
jacg, line 39: Why the trailing underscores?

invisible_cities/cities/maurilia.py
```diff
@@ -0,0 +1,123 @@
...
+                          tbl.filters("NOCOMPR"))
+
+        # loop over input files
+        first=False
+        for ffile in self.input_files:
+
+            print("Opening file {}".format(ffile))
+            filename = ffile
+            try:
+                with tb.open_file(filename, "r+") as h5in:
```
jacg, line 88: Is there any good reason why the file mode is `r+` rather than just `r`?

PR #125, opened 2017-02-10
Added city Maurilia and corresponding data file and test

jerenner:
- The previous pull request was accidentally closed by deleting the branch
- The file NEW_se_mc_reco_pmaps_1evt.h5 is not written by the test but is data used to run the test.  The file NEW_se_mc_1evt is now written to a temporary directory in the test, as is the configuration file itself.
- The trailing underscores in for example  self.set_input_files_  are to avoid conflict with the corresponding function definition names


jmbenlloch:
@jerenner, you were creating the conf file in a tmp dir, but you were not passing that file to the configure module, I have fixed that for you. Due to that error, your test was failing. Remember that you should run the test before submitting a pull request, you can do that locally by running `pytest` and also in your fork's travis account. You can see in your [branch commits](https://github.com/jerenner/IC-1/commits/maurilia) that the test were failing (the red X). In that [link](https://travis-ci.org/jerenner/IC-1/builds/200326525) you can see all the builds that travis has tried and their error messages.

Apart from that, I think you forgot to include an configuration file example, maybe it was lost when we tried to fix your repository. Please include it in a new commit to your branch, it will automatically appear in this pull request.

jacg:
> The trailing underscores in for example self.set_input_files_ are to avoid conflict with the corresponding function definition names

Yes, I figured it out eventually.

You seem to be leaving a couple of instance attributes unset in the constructor (`input_files` and `h5out`) and then tracking whether they have been set with two more attributes (`set_input_files_` and `set_output_file_`). I suggest you replace this scheme with one of the following two. In both cases you should scrap `set_input_files_` and `set_output_file_`, and then 

1. use `hasattr(self, "input_files")` and `hasattr(self, "h5out") to check whether they have been set yet.

1. Initialize `input_files` and `h5out` to `None` in the constructor, and use `<whichever> is None` to check whether the have been set yet.

jjgomezcadenas:
OK, this suggestion affects not only Josh’s code, but also every state machine in IC, so we need to look at it with some detail.

invisible_cities/cities/maurilia_test.py
```diff
@@ -0,0 +1,71 @@
...
+COMPRESSION {COMPRESSION}
+"""
+
+config_file_spec = dict(PATH_IN  = '$ICDIR/database/test_data/',
+                        FILE_IN  = 'NEW_se_mc_reco_pmaps_1evt.h5',
+                        COMPRESSION = 'ZLIB4')
+
+config_file_contents = config_file_format.format(**config_file_spec)
+
+@mark.serial
```
jacg: The `serial` mark should be used in tests that cannot be run concurrently with other tests. The only reason for this that we have identified so far in IC is those tests which form a chain where some test uses files produced by another one.

I do not see why this test should be marked as serial.

jerenner:
City Maurilia has been modified to derive from class City

jacg:
Once again, please do not mark tests as serial, unlsess thery really are serial.

jacg:
The work proposed here has been included and extended in #127.

PR #135, opened 2017-02-21
Copy SIPM waveforms in Isidora

jmbenlloch:
Reading back from ART needs SiPM data in the file. This pull request copies SiPM waveforms from input file(s) to output file.


jjgomezcadenas (approved):
The additions appear to be correct, but a test is missing. Since the test itself is not obvious, I approve the PR, but we should try to find a suitable test for this case.

PR #138, opened 2017-02-23
Avoid repetition of namedtuple name

jacg:
Standard usage of namedtuples violates the DRY (Do not Repeat
Yourself) principle:

    Name = namedtuple('Name', 'a b c d')

requires the repetition of `Name`.

With isolated uses of `namedtuple` it is not worth worrying about this
too much, but in `core/params.py` we create a whole bunch of
`namedtuple`s in succession, and expect more in the future.

So we get rid of the name repetition with a little metaprogramming:
specify the name and the attributes of the `namedtuple` and loop over
these specs to create them and bind them to the module.

Note that the original code already contained a mismatch in the
repeated information for SensorParam(s)!

### DIFF
--- a/invisible_cities/core/params.py
+++ b/invisible_cities/core/params.py
@@ -1,8 +1,19 @@
+import sys
 from collections import namedtuple
 
-S12Params       = namedtuple('S12Params', 'tmin tmax stride lmin lmax rebin')
-SensorParams    = namedtuple('SensorParam', 'NPMT PMTWL NSIPM SIPMWL')
-ThresholdParams = namedtuple('ThresholdParams',
-                             'thr_s1 thr_s2 thr_MAU thr_sipm thr_SIPM')
-CalibratedSum   = namedtuple('CalibratedSum', 'csum csum_mau')
-PMaps           = namedtuple('PMaps', 'S1 S2 S2Si')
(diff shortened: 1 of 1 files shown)


jjgomezcadenas (commented):
Indeed, elegant meta-programming. Perhaps one should say in favour of the old style that the naked eye (at least the naked eye of the unexperienced python programmer) "sees" the old version but needs to parse the new version (at least I had to parse it and it took me a while). I have no problem with approving this PR, but would like to hear the Jedi's reaction to the question of "seeing" versus "parsing".

jacg:
Yes, that's what I dislike about the new version: it requires a bit more reading than the original. But because it

+ is confined to just one place
+ is the *whole* implementation of the module
+ reduces repetition
+ has already eliminated an inconsistency present in the original
+ makes it impossible to make such a mistake again

I thought it would be worthwhile.

Actually, I have pushed an alternative version which should make it easier to read. See what you think of it.

jjgomezcadenas (approved):
Yes, this new version reduces the parsing effort, the reviewers (peras @jmbenlloch & @jjgomezcadenas) find it's a clear improvement.

jacg:
OK, I'll squash and then merge into master.

PR #142, opened 2017-02-25
fix unintended behaviour in pmaps_functions.py

gonzaponte:
The function s12df_to_s12l intented to convert evt_max events from a DataFrame to a dictionary failed to stop when evt_max index is missing in the input DataFrame. This bug is fixed and some functionality added:
- Passing a negative number as evt_max results in a “read all” behaviour.
- Default value for evt_max is None (which is converted to -1 for the Cython function).
- evt_max is implemented as long int to avoid precision limitations.


jjgomezcadenas (commented):
This commit fixes a bug in s12df_to_s12 and adds a test that checks that the function behaves as expected. I recommend taking the PR

jjgomezcadenas (commented):
This PR fixes a bug in function s12df_to_s12l. The argument evt_max was set to a small number (10) introducing two problems. i) it did not take into account that it was not guaranteed that the event number is found in the table and ii) it failed to specify a correct default behaviour (which should be to read the full table). This PR fixes both problems and add a test to demonstrate correct behaviour of the function

jacg:
I have reorganized this into three separate commits.

1. Add your original monolithic test.
1. Split this monolithic test into four separate tests, which share a fixture.
1. Add the fix.

A few things remain to be done

1. Have a look at how the fixture mechanism works, and incorporate it into your test-writing workflow.
1. Check whether I have left the code in a state that you are happy with. Do any minor tweaks you want to.
1. Squash it all into a single commit.

After that I'll be happy to accept it.

jacg:
Hmm, I see that the comments inside the tests could be improved.

jacg (approved):

PR #144, opened 2017-02-26
Fit functions

gonzaponte:
New module to deal with data fitting. Includes several functionalities and defines the most common functions.

### DIFF
--- a/invisible_cities/core/fit_functions_test.py
+++ b/invisible_cities/core/fit_functions_test.py
@@ -1,0 +1,290 @@
+"""
+Tests for fit_functions
+"""
+
+import numpy as np
+from pytest import mark
+from numpy.testing import assert_array_equal, assert_allclose
+from hypothesis import given
+from hypothesis.strategies import integers, floats
+from hypothesis.extra.numpy import arrays
+
+import invisible_cities.core.fit_functions as fit
+
+
+# TODO: Move to a different module {
+def float_arrays(size       =   100,
+                 min_value  = -1e20,
+                 max_value  = +1e20,
+                 mask       =  None,
+                 **kwargs          ):
+    elements = floats(min_value,
+                      max_value,
+                      **kwargs)
+    if mask is not None:
+        elements = elements.filter(mask)
+    return arrays(dtype    = np.float32,
+                  shape    =       size,
+                  elements =   elements)
+
+
+def FLOAT_ARRAY(*args, **kwargs):
+    return float_arrays(*args, **kwargs).example()
+
+
+def random_length_float_arrays(min_length =     0,
+                               max_length =   100,
+                               **kwargs          ):
+    lengths = integers(min_length,
+                       max_length)
+
+    return lengths.flatmap(lambda n: float_arrays(       n,
+                                                  **kwargs))
+# }
+
+@mark.slow
+@given(random_length_float_arrays())
+def test_in_range_infinite(data):
+    assert fit.in_range(data).all()
+
+
+@mark.slow
+@given(random_length_float_arrays(mask = lambda x: ((x<-10) or
+                                                    (x>+10) )))
+def test_in_range_with_hole(data):
+    assert not fit.in_range(data, -10, 10).any()
+
+
+def test_in_range_positives():
+    data = np.linspace(-10., 10., 1001)
+    assert np.count_nonzero(fit.in_range(data, 0, 10)) == 500
+
+
+@mark.slow
+@given(random_length_float_arrays(max_length = 1000))
+def test_in_range_right_shape(data):
+    assert fit.in_range(data, -1., 1.).shape == data.shape
+
+
+def test_get_errors():
+    data = np.array([[ 9, -2,  1],
+                     [-1,  4, -1],
+                     [ 2, -4,  1]], dtype=float)
+    assert_array_equal(fit.get_errors(data), [3., 2., 1.])
+
+
+@given(float_arrays(min_value=-10,
+                    max_value=10))
+def test_gauss_symmetry(data):
+    assert_allclose(fit.gauss(data, 1., 0., 1.),
+                    fit.gauss(-data, 1., 0., 1.))
+
+
+#def test_gauss_breaks_when_sigma_is_negative():
+#    assert_raises(ValueError,
+#                  fit.get_from_name,
+#                  funname)
+
+
+@given(integers(min_value = -100,
+                max_value = +100),
+       integers(min_value = -100,
+                max_value = +100),
+       integers(min_value = -100,
+                max_value = +100),
+       integers(min_value = -1  ,
+                max_value = +100))
+def test_gauss_works_with_integers(x, a, b, c):
+    args = [x, a , b, c]
+    assert_allclose(fit.gauss(*args),
+                    fit.gauss(*map(float, args)))
+
+
+@given(random_length_float_arrays(min_length = 1,
+                                  max_length = 10,))
+def test_polynom_at_zero(pars):
+    assert_allclose(fit.polynom(0., *pars),
+                    pars[0])
+
+
+@given(random_length_float_arrays(min_length = 1,
+                                  max_length = 10,
+                                  min_value  = 1e-8))
+def test_polynom_at_one(pars):
+    assert_allclose(fit.polynom(1., *pars),
+                    np.sum(pars), rtol=1e-5)
+
+
+@given(floats(min_value = -20.,
+              max_value = -10.),
+       floats(min_value = -20.,
+              max_value = -10.))
+def test_expo_double_negative_sign(x, mean):
+    assert fit.expo(x, 1., mean) > fit.expo(x+1., 1., mean)
+
+
+@given(floats(min_value = +10.,
+              max_value = +20.),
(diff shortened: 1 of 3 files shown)


jacg:
1. This branch contains two merge commits. Please remove them.
1. Run the tests with `pytest -k fit_functions --durations=0` and  mark the slow ones with `@mark.slow`. There seems to be one which takes about 40s, and three which take around 1s, and the rest are all faster than a third of a second. You seem to have marked three tests as slow: one of them is not slow, and you missed two that are. This is a good example of why you should *always* profile before optimizing.


1. Travis is complaining about some nose version problem, which I cannot reproduce on my machine. (We do not use nose at all, though testing matplotlib might force us to tepend on nose in the future.) However, it appears that you are using `assert_raises` from `numpy.testing`. Please do not use that: use `pytest.raises` in its context manager form, for exception tests.


1. At first glance `parse_to_function` looks like pure unmitigated evil! (If you ever catch yourself writing a parser while doing physics, **STOP!**: you are almost certainly doing it wrong, almost certainly re-inventing some wheel very, very badly indeed!) Functions are perfectly good first-class objects in Python, far better at representing themselves than strings are. If you want to specify a function, then use the function itself, not some string. Specifically, the first parameter of `test_fit` should be `fn` rather than `fname` and the arguments you pass to it should be *the actual functions themselves* rather than strings representing those functions. Unless there is some valid reason for the existence of `parse_to_function` (which I doubt very strongly) please remove it. Also, hunt down any other occurrances of passing strings representing functions anywhere else in any of your other code, and remove it right now!

   I haven't got the time to look at `parse_to_function` in any depth right now, but there is a > 99% chance that it is a horrible idea which needs to be killed *right now*. The surrounding area should probably be sterilized by nuking from space.

More later ...

jacg:
I couldn't resist coming back for a closer look at `parse_to_function` a bit more. It's fascinating, in the way that road-traffic accidents are fascinating.

It seems that `parse_to_function` belongs to a class of ideas that are somewhat related to [Greenspun's Tenth Rule](https://en.wikipedia.org/wiki/Greenspun%27s_tenth_rule), in that it is an ad-hoc, informally-specified, bug-ridden, slow implementation of half of Python's lambda. The fact that it *uses* Python's lambda in its implementation is particularly funny!

Let us have a minute's silence for the time that was wasted on writing this, and could have been better spent elsewhere. These poor minutes (hours?) died in vain.

For the love of all that is good, unless your job is writing compilers, if you ever find yourself writing a parser (in production code, rather than as a fun side-project), please **STOP!**. Do not do it! Whatever the reason you think you might need it, it's wrong. Please take this as a fundamental rule of scientific programming: You *never* need to write a parser, however much it might seem to you that you do!

In this case, it *seems* that the motivation for writing `parse_as_function` is to give you two possibilities:

1. Pass functions
1. Pass arbitrary expressions, as if they were functions.

In the first case, use *the function itself*. In the second case, use `lambda` itself. That't the whole point of `lambda`!

jacg:
Actually, I now have a suspicion that `parse_as_function` is trying to offer something beyond the two features I identify above. Of course, as there are no examples and **no tests** anywhere

1. I can't see what it is *supposed* to be doing,
1. it's almost certainly *not* doing what it is supposed to be doing.

jacg:
It has just dawned on me ... you are probably trying to reproduce a function fitting interface that you have seen in Root, right?

jacg:
Yes, indeed, looking at the [relevant section of the Root docs](https://root.cern.ch/root/htmldoc/guides/users-guide/FittingHistograms.html#the-tf1-function-class), it becomes clear that you are indeed inspired by Root. It takes many years of experience and dedication in the field of horrendous ideas before one can come up with something quite this abominable, so I knew it couldn't be your original work :-)

Nuke it from orbit!

Just pass Python functions (which might be a labmdas) directly to `scipy.optimize.curve_fit`. You don't need to write any wrapper code around it yourself! And you *certainly* don't need to make Python look more like Root!

gonzaponte:
1.- I don’t know where those merges come from, how do I safely remove them? Squashing? (sorry I forgot).
2.- I think those mark.slow may be a mistake from copy-pasting, I’ll fix it
3.- I forgot that nose was needed for that particular test. I’ll fix it.
4.- Well, I was expecting that… I know it is wrong, I know it is nasty, I know it is an awful practice. The point is how comfortable it is to play around with different combinations of functions when trying to fit some data and how annoying it is to be forced to write all those combinations in lambdas. For example, you can just try “gauss” or “gauss + expo” or “gauss + polynom” in a few seconds.

- In recent commits, I had already included the possibility of passing the function itself, but it I have not pushed it yet.
- The documentation and the tests are still incomplete, but I was planning to extend them.
- The fit function idea was probably copied from ROOT, although not intentionally. Too many years in hell can make you act like satan.

So… I think I may be too lazy (although you taught me to be so!). I will force the user to write lambdas and nuke the parser evil from this planet. Sorry if it made you sick, but it was hilarious to read your comments.

jacg:
> For example, you can just try “gauss” or “gauss + expo” or “gauss + polynom” in a few seconds.

Yes, that's the 'added value' that I thought I had recognized buried in the noise. Let's talk about more sensible ways of doing it, in person.

> I think I may be too lazy (although you taught me to be so!)

I advocate being *virtuously* lazy! You seem to have been quite the opposite: skimp on important stuff (like tests) while expending lots of effort on stuff that didn't need to be done at all!

We should work on this together. If you haven't received my email proposing this, then please check your spam filter.

jjgomezcadenas (commented):
The branch has conflicts. Any idea why?

gonzaponte:
Conflicting files
invisible_cities/core/params.py

Jacek made a [commit](https://github.com/nextic/IC/commit/48b8773fe43f3c94de7a876bf974e2b06e53863e) changing the way namedtuples are created. I must rebase onto nextic/master and resolve the conflict.

jjgomezcadenas (commented):
As a matter of fact, the conflicts are minimal, only in params. 

I see nothing wrong with the code, which has been simplified and clean up (except a few fossils functions still commented out, can you get rid of them?). However I wonder if it would be too much work to provide a notebook (put it in ICARO) with examples for a few canonical fit cases (gaussians and such) as well as profiles.

gonzaponte:
> I see nothing wrong with the code, which has been simplified and clean up (except a few fossils functions still commented out, can you get rid of them?). 

Sure, they will be removed before the final version.

> However I wonder if it would be too much work to provide a notebook (put it in ICARO) with examples for a few canonical fit cases (gaussians and such) as well as profiles.

Done. You can already [take a look](https://github.com/gonzaponte/ICARO/blob/fitfundemo/invisible_cities/demos/fit_functions_demo.ipynb) and comment.

Actually, this branch has been frozen since I visited the master last week. If I recall correctly, he was happy enough with the implementations, but not so with the tests, where we fought against hypothesis... Can you comment on the status of this PR, @jacg?

jjgomezcadenas:
@gonzaponte can you comment on why the test fail?

gonzaponte:
a `from __future__ import division` is missing. This affects python version 2.7 only. I will commit the change. Done.

invisible_cities/core/fit_functions_test.py
```diff
@@ -225,18 +225,20 @@ def test_profile1D_full_range_x(func, xdata, ydata, xrange):
     xp, yp, ye = func(xdata, ydata)
     assert np.all(fit.in_range(xp, *xrange))
 
+#
```
jacg: These should be marked with skip rather than commented out.

gonzaponte: I have just tried
```
@mark.skip(reason="There is something wrong with hypothesis")
```
and it crashes anyway because `FLOAT_ARRAY` is a function that returns an actual value, not a generator.

jacg:
Yes, skipping the hypothesis problem doesn't work, because skipping does not stop the test being *collected*, it stops the test from being *run*, and the hypothesis problem arises at collection time.

I'm failing to come up with a good workaround for now.

How about marking it with `skip` and commenting out just the parametrization? That way the skip will show up in our test runs, and we won't forget about it completely.

Also, be more explicit in the reason: "Hypothesis can't handle sequences longer than 2**10".

gonzaponte:
Forgot to mention that I have already correctly skipped the test.

Also, I have rebased onto nextic/master.

jacg:
What is more important to us in this PR? getting this into master quickly, or ensuring high code quality?

If I don't look at this this week, then I won't look at it until April.

gonzaponte:
Probably a linear combination. If the question is now or April, I would say merge now and improve later.
If you can devote some time to this PR this week and improve quality, good! I am available remotely.

jjgomezcadenas:
It would be nice to have this functionality available in the release for general use in the analysis coming up...

jacg:
Something urgent has come up, so I'm essentially out of the picture until April, as of right now.

I'm happy for this stuff to be squashed and merged as it is, if it is going to be useful it the meantime.

jjgomezcadenas (approved):
As stated in the discussion thread, the functionality here is needed and @jacg not too unhappy, thus I am approving the PR.

PR #147, opened 2017-03-01
New test in irene for data files. Check run & events info are copied

jmbenlloch:
Ensure run number & events number/timestamp are copied in output file.

### DIFF
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -129,8 +129,8 @@ class City:
     config_file_format = dedent(config_file_format)
 
     default_config = dict(PATH_IN  = '$ICDIR/database/test_data/',
-                          FILE_IN  = 'electrons_40keV_z250_RWF.h5',
-                          FILE_OUT = 'electrons_40keV_z250_CWF.h5',
+                          FILE_IN  = None,
+                          FILE_OUT = None,
                           COMPRESSION = 'ZLIB4')
 
     @classmethod
--- a/invisible_cities/cities/irene.py
+++ b/invisible_cities/cities/irene.py
@@ -164,7 +164,7 @@ class Irene(PmapCity):
             row     = self.evtInfot.row
             evtInfo = self.eventsInfo[evt]
 
-            row['evt_number'] = evtInfo[1]#.evt_number # TODO Write test !!!!!
+            row['evt_number'] = evtInfo[0]
             row['timestamp']  = evtInfo[1]
             row.append()
             self.evtInfot.flush()
--- a/invisible_cities/cities/irene_test.py
+++ b/invisible_cities/cities/irene_test.py
@@ -8,29 +8,102 @@ package: invisible cities. See release notes and licence
 last changed: 01-12-2017
 """
 import os
+import tables as tb
+import numpy  as np
 from pytest import mark, fixture
 
-from   invisible_cities.cities.irene import Irene, IRENE
+from invisible_cities.cities.irene import Irene, IRENE
 
 @fixture(scope='module')
-def conf_file_name(config_tmpdir):
-    conf_file_name = str(config_tmpdir.join('irene.conf'))
+def conf_file_name_mc(config_tmpdir):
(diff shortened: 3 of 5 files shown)


jacg:
I have pushed a commit with some minor changes, mostly cosmetic. These
should be squashed in before merging.

1. It's not clear to me why we need to make so much noise about
   FILE\_IN and FILE\_OUT. If FILE\_IN is written in the config file (on
   line 40), then why do we have to reconstruct it from the variable
   in which we stored its name and a default Irene config on line 46?
   Something smells very fishy around here!

1. This test seems to be the second slowest test that we have at the
   moment. Could it be made a bit lighter without too much effort?

jmbenlloch:
1. The input file is not the default one, therefore I need to change FILE_IN before writing config file. Later, after Irene has run, I need to open both input and output files. To do this I need the full path, for the input this will mean `default_path + filename_in` and for output this is `tmpdir + filename_out`. I don't know how to that in other way, we could get it from the config file we have just written, but that would mean parsing the file again.

2. Now we are running over 3 events, maybe we could process only one or two. Probably if it works for one event it should work for any event, so we could remove numpy assert and check only 3 numbers (run number, event number and timestamp).

jacg:
> The input file is not the default one, therefore I need to change FILE_IN before writing config file.

Why can't it be passed as a command line option (`-i`) to Irene?

Indeed, you are writing a file called `irene.conf` into the tests' tempdir, both in the `conf_file_name` fixture *and* in this specific test. So you are overwriting the config file created by the fixture! This is wrong.

Actually, I now notice that the `conf_file_name` fixture does not have an explicit scope, which probably means that it has function scope, and that we are writing the conf file repeatedly as soon as we have more than one test which uses it, which probably means it will fail, sooner or later, when running the tests in parallel, and may cause problems in other situations. I have pushed a quick fix to that problem, please rebase on top of it.

I'm planing to refactor the `conf_file_name` fixtures which appear in all the city tests (and the command line tests that use them), because there is still too much copy'n'paste going on. So these things *will* change, but I don't have time for it right now.

In summary.

1. Rebase on the explicitly scoped fixtures commit.

1. Use the `conf_file_name` fixture in the test, and the conf file that was written by the fixture, setting non-default input file with `-i` on the command line.

Then I would like @jjgomezcadenas to have a look at the test.

jjgomezcadenas (approved):
@jacg raised the following issues:

---Why can't it be passed as a command line option (-i) to Irene?

This is done now in all tests (in fact, tests use -I, -o, -n, -r, which specify in a more general way than a default config file what the test wants to do

Indeed, you are writing a file called irene.conf into the tests' tempdir, both in the conf_file_name fixture and in this specific test. So you are overwriting the config file created by the fixture! This is wrong.

This has been fixes. Two fixtures write now a default config file, one for data and one for mc. This may not be a entirely satisfactory solution, but is progress. 

Actually, I now notice that the conf_file_name fixture does not have an explicit scope, which probably means that it has function scope, and that we are writing the conf file repeatedly as soon as we have more than one test which uses it, which probably means it will fail, sooner or later, when running the tests in parallel, and may cause problems in other situations. I have pushed a quick fix to that problem, please rebase on top of it.

Done

I'm planing to refactor the conf_file_name fixtures which appear in all the city tests (and the command line tests that use them), because there is still too much copy'n'paste going on. So these things will change, but I don't have time for it right now.

In summary.

Rebase on the explicitly scoped fixtures commit.

done.

Use the conf_file_name fixture in the test, and the conf file that was written by the fixture, setting non-default input file with -i on the command line.

Done and improved (I believe). 

All put together, I think the PR can be approved.

jacg:
@jjgomezcadenas When quoting other people in your comments, it would be helpful if you used the MD quoting markup syntax to indicate that you are quoting.

Over half of your last comment is quotes of me. The only reason I managed to follow what was going on was that I recognized things that I had written myself and deduced that you must have been quoting me. I pity the third party trying to understand what on earth is going on in your comment.

In short, please prepend `> ` to any text that you are quoting, and everybody will find it easier to understand what is being said and by whom.

Issue #151, opened 2017-03-04
bug in irene (store_pmaps)


jjgomezcadenas:
A bug has been found in irene (store_pmaps), which caused that S2 were not written in the PMAPS file. The bug was introduced when re-factoring the run method, creating the store_pmaps functions.   A fix (and a test) will be submitted shortly

PR #152, opened 2017-03-04
Bug fix and test for issue-151.

jjgomezcadenas:
The bug reported in issue-151 (S2 and S2Si were not
being correctly written to file) is fixed. A test
to ensure that PMAPS tables are correctly saved is added.
The test checks that the PMAPS tables created in memory
(s1t, s2t, s2sit) are correctly written in the PMAPS file.

In addition, a test in pmaps_functions_test is temporarily commented
out. The test commented was failing in python2.7 and
python3.5 (ok in 3.6). The tests in pmaps_functions_test are
improved and expanded in a different branch (issue-149), that will be
rebased on top of the current work.

### DIFF
--- a/invisible_cities/cities/irene.py
+++ b/invisible_cities/cities/irene.py
@@ -116,8 +116,8 @@ class Irene(PmapCity):
                 tbl.filters(self.compression))
 
 
-    def _store_s12(self, S12, event, evt):
-        row = self.s1t.row
+    def _store_s12(self, S12, st, event, evt):
+        row = st.row
         for i in S12:
             time = S12[i][0]
             ene  = S12[i][1]
@@ -133,10 +133,10 @@ class Irene(PmapCity):
                 row["time"] = time[j]
                 row["ene"]  =  ene[j]
                 row.append()
-        self.s1t.flush()
+        #st.flush()
 
-    def _store_s2si(self, S2Si, event, evt):
-        row = self.s2sit.row
+    def _store_s2si(self, S2Si, st, event, evt):
+        row = st.row
         for i in S2Si:
             sipml = S2Si[i]
             for sipm in sipml:
@@ -155,7 +155,7 @@ class Irene(PmapCity):
                         row["nsample"] = j
                         row["ene"]     = E
                         row.append()
-        self.s2t.flush()
+        #st.flush()
 
     def _store_pmaps(self, event, evt, S1, S2, S2Si):
         """Store PMAPS."""
@@ -167,11 +167,11 @@ class Irene(PmapCity):
             row['evt_number'] = evtInfo[0]
             row['timestamp']  = evtInfo[1]
             row.append()
-            self.evtInfot.flush()
+            #self.evtInfot.flush()
 
-        self._store_s12 (S1,   event, evt)
-        self._store_s12 (S2,   event, evt)
-        self._store_s2si(S2Si, event, evt)
+        self._store_s12 (S1, self.s1t,  event, evt)
+        self._store_s12 (S2, self.s2t,  event, evt)
+        self._store_s2si(S2Si, self.s2sit, event, evt)
 
     def run(self, nmax, print_empty=True):
         """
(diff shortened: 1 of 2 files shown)


jacg:
> a test in pmaps_functions_test is temporarily commented out. The test commented was failing in python2.7 and python3.5 (ok in 3.6).

1. In such cases the test should not be commented out but marked with `@xfail`, parametrizing the `xfail` to specify the platforms on which it fails (and, preferably, stating the reason). [Here](http://doc.pytest.org/en/latest/skipping.html) are the relevant docs.
2. The test passes on all 3 Python versions on my machine.

I would like to establish the following practice for PRs which provide a test and a solution:

+ At least two commits: one adding the test (without touching the implementation), thus demonstrating the problem; another providing the fix (without touching the tests), thus demonstrating the solution.
+ We will probably want to squash these together before merging into master, but for review purposes I think that such a presentation would be a big help.

jacg:
Following my advice above, I tried to present this PR in the desired format. On my machine, `test_pmaps_store_issue_151` passes (on all 3 Python versions), *before* the proposed solution is implemented! The same [happens on Travis](https://travis-ci.org/jacg/IC/builds/208120284), on OS X and Linux!

The possible conclusions are:

1. There is no problem that needs fixing
1. There is a problem, but the proposed test does not demonstrate it.
1. I have completely lost the plot.

If I were a betting man, I'd put my money on number 2.

For completeness, you can find my re-organized version of this PR in my fork on the `issue151` branch. The branch contains the test in the first commit, and the solution in the second commit. But do note that the Travis build linked above as done *before* the solution was added.

I still have no idea why `test_read_max_events` needed commenting.

jjgomezcadenas:
Quick comments:

1) I had indeed two commits exactly as you suggest, then I decided to over-smart myself and squashed them. Sorry, will do next time.

2) The bug that is fixed is a bug in the written file. What was happening is that *wrong* S1 pmaps were being written in file, but that does not result in any type of failure. On the other hand, once the solution is added, if you resurrect the bug, then the test will fail (because I check now that what is written is what is calculated, before I calculated one thing and wrote a different thing). I can submit an example if you want.

3) I will take a look to test_read_max. 

I will work in this after lunch.

jjgomezcadenas:
The PR should be now in the format suggested by @jacg. A first commit demonstrates the problem (thus Travis failure) by adding *first* the test, then a second commit fixes the bug (happy Travis). All test pass correctly in Travis.

jacg:
Does this test really require so many Irene parameters to be set by hand? Can we not pick up most of these up from a config file?

It would be good to see only those parameters specified explicitly, which are specific to *this* test.

jacg:
As it is relevant to this work, I'll bring it up here:

In work that I as doing with @neuslopez this afternoon, we came across the need to write some pmaps files, and found the ability buried deep inside Irene. We were wondering whether this capability could not be factored out and made available independently of Irene.

jjgomezcadenas:
The test sets Irene parameters for a specific data run, while the default Irene parameters are for MC data. The parameters could be placed somewhere else if we identify another test that may use them (for the time being we haven't). In short, we think that the parameters being set are relevant for the test (they define exactly the conditions in which the S1, S2 and S2Si are found).

jjgomezcadenas:
In response to the last question 

> we came across the need to write some pmaps files, and found the ability buried deep inside Irene. We were wondering whether this capability could not be factored out and made available independently of Irene.

It turns out that the role or Irene is precisely to write PMAPS. Irene is a concrete city, derived from PMAPS city whose role in this life is to compute and write PMAPS. If one needs to produce PMAPS, it is supposed to call Irene. Factorising out Irene functionality appears to us tantamount to revise the scheme of cities.

jacg:
There is a world of difference between 'compute and write pmaps' and 'write pmaps'. Computing pmaps and writing pmaps are orthogonal activities. Tightly coupling them together in a single unit is not a very good idea.

We wanted to write a roundtrip test, ensuring that reading and writing data (or doing it the other way around) gets us back to where we started. As it stands we have two choices:

1. Rewrite the pmap-writing functionality ourselves
1. Instantiate Irene and pretend that we are calculating some pmaps, with the concomitant overhead and noise of messing with config files etc.

The first is absurd, the second is very bad.

> tantamount to revise the scheme of cities

Don't tempt me.

(Or rather, do tempt me, and we'll clean this up sooner rather than later :-)

jacg:
> we think that the parameters being set are relevant for the test

If they are specific to this test, then explicitly specifying them in the test is the right thing to do.

jjgomezcadenas:
Can we please decouple this discussion from #151 ? In this PR we have identified a clear bug, have fixed it and have proceeded by the book through the steps of demonstration and fixing. This is totally independent of whether Irene should factorized the writing of PMAPS or not. Please, let's complete this PR, if at all possible and open a new issue for the PMAPS refactoring.

jjgomezcadenas:
We have open #158 to carry on with discussion/work related to PMAPS writer.

jacg:
It looks like you pair programmed it. I can't see any obvious problems with it. You have demonstrated that the test does indeed show a problem. If the pair is happy with what is there, then then go ahead and approve it, and merge. Just make sure you squash beforehand: we don't want to have that failing test in `nextic/master`. Also, put something like `fixes #151` in the commit message or the PR message, and it will automatically close the issue when it gets merged.

jacg (approved):
I am trusting that this pair programmed work has been adequately reviewed by the pair. I see no obvious problems myself, and I am sure that the pair has analysed this far more than I possibly could from where I am.

PR #153, opened 2017-03-06
Make manage.sh clean take care of .pyc files

jacg:
Before accepting this PR, I would like it to be improved. The current implementation makes too much noise: it mentions every single `.pyc` file in the `__pycache__` directories.

@jmbenlloch please check that I'm not doing something dangerously stupid.

Feel free to add an improvement which is less noisy. I'll get around to doing it myself eventually, but the need for this functionality arose in the middle of doing something else, so I've done the bare minimum, for now.

### DIFF
--- a/manage.sh
+++ b/manage.sh
@@ -165,20 +165,21 @@ function compile_and_test {
 
 function clean {
     echo "Cleaning IC generated files:"
-    C_FILES=`find . -name '*.c'`
-    SOFILES=`find . -name '*.so'`
-    REMOVE="$C_FILES $SOFILES"
-    if [ ! -z "${REMOVE// }" ]
-    then
-        for FILE in $REMOVE
-        do
-            COMMAND="rm $FILE"
-            echo $COMMAND
-            $COMMAND
-        done
-    else
-        echo Nothing found to clean
-    fi
+    FILETYPES='*.c *.so *.pyc __pycache__'
+    for TYPE in $FILETYPES
+    do
+		echo Cleaning $TYPE files
+        REMOVE=`find . -name $TYPE`
+        if [ ! -z "${REMOVE// }" ]
+        then
+            for FILE in $REMOVE
+            do
+               rm -rf $FILE
+            done
+        else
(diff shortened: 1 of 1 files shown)


jmbenlloch:
I have change the code a little bit so it does not print every file. Now a message is printed for each filetype:

```
$ bash manage.sh clean

Cleaning IC generated files:
Cleaning *.c files
Cleaning *.so files
Cleaning *.pyc files
Cleaning __pycache__ files
```

If there are no such files it will say so:

```
$ bash manage.sh clean

Cleaning IC generated files:
Cleaning *.c files
Nothing found to clean in *.c
Cleaning *.so files
Nothing found to clean in *.so
Cleaning *.pyc files
Nothing found to clean in *.pyc
Cleaning __pycache__ files
Nothing found to clean in __pycache__
```

If ok, I will squash the two commits so we can rebase onto master.

jjgomezcadenas (approved):
This appears to address the requirement made by @jacg

jacg:
> This appears to address the requirement made by @jacg

Actually, it has almost exactly the opposite effect from what I was looking for.

Perhaps it would have been better to let me approve it, given that I was the one who had the requirements.

Let me remind you that there is no karma leakage caused by open PRs. There *is* karma loss as a consequence of merging work that is not complete, into `nextic/master`.

The whole point is that I want to be informed which files are being removed. By adding pyc files to the targets, this signal was completely drowned out by the noise of every single file in `__pycache__` being mentioned. The thing with `__pycache__` is that

1. It's a cache, so it contains nothing of interest
2. It contains a *lot* of crap.

What needed doing was silencing `__pycache__` so that the signal became visible once more. What happened was that the signal that I wanted to recover was completely silenced ... and then you approved the PR and it got merged into `nextic/master`.

Sigh.

Less haste, more speed.

jjgomezcadenas:
Sorry, it was my fault and not really a problem of haste, I just missunderstood you and thought the issue was so trivial it could be approved. Apologies

PR #154, opened 2017-03-06
Remove old version of Diomira

jacg:
This seems to be dead code, which is only polluting our source
directories. If it contains anything of historical interest, it can
always be found in the history of the repo.


jjgomezcadenas (approved):
Yes, this was a fossil which deserved passing to a better life.

Issue #150, opened 2017-03-04
Splitting *core* in two directories (*core* and *reco*)


jjgomezcadenas:
The directory core lumps together rather different animals. I propose to split of core in two separate directories:
a) core which would keep what I think are really core modules: configure, core_functions, exceptions, log_config, random_sampling and system_of_units, mpl_f (plus testing modules and cython modules)
b) reco which would, for now, get all the rest: nh5, params, peak_functions, pmaps_functions, tbl_f, wfm_f, mctrk_f (plus testing and cython)
Opinons?

jacg:
Seems reasonable.

I'll try to knock together a branch that does this, quickly.

jacg:
How about moving `mpl_functions.py` to something like `plot`?

jjgomezcadenas:
Yes i have thought of something like that. Perhaps util so that we can stick there some other utilities?

Enviado desde mi iPhone

jacg:
`mctrk` seems not to be used anywere.

jjgomezcadenas:
Let me check this before removing

Enviado desde mi iPhone

jjgomezcadenas:
mctrk_functions.py contains functionality to draw MC tracks. Although no one is using this right now, I expect use very soon, with the new city being written by Alex. In general these are important plotting functions. They could stay in core. Alternatively, if we create a directory *plot* or *util* this module could be moved there.

PR #155, opened 2017-03-06
Core-reco split

jacg:
Implements the suggestion made in #150.

The first commit creates a new `reco` package. Subsequent commits, move one module from `core` to `reco`. These latter commits should commute, so it should be easy to remove/add modules from the set that we want moved.


jacg:
Obviously, we will want to squash once we're happy with the set.

jjgomezcadenas (approved):
With this PR the IC directory becomes better organised.

PR #159, opened 2017-03-06
Replace brackets and quotes with a dot

jacg:
Work in progress: there may be more places where this should/could be done.

The interface `foo.bar` is much cleaner than `foo['bar']`. The former is the natural interface presented by `argparse` and yet `config.py` insists on turning this convenient interface into the uglier one. These changes reverse this choice.

A similar change is made in `read_FEE_table`.


jjgomezcadenas (commented):
Tests are failing... any idea why?

jacg:
Needed rebasing, because of the core-reco split. Seems to be working now..

jjgomezcadenas (approved):

PR #160, opened 2017-03-06
Fix dependences in irene_test.py after reco split

jjgomezcadenas:


jacg (approved):
Interesting. I remember making these very changes in this very file, and testing the whole lot thoroughly. So it's a bit of a mystery as to how they got lost along the way.

jjgomezcadenas:
I believe I know what happened. The file in question was augmented with the dependencies in the PR which added the latest test. These two PR were in parallel so each one passed their own tests (the irene_test in your reco_split branch did not have reco dependencies and my irene_test in my branch run over unsplitted directories). After rebasing everybody, it was necessary to fix the dependencies. All normal operations, I think.

jacg:
Right, and because I ran out of patience with Travis (I had a zillion branches to merge and Travis was running *none* of them for over two hours), I merged before getting an answer from Travis, so Travis didn't catch it. Urgh!

Issue #149, opened 2017-03-03
Non-existing event introduced when converting pmaps DataFrames to dictionaries


gonzaponte:
The function cdf_to_dict in pmaps_functions_c.pyx always returns event #0 even when it is not present in the input DataFrame.

gonzaponte:
I have in my fork (gonzaponte/issue149) a demonstration of the failure. I had to modify the test file so it does not include event #0. Although I have written a new test to demonstrate the bug, all tests fail because of the bug. However, since the new test is more exhaustive, I will rewrite the previous ones to perform as the new one.

jacg:
>  I had to modify the test file so it does not include event #0.

Would it be possible/better to do this by adding another test file? How big a new data file would you need to be able to write this test?

Please do a whitespace cleanup. `pmaps_functions_c.pyx`, for example, contains a number of spurious whitespace additions.

You have removed the initializations for `evt` and `pk` in  `cdf_to_dict`. Is there any good reason for this? A quick search into uninitialized variables in Cython does not inspire confidence. Please initialize these variables, unless there's some good reason not to (which I find hard to imagine).

Please squash into two commits: one adding the tests in their final form, leaving the implementation alone; one implementing the solution, leaving the tests alone.

@jjgomezcadenas Please have a quick look at this, and comment on the how this fits in the grand scheme of things. Also please comment on the desirability of adding a new test data file vs. modifying the existing one.

gonzaponte:
> Would it be possible/better to do this by adding another test file? How big a new data file would you need to be able to write this test?

It is possible, but maybe redundant. It will not be very large (few tens or hundreds of kb).

> Please do a whitespace cleanup. pmaps_functions_c.pyx, for example, contains a number of spurious whitespace additions.

Ok.

> You have removed the initializations for evt and pk in cdf_to_dict. Is there any good reason for this? A quick search into uninitialized variables in Cython does not inspire confidence. Please initialize these variables, unless there's some good reason not to (which I find hard to imagine).

No good reason, it is just not necessary as they are always assigned when needed. Maybe initializing them to 0 can be misleading, but it is not a big deal. I will set them to 0.

> Please squash into two commits: one adding the tests in their final form, leaving the implementation alone; one implementing the solution, leaving the tests alone.

Ok.

jacg:
> Maybe initializing them to 0 can be misleading, but it is not a big deal. I will set them to 0.

[...]

PR #162, opened 2017-03-07
Issue149

gonzaponte:
Fix issue #149. The function cdf_to_dict in pmaps_functions_c.pyx has been restructured so it does not include unexisting events.


gonzaponte:
I still have the squash pending.

jjgomezcadenas (commented):
Would it be possible to  benchmark your current function and compare it wit the old one?

gonzaponte:
What do you mean?

jacg:
I have included two commits. The first one uses `xxx` as the implementation of `df_to_pmaps_dict`. The second one replaces that implementation with `yyy`. I would like to keep them both, as this will be the only record remaining of `xxx` once I have cleaned up my local branches.

The namedtuple used to store the pair of time and energy arrays is called `Peak`. Any better suggestions?

jacg:
We have just noticed that the `df_to_pmaps_dict` tests do not exercise the namedtuple within, and that the namedtuple is not being used at all in the first commit. Fixes coming up ...

gonzaponte:
> The namedtuple used to store the pair of time and energy arrays is called Peak. Any better suggestions?

Waveform?

jjgomezcadenas:
Peak is correct. We want to keep the name Waveform for the whole thing. A Waveform contain a number of Peaks…

jacg:
We just have some doubts about `test_df_to_pmaps_dict_limit_events`: It asks for a maximum of 10 events, and then asserts that there are 4. I understand how this fits with the data presented in the test, but it smells fishy. Is this really what we mean/want?

gonzaponte:
After taking directly, we've concluded that this ***is*** what we want

jacg:
We have removed the comment which expressed our doubts, from the test.

As far as we are concerned, this is now ready.

jjgomezcadenas:
PR coming then?

jacg:
This *is* a PR.

(If you did all this via GitHub, rather than your mail, all these things would be obvious ... which is a topic for another discussion that we need to have sometime, but not here)

jjgomezcadenas:
Sorry! Just got sidetracked. Review coming

jjgomezcadenas (commented):
Quite impressive. I have only a suggestion, for your consideration. I believe that the "normal user" (whoever this guy is) should work always with the "dictionary" PMAPS, and never with the "DF" PMAPS. This is to avoid proliferation of PMAPS analysis code based in two different data structures. In short, I see "DF" as internal representation of the PMAPS (a mapping of the potables in file) and "dictionaries" as the "analysis" representation. 

If this is agreed, I believe that read_pmaps() should return directly the dictionaries, thus, load the DF functions, then call the df_to_pmaps_dict(). 

This, BTW, may bring an interesting argument in favour of implementing a class PMapsReader(), as a convenient way to hide functions that we do not want to expose publicly. In this case I would consider df_to_pmaps private to the class while read_pmaps would be public. 

In any case, no need to address the PMapsReader now, just consider that read_pmaps() could return the dictionaries straight away.  Let me know what you think (implement change if you agree) and then I will approve the PR

jacg:
> If this is agreed, I believe that read_pmaps() should return directly the dictionaries, thus, load the DF functions, then call the df_to_pmaps_dict().

Yes, we were just making similar observations here.

Shall we do it now, or as a separate PR?

gonzaponte:
> If this is agreed, I believe that read_pmaps() should return directly the dictionaries, thus, load the DF functions, then call the df_to_pmaps_dict().

The problem is that the SiPMs need a different function (and read_pmaps returns them as well). I am working on the implementation of "s2sidf_to_dict" on a different branch.

> Shall we do it now, or as a separate PR?

Because of what I've just said, it should be implemented in a separate PR.

jacg:
Then all that remains in this PR is approval and merging.

jacg:
No, one more thing. Elsewhere @gonzaponte says

> >      I suggest df_to_pmaps_dict as the new name

> Since it is already in pmaps_functions.py, I would avoid naming it pmaps (simply df_to_dict?).

I like this idea. Shall we make the change?

jjgomezcadenas:
I like the current name is expressive and clear

Enviado desde mi iPhone

jacg:
The version with the name changed is available in *my* fork (branch name: issue149). I have left it out of this PR so far.

jjgomezcadenas (approved):

PR #163, opened 2017-03-09
Add function to create dictionaries from S2Si dataframes

gonzaponte:
I have made a new cython function `df_to_namedtuples` to convert S2Si dataframes to python dictionaries. It has been implemented in the same fashion as `df_to_pmaps_dict`. The output contains the following structure:
- Dictionary with event numbers as keys and dictionaries as values.
- Inner dictionaries have peak numbers as keys and lists as values.
- The lists contain instances of the SiPM namedtuple defined in `reco/params.py`.
- The SiPM namedtuple contains three attributes:
  1. ID (int)
  2. sample (np.array)
  3. ene (np.array)

I have extended the test suite so this new function is tested against the same kind of problems as `df_to_pmaps_dict`.

Check [my fork](https://github.com/gonzaponte/IC/tree/S2Sidicts) for the implementation details.

To be discussed:
1. Do we like this structure? Another possibility is to have the dicts (evt) holding dicts (peak) holding dicts (sipms) with SiPMs numbers as keys and namedtuples (with just "sample" and "E" as attributes) as values. I rather use the latter, but IIRC, @jjgomezcadenas prefers the current implementation. Please comment on that.

2. Name of the function is probably not very good. Suggestions? What about df_to_s2si_dict?

### DIFF
--- a/invisible_cities/reco/pmaps_functions.py
+++ b/invisible_cities/reco/pmaps_functions.py
@@ -4,17 +4,14 @@ JJGC December 2016
 """
 from __future__ import print_function, division, absolute_import
 
-import math
 import numpy as np
 import pandas as pd
 import tables as tb
 import matplotlib.pyplot as plt
-import invisible_cities.core.system_of_units as units
-from   invisible_cities.reco.pmaps_functions_c import df_to_pmaps_dict
+from   invisible_cities.reco.pmaps_functions_c import df_to_pmaps_dict, df_to_s2si_dict
 import invisible_cities.core.core_functions as cf
 from   invisible_cities.database import load_db
 from   invisible_cities.core.mpl_functions import circles
-
 
 
 def read_pmaps(PMP_file):
@@ -27,6 +24,7 @@ def read_pmaps(PMP_file):
         return (pd.DataFrame.from_records(s1t  .read()),
                 pd.DataFrame.from_records(s2t  .read()),
                 pd.DataFrame.from_records(s2sit.read()))
+
 
 def s12df_select_event(S12df, event):
     """Return a copy of the s12df for event."""
--- a/invisible_cities/reco/pmaps_functions_c.pxd
+++ b/invisible_cities/reco/pmaps_functions_c.pxd
@@ -6,3 +6,4 @@ cimport numpy as np
 import numpy as np
 
 cpdef df_to_pmaps_dict(df, max_events=?)
+cpdef dict df_to_s2si_dict(df, max_events=?)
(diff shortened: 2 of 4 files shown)


jjgomezcadenas:
My five cents:

1. I am not sure that we need the sample array, since the S2Si samples (times) are already defined by the S2s. 

2. I like your proposal, for the structure

{event:{peak{sipm:[named_tuple]}}} 

3. I like df_to_s2si_dict

I can review the PR tomorrow unless @jacg plans to do it (as I suspect).

gonzaponte:
> 1. I am not sure that we need the sample array, since the S2Si samples (times) are already defined by the S2s.

We either keep the sample array or we fill empty slices with zeros.

> 2. I like your proposal, for the structure
>
>{event:{peak{sipm:[named_tuple]}}}
>
> 3. I like df_to_s2si_dict

Ok. I will just wait to see if @jacg wants to comment on that.

jacg:
Yes, I do want to have a *good* look at this, and related issues, because I have discovered a lot of things in the codebase relating to this sort of stuff. The trouble is

+ I don't know where to begin, it's all intertwined.
+ I don't know when I'll have the time. Probably not until tomorrow afternoon. Maybe tonight, but unlikely.

How urgent is this?

gonzaponte:
> How urgent is this?

I need it (and probably Neus) to keep going with the analysis, but not sure if it is mandatory for it to be in the main repository right now.

jjgomezcadenas:
It is done, array of energy no longer sparse.

Enviado desde mi iPhone

jjgomezcadenas:
This is becoming urgent, since the first data from Canfranc will arrive... today! The data will be very preliminary for one week or so, but then good data will come. I am hoping that we can manage to stabilise the code to the level of PMAPS including read/write.

jacg:
Nice work.

I approve in principle. I won't approve formally, because I haven't really been following the various discussions that have been taking place elsewhere about how we want the various formats to change, so I am currently not in a position to judge whether the present state is in line with wherever those discussions have taken us.

Apart from that minor reservation (entirely due to my own ignorance) I am perfectly happy to approve this code. Anyone who has been following those discussions, and sees that this code is in line with them, has my full blessing to approve this formally.

jjgomezcadenas (approved):
The functions return the correct data format:

For S1 and S2:
S1 = {event:{peak:Peak(t,E)}} Where Peak('t' 'E') is a named tuple and [t], [E] are numpy arrays.

For S2Si

S2Si = {event:{peak:{nsipm: [E]}}} where E is an array of energies with the same length than the corresponding arrays in S2. 

Thus, I think the PR can be approved.

jjgomezcadenas:
Is there some squashing to do, or can we rebase and merge straight away?

gonzaponte:
Needs squash! But I can do it in a few minutes.

jjgomezcadenas:
Great. Go for it

gonzaponte:
Done!

Issue #146, opened 2017-03-01
Wrong type in last evt in s12df_to_s12l


gonzaponte:
The function is supposed to return a dictionary with dictionaries holding two np.ndarrays. When looping over the full input DataFrame the last event is not cast to np.ndarray and remains as a list.

jacg:
Could you provide a test that demonstrates the problem?

gonzaponte:
Done in my fork gonzaponte/issue146.

jacg:
In Python we usually do not care about the concrete types of objects. Can you reframe this test in terms of some *capabiltiy* of the objects? In other words, as it stands, this test looks too much like it is inspired by a failure to understand Duck Typing. And we wouldn't want future readers of our code to get the impression that we don't understand Duck Typing.

gonzaponte:
The problem is that the difference of capabilities of a duck and a person are too many. I want to have full functionality of np arrays, so the easiest way is to check it behaves as a np array is to ask for its type, isn’t it?

jacg:
But you don't want the full functionality of numpy arrays; you only need that functionality which you are actually using.

Why don't you start with *one* bit of numpy functionality that is missing from whatever is being returned right now, that is stopping you from achieving your broader goals.

Besides, I don't know why we are talking about numpy arrays, when the test fails because it demands a list in a situation where it gets a generator. This in itself is a good illustration why you should write tests in terms of capabilities rather than types! And these capabilities should be ones you *actually* need, not ones you think you need, because they happen to be provided by the type you happen to be thinking of.

Hint: you will find it almost impossible to persuade me to accept code containing a type check demanding a list (or any other contrete type or hierarchy). Sequence-like behaviour? OK; iterator-like behaviour? OK; broadcasting? OK; the concrete list type? NO WAY!

Checks for concrete types have no place in Python programming (except when talking to low-level components). The sooner you accept this, the happier your Python experience (and that of other users of your code) will be.

gonzaponte:
It is true the test fails there, although that was just a secondary check (and that’s why we’re talking about np arrays: the primary check that is currently implemented within the same test).

When JJ first wrote that function it was supossed to return {evt_idx: {pk_idx: [np.array, np.array] } }. For the last event, however, it returns  {evt_idx: {pk_idx: (list, list) } }. So

1.- For consistency I find annoying that the last event has a tuple instead of a list. I don’t care much about this because, as you pointed out, the functionality is the same. Do you think consistency is not that important to change this?

[...]

PR #164, opened 2017-03-09
Issue146

gonzaponte:
Extend tests to check the behaviour of the data in the namedtuples.

### DIFF
--- a/invisible_cities/reco/pmaps_functions_test.py
+++ b/invisible_cities/reco/pmaps_functions_test.py
@@ -90,6 +90,9 @@ def test_df_to_pmaps_dict_structure(s12_dataframe_converted):
         for peak_no, peak_data in subdict.items():
             assert hasattr(peak_data, 't')
             assert hasattr(peak_data, 'E')
+            for element in peak_data:
+                assert type(element) is np.ndarray
+
 
 def test_df_to_pmaps_dict_events_contain_peaks(s12_dataframe_converted):
     converted, _ = s12_dataframe_converted


jjgomezcadenas (commented):
I can't see anything wrong with this PR. If @jacg has no issues, I recommend approval (but I let pass the ball to @jacg so that he has a chance to look at the stuff and approve)

invisible_cities/reco/pmaps_functions_test.py
```diff
@@ -110,20 +118,3 @@ def test_df_to_pmaps_dict_events_data_correct(s12_dataframe_converted):
...
     assert (converted_time == original.time).all()
     assert (converted_ene  == original.ene ).all()
-
-
-def test_df_to_pmaps_dict_duck_types(KrMC_pmaps):
-    (s1s, s2s, _), _ = KrMC_pmaps
-    s1dict = df_to_pmaps_dict(s1s)
-    s2dict = df_to_pmaps_dict(s2s)
-
-    for d in [s1dict, s2dict]:
```
jacg: Minor point: this should be a tuple, not a list.

gonzaponte: That's from the first commit. The second one removes that test (`test_df_to_pmaps_dict_duck_types`) because the code fits better inside `test_df_to_pmaps_dict_structure`

jacg (commented):
Back in the mists of time, when we were discussing this point, I said that you should write a test which checks the duck type by *actually using the object* in some way that a real user might. I can't imagine why a real user would ever want to use `hasattr` so many times without actually *using* any of the attributes.

Essentially, this is just a long-winded way of writing `assert type(element, np.array)` without using `type` or `isinstance`. So it's kind of missing the point I was trying to make. It tries to follow the letter of my advice, while completely ignoring the spirit.

TBH, I would *prefer* `assert type(element) is np.array`, as it states much more clearly and robustly what it is that you mean.

Of course, I would prefer it even more if there were a genuine duck type test. But I appreciate that it's not obvious how to do that, and checking that this is a numpy array is kinda useful.

+ Replace the `hasattr`s with a type check
+ Replace the tuple with a list, as suggested in the line comment.

Then I'll merge it ... with the caveat that I will remove this check if it ever gets in the way of any mocking or faking tests that anyone might ever want to write in the future, or if any similar situations arise.

invisible_cities/reco/pmaps_functions_test.py
jacg: Ah, yes ... I thought I was in the second commit at that point. Sorry.

gonzaponte:
> So it's kind of missing the point I was trying to make. It tries to follow the letter of my advice, while completely ignoring the spirit.

That was not my intention. I asked in #146 whether it was ok to do this (use `hasattr` as duck type checking), but didn't get an answer which I (incorrectly) took as an affirmative answer. Over that, I probably misinterpreted the use you gave to this function in `test_df_to_pmaps_dict_structure`.

> Of course, I would prefer it even more if there were a genuine duck type test. But I appreciate that it's not obvious how to do that, and checking that this is a numpy array is kinda useful.

I am willing to do proper duck type checking. Do you prefer to forget about it for the time being or to devote some time now? If the answer is the latter one, should I incorporate such duck type checking into `test_df_to_pmaps_dict_structure` or in a separate function?

jacg:
I didn't mean to imply that you are deliberatly avoiding my instructions. What I mean is that I have failed to make my point adequately.

I think that you got no answer, because I had to unplug from Next at that time. As I will have to do once again, very soon.

If you are not desperate to get this in right now, I would prefer to think about it properly later. The problem, is that I have plenty of non-Next things to do next week, and I will have essentially no time at all for Next in the two weeks following that. Not to mention plenty of Next stuff which seems more urgent.

So It would be OK if you just stuff in the type check right now, and we get on with our lives.

> should I incorporate such duck type checking into `test_df_to_pmaps_dict_structure` or in a separate function?

`test_df_to_pmaps_dict_structure` is the perfect place for a concrete type check. A duck type check would almost certainly complicate and obfuscate that test too much, so I would definitely put it elsewhere.

gonzaponte:
Ok, done. If there is no other issue, I will squash this into a single commit.

jacg (approved):

Issue #145, opened 2017-03-01
Event ID numbering in Monte Carlo


neuslopez:
The need of having unique event ID numbers for a MC production bunch.
Now we have event IDs starting at '0' for every file in the same production, therefore we cannot merge files.

jacg:
Could you write an automated test that somehow communicates this need and demonstrates the shortcoming?

jjgomezcadenas:
The point raised by @neuslopez is related with data organisation and is not obvious to me that a test is relevant here. In short, the situation is as follows:

1. Data events have a unique "event number" generated by the DAQ. This event number is extremely useful when you need to locate a particular data event. It is, so to speak, an analysis feature, a commodity.

2. MC events in IC, so far do not have a unique event number, since this number is not, as far as I know, generated by the input Monte Carlo. What we do is to give MC events a number associated with the position of the event in the set of input files that are processed by IRENE. But as a consequence, next time we process a different bunch of files, the numbers repeat. This is not good.

The best solution for this would be that event numbers are propagated to us from NEXUS. That should be possible, no? Please @paolafer, @neuslopez, @jmbenlloch, say your opinion.

Alternatively, we can call a random number as a seed each time we process a bunch of files and create a "unique event number"  (although then we need bookkeeping to ensure that there are no repetitions).

The issue is not trivial, since presumably we will find events we want to isolate and study/reprocess. Thus we need to settle it. 

The need to carry the event_DAQ as unique identifier of the PMAP is another related issue. Notice that in the current implementation (@jacg, @gonzaponte), we are returning:

{event: {peak: ([t], [E])}

But we use as reference for "event" the field "event" in the tables:

class S12(tb.IsDescription):
    """Store for a S1/S2
    The table maps a S12:
    peak is the index of the S12 dictionary, running over the number of peaks found
    time and energy of the peak.
    """
    event  = tb.Int32Col(pos=0)
    evtDaq = tb.Int32Col(pos=1)
    peak   = tb.UInt8Col(pos=2)  # peak number
    time   = tb.Float32Col(pos=3) # time in ns
    ene    = tb.Float32Col(pos=4) # energy in pes

This is fine if we work with a bunch of PMAPS derived from a single "file bunch" processing, but would result in repeated keys (for event) if we merge to PMAPS files. Thus, the key must be evtDAQ, not event (in fact I believe that we just need one field and it must be what we call evtDaq, do you agree, @jmbenlloch?) 

[...]

PR #168, opened 2017-03-13
Write run info & evt number in Diomira

jmbenlloch:
Diomira writes run rumber in one table and event number+timestamp in
another table. There is a new option for configuration files (or
command line) to specify which number should be assigned to the first
event in the file.

This is useful for MC productions where we want to keep an absolute
numbering for all the events in one MC run.

This PR addresses #145


jjgomezcadenas:
I cannot manage to see where you call set_output_store()

jjgomezcadenas:
@jacg This is the PR that addresses the comments "COMING SOON" in IRENE. It extends DIOMIRA (which produces MC events) so that it also writes a table with absolute event numbers, etc., so that we can treat data and MC in an uniform way.

jacg:
> I cannot manage to see where you call set_output_store()

Line 101 in `diomira.py` ?

jjgomezcadenas:
Concerning your text:    
                 
                       # add evt number
                        evt_row = self.evtsInfot.row
                        evt_row['evt_number'] = evt + self.first_evt
                        evt_row['timestamp'] = 0
                        evt_row.append()

This could go to the base class, no? 

Then

                        n_events_tot +=1
                        if n_events_tot % self.nprint == 0:
                            print('event in file = {}, total = {}'
                                  .format(evt, n_events_tot))

In the branch pmaps-writer I have included a small function print_conditional in the base class (City), this would save a few lines of code.

jjgomezcadenas (commented):
Aside of the small comments I have done, my limited eyes cannot see much wrong here either. From my side this can go, after @jacg is happy (which means he approves).

jmbenlloch:
> This could go to the base class, no?

I don't see why that code should go to the base class. I understand that in MC we start at Diomira where we generate the event number. In data there is already a event number from DAQ. Following cities, both in MC and data, should copy the event number that is already in the file, not compute a new one. To propagate the event number we need a different code.

jjgomezcadenas:
But Diomira is derived from a base class (SensorResponseCity). Should the code go there?

jmbenlloch:
There are 3 pieces here:

1- **Create tables**. This is done in `set_output_store`
2- **Write run number**. Done in the run method after opening output file.
3- **Write event number & timestamp**. Also done within `run` method.

Steps 2 and 3 need step 1 to be executed before. Since all the output is managed within Diomira I don't see why we should put 2 or 3 in `SensorResponseCity`, it is just another Diomira output.

Besides that, if you want to encapsulate those 4 lines in one function it is done in my last commit. Feel free to choose any alternative, I don't have strong feelings about it.

jacg (approved):
I won't have time to look into this deeply before I disappear. Superficially, I see nothing wrong. I don't want to delay this any longer, so I'm approving and merging ...

PR #169, opened 2017-03-13
Use dictionary instead of seqence of ifs

jacg:
The signal to noise ratio in the original was pretty low: endless
repetition of

    if name == ????? return tp.Filters(complevel=?, complib=????)

where the ???? represent signal, and the rest is boilerplate.

Dictionaries exist for good reason, and we are all familiar with
them. Let's not re-invent this particular wheel.

It is worrying that no current test spots serious and obvious problems
with this function. For example, try removing the return altogether,
and all our tests pass!

### DIFF
--- a/invisible_cities/reco/tbl_functions.py
+++ b/invisible_cities/reco/tbl_functions.py
@@ -34,21 +34,18 @@ def filters(name):
     filt : tb.filters.Filter
         Filter mode instance.
     """
-    if name == "NOCOMPR":
-        return tb.Filters(complevel=0)  # no compression
-    if name == "ZLIB1":
-        return tb.Filters(complevel=1, complib="zlib")
-    if name == "ZLIB4":
-        return tb.Filters(complevel=4, complib="zlib")
-    if name == "ZLIB5":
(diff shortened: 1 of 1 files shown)


jjgomezcadenas (approved):
Elegant!

jacg:
Any comments about the lack of tests?

jjgomezcadenas:
True, I just looked at the code, which is a much better imp that what we had, a test would do no harm. This feature have never been really exercised since we always use ZLIB4, so it's a perfect place for a bug to lurk (although looking at the elegant dict it appears difficult that the bug finds a hideout...)

jacg:
Let me just remind you: while refactoring this, *I completely forgot to return anything from this function* and **not a single test spotted it**!

That's bad. That implies that nothing that depends on this function has any tests that care when it returns `None`. I guess that means that we don't care if our compression requests are ignored and we write everything uncompressed.

jjgomezcadenas:
Well, what it really means is that since we are all the time using ZLIB4, we had not identified the need for a test. It is clear that we must add one and I am sure @jmbenlloch volunteers for it. Roma no se hizo en una hora...

PR #171, opened 2017-03-13
Refactoring of IRENE, new writer, new S1 S2, Si classes, new tests.

jjgomezcadenas:
This PR includes:

1) an interface for pmaps writing (context capable).
2) Refactoring of S1, S2, S2Si as dict subclasses (with self-writhing capabilities)
3) New format for S2Si 
4) Refactoring of IRENE, using new classes and writing interface.
5) New tests to exhaustively check IO and PMAPS structure
6) Reworking of existing tests
7) Some cleanup

### DIFF
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -31,6 +31,7 @@ from   invisible_cities.core.random_sampling \
 from   invisible_cities.core.configure import print_configuration
 from   invisible_cities.reco.params import S12Params, SensorParams
 
+import invisible_cities.reco.pmap_io as pio
 
 if sys.version_info >= (3,5):
     # Exec to avoid syntax errors in older Pythons
@@ -102,6 +103,11 @@ class City:
     def set_compression(self, compression):
         """Set the input files."""
         self.compression = compression
+
+    def conditional_print(self, evt, n_events_tot):
+        if n_events_tot % self.nprint == 0:
+            print('event in file = {}, total = {}'
+                  .format(evt, n_events_tot))
 
     def display_IO_info(self, nmax):
         print("""
@@ -502,7 +508,7 @@ class PmapCity(CalibratedCity):
         """Return S2Si."""
         SIPM = cpf.select_sipm(sipmzs)
         S2Si = pf.sipm_s2_dict(SIPM, S2, thr = self.thr_sipm_s2)
-        return S2Si
+        return pio.S2Si(S2Si)
 
 
     config_file_format = CalibratedCity.config_file_format + """
--- a/invisible_cities/cities/irene.py
+++ b/invisible_cities/cities/irene.py
@@ -4,27 +4,19 @@ import sys
 from glob import glob
 from time import time
 
-import numpy as np
+import numpy  as np
 import tables as tb
 
-from   invisible_cities.reco.nh5 import S12, S2Si, EventInfo, RunInfo
-import invisible_cities.reco.tbl_functions as tbl
-from   invisible_cities.core.configure \
-       import configure, define_event_loop, print_configuration
 
-from   invisible_cities.core.system_of_units_c import units
-from   invisible_cities.cities.base_cities import PmapCity, SensorParams
-from   invisible_cities.cities.base_cities import S12Params as S12P
-from   invisible_cities.core.mctrk_functions import MCTrackWriter
+from invisible_cities.core.configure import configure, print_configuration
+from invisible_cities.core.system_of_units_c import units
 
+from invisible_cities.reco.pmap_io import pmap_writer, S12, S2Si
+from invisible_cities.core.system_of_units_c import units
+from invisible_cities.cities.base_cities import PmapCity, SensorParams
+from invisible_cities.cities.base_cities import S12Params as S12P
+from invisible_cities.core.mctrk_functions import MCTrackWriter
 
-# Parameters for S1/S2 searches
-# tmin and tmax define the time interval of the searches
-# lmin and lmax define the minimum and maximum width of the signal
-# stride define the tolerable size of the "hole" between to samples above
-# threshold when performing the search
-# S12Params = namedtuple('S12Params', 'tmin tmax stride lmin lmax')
-# S12P = S12Params
 
 class Irene(PmapCity):
     """Perform fast processing from raw data to pmaps.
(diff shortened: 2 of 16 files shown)


jacg:
I have rebased onto latest master, as there were conflicts.

I have added a few mostly trivial commits. You had forgotten to pass Irene's compression setting to `write_pmaps`.

Irene's `run` method is still well over 100 lines long (though I managed to remove a few with trivial changes). Given fix this now, given that we are here?

JJ replies: Possibly, although I believe that is quite concise already

jacg:
# produce a fake timestamp for MC
    # TODO This will be fixed shortly
    timestamp = int(time())
    # For MC, event is the absolute number in file group
    # TODO: this will be fixed shortly
    event = n_events_tot

If the comments are to be believed, a number of things will be fixed shortly. Has this been done? Will it be done any time soon?

JJ replies: Yes, this has to do with the work that Jose Maria is doing so that MC events are already produced with the same event/run tables than data events.

jacg:
self.display_IO_info(nmax)
    
    print("""
             S1 parameters {}""" .format(self.s1_params))
    
    print("""
             S2 parameters {}""" .format(self.s2_params))
    
    print("""
             S2Si parameters
             threshold min charge per SiPM = {s.thr_sipm} pes
             threshold min charge in  S2   = {s.thr_sipm_s2} pes
                      """.format(s=self))

I suspect that these prints can go into `display_IO_info`, extended in `Irene`.

JJ replies: agreed!

jacg:
`IOError` is being raised in a place where some kind of `ICException` should.

JJ replies: Agreed!

jacg:
s1, s2 = self.find_S12(s1_ene, s1_indx,
                           s2_ene, s2_indx)
    s2si = self.find_S2Si(s2, sipmzs)
    
    # make S1, S2 and S2Si objects (from dicts)
    S1 = S12(s1)
    S2 = S12(s2)
    Si = S2Si(s2si)

Is there any reason why the `find_S*` functions can't return `S12` and `S2Si` instances?

JJ replies: Certainly not! We can do this!

jjgomezcadenas:
I have answered your points editing the comments themselves (is this correct practice?) In short, agree with all your suggestions. Do you want to go ahead and do it? I am running short of time, since I am travelling tomorrow, and will be pretty useless over the next few days.

jacg:
> I have answered your points editing the comments themselves (is this correct practice?) 

I don't think so:

+ Nobody gets notifications of the edits, so it could be sitting there forever without anyone noticing.
+ There is no way of telling who wrote what, so a third party reading it (or indeed one of us, after a significant time has passed) will be *hopelessly* lost as to what it all means.

Quote the part to which you are replying in your comment (by prepending `>`) in a separate comment of your own.

jacg:
> Do you want to go ahead and do it?

Yes, I'll knock them off sooner or later.

jjgomezcadenas:
>Quote the part to which you are replying in your comment (by prepending >) in a separate comment of your own.

OK, got it.

jacg:
>> method is still well over 100 lines long 

> I believe that is quite concise already

Spoken like a true physicist! :-)

Remember the criterion you yourself invented? Does it fit on your screen in a font legible to *you*?

(We're at the next phase now, it's no longer my eyesight that sets the bar!)

jacg:
> OK, got it.

Yeah! That's it.

jjgomezcadenas:
>Spoken like a true physicist! :-)

Me thinks I am becoming a (bad) programmer way too fast

In any case it is clear that the noise can be reduced by shortening or sending elsewhere some of the trivial stuff, I believe you have already identified a few issues. There are also a few prints that have to go and made it by accident to the method.

jacg:
> There are also a few prints that have to go and made it by accident to the method.

I was hoping that were the case.

Could you please remove those, as you know which ones should be there and which should not.

Do it on top of the lastest commit in this PR, which is *your* `pmap-writer` branch in *your* fork. So you will have to fetch it.

jjgomezcadenas:
>Do it on top of the lastest commit in this PR, which is your pmap-writer branch in your fork. So you will have to fetch it.

Done. Shall I rebase my map-writer into origin/pmap-writer (just want to be sure I don't do something stupid)

jacg:
You should have 

1. fetched `origin/pmap-writer`,
1. (either) reset your local `pmap-writer` (or a new local branch) to `origin/pmap-writer` and done the work on it,
1. pushed it back to `origin/pmap-writer` (I still haven't seen it appear)

I don't think there's any rebasing involved. I don't think I rebased anything here, so your `origin/pmap-writer` should simply have stuck a few more commits on top of your local `pmap-writer`. Ah, unless you did some other work on your local `pmap-writer` in the meantime.

In that case, it depends on whether you want to share that work as part of this PR or not.

jjgomezcadenas:
I have done the reset, but a good number of tests that used to work now fail. Are all your tests passing?

jacg:
> Are all your tests passing?

Yes, as indicated by the green tick next to my last commit, between four and five comments higher up this page.

jjgomezcadenas:
OK, the problem in some of the tests is related with returning instances of S1, S2, Si instead of dictionaries from the functions. In some tests, I call the functions and use directly the dictionaries to compare with the data that I read from file. I have not looked at the test in detail yet, but I know that is is related with finding an instance rather than a dict from the messages and because when I reset right before that change all the test pass.

jacg:
I'm signing off for the night.

jjgomezcadenas:
I think the problem is here:

    def find_S12(self, s1_ene, s1_indx, s2_ene, s2_indx):
        """Return S1 and S2."""
        S1 = cpf.find_S12(s1_ene,
                          s1_indx,
                          **self.s1_params._asdict())

        S2 = cpf.find_S12(s2_ene,
                          s2_indx,
                          **self.s2_params._asdict())
        return S1, S2

    def find_S2Si(self, S2, sipmzs):
        """Return S2Si."""
        SIPM = cpf.select_sipm(sipmzs)
        S2Si = pf.sipm_s2_dict(SIPM, S2, thr = self.thr_sipm_s2)
        return pio.S2Si(S2Si)

Notice that you are returning the class for S2Si but only the dicts for S1 and S2. Thus, my test fail, as they should. I suspect that you have missed a commit.

jacg:
GitHub usage point: Code blocks should be indented four spaces, or triple-backquoted. The first `def` in your last code example was at column zero, so the first block was not highlighted as code. Note that you have a preview button on top of the editing area, so you can see what you have written in all its glory before sending it off.

I have edited it, so it's fixed now.

On to the actual code ...

> Notice that you are returning the class for S2Si but only the dicts for S1 and S2. Thus, my test fail, as they should. I suspect that you have missed a commit.

I used all available tests to guide me in my implementation. All tests passed on my machine and on Travis, which let be to believe that I had finished.

I have pulled your solution and added the line

    return         S1 ,         S2

just before

    return pio.S12(S1), pio.S12(S2)

thereby effectively removing your solution. All the tests I have still pass. We are clearly not running the same code. Which tests fail for you?

jjgomezcadenas:
>All the tests I have still pass. We are clearly not running the same code. Which tests fail for you?

    invisible_cities/cities/irene_test.py::test_command_line_irene_electrons_40keV FAILED
    invisible_cities/cities/irene_test.py::test_command_line_irene_run_2983 FAILED
    invisible_cities/cities/irene_test.py::test_command_line_irene_runinfo_run_2983 FAILED

The tests *must* fail, since they try to write the objects S1 S2 which they cannot (since they are dicts, nor instances of classes with write method).

```
invisible_cities/cities/irene.py:136: in run
    write(self.run_number, event, timestamp, S1, S2, Si)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <invisible_cities.reco.pmap_io.pmap_writer object at 0x114d1b518>, run_number = 0, event_number = 0, timestamp = 1489481840
s1 = {0: [array([ 100050.,  100075.,  100100.,  100125.,  100150.,  100175.,
        100200.,  100225.,  100250.]), array([...4186,  0.77949337,  1.31605857,  0.86779838,  0.73778522,
        0.46606872,  0.42582626,  0.50926897,  0.33664909])]}
s2 = {0: [array([ 124662.5,  125662.5,  126662.5,  127662.5,  128300. ]), array([  596.40477265,  3087.31824433,  3380.07039472,   920.54217065,
          25.12389311])]}
s2si = {0: {}}

    def __call__(self, run_number, event_number, timestamp, s1, s2, s2si):
>       s1  .store(self._pmp_tables[0], event_number)
E       AttributeError: 'dict' object has no attribute 'store'
```

jacg:
Just to be sure: have you recompiled the cython modules before running the tests?

jacg:
Yes, the test failure you present appears to be consistent with not having recompiled `peak_functions_c.pyx`.

jjgomezcadenas:
No, I had not. I have recompiled now and it passes all the tests. Now I understand, the functions returning classes are the python functions (I was looking at the interface) and thus, not being recompiled, they were still sending dicts. Mystery clarified, sorry about compilation. Shall I fetch again from origin?

jjgomezcadenas:
I have pushed to origin, undoing my changes. Now I see that there is a conflict in Irene_test.py...

jacg:
> there is a conflict in Irene_test.py...

Wait. I am rebasing on top of #172.

I notice that the whole test `test_pmaps_store_issue_151` has gone. Is that intentional?

jacg:
OK, done.

**Careful!** I have pushed (con mucha fuerza) over your last push. The last commit should be equivalent to yours, though.

It is now rebased on top of #172. We can take it from here.

jacg:
I have gone with the removal of `test_pmaps_store_issue_151`. Please confirm that you really meant it, and that something equivalent exists somewhere else.

jjgomezcadenas:
>Please confirm that you really meant it, and that something equivalent exists somewhere else.

Yes I meant it. A much more exhaustive test exists now in pmaps_io_test

jacg:
What is the relationship between the timestamp written into the event info table, and the times that appear in the `S12`s?

jjgomezcadenas:
None. The times in the PMAPS refer to relative sample in the waveform. The timestamp marks when the whole event was recorded. 

BTW, signing off for the day, travelling to Venice, probably out of the game for 2-3 days.

jacg:
What is the time resolution of the detector? It's not sub-nanosecond, IIRC. If so, why is the `S12` time (in ns, according to the comment) being stored as a float?

jjgomezcadenas:
Resolution is order 1 ns, although never used in practice. Samples are 25 ns for the DAQ, times are often rebin to microsecond (in S2).

jacg:
So would it make sense to change the type of `time` in

    class S12(tb.IsDescription):
        ...
        time   = tb.Float32Col(pos=3) # time in ns

to some kind of `int`?

jjgomezcadenas:
Not sure. We want to rebin and need to understand the omplications. Lets discuss this when i an back in Geneve

Enviado desde mi iPhone

jacg:
Here is something that seems to work. I warn you, it's not pretty!

The interfaces relating to pmaps (and some other decisions) need major review, but we do not have time for that at present.

I have pushed a single, fat commit, containing all the changes. I have kept the very noisy unsquashed history on the [unsquashed-pmap-writer](https://github.com/jacg/IC/tree/unsquashed-pmap-writer) branch in my fork.

jmbenlloch (approved):
@jacg has given approval

PR #172, opened 2017-03-14
Add ICDIR fixture

jacg:
`os.environ['ICDIR']` appeared lots of times in lots of tests. It's a
pain to write. So provide an `ICDIR` fixture to ease the pain.

### DIFF
--- a/invisible_cities/cities/diomira_test.py
+++ b/invisible_cities/cities/diomira_test.py
@@ -31,9 +31,9 @@ def conf_file_name(config_tmpdir):
     return conf_file_name
 
 
-def test_diomira_fee_table():
+def test_diomira_fee_table(ICDIR):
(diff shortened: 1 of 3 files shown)


jacg:
An alternative approach would be to provide an `ICDIR()` function in some module, maybe in the `config` package.

+ This would require the function to be imported only once per module, rather than being mentioned in the parameter list of every test that uses it ... but using it (`ICDIR()`) is slightly more annoying that the fixture (`ICDIR`)

+ It could be used in *any* code, not just the tests ... but so far, it is *only* needs seem to be in the tests.

jjgomezcadenas (approved):
Very useful. Preferable, I think, to ICDIR() for the time being (only used by tests and shorter to write).

PR #173, opened 2017-03-15
Update DB to use the same tables used in ART software.

jmbenlloch:
So far we have been using different databases in NEXT software. This PR changes our code to use the tables structure used in ART, so in the future we don't have two keep updating two databases whenever there is a new calibration, etc.

### DIFF
--- a/invisible_cities/database/download.py
+++ b/invisible_cities/database/download.py
@@ -17,6 +17,7 @@ def loadDB():
     connSql3 = sqlite3.connect(dbfile)
     cursorSql3 = connSql3.cursor()
 
+
     connMySql = MySQLdb.connect(host="neutrinos1.ific.uv.es", user=dec('am1iZW5sbG9jaA=='),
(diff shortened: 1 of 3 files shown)


jjgomezcadenas (commented):
This is a much needed technical upgrade. I see that no new tests are added. Are u satisfied with the level of testing here?

jacg:
Unless you have some very specific questions for me, I will ignore this PR. I am too unfamiliar with this part of IC, and I have too much on my plate before I disappear for two weeks at the end of this one.

jmbenlloch:
@jacg no problem.

@jjgomezcadenas you are right. I have modified existing tests to check not only the number of sensors but also the properties appearing in the pandas dataframe.

jjgomezcadenas (approved):
OK, we go ahead with this.

jmbenlloch:
I forgot to remark one important thing. Everybody has to download again the database. So please execute: `source manage.sh download_test_db` after fetching nextic/master.

PR #176, opened 2017-03-16
Read MC event info in Irene

jmbenlloch:
Irene now reads event number & timestamp from file both in MC and data.

There is also one detail, in MCRD and Diomira we have:
```
/Run                     Group
/Run/events              Dataset {10/Inf}
/Run/runInfo             Dataset {1}
```

while in Irene we have:
```
/RunEvtInfo              Group
/RunEvtInfo/eventInfo    Dataset {10/Inf}
/RunEvtInfo/runInfo      Dataset {10/Inf}
```

At some point we should put the same names in all cities.


invisible_cities/cities/irene_test.py
```diff
@@ -70,6 +70,12 @@ def test_command_line_irene_electrons_40keV(conf_file_name_mc, config_tmpdir, IC
             mctracks_out = h5out.root.MC.MCTracks[0]
             np.testing.assert_array_equal(mctracks_in, mctracks_out)
 
+            # chech events numbers & timestamps
+            evts_in = h5in.root.Run.events[:3]
```
jacg: Could we get rid of these magic numbers? There are a couple of 3s here, and a couple of 0s a bit higher up.

jjgomezcadenas:
Can you go ahead and fix Irene so that the table is the same than the table in DIMIRA? I believe that this PR is the right place to do it!

jacg:
Whilst we're here, could we deal with the `0` magic numbers a few lines above the ones that you removed?

jmbenlloch:
That is just to read first row in both tables and checking the have the same content. I have added a variable for that number.

jjgomezcadenas:
@jmbenlloch I am happy to approve when @jacg is happy. This PR is important to have unified the tables of DIOMIRA and IRENE (and to get rid of magics). @jacg let @jmbenlloch know of any new suggestion, since he will be off tomorrow for a few days.

jacg (approved):
Needs squashing before merging.

jacg:
Don't forget to talk about the table unification in the squshed commit message!

jmbenlloch:
Done!

jacg:
How does this change of table names affect us in terms of data files *that have already been written*?

jmbenlloch:
I don't think it is a problem since we are regenerating all files now. There is a small diomira production (200k evts) going on right now, after that I will run Irene and if there is now more problems, we will run eventually the whole production (4M evts)

jjgomezcadenas:
Not seriously. The point of pushing all this PR work was that we are going to reprocess all Monte Carlo data and we are getting new data from the detector... this afternoon! Old data can be ignored (or reprocessed if needed).

jacg:
> Old data can be ignored (or reprocessed if needed).

I'm just trying to have some awareness of where we stand in terms of having to regenerate, reprocess, or throw away old data, and how much it's costing. And how this cost is likely to grow, in case we feel the need to do something like this again!

>  we will run eventually the whole production (4M evts)

What does this mean in terms of CPU hours?

jjgomezcadenas:
>I'm just trying to have some awareness of where we stand in terms of having to regenerate, reprocess, or throw away old data, and how much it's costing. And how this cost is likely to grow, in case we feel the need to do something like this again!

Old data is not very useful, we had a lot of problems with it. The most likely scenario is that we ignore it and carry on with new data. Reprocessing MC is common practice and in this case was mandatory, since we had to add the new MC tables.

jmbenlloch:
> What does this mean in terms of CPU hours?

Around 800 hours for diomira and a little bit less for Irene.

jacg:
> Around 800 hours for diomira and a little bit less for Irene.

Has anybody ever profiled any of this?

jmbenlloch:
Well, I can tell you the times from last time. First we have to copy the input file from the network filesystem to the computing node, then run diomira and in the end copy the output. This are the times in seconds for jobs with 10k evts each, 400 jobs in total:

```
copy input file
avg: 40.8040201005,	 std: 50.7643762238

diomira
avg: 11647.7110553,	 std: 5665.84610538

copy output file
avg: 73.675879397,	 std: 96.8217705237

total
avg: 11762.1909548,	 std: 5711.30528197
```

So in the end, is more than the 800 hours I told you before (that was a quick estimation I underestimated the time per event a little bit).  Not all machines in the CERN cluster have the same CPU, in some machines jobs take longer than in others. For more details I guess one should profile diomira.

jjgomezcadenas:
Last time I tried, Diomira run at 1 second per event and Irene a 0.5 second per event.

jacg:
Yes, yes, but I mean actually *profile*, you know, with a profiler.

If we are running stuff for the order of 1000 hours, then it might be worthwhile to optimize things a bit.

If it's not holding us back so far, then I wouldn't prioritize it, but we should aim to do it sometime in the not too distant future.

jjgomezcadenas:
I did profile both DIOMIRA and IRENE. In the case of DIOMIRA all the time was consumed by Scipy functions, which are already highly optimised, so I am not sure we can optimise much more. In the case of Irene I found the bottlenecks and re-wrote the functions in cython. I think that the performance was quite optimised but I am happy to look at the thing together again! Of course it would be nice if we can scratch some extra time!

jacg:
> all the time was consumed by Scipy functions

That's encouraging

> which are already highly optimised, so I am not sure we can optimise much more.

Well, if (for example) you are calling some scipy function N<sup>2</sup> times where N times would do then that would

+ be consistent with the profile output you describe, **and**
+ still leave room for improvement.

We should aim to look at it sometime when we're not too pressed by other things.

jjgomezcadenas:
Agreed! Also I don't have much experience with profiling, thus looking forward to the peras!

PR #180, opened 2017-03-21
Issue174

gonzaponte:
There was some mismatching between the number of slices in the cathode and in the anode. This was caused by a difference in the time value of consecutive slices smaller than one mus. Although this is still the case, the number of slices has been forced to be the same in both planes.

These changes require rerunning irene.


jjgomezcadenas:
how big is your pmaps file?

gonzaponte:
204k

jjgomezcadenas (approved):
As agreed, this fix provides a practical solution to the problem of unmatched slices.

gonzaponte:
I will squash it, then

jjgomezcadenas:
I have approved the PR, but I believe we should squash. Will you do it?

gonzaponte:
Done.

PR #181, opened 2017-03-21
Peak functions

jjgomezcadenas:
This PR addresses modifications in various functions needed to deal with the fact that one can have a number of dead PMTs (in real data). This is handled by defining a list of active PMTs which is passed around the functions that operate with PMTs. If the list is empty, the default is all PMTs active (this guarantees backward compatibility, e.g, preserves the signatures of functions). Some tests have been added to check the new functionality.

### DIFF
--- a/invisible_cities/core/mpl_functions.py
+++ b/invisible_cities/core/mpl_functions.py
@@ -66,7 +66,8 @@ def plot_signal_vs_time_mus(signal,
                             t_min      =    0,
                             t_max      = 1200,
                             signal_min =    0,
-                            signal_max =  200):
+                            signal_max =  200,
+                            label=''):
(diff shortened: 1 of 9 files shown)


jmbenlloch (commented):
This is a very important update since we are running with only 11 PMTs instead of 12. With this changes IC can process only active PMTs, fixing the errors we were getting when we had (inactive) PMTs with gain 0.

This PR includes some test for the new functionality and they work correctly. The only thing needed before approval is to squash the commits. @jjgomezcadenas can you do that?

jjgomezcadenas:
This has been squashed (but somehow the conflict has reappeared)

jmbenlloch (approved):

jmbenlloch:
I have solved the conflict and approved this PR.

PR #182, opened 2017-03-21
Irene masked pmts

gonzaponte:
This PR updates the cities machinery to work with the new functionality introduced in the last PR. 

The Deconvolution and Calibrated cities have been extended to work with a reduced set of PMTs. This is done by adding a new configuration parameter PMT_ACTIVE which is a list of the PMTs that are alive in the detector.

Since this new parameter is a list, the `read_config_file` function has been also updated to work with lists.

A new test using Irene with a reduced number of PMTs has been introduced.

### DIFF
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -72,6 +72,7 @@ class City:
         # This is JCK-1: text reveals symmetry!
         self.xs              = DataSiPM.X.values
         self.ys              = DataSiPM.Y.values
+        self.pmt_active      = np.nonzero(DataPMT.Active.values)[0].tolist()
         self.adc_to_pes      = abs(DataPMT.adc_to_pes.values).astype(np.double)
         self.sipm_adc_to_pes = DataSiPM.adc_to_pes.values    .astype(np.double)
         self.coeff_c         = DataPMT.coeff_c.values        .astype(np.double)
@@ -321,6 +322,7 @@ class DeconvolutionCity(City):
         return blr.deconv_pmt(RWF,
                               self.coeff_c,
                               self.coeff_blr,
+                              pmt_active  = self.pmt_active,
                               n_baseline  = self.n_baseline,
                               thr_trigger = self.thr_trigger)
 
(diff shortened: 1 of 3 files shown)


gonzaponte:
This is still Work in Progress. Remaining tasks:

- [ ] Update/Add test for `read_config_file` function so it uses some list.
- [ ] Discuss the "nastiness" introduced.
- [ ] ...

jjgomezcadenas:
First look (at a late hour, though) appears as if the machinery of the city is correctly updated. Difficult to judge (for me) if there is a better way to parse the list. We need a test for sure to demonstrate that the list is correctly parsed. After you add it I will review again. 

I suspect that @jacg will have comments at least in what refers to the parsing of the list, but since we need the updated version of IRENE for the upcoming Krypton production (millions of krypton events already waiting), let's focus now in ensuring correct behaviour (tests) and we will review with @jacg technical debt asap.

jmbenlloch:
I don't really get this parameter for the config file. Shouldn't we read the active pmt list from the database? I will have the fix for the DB (both in the server and in the code) later today.

gonzaponte:
I agree. @jjgomezcadenas?

O 22 mar, 2017 10:32 a.m., "jmbenlloch" <notifications@github.com> escribiu:

jjgomezcadenas:
Much better! indeed, from data base!

gonzaponte:
JM is so right I feel stupid now!

O 22 mar, 2017 10:42 a.m., "jjgomezcadenas" <notifications@github.com>
escribiu:

jjgomezcadenas:
Don't worry, no harm to your Karma. Let's fix the code, removing noise associated with configure, prepare for reading data base etc. Namaste!

gonzaponte:
I have rebased onto JM's work #183.

The only thing missing is a test to check this new feature, but I can't think of one... What we need to test now is that the actual number of PMTs is the one we expect and that it is carried on to the relevant functions. The former is easy, but the latter, not that much. Because we do not store the cwf for each PMT, I cannot check the cwfs of the missing pmts are empty. Suggestions?

jjgomezcadenas (approved):
The changes are straight forward and the additional test sufficient to exercise them. Thus, I approve.

PR #184, opened 2017-03-24
Separate run-event tables from pmp tables.

jjgomezcadenas:
This module provides a writer interface for PMAPS which includes
writing run and event info. By separating the PMAPS tables and the
event-run tables it is possible that other modules reuse the run-event
table functions, which will be common to other writers.

### DIFF
--- a/invisible_cities/reco/pmap_io.py
+++ b/invisible_cities/reco/pmap_io.py
@@ -8,7 +8,9 @@ class pmap_writer:
 
     def __init__(self, filename, compression = 'ZLIB4'):
         self._hdf5_file = tb.open_file(filename, 'w')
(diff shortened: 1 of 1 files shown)


jmbenlloch (approved):
This change make it easier to reuse the functions to write those tables. All tests are passing, so I approve.

jacg:
I haven't looked at the code, but just from the title, I approve. This was something on my list of ''things I have to convince you to do", so I'm glad to see that you have reached this conclusion independently.

jjgomezcadenas:
>This was something on my list of ''things I have to convince you to do", so I'm glad to see that you have reached this conclusion independently.

Right now I feel my Karma levels skyrocketing...

PR #185, opened 2017-03-24
Remove unnecessary parameter in event_and_timestamp

jjgomezcadenas:
Parameter n_events_tot was being passed to function but not used.


jmbenlloch (approved):
You are right indeed, that argument is not needed. This can be approved.

PR #186, opened 2017-03-27
Add function load_pmaps() to return transient rep of PMAPS

jjgomezcadenas:
Function read_pmaps() returns data frames (e.g, a "memcopy" in the
form of a dataframe of the table used for persistent rep of PMAPS. The
user needs to call extra functions to obtain the transient rep. This
addition allows the use to obtain the transient rep (e.g, dicts) with
a single call, simplfying the interface and hiding the low level
stuff.

A test has been added to ensure correct behaviour.


jmbenlloch (approved):
This seems to be an useful function, it includes a correct test and all of them have passed. I'll approve and rebase onto master.

PR #187, opened 2017-03-28
Fix error in MCTracks event numbers

jmbenlloch:
When we change event numbers in diomira using the parameter `FIRST_EVT`, where are not updating those numbers also in MCTracks table. This PR fixes that mismatch.

The first commit include a test showing the error and the second commit contains the fix. If everyone is happy I'll squash them so we can rebase onto master.


jjgomezcadenas (approved):
Bug fix with tests, approved

PR #188, opened 2017-03-29
S1mau

gonzaponte:
We use the MAU-corrected waveform to find the S1 signals, but this affects the reconstructed energy, so we need to take the same time window in the non-corrected waveform. This PR addresses this issue.


jjgomezcadenas (approved):
This addition is needed since the MAU biases slightly the energies. The implementation follows the same guidelines than the rest of the coding, using an interface in base_cities that calls a fast CYTHON function in peak_functions_c. The test is convincing, so I approve.

PR #189, opened 2017-03-30
Change event index type from int16 to int32 in MCTracks table

jmbenlloch:
With int16 we have can only store 2^15 (16384) events without overflow in event number. Changing to int32 we would get up to 2147483648 events per MC run.

In the last reprocessing of MC Kr many files have wrong event numbers in `MCTrack` table due to overflow. We need this change in the code and rerun MC again at CERN.

This same error could happen in `nh5.py` in this table:

```
class DECONV_PARAM(tb.IsDescription):
    N_BASELINE            = tb.Int16Col(pos=0)
    THR_TRIGGER           = tb.Int16Col(pos=1)
    ACUM_DISCHARGE_LENGTH = tb.Int16Col(pos=2)
    ACUM_TAU              = tb.Int16Col(pos=3)
    ACUM_COMPRESS         = tb.Float32Col(pos=4)
```

Taking into account the sign bit, `N_BASELINE` could go up to 32768 and I think sometimes we use higher values (like 48000). I have search in our code and I have not found any place where we use `DECONV_PARAM`. I think we should either remove it from the code or change also `N_BASELINE`to `Int32Col` just in case we use it in the future.

I have corrected the bug in `MCTracks` and update one test to check this issue. In the test we check that event number 100000 is written to the hdf5 file and read back correctly. This also required regenerating the one of the test files to change the format of its `MCtracks` table.


gonzaponte (approved):
Maybe you should write the `DECONV_PARAM` stuff as an issue so we can easily keep track of it when this PR is closed.

The PR itself is straightforward, so I am approving it.

jjgomezcadenas:
Deconv Params should be written by IRENE in an independent table, so that we can keep track of what we did in deconvolution 

@jmbenlloch  please, change Int16 to Int32 in the table, let's get that fixed first. Then @gonzaponte, once you have time (I guess this afternoon or tomorrow), add another table to the output of IRENE with this info and write it.

PR #193, opened 2017-03-30
Update deconv_param table and store it in output file

gonzaponte:
squash


jjgomezcadenas (approved):
This squashed the previous commit. I am afraid there is a better way to do it, but I guess this should do

PR #197, opened 2017-04-05
Add several plot functions

jjgomezcadenas:
Several plot functions, hitherto living in notebooks have been added
to the mpl and pmaps modules.


jmbenlloch (approved):
Useful plotting functions. As it is difficult to test these functions I'll give the approval since this have been already used in notebooks and they worked.

PR #198, opened 2017-04-05
City to simulate trigger

paolafer:
This is a city which simulates trigger in MC. For now, it allows trigger in S1 and S2, but not in a coincidence with external scintillator. This is just a first version, I'm doing a PR in order for people to look at the code and comment. I've implemented only two tests so far, this is also to be extended.


jacg:
This requires review con peras. I volunteer if you are geographically near to me sometime in the next week.

paolafer:
I'm afraid I'm in Valencia... The sooner I could go to Geneva (if Juanjo
thinks it's ok) is after the 26th of April.
On the other hand, here in Valencia there are Alex and Josh.

Il giorno mer 5 apr 2017 alle 22:58 jacg <notifications@github.com> ha
scritto:

jacg:
> here in Valencia there are Alex and Josh.

Fine people that they are, they have not yet gone through the process of having their own code accepted into IC enough times to be up to speed on the points that need addressing here.

So, I'm afraid you need pera review from @gonzaponte, @jmbenlloch, @jjgomezcadenas or @jacg on this one.

paolafer:
@jmbenlloch <https://github.com/jmbenlloch> will be in Valencia from the
18th of April... Maybe he could be the chosen one 😉?

Il giorno gio 6 apr 2017 alle 08:42 jacg <notifications@github.com> ha
scritto:

jacg:
I recommend that, rather than replying by email, you follow the link that appears at the very bottom of the notification email message you get from GitHub, and then post your reply from GitHub itself. All sorts of things go better if you do that.

Please treat email just as a notification that something has arrived in your GitHub inbox, not as a means of posting content on GitHub.

jjgomezcadenas:
Cecilia needs some serious thinking. In its current implementation Cecilia reads BLR waveforms, apply some trigger condition and writes BLR waveforms. I think both the input and the output are wrong.

1.- The input should be RWF, which is our standard way to store raw data. 

2.- This means that Cecilia has to deconvolute the waveform. But looking at @paolafer code, one also realises that she is in fact computing zero surpassed waveform and finding S2... of course, since all this is needed for the trigger condition. It follows that Cecilia is not a "simple" city, but in fact a full-fledged PMAP city. @paolafer is reinventing the wheel and calling "by hand" functions which are already available in the PMAP city interface. 

3.- But why should the output be a BLR, in particular given the fact that we have taken the effort to compute PMAPS (needed for the trigger?). The natural output should be PMAPS...

4.- But then, Cecilia is identical to IRENE if we add a TriggerCondition to Irene. 

5.- In short, I think that what we really want is to add a TriggerFilter to Irene.

jjgomezcadenas:
Actually, the problem is even a bit more complex than described before, since the trigger works on individual PMTs.  So one has to run the S2 search on individual PMTs, as @paolafer is doing. However, the input should still be RWF and the output, I think PMAPS. To be discussed.

jacg:
Done in #248.

PR #200, opened 2017-04-07
Ignore inline comments in config files

gonzaponte:
Example:
S1_LMIN 0 # This comment will be ignored


jjgomezcadenas:
Very useful! Can you add a test, please?

jacg:
Very quickly: I would recommend an approach based on `line.split('#',1)[0]`.

gonzaponte:
I have implemented @jacg suggestion and use some inline comment in the config file used for testing.

jjgomezcadenas:
Please squash and we can approve this PR

gonzaponte:
Done

jmbenlloch (approved):
Again, tests are passing, but in travis there is an error with the installation of pytest-xdist 1.15.0 (issue #205). We can approve this.

PR #204, opened 2017-04-07
Dorothea

gonzaponte:
A new City, Dorothea, devoted to producing a Kr DST. It includes several other cities in order to structure the process.

`FilterCity` is designed to provide some tools for peak selection given some criteria.
`S12SelectorCity` is designed to specialize `FilterCity` to S1/S2 signals creating a structured Event class when the event in question passes all selections.
`Dorothea` is the one that takes the data, uses the tools and dumps it to file using the Kr_writer.

`DST_writer` is somehow a generic table writer.
`Kr_writer` specializes `DST_writer` to write the DST as we want it for Kr-like events.

`Event` is a generic class to structure the relevant event info.
`KrEvent` extends `Event` with a method to store it in a table assuming it is Kr-like.

Everybody, please, criticize.


jjgomezcadenas:
Just discussed with @gonzaponte. He has added new tests including a "running city" test. We have also checked that the contents of the DST make sense after processing a very large run (0.5 M events). I am OK with approving this PR, if @jacg has no extra comments.

jacg:
@jacg is unlikely to have any comments in the near future.

Let me know once it's squashed into a sensible set of clean commits, and I might have a look.

jjgomezcadenas:
@gonzaponte please squash a per suggestion of the Master and let's wait today to see if he has extra input for us.

gonzaponte:
Squashed in two commits:
- Main code
- Tests and fixes

jacg:
I am going to insist on high quality of orthogonal and self-contained commits.

Why is the second commit entitled 'Tests and fixes'? If the only other, preceding, commit is called 'Main code', why isn't it correct in the first place? Whatever fixes there are, should have been squashed into the first commit.

There seems to be some shuffling of code from one module to another. This needs to be isolated into a separate commit. Anyone looking at the commit to understand what is being added in Dorotherea, shouldn't be distracted by orthogonal administration activities.

jjgomezcadenas:
The points are all correct, but you need to bear with us here. @gonzaponte is writing code for two new cities that we need for analysis of data, and running analysis himself. He has implemented all the points that we suggested in the first review, plus added some tests, plus we have exercised the code with analysis data and it works. Re-organizing all the commits may be difficult for him (we are not as experts using Magit as we should be) and we are really very short of time. Would that be possible for you to take a look to the PR as it is?

gonzaponte:
I didn't know how to do it properly. The problem is that I made self-contained commits (something like move this function to some module) in between other things (like fixing bugs in the code). I saved the unsquashed version in a branch, so I can undo and I am happy to repeat the squashing, but I would need some instructions there.

jacg:
If you need the code urgently, then just approve it. You don't need me for that.

jjgomezcadenas (approved):
The code includes all suggestions expressed in the review and adds a significant number of tests. The resulting DST appears to be bug free after several productions involving many hundred thousands of events. Thus, it appears robust enough for inclusion in IC.

PR #207, opened 2017-04-15
Accept None as value in config file

gonzaponte:


gonzaponte:
I don't think we actually need this, so I am closing it. However, this arose as part of a more general question, which I will open for discussion shortly.

Issue #208, opened 2017-05-02
Build broken, probably because of database change


jacg:
I just tried to push a trivial commit, and Travis/pytest gave 8 failures and 56 errors.

Rerunning builds which had passed earlier, now gives a similar failure.

Downloading the latest version of the database makes the tests fail in a similar way on my local machine.

Does anyone have a quick fix?

jmbenlloch:
I am checking what happened to find out where is the error. Until we manage to fix it, maybe you can take the previous DB version and run the tests locally, You can download the file here: http://jobenllo.web.cern.ch/jobenllo/localdb.sqlite3

PR #209, opened 2017-05-02
Fix DB code to include external PMT

jmbenlloch:
For the processing of some sodium runs we need to include the external PMT used in the detector to trigger in coincidence. This PMTs has a different label (PmtID), which is NaI. This PR updates load_db.DataPMT to include only PMTs with names like PMT* (from PMT1 to PMTC). 

The test has been updated to check that the returned dataframe includes only those kind of sensors.

Regarding the tests there is one small issue, DataPMT() has an optional argument, the run number. For MC we call DataPMT(0), if we run DataPMT() we will get the latest calibration available. Apart from that, tests test_pmts_pd and test_pmts_MC_pd are the same. Probably this could be parametrize, but I don't know how to parametrize it giving one argument in one case and no argument in the other.

When everything is ok, this should also be squashed before rebasing master.

### DIFF
--- a/invisible_cities/database/load_db.py
+++ b/invisible_cities/database/load_db.py
@@ -67,7 +67,7 @@ as msk ON pos.SensorID = msk.SensorID LEFT JOIN
 as gain ON pos.SensorID = gain.SensorID LEFT JOIN
 (select * from PmtBlr where MinRun < {3} and (MaxRun >= {3} or MaxRun is NULL))
 as blr ON map.ElecID = blr.ElecID
-where pos.SensorID < 100 and pos.MinRun={0} and map.MinRun={2}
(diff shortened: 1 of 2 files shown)


jmbenlloch:
This PR solves #208

jacg:
Even though these commits are now included in #210, I think that this should go in before #210. If we decide to accept #210, it can be rebased on top of this.

I'll leave approval to someone who is more in touch with how this relates to what is going on in the detector.

jjgomezcadenas (approved):
This is a crucial (and rather straight forward) fix to work with the extra PMT added in Run I (for tagging the Na-22 source). The added test is OK, although there is some redundant code that we may try to fix in the future.

PR #210, opened 2017-05-02
WIP: Use Git Large File Store (LFS) to version binaries.

jacg:
The local copy of the database (`invisible_cities/database/localdb.sqlite3`) is not version controlled. This leads to annoying test failures whenever the database is updated, and makes it difficult to run tests for older revisions. It also burdens the user with the need to download the database manually every now and then. This is all very unsatisfactory.

Enter Git Large File Storage: The idea is that, rather than storing large binary files in the repository itself, you store pointers to those files, which really live on some server. With `git-lsf` enabled, these pointers are automatically replaced with the actual files themselves, but only when you check out a revision that contains one of the files. Similarly, if you modify the file, and push it, `git-lsf` takes care of automatically uploading the file to the server. The file will be automatically downloaded whenever a commit that contains it is checked out in some other repository.

The big advantage of this is that these large files do not weigh down the history of the project, and yet they are still versioned. Specifically, it should avoid problems like #208.

If we are happy with how this works, then we should consider using it for the `h5` files which we use in the tests.

Please do not merge this yet !


jacg:
The first commit in #209 changed the code that matches the change in the DB that is described in its commit message, but the database itself had to be updated by hand. (Though the test which is added in the second commit of #209 still passed even with the old database! @jmbenlloch: Is that expected?)

I have added those two commits to this PR, but I modified the first one to include the modification of the database itself in LFS. Consequently, checking out revisions before/after this commit automatically updates the version of the database in the working directory.

+ No more need to remember to download the latest version of the database.
+ Running earlier tests (starting from this point) is now trivial: no need to hunt down the state of the database as it was at the time: versioning via Git LFS takes care of downloading the corresponding version of the database automatically.

Please play with this and tell me what you think. Please *DO NOT MERGE YET!*

jmbenlloch:
One quick question, I understand that with this new system we should make a new commit (new PR) for each update we make in the database right?

Regarding the test, yes, with the previous database the test should pass. The issue is that a new row was added in one table breaking the output of DataPMT (adding one more sensor). The test should pass with the previous version and with the new one, it will only fail with the new db (having the extra row) if you don't apply the change to the sql query that I put in one of the commits.

jacg:
> with this new system we should make a new commit (new PR) for each update we make in the database right?

Yes, I think that's probably the case. It's not obvious to me why we would want to do anything else.

Git should certainly be aware of which version of the database goes with which version of the code.

Remember, unlike normal git-managed files (where *all* versions that ever appeared in the history of he repo) are downloaded when you clone, only those versions of LFS files that you *check out* will ever be downloaded, and the download will happen at the time of the chekckout.

jacg:
Note that there are storage and bandwidth quotas for LFS on GitHub:

  https://help.github.com/articles/about-storage-and-bandwidth-usage/

> Git LFS is available for every repository on GitHub, whether or not your account or organization has a paid plan.
>
> All personal and organization accounts using Git LFS receive 1 GB of free storage and 1 GB a month of free bandwidth.
>
> If the bandwidth and storage quotas are not enough, you can choose to purchase an additional quota for Git LFS.

The DB is currently at 6M. Which would give us a limit of 160 Travis builds per month.

We also appear to have around 20M of HDF files in `test_data`:

> find . -name "*.h5" | xargs du -ch 
> 2.0M	./invisible_cities/database/test_data/electrons_2500keV_z250_RWF.h5
> 80K	./invisible_cities/database/test_data/KrMC_pmaps.h5
> 1.5M	./invisible_cities/database/test_data/electrons_40keV_z250_RWF.h5
> 208K	./invisible_cities/database/test_data/irene_bug_Kr_ACTIVE_7bar_MCRD.h5
> 1.8M	./invisible_cities/database/test_data/electrons_511keV_z250_RWF.h5
> 12K	./invisible_cities/database/test_data/Kr_dst.h5
> 316K	./invisible_cities/database/test_data/irene_bug_Kr_ACTIVE_7bar_RWF.h5
> 4.6M	./invisible_cities/database/test_data/cwf_electrons_40keV_z250_RWF.h5
> 640K	./invisible_cities/database/test_data/electrons_40keV_z250_MCRD.h5
> 80K	./invisible_cities/database/test_data/NEW_se_mc_reco_pmaps_1evt.h5
> 1.9M	./invisible_cities/database/test_data/electrons_1250keV_z250_RWF.h5
> 6.7M	./invisible_cities/database/test_data/run_2983.h5
> 1.6M	./invisible_cities/tests/electrons_40keV_z250_RWF.h5
> 22M	total

jacg:
I propose that we try this and see how it goes.

+ I have no doubt that it's a huge improvement over the manual DB download, *in principle*.
+ I am worried about the bandwidth restriction, *in practice*.

If the bandwidth proves to be a problem, we can always revert to the old policy.

jacg:
@jmbenlloch I'm trying to find out our bandwidth usage, following the instructions at https://help.github.com/articles/viewing-storage-and-bandwidth-usage-for-an-organization/ but what is described on that page does not match reality as seen through my GitHub account. Maybe it's somehting to do with my permissions within the nextic organization. Can you see whether it works for you?

jmbenlloch:
Yes, I have followed the instructions and this is what I see:

bandwidth

There are no more details, the question mark symbol appearing near the bandwidth just says "Bandwidth quota resets every billing cycle".

I don't where those ~220MB are coming from, but it seems that we would really need more bandwidth for this project. The cost for extra bandwidth is $5 for 50GB (and any multiple of that, i.e. $10 for 100GB, etc.). I think if we don't get the extra it won't be enough.

jacg:
Hmm, 220MB.

There are 6 Travis builds per push. Each one sucks down the 6MB DB from LFS, which makes 36MB per push. So, six pushes of branches containing the LFS-versioned DB would take us to about 220MB.

That's possible, if you take into account the fact that LFS usage by forks is accounted against the parent repo. IOW, when I pushed this feature branch to my fork, that's already 36MB accounted against nextic. When I made a PR, that's another 36MB accounted against nextic. When I pushed your #209 commits on top, that caused Travis to run on my fork aswell as on nextic/master, costing another 72MB. Then I rebased the PR on top of the merged #209, that's another 2 builds, which takes us up to the 6 we need for 220MB.

This makes it look like LFS and Travis aren't really compatible. The trouble being that Travis *clones* the repo *every* build (6 times!). Contrast this to a normal user's usage, where LFS bandwidth is only used when the DB changes.

So it looks like we can't really use this, unless

+ we can use our own LFS server, or
+ drastically reduce Travis' LFS usage

jacg:
I am looking into alternatives, but, in case we consider paying the GitHub LFS bandwidth rates, here is an estimate of the cost. In summary: **USD 10 per month**.

+ Nextic has been running Travis for about 5 months, and has over 550 builds. That's 100 builds per month. Expect this to increase as more developers become more active. Let's assume **200 jobs per month**.

+ Our build matrix runs **6 builds per job**: 3 versions of python x 2 OSes.

+ Each push to a Pull Request triggers *two* builds: one on nextic, one on the developer's fork. In addition to this, developers also trigger builds by pushing to their forks outside of the context of PRs. These fork-builds do not show up in the 550 builds mentioned above, but *are* accounted against nextic in the LFS bandwidth billing calculation. Let's assume a **factor of 3**.

+ GitHub gives 1 GB LFS bandwidth for free every month. Beyond that they charge **USD 0.1 per GB** (in chunks of USD 5 / 50 GB).

The first three points above turn 1MB of LFS-stored data into (1MB x 200 x 6 x 3 =) 3.6GB of LFS-bandwidth per month. The last point turns this into **USD 0.36 per MB per month**.

This PR tracks a 6MB database in LFS, but we also have about 20 MB of `.h5` files that belong in LFS. The sizes of these files are likely to increase. Let's assume 30MB total for now. That gives us USD 0.36/MB x 30 MB = **USD 10 per month**.

In the grand scheme of things, this is peanuts. But given my experience with *los estimados señores administradores*, I suspect that Hell will be a popular skiing destination<sup>[[1]](#hellfreeze)</sup> and pigs a commonly-used physical layer in [IP protocols](https://en.wikipedia.org/wiki/IP_over_Avian_Carriers) <sup>[[2]](#pigsfly)</sup> before this can be approved and made to work without grinding our Travis builds to a halt on the second day of every month (by not paying the GitHub fees on time).

I'm looking for alternatives.

Whatever we do, **the database needs to be versioned**!

<a name="hellfreeze">[1]</a>: When hell freezes over
<a name="pigsfly">[2]</a>: When pigs fly

jjgomezcadenas:
10 Usd is doable. We have a trick that may work with los señores administradores.
J

Enviado desde mi iPhone

jacg:
Even if USD 10 is an acceptable cost in general, it's not really an acceptable cost for *this service*. GitHub changes for LFS bandwidth, but not for bandwidth used for other kinds of traffic. So, if we just stuck these files in our repo without LFS, we would use up much more bandwidth and get it all for free!

Something like [git bigstore](https://github.com/lionheart/git-bigstore) would let us use something like Amazon S3 to host the files, and Amazon would charge USD 0.10 (a factor of 100 less) for that sort of bandwidth.

In short, GitHub's policy is ridiculous, and it is *morally* wrong to agree to it.

I'm looking at other possibilities.

jjgomezcadenas:
Ok!

Enviado desde mi iPhone

jacg:
@jjgomezcadenas Just a reminder that this is becoming more urgent as we are discovering more use-cases for it such as CERES and productions ... and that I do not have the time or the means to investigate possible solutions, including the possibility of running our own LFS server.

jjgomezcadenas:
Here you loud and clear. This is in my priority list for discussion with @jmbenlloch next week

jacg:
And, was any progress been made on this last week?

In the meantime, here's a heads-up to me (and anyone else who cares) to check out https://dataversioncontrol.com/ to see whether it provides an alternative to LFS.

jjgomezcadenas:
Yes, there is progress. J Vicente says that he has a complete scheme to migrate (if we wish) everything to our own gitlab local server including LFS. I thought we could use some time in January, when you are here to discuss in detail.

jacg:
I want to leave an explicit record here of the fact that **once again** (#461) people's time is being wasted becasue this is not implemented yet.

jjgomezcadenas:
And I will take the occasion to state why this is being neglected. No man power. In oder to implement LFS, I need a team of dedicated developers to try and test the system. Right now, I can think only of two people, @jmbenlloch and @gonzaponte. Both of them are already in saturation mode. Thus, unlike some volunteers with enough technical expertise step forward, this issue has to wait until the usual suspect can breath again.

jmbenlloch:
I have done some tests with and here is my update:

Server configuration
--------------------

We can use a gitlab server as an external LFS server. To do that, we need to create a repository that will hold the files, for example: https://nextdesk1.ific.uv.es:4433/jmbenlloch/IC. There are 3 requisites for this repository:

- Gitlab must have LFS enabled (https://docs.gitlab.com/ee/workflow/lfs/lfs_administration.html)
- The repository must have public visibility if we want people to be able to checkout anonymously (this is also useful for travis).
- Every developer will need an account on this server with writing permissions for this repository to push commits (besides their github account).


Client configuration
--------------------

To configure LFS for the repository every user must have git lfs installed on their machines. On debian-like linux can be done with:
```
curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | sudo bash
sudo apt-get install git-lfs
```
For Mac OS X it is possible to install it using brew:
`brew install git-lfs`

For other systems or alternatives downloads check this: https://git-lfs.github.com/

For git lfs to work there are some hooks that must be setup by running the following command:
`git lfs install`

Setting up the external server
------------------------------
We need to add a `.lfsconfig` file with the address of our external lfs server:
```
[lfs]
    url = https://nextdesk1.ific.uv.es:4433/jmbenlloch/IC.git/info/lfs
```
Issues:
- So far we are using a self-signed certificate for the gitlab server, so unless we fix that (maybe https://letsencrypt.org/) we need to add this to `.bashrc`: `export GIT_SSL_NO_VERIFY=1`
- Git LFS does **NOT** support ssh yet (https://github.com/git-lfs/git-lfs/issues/1044). So developers will need to input their gitlab's user&pass after every commit.


Tracking files
--------------
To track files the command is:
`git lfs track '*.h5'`

that will add a new line to `.gitattributes` like `*.h5 filter=lfs diff=lfs merge=lfs -text`

Migration
---------
In principle it is possible to migrate all the binary files in an existing repository using a tool provided by the git-lfs client:
`git lfs migrate import --include="*.h5" --include-ref=refs/heads/master`

That will migrate all the `h5` files from the master branch history and will rewrite all the commits. There is a caveat, though: since we are using an external server for lfs, the file `.lfsconfig` is needed to checkout the files and that file won't be present in older commits, so in principle it wouldn't work unless we put that file somehow in the past commits.

In any case, we could use LFS for the new files and keep the old ones in the repository if there is no simple solution for this issue.

You can find an already-migrated branch here: https://github.com/jmbenlloch/IC-1/tree/lfs-migrate

Travis is working well with that branch as you can see here: https://travis-ci.org/jmbenlloch/IC-1/builds/370972231

PR #212, opened 2017-05-03
Add Correction class

gonzaponte:
An interface for accessing geometrical corrections.


jjgomezcadenas (commented):
The code is elegant and concise but difficult to read (at least for this reviewer). Could you please add comments?

jacg:
Usage examples in the form of tests would also be helpful.

jjgomezcadenas:
@gonzaponte can you check why Travis complains that there is a pending check?

gonzaponte:
I have rebased recently. Perhaps it was stuck in the queue and start to run just now.

jjgomezcadenas (dismissed):
Class Correction accepts a matrix of energies (or previously existing factors) used to produce normalisation factors for the energy. The class comes with extensive testing and has been thoroughly revised by @jacg  and @jjgomezcadenas

invisible_cities/reco/corrections_test.py
```diff
@@ -154,35 +155,40 @@ def test_correction_attributes_2d(toy_data_2d):
...
+                         interp_strategy = interp_strategy)
 
-    # Combine x and y so they are paired
-    X_test = np.repeat(X, Y.size)
-    Y_test = np.tile  (Y, X.size)
-    F_test, U_test = correct(X_test, Y_test)
-    assert_allclose(F_test, F )
-    assert_allclose(U_test, Fu)
+    # create a collection of (x,y) point such that the
+    # x coordinates are stored in X_sample and the y coordinates in Y_sample
```
gonzaponte: Just so I can keep my training as a padawan... why is this better than using np.repeat and np.tile?

jacg:
> Just so I can keep my training as a padawan... why is this better than
> using np.repeat and np.tile?

I assume that you are talking about

    xs = np.repeat(X, Y.size)
    ys = np.tile  (Y, X.size)

as opposed to

    xs = np.array([x for x in X for _ in Y])
    ys = np.array([y for _ in X for y in Y])

Any Python programmer should know that multiple `for`s in list
comprehensions / generator expressions / dict comprehensions / set
comprehensions (and any members of this family that might appear in the
future) result in a combination of all elements in one collection, with
all elements in another collection: a Cartesion product, if you
like. Fewer people are likely to be immediately familiar with
`np.repeat` and `np.tile`.

The second version shows the symmetry between the two cases: it's clear
that `xs` and `ys` are created from exactly the same logical structures,
the only difference between them being that in one case we keep `x` and
ignore `y`, while in the other we keep `y` and ignore `x`. This is far
from clear in the first.

Once you've understood the double-for in the second calculation of `xs`,
you've already understood the structurally-identical calculation of
`ys`; once you've understood the use of `repeat`, well ... you now have
to figure out the use of `tile`.

PR #214, opened 2017-05-04
Remove obsolete coveralls badge

jacg:
We do not use coveralls any more, but the badge is still there on our
main readme page, constantly reporting 'unknown'.

Let's get rid of it!


jjgomezcadenas (approved):
Needed and obvious.

PR #215, opened 2017-05-04
Penthesilea

gonzaponte:
Penthesilea is a new City that saves a DST with collections of hits. It works on PMaps, looping over the time slices, correcting them (in z and x, y) and turning the tracking signal into a collection of points. Each point gets its x, y from some reconstruction algorithm managed by the city, and its z and energy from the pmaps.

Persistence is managed by a new writer, HitCollection_writer.

This is still very much work in progress.


jerenner:
Paola and I have spotted a bug in the current implementation of the construction of hits.  In the select_events method of HitCollectionCity some energy is lost when no clusters are reconstructed for a given slice. To fix this, we need to add that energy to the following slice or, if it happens in the last slice, add it to the previous one.

jacg:
Done in #250.

PR #216, opened 2017-05-04
Run the tests in parallel on Travis

jacg:
The parallel tests keep getting broken, because nobody runs them,
because they are still not worth while: serial execution is still
faster. But we seem to be getting close to the break-even point.

The serial tests should pass if the parallel ones do; the reverse is
not necessarily true. So let's get Travis to help us not break the
parallel tests.


jjgomezcadenas (approved):
God bless Travis

PR #220, opened 2017-05-05
Chi2ndof

gonzaponte:
Fix degrees of freedom in pvalue + normalize chi2


jjgomezcadenas (approved):
Normalization of chi2 to degrees of freedom. Already agreed.

PR #222, opened 2017-05-09
Switch to relative imports

jacg:
The absolute imports we were using were very noisy. Replace them with
the much less tedious relative imports.

Whilst we're at it, we take the opportunity to:

+ Remove unused imports

+ Conform to PEP8's one import per line rule


jjgomezcadenas (commented):
Approve and thank you for taking the time to fix this.

jjgomezcadenas (approved):

PR #223, opened 2017-05-16
New version of Paolina

jacg:
+ Test-driven.

+ Contains more tests. 

+ Functions (and tests) more orthogonal and self-contained.

+ Code tries to express ideas on a higher level.


jjgomezcadenas (approved):
This version of Paolina, by master @jacg is characterised by a test-driven philosophy and very high-level implementation. It also used advanced tools (hypothesis, parameterisations) for tests. A good example for developers! 

The only problem with it is that the enemy does not get confused. It is programmed as if the author could not write a comment under penalty of flogging, thus clear variable names, expressive functions and the like. Too clear!!!

PR #224, opened 2017-05-17
initial barycenter and corona with tests

abotas:
Initial code for barycenter and corona algorithms with tests and a one line addition to reco/params.py  (barycenter de gonzalo)

### DIFF
--- a/invisible_cities/core/exceptions.py
+++ b/invisible_cities/core/exceptions.py
@@ -15,3 +15,10 @@ class NoOutputFile(ICException):
 
 class ParameterNotSet(ICException):
     pass
+
+class SipmEmptyList(ICException):
+    pass
+
+class SipmZeroCharge(ICException):
+    pass
+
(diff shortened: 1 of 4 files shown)


jjgomezcadenas:
Thanks @abotas and @gonzaponte  I see that you use the type Cluster. Where is it defined?

abotas:
It is defined in reco/params.py
https://github.com/abotas/IC-1/blob/ClusterFunc/invisible_cities/reco/params.py

jjgomezcadenas (approved):
This PR submitted barycentre and corona algorithms, two basic pieces of IC software. It has been extensively reviewed (with melones rather than peras). The review has identified a number of issues: a) a few reinventions of the wheel (a favourite of the house) which have been un-invented; b) a few semantic encryption's in names (have decrypted for the benefit of the enemy) and otherwise very minor issues. Concerning the tests, the initial set was quite satisfactory but has been broken down in smaller tests (as in the previous PR). Furthermore hypothesis and parameterisations have been used for the sake of generality. Travis is happy at the time of writing this review.

PR #225, opened 2017-05-17
Add N_PROC parameter to parallel test execution

jacg:
By default pytest and the `*_par` functions in manage.sh use all the
available processors. This is (currently) a sub-optimal choice, and
makes doing anything else on the machine in the meanitime a bit
sluggish.

This change makes `run_tests_par` and `compile_and_test_par` accept an
optional argument which specifies the number of processors to use.

### DIFF
--- a/manage.sh
+++ b/manage.sh
@@ -1,9 +1,8 @@
 #!/usrbin/env bash
 
 COMMAND=$1
-PYTHON_VERSION=$2
+ARGUMENT=$2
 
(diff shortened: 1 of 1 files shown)


jjgomezcadenas (approved):
Useful and straight forward

jacg:
> straight forward

Hmmmm ... it's `bash` and there are no tests for it, which makes me nervous.

PR #229, opened 2017-05-18
Clean commit history

jacg:
The new merge policy (rebase + merge --no-ff) seems to be popular. One thing that was initially overlooked is that, when viewing the log with `-p` (`--first-parent`) only the name on the merge commit will be visible, which is not very useful: it should be the name of the principal author of the branch. Therefore we have adopted a policy that the merger should ensure that the merge commit bears the name of the principal author of the branch.

This leaves us with a few of mis-labelled merge commits in history.

Additionally, there are a few commits with the author is `Your Name`.

This PR contains a branch called `new-master` in which such commiter name errors have been fixed. I propose that we reset `nextic/master` to it. This will imply rebasing our local branches, but

+ most of our local branches are behind master already, and will need rebasing anyway

+ the code in `master` and `new-master` is identical, so it will not create any conflicts

so the cost should be essentially zero.


jacg:
Looking into this a bit more, I have decided that it is not worth the trouble. It would take too much effort to verify that nothing else in the history has been broken, for the benefits that it gives us.

PR #230, opened 2017-05-19
Test speed

### LINKED ISSUE #227: Hypothesis speed
Recent work has shown that hypothesis array generation is very slow (typically a factor 30) wrt to NUMPY random number generation. This brings about the following dilemma: a) speed: NUMPY would be preferable to hypothesis; b) simplification of counter-examples: hypothesis would be preferable to NUMPY.

A possible solution is to use NUMPY by default and switch to hypothesis when tests fail.

jacg:
#227 suggests switching between hypothesis generation and numpy random number generation. The downsides of this are that it would require

+ Writing two generators for every test to which this applies

+ Having to invent a switching mechanism

This PR uses hypothesis profiles to change the number cases that are generated. The default profile that comes with hypothesis uses 200 cases. Here we add a `dev` profile which uses 10 cases (for speedy turnaround during development) and a `hard` profile which use 1000 cases and is used on Travis.

The profile can be chosen on the command line in two ways:

+ `pytest --hypothesis-profile <profile name>`
+ `HYPOTHESIS_PROFILE=<profile name> <general command>`

With our current test suite, we observe the following times

+ `dev`: 25s
+ `default`: 40s
+ `hard`: 90s 

Close #227.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,4 +22,4 @@ install:
   - source manage.sh install ${IC_PYTHON_VERSION}
 
 script:
-  - HYPOTHESIS_PROFILE=hard bash manage.sh run_tests_par
+  - HYPOTHESIS_PROFILE=hard bash manage.sh run_tests_par 4
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,4 +22,4 @@ install:
   - source manage.sh install ${IC_PYTHON_VERSION}
 
 script:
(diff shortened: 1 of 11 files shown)


jmbenlloch (approved):
Important change to increase test's speed. Creating a new `dev` profile with lower samples for hypothesis will make easier for developers to run the test suite while updating the code.

PR #231, opened 2017-05-20
Corrections

jacg:
We propose a simple alternative to the Correction class suggested in #212.

The advantages of this proposal are:

+ Simple and legible implementaiton.
+ Simple and legible tests.
+ Generalizes the `centre` strategy to `index`.
+ Considers only x-y corrections, which are the ones really relevant for matrix corrections.
+ It defines a simple wrapper `FCorrection` with a tests which documents its application for the z-corrections associated with exponential attachment.

@gonzaponte should review this.

### DIFF
--- a/invisible_cities/reco/corrections.py
+++ b/invisible_cities/reco/corrections.py
@@ -22,3 +22,7 @@ class XYCorrection:
 
     def E(self, x, y): return self._Enorm[(x,y)]
     def U(self, x, y): return self._Unorm[(x,y)]
+
+
+def FCorrection(E, z, fn, *args):
+    return E * fn(z, *args)
--- a/invisible_cities/reco/corrections.py
+++ b/invisible_cities/reco/corrections.py
@@ -3,19 +3,22 @@ import numpy as np
 
 class XYCorrection:
 
-    def __init__(self, xs, ys, E, U, strategy):
-        E_max = E.max()
-        dE = E / E_max if strategy else np.ones_like(E)
-        dN = U / E_max
+    def __init__(self, xs, ys, E, U, strategy, ni=None, nj=None):
+        if   strategy == 'max'  : E_ref = E.max()
+        elif strategy == 'index': E_ref = E[ni,nj]
+
+        dE = E / E_ref if strategy else np.ones_like(E)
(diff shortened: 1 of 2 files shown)


gonzaponte:
Interpolation must be an option. We have found that it works worse than getting the correction from the nearest point. Unfortunately, "nearest" is not a valid option for the "kind" argument in inter2p (it is in interp1d), so we need to invent the wheel in this case. I do so in my (updated) proposal quite neatly, I think.

We need the function to manage both arrays and numbers as input. I think the interp2d can do so, but any other implementation should have this property.

Uncertainties are not propagated correctly. It is not critical, but it is not correct either. The correct formula is: u(f) = f · √(u(e)^2/e^2 + u(e_ref)/e_ref^2).

Is the order of `nj, ni` in `E_ref = Et[nj,ni]` intended? We tend to use it the other way around.

We should discuss whether the factors should be E/E_ref or E_ref/E. I prefer the latter because we can manage the case E=0 more elegantly. Notice that in the former you will attempt a division by 0!

jjgomezcadenas:
--I believe interpolation will always work better than taking the nearest neighbour provided the grid is sufficiently fine-grained, but this is of course not guaranteed. Thus, I believe that indeed, it makes sense to define two strategies: "interpolate" and "nearest-neighbhoor", which is nothing but a king of digital interpolation. Thus, we could use a variable: interpolation_strategy, which could define several interpolation strategies, for example: "lineal", "spline", etc. (many of them are defined by SciPy), OR "digital" (meaning nearest neighbour). 

--As for uncertainties you are right.

---Correction factor: it is a matter of taste: E_ref cannot be zero and if it is one should complain, thus nothing wrong with /E_ref provided that one catches the exception if needed. Nothing wrong with the other way around either, really.

---arrays vs numbers and/or extrapolation function: I think that indeed interp2d lets you use both, but I worry about performance (it is explicitly stated in SciPy that there is a faster version). The faster version, indeed, is optimised for rectangular arrays which is our case. Thus, perhaps a better alternative is to fill vectors first and correct them as arrays always. It is after all more efficient, in the NUMPY style.

jacg:
We ended up going with the version in #212.

PR #233, opened 2017-05-24
Add hypothesis profile: noshrink

jacg:
When hypothesis needs to generate complex examples, the generation may
take some considerable time. If the test fails, the shrinking process
generates many more examples of similar complexity in the attempt to
find the simplest case. In IC we have observed wait times of over a
minute!

If, at the end of this process, we get some useful information, then
the wait is worthwhile. If the information we get at the end of the
shrink is that we made a typo in the source code, then the wait was a
complete waste of time.

The `noshrink` profile switches off the shrinking process completely,
by setting `max_shrinks` to zero.

To use it:

    pytest --hypothesis-profile noshrink

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -4,8 +4,10 @@ from hypothesis import settings
 from hypothesis import Verbosity
 
 # In addition to the 'default' profile, we provide
-settings.register_profile("hard" , settings(max_examples=1000))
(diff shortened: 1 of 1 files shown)


jjgomezcadenas (approved):
Needed, in order to work properly with hypothesis

PR #235, opened 2017-05-28
Introduce LifetimeCorrection function

gonzaponte:
This is a thin layer of code so the construction of a lifetime correction function is straightforward. It also helps consistency among users.


jjgomezcadenas (approved):
Useful and straight forward.

PR #236, opened 2017-05-29
Refactor corrections persistency

gonzaponte:
The code for storing the XY corrections contained too much noise. It has been cleaned up.

### DIFF
--- a/invisible_cities/cities/zaira.py
+++ b/invisible_cities/cities/zaira.py
@@ -4,13 +4,12 @@ import time
 import numpy  as np
 import pandas as pd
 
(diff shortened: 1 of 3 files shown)


jjgomezcadenas (approved):
This PR increases the readability and organisation of Zaira and friends. It does not address some of the issues that came out of the Irene review with @jacg  (such as avoiding that the writers open internally the output file to avoid clashes between concurrent writers). Also, some improvements could be made, such as a lifetime correction which is defined in terms of time stamps (this to be discussed and implemented together with @jmbenlloch). However, all those issues can be implemented separately and do not prevent the approval of this clear improvement of the code.

PR #237, opened 2017-05-30
Add functions to read back corrections tables

gonzaponte:


jjgomezcadenas:
Can you provide a test?

jjgomezcadenas (approved):
We need these functions and I can't see anything wrong with them, so I am approving them, but please try and add a test.

jjgomezcadenas:
Why are tests failing now @gonzaponte ? That should not be the case (was not last time I looked, I think
)

gonzaponte:
The test is there and it does not fail. The failure comes from another (unrelated) test that is not properly tuned with flaky. As a matter of fact, I opened PR #241 to fix it.

PR #239, opened 2017-05-30
Refactor cities

jacg:
Major refactoring of the `run` methods of `Irene` and `Isidora`. These methods are no longer monolithic mostrosities, but have been broken down into smaller component functions.

An attempt has been made to avoid side effects, and make the flow of data explicit by passing and returning any values that are needed, rather than modifying `self`.

Some common functionality has been identified and moved into the base classes. 

Various writers have been decoupled, leaving us, for now, with:

+ `pmap_writer`
+ `run_and_event_writer`
+ `mc_track_writer`

As all of these need to write to the same HDF5 file 'simultaneously', they are no longer responisble for opening this file: they receive it as a construction argument.

In principle, all city methods besides `run` are private, but whether or not they are marked as such with underscore, is a bit random. This needs to be fixed in the future, once the set of methods stabilizes.

There is still a long way to go down the path if simplifying and clarifying this part of the code, but it is already a big step in the right direction.


gonzaponte:
Maybe this is a stupid question, but why do we carry the MC writer around when we can do `file.copy_node(old_location, new_location)` (or something similar) with tables?

jacg:
> why do we carry the MC writer around when we can do `file.copy_node(old_location, new_location)` (or something similar) with tables?

Because (I am reliably told by @jjgomezcadenas) sometimes you will want to filter some of these MC events or use them in some other way, so, in general, we need an MC reader and an MC writer.

In the cities where we simply copy the MC data without touching them, we should indeed do it in the most efficient way that PyTables allows. This will have to be done before this PR is done, but there's still plenty of work to be done before we get that far.

jjgomezcadenas:
Yes, this is precisely the reason. You want to copy exactly the same MC events that pass the data filters. In principle, there are no filters in Irene right now, so one could just go and do a blunt copy, but then nothing guarantees that we don't introduce them one day. In any case the solution wouldn't work for any city with filters. So I prefer to keep the option of writing MC events with writer...

gonzaponte:
I've found a couple of details when diving into the code:
- The parameter `n_MAU_sipm = CFP.NMAU_SIPM` is missing in the Irene driver.
- The units in the driver should match those of the `__init__` method.

jjgomezcadenas (approved):
Major refactoring of key cities, to clean up and re-structure code. In addition to the usual unit tests, we have performed a system test, checking that results produced by the re-factored version of code are identical to results produced with previous version. 

DIOMIRA has not yet been refactored, but is better to do it in a different PR, since the city is different (a MC city) from IRENE, ISIDORA and DOROTHEA.

PR #240, opened 2017-05-31
Deal degenerate blob energies

### LINKED ISSUE #238: Blob energies returned as set in paolina.py function blob_energies
Since the blob energies are returned as a set, if they have the same value, the set will contain only 1 element with that value.  Therefore if the user attempts to obtain the energies expecting 2:

```
eblobs = blob_energies(trks[itmax],blob_radius)
iter_eblobs = iter(eblobs)
Eblob1, Eblob2 = next(iter_eblobs), next(iter_eblobs)
```

a `StopIteration` error will occur.  One could get around this by checking the length of the returned set `eblobs` but Paola disagrees because the user should always expect two blobs.  Note that the blob energies being the same value is not necessarily an error - it may happen if the extremes are found close by and therefore include the same energy or of course if the exact energy by chance is found in both extremes.

paolafer:
Fix #238
Two cases:

+ single-voxel
+ degenerate blob energies

### DIFF
--- a/invisible_cities/cities/paolina.py
+++ b/invisible_cities/cities/paolina.py
@@ -7,6 +7,8 @@ from collections import namedtuple
 import numpy    as np
 import networkx as nx
 
+from .. core.exceptions import NoHits
+from .. core.exceptions import NoVoxels
 
 class Hit:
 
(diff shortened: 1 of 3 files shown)


jjgomezcadenas (dismissed):
Solves the situation in which both voxels were identical

jerenner (commented):
We should consider in addition the cases in which no voxels are produced (this would cause an error in find_extrema as it is currently written).

jerenner (approved):

jacg (approved):

PR #241, opened 2017-05-31
Fix occasional failure of corrections tests

gonzaponte:
Sometimes test_corrections_1d or test_corrections_2d failed due to
a bad fit. It has been solved by using non-static fixtures and
modifying the default flaky parameters.


jjgomezcadenas (approved):
This was much needed, the failing test was a pain.

PR #243, opened 2017-06-01
Command line interface to cities

jacg:
+ The command line interface for city use has been changed from 

    `python invisible_cities/cities/whatever_city.py -c <config file> <options>`

   to

    `city whatever_city -c <config file> <options>`

  As a side-effect, problems with relative imports on the command line, are fixed.

+ Genuine command tests have been added.

  The tests which claimed to be command line tests, didn't really exercise the command line at all. (That is why we didn't notice when relative imports broke the command line interface.

  The pretense and complication of using the command lines in the latter, has been removed. Consequently ...

  + Many obsolete test fixtures have been removed.

  + The config file structure and default value specifications have been removed from the base cities.

+ The config files in `invisible_cities/config/` have been cleaned up. The new command line tests use them (and uncovered a few bugs in the config files). They should now serve as more useful documentation for the user.


jmbenlloch (approved):
This PR fixes an important problem, cities were not working when executed from command line. Also it updates tests and remove old code making everything clearer.

PR #245, opened 2017-06-07
Refactor diomira

jacg:
Follow-up to the Cities refactor: the Diomira run method has been broken down into the structure established there:

+  `run()`
+ `_file_loop()` (previously known as `_main_loop()`
+ `_event_loop()`


jjgomezcadenas (approved):
This refactoring follows up the work started with Irene, Isidora, Dorothea and Zaira. It has been developed with Peras, and passes all tests. Experimental production will be required.

PR #246, opened 2017-06-07
Add minmax type and use it in S12 and Pmap Params

jacg:
The type `minmax` (a named tuple with attributes `min` and `max`) is
used to store pairs of values which together place limits on some
quantity.


jjgomezcadenas (approved):
useful named tuple to name variables which have a minimum and a maximum (defining typically a cut)

PR #248, opened 2017-06-07
Cecilia

jacg:
The purpose of Cecila is to implement the emulation of the FPGA
trigger on MC data, aswell as trigger selection.

Cecila reads RWFs generated by Diomira, filters them by the trigger
response and writes them out.


jjgomezcadenas (approved):
Cecilia has been eaten with Peras and not little pain by @jacg and @jjgomezcadenas, starting from @paolafer version. The city has been totally re-factored to conform to other cities. A couple of tests have been added. There are not enough tests yet for this city, those are to be added by @paolafer. In addition, we need test productions of MC data to check data integrity (@jmbenlloch, @gonzaponte, @paolafer).

PR #250, opened 2017-06-09
Penthesilea

jacg:
The functionality of Penthesilea (at this moment) is

+ filter PMAPs
+ compute a hit collection from from remaining PMAPs
+ write to disk

A number of simplifications and some refactoring has been done WRT the
original class proposed by Gonzalo:

+ Penthesilea no longer makes energy corrections, which are deferred
  further down the reconstruction line.

+ Hit classes have been simplified

+ The run method has been made to conform to current standards, in
  particular Dorothea and Penthesilea have essentially identical
  run structure.

Additionally, some pre-existing code has been modified

+ `barycenter` now returns a list containing just one Cluster ()

+ The quick-fix implemented by Gonzalo (`select_si_slice`), which
  addressed the occasional mismatch between S2 and Si slices, is
  currently bypassed. The reason for this is twofold

  - to allow us to progress with implementing Penthesilea without
    dealing with this distraction

  - to allow us to understand the problem being addressed by
    `select_si_slice` further down the line, and decide whether this
    is the right approach: the point is that we *want* the code to
    fail, to highlight the situations in which the issue arises.

+ The `pos` attribute of `Cluster` is now filled with an `XY` object
  instead of a plain 2D numpy array.

### DIFF
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -31,6 +31,7 @@ from ..io.dst_io          import KrEvent
 from ..reco               import peak_functions_c as cpf
 from ..reco               import peak_functions   as pf
 from ..reco               import pmaps_functions  as pmp
+from ..reco               import dst_functions    as dstf
 from ..reco               import tbl_functions    as tbf
 from ..reco               import wfm_functions    as wfm
 from ..reco.params        import SensorParams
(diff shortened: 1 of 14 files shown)


jjgomezcadenas (approved):
Extensive (and exhaustive) work done with Peras and cafe con leche. Approved so that we have a baseline for further development.

PR #254, opened 2017-06-19
Penthesilea plus

jjgomezcadenas:
This PR proposes a number of improvements in the code, originated by the need to expand Penthesilea so that it can write also Tracks (and not only HitColelctions as of today).

Specifically:
1. The code computing Tracks (paolina) was placed in the wrong place (cities) and has been moved to Reco.
2. The code used a class Hit which was unncessary (it was a copy of Voxel and it could interfere with the class Hit in the event model).
3. The filtering in Penthesilea needed cleanup 
4. Some dead code in base cities has been identified and removed.

On the other hand, this PR also proposes a first attempt to define a decent data model. To that end:

1. A module ic_types containing "IC specific types" have been added to core. It includes types "xy" and "minmax"
2. A module event_model has been added to reco. All transient and persistent classes defining the data model have been moved there. The clases themselves have been re-factor and new transient classes (Cluster, Waveform, S12) have been added. 
3. Writers in io have been refactor to eliminate some of the duplications (eg, in creation of tables) and indexing of tables have been added (this is an essential feature that was added to PMAP tables but was missing in other tables). 

Finally, module peak_functions has been refactor, separating functions in public and private (those starting by  _). Private functions are used by tests, but should not be in the API (e.g, there are faster cython functions available). 

This is work in progress. The goal is to complete a first version of the event model (almost done) and to prepare the code to write tracks. 

Several tests have been added, but more tests are expected from @gonzaponte


gonzaponte:
Attributes of some classes (`xy` in ic_types.py for instance) are public and, at the same time, there is an interface to access those attributes. Is this intended, @jjgomezcadenas?

jjgomezcadenas:
Yes it is intended. You can access xy attributes as .x or .y but also as .X, .Y and .pos. There is no repetition (.x is not the same than .X)

jjgomezcadenas:
The PR was closed by a glitch. I fetched next/master and then did a reset of the branch to nexic (which was ages behind my branch) rather than a rebase. Fortunately, @gonzaponte had a copy of the branch. The PR is open again

gonzaponte (commented):
Some comments:
- I would use a more descriptive name in `S12L`. `SignalType`, for example.

- I don't see the need for the `_S12` to be "private", why not `S12`?

jacg (approved):
A miscellaneous collection of improvements. It's not worth our time to try to split it up into coherent chunks at this point. Let's just merge it and proceed to build on top of it.

PR #255, opened 2017-06-20
Get first event number from input filename

jmbenlloch:
We use the structure of the filename to compute first event number. Filename needs to have a particular structure including software version, file number and number of events. For more details look at the regular expression.

Sample filename: dst_NEXT_v0_08_09_Co56_INTERNALPORTANODE_74_0_7bar_MCRD_10000.root.h5

If the filename does not match the pattern (for files not in the official production), gives 0 as first event number.

@jjgomezcadenas what do you think?

### DIFF
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -185,13 +185,7 @@ class City:
 
     @staticmethod
     def event_number_from_input_file_name(filename):
-        file_base_name = filename.split('/')[-1]
-        base_hash = hash(file_base_name)
-        # Something, somewhere, is preventing us from using the full
-        # 64 bit space, and limiting us to 32 bits. TODO find and
-        # eliminate it.
-        limited_hash = base_hash % int(1e9)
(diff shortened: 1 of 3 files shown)


jjgomezcadenas:
Please submit the proposed function as a "free function" (place it in tbl_functions) and then call the free function from the base_city providing the interface to the concrete functions. Look  at my branch penthesilea-plus for examples (in fact you could probably rebase to it).

jmbenlloch:
> Please submit the proposed function as a "free function" (place it in tbl_functions) and then call the free function from the base_city providing the interface to the concrete functions.

I have already move the function to tbl_functions.

> in fact you could probably rebase to it

If I do that we won't be able to approve this PR until we approve your PR before.

jjgomezcadenas (approved):
This is a useful function that allows a simple numbering scheme (possibly faster and simpler than the hash-based scheme). It comes with a test.

jjgomezcadenas:
@jmbenlloch you can merge...

PR #256, opened 2017-06-21
Lifetimefunctions

gonzaponte:
Extend Fcorrection to multiple variables and add radial-dependent lifetime correction function.


jjgomezcadenas (approved):
Straight forward functions, with test.

PR #257, opened 2017-06-24
fix bug in hist2d_profile

gonzaponte:
There was a confusion between two variables that was visible only when the number of bins in the two coordinates was different. This PR fixes it.

### DIFF
--- a/invisible_cities/icaro/hst_functions.py
+++ b/invisible_cities/icaro/hst_functions.py
@@ -102,22 +102,13 @@ def hist2d_profile(x, y, z, nbinx, nbiny, xrange, yrange, **kwargs):
     """
     Create a profile 2d of the data and plot it as an histogram.
     """
-    if "new_figure" in kwargs:
-        del kwargs["new_figure"]
-    else:
-        plt.figure()
 
     x, y, z, ze = fitf.profileXY(x, y, z, nbinx, nbiny, xrange, yrange)
-    plot_output = display_matrix(x, y, z)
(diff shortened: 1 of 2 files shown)


gonzaponte:
I can't reproduce the error in my computer. I think it has something to do with matplotlib. Ideas?

jjgomezcadenas:
The problem is related with the back-end of matplotlib. In short, **do not involve matplotlib in tests** at least for the time being. The best approach here would be to decouple the matplotlib part from the math part, and just test the math part.

gonzaponte:
The problem in this case is that the test uses the output of the matplotlib part... The workflow is:
1. Do a profileXY (this function is already tested)
2. Reorganize the output of this function
3. Input the reorganized data to hist2d (which is just a wrapper of matplotlib's hist2d)

And the test aims to:
1. Check that the output of profileXY makes sense. I am trying to check that the function has been coded correctly, independently of profileXY working fine. I may have entered into the reinventing-the-wheel zone, though.

2. Check that the output of hist2d makes sense. We actually would only like to check that the data is input properly (provided that the function is already tested), but we can't do that that so we check the output instead.

jjgomezcadenas:
Well, we have to find away to get rid of matplotlib in tests.

jjgomezcadenas:
What about this @gonzaponte ?

gonzaponte:
I have plans to split the function in two. I can do it this week, I think.

gonzaponte:
I think this is a bit obsolete. I will create a new PR with many miscellaneous functions for icaro that include issues from this PR.

PR #258, opened 2017-06-27
Drop support for Python 3.5

jacg:
We keep getting annoying test failures because `path.join` cannot handle paths (!) in Python 3.5.

The solution is simple (convert the path to a string), but at this stage it's difficult to see what benefit we get in return for making our code uglier in this way.

It's time to move on.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ matrix:
   - os: osx
 
 env:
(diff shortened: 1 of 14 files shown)


jmbenlloch (approved):
Very convenient update to the code, there is no need to keep version 3.5.

Issue #203, opened 2017-04-07
Cities inheritance structure


gonzaponte:
I have noticed something in the inheritance chain of the base cities that may not be structurally correct.

All base cities inherit from City, which seems to be designed to work as the general manager which controls the job-related parameters (such as IO files, number of events, etc.), but it also holds information from the database. However, base cities do not care about IO at all as they simply provide tools and interfaces to handle the data. They may require, however, the database info, which is the reason they derive from City, I think. The merging of database- and IO-related stuff in City causes a lot of noise in the base subcities (recursively passing several parameters they don't care about to the constructors).

My proposal is to split City into two separate abstract cities, namely DatabaseCity (for run-related stuff) and IOCity (for IO stuff, obviously). This way, base cities can derive from DatabaseCity only and concrete implementations (Irene, etc.) can subclass **both** DatabaseCity and IOCity because these are the ones that actually use IO operations.

Hope it is clear, I have experienced some pain trying to explain myself properly.

jjgomezcadenas:
I like this. What do you think @jacg?

jacg:
Forming an opinion worth sharing would take far more time than I have available at present.

jjgomezcadenas:
My opinion is then to postpone the split of the cities until I have some time to discuss with @jacg, since we have more urgent things to do, in particular to review Dorothea. Agreed, @gonzaponte?

gonzaponte:
> Agreed, @gonzaponte?

Sure, I opened this issue just to start the discussion so that @jacg can meditate (because Jedis meditate) about it!

jjgomezcadenas:
I believe most of this issues are addressed now.

PR #259, opened 2017-06-28
Eliminate drivers and refactor configuration

### LINKED ISSUE #143: Cities init method arguments
We need to discuss whether it is more convenient for the cities constructors to list just their own arguments and take the ones belonging to the classes they inherit from as keywords or to list them all and make it more explicit.

jacg:
1. The driver functions for each city (the city name in uppercase) have been removed. They are all replaced with a single `drive` classmethod in `City`. This simplifies the code considerably and removes the copy-n-paste mess that has been growing an each new city was added.

2. The city constructors no longer take and pass arguments explicitly by name, this is now achieved with `**kwds`. This closes #143 and has a number of consequences

   + Significant noise reduction in constructor definitions.
   + The config files become the sole source of
      - values of the parameters
      - documentation of what parameters are needed by each kind of city

3. The config files are now written in Python syntax and parsed with Python's built-in parser.  Our custom config parser for our ad-hoc config syntax has been thrown away and replaced with Python's parser. Do not re-invent the wheel.

4. The ability to inculde config files within other config files has been added, allowing the use of hierarchical configurations. Thus the default configuration of some base city does not need to be repeated in each concrete city's config file, as in can be included.

5. A new configuration printer has been added with the key abilities to

   + report the config file which provides any given value in an active configuration
   + highlight cases where a value set in one config file is overridden by a setting in another one.

   Improvements to the config printer are envisaged:

   + Displaying the comments from the config files.
   + Displaying the units used in the config files.

6. The mandatory `-c` (config file) command line option has been turned into a positional argument.

(#203 has not been addressed at all here.)


jjgomezcadenas (approved):
A major advance in the code. Good and well!

PR #260, opened 2017-06-28
Rebin waveform

abotas:
These changes primarily address a bug found in find_S12 and rebin_waveform in peak_functions.py/.pyd. We now:
1) avoid discarding photoelectrons suppressed by CalibratedCity.csum_zs before finding S12s. We give find_S12/rebin_waveform to access to the indices of the above-threshold waveform, but also access to the unsuppressed waveform. 
2) ensure S2 time bins can be properly aligned with corresponding S2Si time bins. 

These changes are implemented in python and cython in parallel (peak_functions). They greatly simplified find_S12, but required a much more careful rebinning in rebin_waveform.

Slight modifications were made to irene.py, base_cities.py, pmap_io_test.py to coordinate sending slightly different arguments to find_S12 and rebin_waveform. 

Additional tests were added to peak_functions_tests.py to cover the new changes.

Some insignificant changes were made to other test.py files during a rebase on nextic/master.


jjgomezcadenas (approved):
This development has been done while eating peras (JJ and Alex). Numerous tests show that the long-standing rebinning-bug is now dead for good (toca madera).

jmbenlloch:
There were some commits on this branch from @jacg previous (and already merged) PR. I have updated the branch to include only the new commits and not those already in master.

Tests passed locally, so if travis is also ok I will merge afterwards.

Issue #262, opened 2017-06-30
Too many things called S2


abotas:
If no one tells me not to, I will make a quick, tiny PR to solve this:

We have a variable in the code, S2, that means different things in different places. These idiosyncrasies are not a big deal I tried to ignore them, but now as reuse IC functions/objects the ambiguity infects the code I am developing. 

1) In the majority of places in the code, we call S2: a dictionary mapping peak numbers (keys) to their respective peaks, S2 times and S2 pmt energies np.arrays.

2) However, we in reco.params.py, in PMaps, we call S2 a a dictionary mapping event number to the dictionary described in 1). 

3) And, in reco.peak_functions, in sipm_s2_dict() and sipm_s2(), we call S2 one peak containing the times and pmt energies, from one event. (In these two functions the S2 described in 1) is called S2d).

I SUGGEST:
a) in reco.params PMaps should contain S1s, S2s, and S2Sis instead of S1, S2, Si.

b) we distinguish between the dictionary containing the various S2 peaks for an event and a single S2 peak in an event. There are a couple different ways we can do this. 

Any input, particularly suggestions as to how to implement b) would be very much appreciated. I'm starting to feel bad for our pobres enemigos. They are hopelessly lost now...

jjgomezcadenas:
Agreed, pobres enemigos!

To complicate matters further, we also call S2 (in event_model.py) the high level interface!!!
So a way out of it is:

PMaps as you should S1s, S2s, S2Sis.
S1, S2 and S2Si ---> instances of high level class (in event_model.py)
dictionary of peak number (key) to peak --> S1d, S2d, S2Sid. 
The individual elements of the S1/S2 should be called peaks, where needed.

Does this help??

abotas:
Yes, sounds good. I put this in a tiny PR

jjgomezcadenas:
All this is addressed in #265

PR #265, opened 2017-07-10
S12 devel

jjgomezcadenas:
This is a major re-factoring of code related with PMAPS. The main changes are:

1) high level objects have been defined for S1, S2 and S2Si. These are classes with persistent capabilities that provide a high level interface to the underlying objects hitherto called by the same name. 

2) The notation has been clarified, so that now: s1, s2 and s2si are instances of classes S1, S2 and S2Si (we can refer to any of these high level objects as si). Instead,  s1d, s2d and s2sid are the underlying dictionaries previously used to represent the si. Thus, a s1d is  a {peak_no: [t, E]}, where t and E are np. arrays. The sid are still accessible through the si (e.g, s1.s1d and so on). However the si provide a better interface, si.peaks where peaks = {peak_no: Peak(t,E)} and Peak itself is a class that provides a high level interface to the peak waveform. 

3) When PMAPS are read (e.g, function load_pmaps) the objects returned are called si_dict (e.g, s1_dict, s2_dict and s2si_dict). These are: s1_dict = {event_number: s1} (and so on).

4) PMAPS are created by a set of functions in peak_functions_c: namely:
**find_s1**, **find_s2** and  **find_s2si**, which provide a cleaner interface than previous version. The free functions are accessed from the method **pmaps** in base_cities:

def pmaps(self, s1_indx, s2_indx, csum, sipmzs):
        s1 = cpf.find_s1(csum, s1_indx, **self.s1_params._asdict())
        s1 = cpf.correct_s1_ene(s1.s1d, csum)
        s2 = cpf.find_s2(csum, s2_indx, **self.s2_params._asdict())
        s2si = cpf.find_s2si(sipmzs, s2.s2d, thr = self.thr_sipm_s2)
        return s1, s2, s2si

which returns now instances of the high level objects. Analogously, the PMAPS are read back by function **load_pmaps** (now in pmap_io module in io) which in turn call free functions (living in pmaps_functions_c): 

def load_pmaps(PMP_file_name):
    """Read the PMAP file and return transient PMAP rep."""

    s1t, s2t, s2sit = read_pmaps(PMP_file_name)
    s1_dict              = df_to_s1_dict(s1t)
    s2_dict              = df_to_s2_dict(s2t)
    s2si_dict            = df_to_s2si_dict(s2t, s2sit)
    return s1_dict, s2_dict, s2si_dict

The objects returned by **load_pmaps** are si_dict = {event_n0:si}.

5) The high-level classes themselves are created in pmaps.pyx which now lives in a new package called evm (from event model). 

6) The high level pmap classes and all the functions that create and manipulate them are now cython functions, in order to optimize performance and type safety. Previous python functions are kept in some cases (for testing purposes) but removed from the API (eg, by changing their name from name_of_function to _name_of_function). 

7) Tests have been revised and often refactored to describe the changes above. 

8) cities involving pmaps as well as other relevant pieces of code (e.g, s12_selector) have been modified to use the high level PMAP interface rather than the old representation.


jmbenlloch:
We have added a new commit changing variable `event_no` to `peak_no` in `evm/pmaps.pyx`. If we understand the code and your comments correctly the previous name was wrong.

invisible_cities/evm/pmaps.pyx
```diff
@@ -0,0 +1,272 @@
...
+    The s2sid is a dictionary {peak:{nsipm:[E]}}
+    """
+
+    def __init__(self, s2d, s2sid):
+        """where:
+           s2d   = {peak_number:[[t], [E]]}
+           s2sid = {peak:{nsipm:[Q]}}
+           Q is the energy in each SiPM sample
+        """
+        S2.__init__(self, s2d)
```
jmbenlloch, line 164: We don't understand why there are two copies of `s2d`, one in the object `s2` and another here in the `s2si`. If we are going to have `s1`, `s2` and `s2si`, this information seems to be redundant.

invisible_cities/cities/base_cities.py
```diff
@@ -422,17 +414,20 @@ def __init__(self, **kwds):
...
+    #     if self.rebin <= 1:
+    #         return s2d, sid
+    #
+    #     S2_rebin = {}
+    #     Si_rebin = {}
+    #     for peak in S2:
+    #         t, e, sipms = cpf.rebin_S2(S2[peak][0], S2[peak][1], Si[peak], self.rebin)
+    #         S2_rebin[peak] = Peak(t, e)
+    #         Si_rebin[peak] = sipms
+    #     return S2_rebin, Si_rebin
```
abotas: I think we can delete all of this. This code calls a function, `cpf.rebin_S2`, that has not existed in nextic/master for months. `cpf.rebin_S2` seems to be a precursor to the old version of `rebin_waveform` that I recently changed.

invisible_cities/cities/dorothea.py
```diff
@@ -133,9 +144,56 @@ def _create_kr_event(self, evt_number, evt_time, S1, S2, Si):
...
-            dt  = s2time - evt.S1t[0] if len(evt.S1t) > 0 else -1e3
-            dt *= units.ns / units.mus
+            dt  = evt.S2t[peak_no] - evt.S1t[0]
+            dt  *= units.ns / units.mus
             evt.DT   .append(dt)
             evt.Z    .append(dt * units.mus * self.drift_v)
 
         return evt
+
+    # def _create_kr_event(self, evt_number, evt_time, S1, S2, Si):
```
jmbenlloch: Is there any reason for this to be here? You have written a new version of the function and the old one is commented.

invisible_cities/cities/penthesilea.py
```diff
@@ -72,48 +72,58 @@ def _file_loop(self, write_hits):
...
+        # as reference.
 
-        S2, Si = self.rebin_s2(S2, Si)
+        # peak  = S1.peak_waveform(0)
+        # t, e = next(iter(S1.values()))
+        # S1t  = t[np.argmax(e)]
+
+        s1_t = s1.peak_waveform(0).tpeak
+
+        #S2, Si = self.rebin_s2(S2, Si) TODO REVISE
```
abotas, line 118: Again, I think this is the same comment: rebin_s2 calls cpf.rebin_S2 which has not existed in our master repository for months.  

If the idea is that penthesilea wants to rebin the s2 and s2si such that they have time bins larger than those saved in the pmaps, we can write a tiny function to do that. It should be very simple since now the `rebin_waveform` function returns clean usually 1 mus time bins. If not, we can just delete this.

invisible_cities/evm/pmaps.pyx
jjgomezcadenas: It is redundant but nice. Consider the case in which you want to retrieve the waveform of a given SiPM. The underlying s2sid object has only information on E but not on time (since this is own by the s2d object). By keeping a reference to the s2d object one can access the time waveform (and also the energy waveform) of the s2 *directly*, and this is provided by the high level interface without the need for gimmnastics (e.g, go and search for the corresponding information in the s2 object). At the same time, by keeping the reference to s2, we do not need to save the time waveform for each SiPM in the S2Si, thus saving space and speed access.

invisible_cities/cities/base_cities.py
jjgomezcadenas: Yes, but there is a point in having the possibility of re-rebinning S2 and S2Si at Dorothea level. Typically you have rebinned the s2 and s2si in bins of 1 mus when creating the pmaps. This bin is often "too thin" (it corresponds to 1 mm) and when creating hits and tracks we will probably increase to 5 mus or even 10 mus. Thus we must be able to rebin by a factor 5 or even 10 at this level. The idea is to call from here @abotas new rebinning functions. Ale, can you do this?

invisible_cities/cities/dorothea.py
jjgomezcadenas: Yes, the old version was commented while I was trying to pass tests... now it can be removed.

invisible_cities/cities/penthesilea.py
jjgomezcadenas: Same comments as before, I need you to include rebinning (new style here)

invisible_cities/io/pmap_io_test.py
```diff
@@ -101,144 +174,149 @@ def test_pmap_writer(config_tmpdir, s12_dataframe_converted, s2si_dataframe_conv
...
-        s2si = XS2SiL[event]
-        if s2si:
-            sip = S2SiL[event]
-            for peak_number in sip:
-                sipm = sip[peak_number]
-                sipm2 = s2si[peak_number]
-                for nsipm in sipm:
-                    np.testing.assert_allclose(sipm[nsipm], sipm2[nsipm])
+# @mark.slow
+# def test_pmap_electrons_40keV(config_tmpdir):
```
jmbenlloch: Again, if this is useful shouldn't be commented, if not maybe we should remove it. What do you think @jjgomezcadenas ?

invisible_cities/reco/event_model.py
```diff
@@ -54,187 +54,187 @@ def __str__(self):
...
-
-        s2b = ["peak number = {}: sipm number = {}\n    sipm waveform (zs) = {}".format(peak_number, sipm_number, self.sipm_waveform_zs(peak_number, sipm_number))
-               for peak_number in self.peaks
-               for sipm_number in self.sipms_in_peak(peak_number)
-               if len(self.sipms_in_peak(peak_number)) > 0]
-
-        return s + '\n'.join(s2b) + "\n" + "=" * 80
-
-    __repr__ = __str__
+# class Waveform:
```
jmbenlloch: Why is all this commented?

invisible_cities/evm/pmaps.pyx
```diff
@@ -0,0 +1,272 @@
...
+        return  s + ''.join(s2)
+
+    def __repr__(self):
+        return self.__str__()
+
+cdef class S2(S12):
+    def __init__(self, s2d):
+        self.s2d = s2d
+        super(S2, self).__init__(s2d)
+
```
abotas, line 140: One minor issue is that if we want a function `manipulate_si` or `manipulate_s12` that interacts either the dictionary in s1 or s2, we have a small problem because the function does not know whether to use si.s1d or si.s2d, right?

Maybe we use s1.d and s2.d to access their dictionaries, which is also more concise. 

In s2si we can leave things as they are since it has two dictionaries. And, since if a function needs to interact with s2si, it needs to know that it is an s2si since s2sid has a different format than s1d and s2d. What do you think @jjgomezcadenas ?

jmbenlloch (approved):
Important changes to add clarity to the pmaps API. This closes #262.

PR #268, opened 2017-07-11
Fix diomira noisy sipms

gonzaponte:
Some sensors with weird noise pdfs were causing bad responses in the tracking plane. This PR fixes it by skipping masked sensors when simulating the SiPM noise. On top of that, a test has been added to ensure this behavior is under control.


jmbenlloch (approved):
Important update for MC productions, now we take into account which channels are masked.

PR #269, opened 2017-07-11
Reduce range of input data for the shift_to_bin_centers test

gonzaponte:
Reduce range of input data for the shift_to_bin_centers test
This caused the test to fail from time to time.


jmbenlloch (approved):
This test was failing from time to time due to the extreme numerical precision required, now we have more sensible values that won't break the test.

PR #273, opened 2017-07-19
New command to download dev version of database

jmbenlloch:
This PR adds the functionality of downloading a dev version of the
central database. This is needed to introduce and test changes in the
DB without compromising the production version. After we have seen
that the changes are good, the could be propagated to the production system.


jjgomezcadenas (approved):
This is a needed addition, so that we can work with a stable and a development version of the data base. The code is straight-forward and has been tested to work.

PR #274, opened 2017-07-20
cities-debug

abotas:
S2Si sets its peaks to be those present in s2d. This is unacceptable when the sipms do not have signal during the one or more of the peaks in s2d. this has been fixed by `check_s2d_and_s2sid_share_peaks` in `pmaps.pyx`.


jjgomezcadenas (dismissed):
This was indeed a needed fix (which indeed reveals a strong feature of the new interfaces, eg., the capability to ensure consistency between peaks and sipms). The code is clear and is exercised by a test.

jmbenlloch (approved):
This PR fixes some issues with peak numbers in several places and a bug in find_s12.

PR #275, opened 2017-07-21
Daemons

jjgomezcadenas:
Add daemons to IC

Daemons (see Northern Lights) are people's (external) souls in Lyra's
world. In IC, daemons are slave (or companions) objects that can be
invoked from cities.

Consider for example the problem to add monitoring histograms to a
given city. One solution is to write all the monitoring code in the
city itself. Master forbids! The city becomes
immediately cumbersome, and the code booking and filling histograms
proliferates quickly. Plus, monitoring is typically not stable, since
the critical histograms tend to vary with time.

The alternative is that the city calls a monitoring script (ms). At init
level, the city will call the init level of the ms (say to book
histograms). Then, at run level the city calls the run method of the
ms (fill histograms) and finally, at at the end level the city calls
the end method of the ms (save histograms to disk).

The scheme is quite general, and applies to any type of script that a
given city wants to invoke to delegate any task. Thus, a daemon is
defined as an object with an __init__() method, a run method and an
end method.

Daemons live in the new daemons directory (package) in IC. In this
PR we add to IC the functionality to handle daemons.

1. The config files for cities can now define the presence of daemons by adding a variable called (surprise!) daemons which is set to a list of names (see irene_example_daemon_config.conf)

2. The method drive in base city parses the names and calls function invoke_daemons() to return instances of the corresponding classes, which must be defined in the daemon directory (see examples lyra.py and asriel.py). 

3. Also in drive, the instances of the daemons are set as attributes of the cities, which can now call them. 

4. A new method, in base_cities, called end() calls the end() method of the daemons. The run method of the daemons are called by the concrete cities.

### DIFF
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -64,6 +64,7 @@ from ..sierpe                   import blr
 from ..sierpe                   import fee as FE
 from .. types.ic_types          import Counter
 from .. types.ic_types          import NN
+from .. daemons.idaemon         import invoke_daemon
 
 def merge_two_dicts(a,b):
     return {**a, **b}
@@ -125,6 +126,17 @@ class City:
             self.writers = self.get_writers(h5out)
             self.file_loop()
 
+
+
+    def end(self):
+        """The (base) end method of a city does the following chores:
+        1. prints the counter dictionary
+        2. calls the end method of the daemons if they have been invoked.
+        """
+        if hasattr(self, 'daemons'):
+            for deamon in self.daemons:
+                deamon.end()
+
         return self.cnt
 
     def display_IO_info(self):
@@ -159,16 +171,25 @@ class City:
         if opts.print_config_only:
             return
         instance = cls(**conf.as_dict)
+
+        # set the deamons
+
+        if 'daemons' in conf.as_dict:
+            d_list_name = conf.as_dict['daemons']
+            instance.daemons = list(map(invoke_daemon, d_list_name))
         instance.go()
+        instance.end()
 
     def go(self):
         t0 = time()
-        cnt = self.run()
+        self.run()
         t1 = time()
         dt = t1 - t0
-        n_events = cnt.counter_value('n_events_tot')
-        print("run {} evts in {} s, time/event = {}".format(n_events, dt, dt/n_events))
-        print(cnt)
+        n_events = self.cnt.counter_value('n_events_tot')
+        print("run {} evts in {} s, time/event = {}".format(n_events,
+                                                            dt,
+                                                            dt/n_events))
+
 
     def set_up_database(self):
         DataPMT       = load_db.DataPMT (self.run_number)
--- a/invisible_cities/daemons/idaemon.py
+++ b/invisible_cities/daemons/idaemon.py
@@ -1,0 +1,15 @@
+from importlib import import_module
+import traceback
+
+
+def invoke_daemon(daemon_name):
+    """Takes a daemon name and returns an instance of the daemon"""
+    try:
+        module_name = 'invisible_cities.daemons.' + daemon_name
+        daemon_class  = getattr(import_module(module_name),
+                                          daemon_name.capitalize())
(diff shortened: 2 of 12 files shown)


jmbenlloch (approved):
This PR addresses the need of making monitor plots in the cities. It provides the skeleton that will be use by all particular monitors. Done with peras with me.

jacg:
While the choice of name 'daemon' is very nice from the literary point of view, given that the word 'daemon' has quite a specifc meaning in computing (and I'm struggling to see how the usage here matches it) perhaps it wasn't the best choice ...

I still haven't had the time to read and digest what it is that these daemons give us: for now I just need to know which city is expected to know about them. I take it that the answer is "all of them".

jjgomezcadenas:
Daemons are intented as auxiliary processes for cities. For example a deamon could take care of control histograms related with the process that the city operates it, without messing arround with the city itself. 

I think in that sense the word daemon is not absolutely wrong.

invisible_cities/cities/base_cities.py
```diff
@@ -159,16 +171,25 @@ def drive(cls, argv):
         if opts.print_config_only:
             return
         instance = cls(**conf.as_dict)
+
+        # set the deamons
+
+        if 'daemons' in conf.as_dict:
+            d_list_name = conf.as_dict['daemons']
```
jacg, line 178: Whatever was being smoked while writing this, I want some!

Fixes coming up, when I get a bit more spare time ...

invisible_cities/daemons/idaemon.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def invoke_daemon(daemon_name):
+    """Takes a daemon name and returns an instance of the daemon"""
+    try:
+        module_name = 'invisible_cities.daemons.' + daemon_name
+        daemon_class  = getattr(import_module(module_name),
+                                          daemon_name.capitalize())
+    except ModuleNotFoundError:
+        traceback.print_exc()
```
jacg, line 12: Why do we prefer use of an unknown daemon name to continue running the program while polluting standard output?

Surely the correct behaviour would be to fail, because the configuration contains nonsense.

In other words, the `ModuleNotFound` error should *not* be caught here.

invisible_cities/daemons/idaemon.py
```diff
@@ -0,0 +1,15 @@
+from importlib import import_module
+import traceback
+
+
+def invoke_daemon(daemon_name):
+    """Takes a daemon name and returns an instance of the daemon"""
```
jacg, line 6: This docstring contradicts the name of the function. Looking at the code, it seems that the docstring is correct, and the function name is misleading. The function should have a name along the lines of `get_daemon`.

invisible_cities/daemons/idaemon.py
jjgomezcadenas: The idea is that daemons do chores that you don't want to do in the standard flow of the city (for example control histograms). Since those chores may also change, one has the possibility of invoking different daemons from the same city.

invisible_cities/daemons/idaemon.py
jjgomezcadenas: agreed

invisible_cities/daemons/idaemon.py
jacg: I don't see how that addresses the question.

`ModuleNotFound` will be raised if a daemon name is incorrect. In this case we don't want the code to run (while making noise on `stdout`); we want the code to stop, so that the daemon name can be corrected.

jjgomezcadenas: Sorry didn't get the sense of your question, and BTW I don't remember having written this part of code (probably the few neurones I have left are dying fast). Agree, we want the code to stop.

jacg: OK, I've fixed this in a miscellaneous cleanup branch I've got here. Progress is slow, as I only have scraps of time here and there.

PR #276, opened 2017-07-24
Diomira cecilia

jjgomezcadenas:
This PR transfers to Diomira the functionality intended for Cecilia who knows rests in peace until next incarnation. The PR introduces also a number of improvements, whose need has been revealed by the need to emulate the trigger. Specifically in this PR:

Base cities has been refactored to introduce three new cities:

1) RawCity: a city that reads RawData.
2) PCity: a city that reads PMAPS.
3) DstCity: a city that reads DSTs (KDSts), formely called MapCity.

Inheritance from those cities provides common functionality for the
file_loop() method. Thus, Irene, Isidora and Diomira, all are
RawCities (although Diomira reads MCRD while Irene and Isidora read
RWF), Dorothea and Penthesilea are PCities and Zaira is so far the
only example of a DstCity (but there will be likely others).

The cities still keep a linear scheme of inheritance, since
DeconvolutionCity inherits from RawCity propagating the
functionality to Irene and Isidora (both of them Deconvolution cities)
while KrCity inherits from DstCity, propagating the functionality to
Dorothea and Penthesilea.

Furthermore this PR adds a TriggerEmulationCity, which allows
the emulation of the trigger. The base clase formerly called SensorResponseCity has been renamed to MonteCarloCity and made to inherit from 
TriggerEmulationCity (which provides the capability to read Raw
Waveforms). Finally, Diomira inherits from MoteCarloCity (and thus from
TriggerEmulationCity). This scheme allows Diomira to read raw
data (MCRD), simulate sensor response (functionality from
MonteCarloCity) and emulate trigger (functionality from
TriggerEmulationCity).

Configuration files have been refactored to reflect the new
inheritance structure of cities.


invisible_cities/cities/diomira.py
```diff
@@ -73,30 +82,48 @@ def event_loop(self, NEVT, first_event_no, dataVectors):
...
+                                                         self.sipm_adc_to_pes)
+            dataSiPM_noisy = self.simulate_sipm_response(evt, sipmrd,
+                                                         self.noise_sampler,
+                                                         self.sipm_adc_to_pes)
             dataSiPM = wfm.noise_suppression(dataSiPM_noisy, self.sipms_thresholds)
+            RWF = dataPMT.astype(np.int16)
+            BLR = blrPMT.astype(np.int16)
+
+            # simulate trigger
+            peak_data = self.emulate_trigger(RWF)
```
jmbenlloch, line 106: With this new version we are always doing the trigger simulation. Is this what we want? There could be a flag to choose whether to simulate trigger or not. Maybe this is not necessary if we now want to put the trigger always.

jjgomezcadenas:
In the general case we always want to simulate a trigger, since this corresponds to reality. We can always open the trigger cuts so that everybody passes the trigger.

A different matter is that in the current version we have implemented only the S2 trigger. The S1 trigger works in a different way, and will need to be implemented by a different emulation. Which emulation to run then needs to be decided by parameters, and since an option needs to be made we can always decide a "NULL" trigger emulation. 

However, I think this is a matter for a different PR.

jmbenlloch:
Ok, understood, makes sense. I will give you the approval and I'll merge this PR.

jmbenlloch (approved):
Important modification to diomira, now we can simulate the trigger getting much better results to compare with real data.

jjgomezcadenas:
Not merged yet. Can you merge it @jmbenlloch?

jmbenlloch:
Done.

PR #277, opened 2017-07-24
Take negative run numbers for MC

jmbenlloch:
This PR changes the way we use the DB. Instead of using just run number 0 for MC, now the idea is to put a negative number. The configuration obtained will be the same as the one from the same positive run number. The negative sign will indicate that is an MC file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ branches:
 
 install:
   - source manage.sh install ${IC_PYTHON_VERSION}
+  - bash manage.sh download_test_db_dev
 
 script:
   - HYPOTHESIS_PROFILE=hard bash manage.sh run_tests_par 4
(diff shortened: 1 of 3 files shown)


jjgomezcadenas:
Travis is unhappy. Any idea why, @jmbenlloch ?

jmbenlloch:
It is strange, in my fork travis was not complaining... I'm looking at it.

jmbenlloch:
It is already fixed, it was a problem with the dev db. You should redownload dev db again.

jjgomezcadenas (approved):
Updates in the DB needed for the MC to have the same configuration than the data (e.g, for the elecID in the PMTs). The new dev DB also introduces masked SiPMs.

jjgomezcadenas:
please merge @jmbenlloch

jmbenlloch:
Done.

PR #278, opened 2017-07-25
Diomira fix

jjgomezcadenas:
This small PR fixes a minor (major) issue.

DIOMIRA simulates both the response of the sensors and the trigger. When emulating the response of the trigger, one is simulating the search on the FPGA on peaks in individual PMTs. This search is performed with the same algos used to search s1 and s2 in the calibrated sum, but the individual PMTs are neither calibrated, nor rebinned. 

Consequently, the search for s1 and s2 must allow the possibility that the waveform is not rebined (previously we never rebinned s1 but always rebinned s2). This is done in this PR, together with defining proper parameters for DIOMIRA.


jmbenlloch (approved):
This PR fixes an important bug for the MC production. Done with peras with me.

PR #279, opened 2017-07-26
Penthesilea extend

jjgomezcadenas:
This PR includes fixes needed to extend the functionality of Penthesilea (so that it can call Paolina). The most significant change is the re-factoring of the Voxel-Hit-Cluster classes that are now organised consistently. All classes are derived from a BHit class (which defines a base-hit (x,y,z,E) and is identical to Voxel, which keeps a different name for clarity). 

An old-standing issue related with Cluster has been fixed. Cluster had a "pos" attribute which returned an (x,y) vector, while a "pos" attribute of a hit should return an (x,y,z) vector, has it happens now. The old "pos" in Cluster has been renamed to "posxy".

In addition a number of small miscellaneous fixes and improvements have been implemented.


jmbenlloch (approved):
Small fixes and needed changes to include Paolina in Penthesilea.

PR #280, opened 2017-07-26
functions for finding and using s2s from individual pmts

abotas:
These additions allow IC to find and interact with the s2s of individual pmts.

1.
`find_s12` has been broken up into 3 functions `find_peaks`, `select_peaks_of_allowed_length`, and `extract_peaks_from_waveform` (in python and cython). 
`find_s12` was too big. First it found peaks, then it selected peaks of the correct length, then it put those peaks into a dictionary. 
Without changing the inputs/outputs, adding flags, and other technical debt, `find_s12` would not allow us to efficiently find the s12s of individual pmts.
Note finding the s2s of individual pmts is not quite as simple as it first appears since we need to rebin each pmt `cwf` in the same exact way we rebinned the `csum`, so that the rebinned time bins of the s2pmtd match and sum to s2d for each peak.
Also, breaking `find_s12` into modular functions allows for much better testing. 

2. 
a class, `S2Pmt` has been added to `pmaps.pyx`. It is initialized with dictionaries `s2d` and `s2pmtd` which maps `peak_no -- > [[pmt0E], [pmt1E], ..., [pmtnE]]`. People should look at this and make sure they like the basic structure.

3. 
a function `get_s2pmtd` has been added to peak functions. It takes as input `CWF` and `peak_bounds`.  `peak_bounds`  is a dictionary mapping `peak_no --> i_start, i_stop` of the s2 peaks found in csum. Note `peak_bounds` is the output of `find_peaks`, and `get_s2pmtd` makes uses `extract_peaks_from_waveform` to construct s2pmtd --> tearing  `find_s12` apart was worth it. 

4. 
tests for all this stuff

`-----------`
I have only created the free functions, I have not yet started messing around with any of the cities or base cities, so I have not deleted find_s12. If people are happy i or whoever can update `pmap_city` and whatever else to use the new functions.

### DIFF
--- a/invisible_cities/core/exceptions.py
+++ b/invisible_cities/core/exceptions.py
@@ -44,3 +44,6 @@ class NoHits(ICException):
 
 class NoVoxels(ICException):
     pass
+
+class SumNotEqualToSumOfParts(ICException):
+    pass
--- a/invisible_cities/core/exceptions.py
+++ b/invisible_cities/core/exceptions.py
@@ -45,5 +45,5 @@ class NoHits(ICException):
 class NoVoxels(ICException):
     pass
 
-class SumNotEqualToSumOfParts(ICException):
+class InconsistentS2dS2pmtd(ICException):
     pass
--- a/invisible_cities/evm/pmaps.pxd
+++ b/invisible_cities/evm/pmaps.pxd
@@ -47,7 +47,7 @@ cdef class S2Si(S2):
     The s2sid is a dictionary {peak:{nsipm:[E]}}
     """
     cdef public s2sid
-    cdef dict _s2sid
+    cdef  dict _s2sid
     cpdef number_of_sipms_in_peak(self, int peak_number)
     cpdef sipms_in_peak(self, int peak_number)
     cpdef sipm_waveform(self, int peak_number, int sipm_number)
@@ -60,9 +60,11 @@ cdef class S2Si(S2):
 """Given an s2d and an s2sid, return an s2d containing only the peaks shared by s2sid"""
 cpdef check_s2d_and_s2sid_share_peaks(dict s2d, dict s2sid)
 
+
 cdef class S2Pmt(S2):
-  cdef public s2pmtd
-  cdef dict _s2pmtd
-  cpdef pmt_waveform(self, int peak_number, int pmt_number)
-  cpdef pmt_total_energy(self, int peak_number, int pmt_number)
-  cpdef store(self, table, event_number)
+    cdef public s2pmtd
+    cdef  dict _s2pmtd
+    cpdef pmt_waveform(self, int peak_number, int pmt_number)
+    cpdef pmt_total_energy_in_peak(self, int peak_number, int pmt_number)
+    cpdef pmt_total_energy(self, int pmt_number)
+    cpdef store(self, table, event_number)
--- a/invisible_cities/evm/pmaps.pxd
+++ b/invisible_cities/evm/pmaps.pxd
@@ -59,3 +59,10 @@ cdef class S2Si(S2):
 
 """Given an s2d and an s2sid, return an s2d containing only the peaks shared by s2sid"""
 cpdef check_s2d_and_s2sid_share_peaks(dict s2d, dict s2sid)
+
+cdef class S2Pmt(S2):
+  cdef public s2pmtd
+  cdef dict _s2pmtd
+  cpdef pmt_waveform(self, int peak_number, int pmt_number)
+  cpdef pmt_total_energy(self, int peak_number, int pmt_number)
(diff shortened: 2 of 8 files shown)


jjgomezcadenas (commented):
Very good work. Some minor points:

1. find_s1 and find_s2 still use find_s12. I suggest making find_s12 private rather than delete it, then let find_s1 and find_s2 use the new functions and ensure in a test that results using old function and new functions is consistent.

2. in class S2Pmt I would change the function 

pmt_total_energy(self, int peak_number, int pmt_number)

 to

cpdef pmt_total_energy_in_peak(self, int peak_number, int pmt_number) 

and then add a new method:

cpdef pmt_total_energy(self, int pmt_number)

That adds the energy of all peaks.

3. Which of the three functions you have newly defined (if any) needs to be public (meaning leaving outside the python code)? My first reaction is that none of them, since they are all called internally by 
find_s1, find_s2 or get_s2pmtd, right? Make cdef of those functions that are not needed outside their ctyhon clients (unless you need to test them directly in a python code).

4. Have you tested that python and python code yield same results?

Anyway, again, excellent. Consider the above suggestions, rebase on top of the head and I will approve this very professional piece of work.

abotas:
1. y 4.  I have tested to make sure that the modular functions give the same output as `find_s12`. I understand you want to keep `find_s12`, but i don't think we want duplicate code in the repo. I suggest we keep `find_s12`, have it take all the same inputs and outputs, but have it find s12s with the new modular functions. Does this make sense? What do you think??

2. Ok sounds good

3. I think you are right, they can probably all be cdef. Definitely the function that checks the length of the peaks should be private. I made them public really only to make them very easy to test. But I will make the private, accesible via public functions.

jjgomezcadenas (commented):
Excellent. Are you still planning any addition? Otherwise I will approve.

jacg:
Sorry, no time to read in detail, but I have a slight suspicion that there is a misconception about what private means. Specifically, I would like all to agree that a function being written in Cython or Python is completely orthogonal to it being private or public. Both private and public functions can be defined in pure Python, and both private and public functions can be defined in Cython.

abotas:
@jjgomezcadenas let me push a couple more small commits porfa. The branch will be ready tomorrow morning! Unless anyone has complaints. 

@jacg, we all agree! Of course it's still possible there's something I don't understand... What I meant is that I originally wrote `cpdef` functions because I wanted to be able to test them in our `test_whatever.py` files. The plan now is to make those functions cdef, call them with `cpdef` functions, and test the `cpdef` functions. Does this make sense?

jjgomezcadenas:
@jacg No, I don't think there is any confusion here, probably we are not being very clear. The point I was trying to make is that in cython if you want to make a function private (e.g, accesible only to the  cython module in which it leaves) you can use cdef function while to make it accesible to the python API you use cpdef function. 

In python, instead, we use the convention _function to make it "private" (it is still accesible to any one that knows the name, while the cython function is not).

jjgomezcadenas:
@abotas no hurry, take your time!

jacg:
> I originally wrote cpdef functions because I wanted to be able to test them in our test_whatever.py files. The plan now is to make those functions cdef, call them with cpdef functions, and test the cpdef functions. Does this make sense?

Are you aware that `cpdef` creates two funcions? a `def` version *and* a `cdef` version. So, your plan to call `cdef` from `cpdef` seems to be a re-invention of a wheel that is built-in to Cython.

http://notes-on-cython.readthedocs.io/en/latest/function_declarations.html

jjgomezcadenas:
Not sure we are talking about the same thing: This is our case:

in module fulanito_c.pyx

cpdef function_to_be_called_from_python()

     x = cython_function_that_does_not_need_to_be_called_out_of_this_module()
    ...

cdef cython_function_that_does_not_need_to_be_called_out_of_this_module()


in other python modules:

from fulanito_c import function_to_be_called_from_python()

use it...

---

makes sense?

jjgomezcadenas:
Oh, I see. What @jacg says makes sense. @abotas  Of course if you want to test the functions that means exposing them to python and therefore you need to use cpdef (otherwise we are indeed reinventing the wheel). 

Of course, even if we use cpdef we can make them "private" by using _ convention. 

So, to be clear, one uses cdef only if the function is never called by python. In our case, we test all functions with pytest, thus use of cpdef is almost automatic.

abotas:
Ok, from my point of view this is ready. It passes all the tests, and I've run irene over 1k montecarlo events and creates exactly the same pmaps as before / everything looks good. I will open another PR to try to get rid of the rebin flag.

jjgomezcadenas (approved):
Excellent work.

jjgomezcadenas:
Please @jmbenlloch or @jagc merge this when you can

PR #283, opened 2017-07-27
Add exception for slices without clusters

paolafer:
One exception added to be thrown when no clusters are found in a slice.

### DIFF
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -24,6 +24,7 @@ from .. core.exceptions         import FileLoopMethodNotSet
 from .. core.exceptions         import EventLoopMethodNotSet
 from .. core.system_of_units_c  import units
 from .. core.exceptions         import SipmEmptyList
+from .. core.exceptions         import ClusterEmptyList
 from .. core.random_sampling    import NoiseSampler as SiPMsNoiseSampler
 
 from .. database import load_db
@@ -583,6 +584,8 @@ class KrCity(PCity):
             return self.reco_algorithm(np.stack((xs, ys)).T, Qs)
         except SipmEmptyList:
             return None
+        except ClusterEmptyList:
+            return None
 
     def compute_z_and_dt(self, ts2, ts1):
         """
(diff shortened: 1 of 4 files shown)


jjgomezcadenas (commented):
Adding the exception is fine, but why don't you go ahead and throw it in the proper place (otherwise is not very useful)?

paolafer:
I've just done it, but @abotas and I have seen a possible issue: now the corona function in xy_algorithms.py can raise three different exceptions and we want to be able to catch the three of them and distinguish them to treat them differently. In the HitCity class, for instance, I have added a second exception (line 692), but if I return None in both, one cannot distinguish them, right? Maybe we can Skype on this issue, if it's easier.

jacg:
I detect a hint of confusion about exceptions an return values.

> the corona function [...] can raise three different exceptions and we want to be able to catch the three of them and distinguish them to treat them differently.

Yes, like this:

    try:
        # code that might raise any one of your 3 exception types
    except ExceptionType1:
        # What to do if ExceptionType1 is raised
    except ExceptionType2:
        # What to do if ExceptionType3 is raised
    except ExceptionType3:
        # What to do if ExceptionType1 is raised
    else: # This block is optional
        # What to do if no exception is raised at all

> In the HitCity class, for instance, I have added a second exception (line 692), but if I return None in both, one cannot distinguish them, right?

Return `None` in both of what?

When you call a Python function, one (and only one, for any single invocation of the function) of the following 3 things will *always* happen 

1. a value (which may be `None`) will be returned,
2. an exception will propagate out of it,
3. the interpreter will crash.

Consequently, you cannot both return None (or any other value) and raise an Exception as the result of a single function call.

Ah, perhaps you mean that you raise some exception internally, then handle it internally and return `None`? If so, then you are failing to get the point of exceptions. Don't handle the exception in this function, let it propagate out to the part of the code which knows how to deal with the situation.

Returning values communicates between some stack frame and the one *directly* below it (between some function and its direct caller). Exceptions allow for communication between some stack frame an some other stack frame arbitrarily far below it.

So, I suspect that your problem is that you are handling your exception too close to its source.

paolafer:
Yes, I understand and agree. I'm going to change the code in the sense you suggest.

PR #284, opened 2017-07-28
get rid of redundant rebin flag, we already have rebin_stride

abotas:
Our find s1/s2 algorithms took as input both `rebin_stride`, and a flag, `rebin`. The flag is redundant. 

Unfortunately this redundancy invaded many corners of IC including the config files for diomira and pmap_city, so I had to make a bunch of small changes. ^ This might be of particular interest to @jmbenlloch if he is in charge of the productions. 

The code passes the tests, I've checked to make sure that diomira does not try to rebin s2s, and again, I've run Irene over 1k events in a notebook and everything looks good. 

Now when `rebin_stride == 1` there is no rebinning, and we don't pass extra arguments to our peak finding algorithms.

### DIFF
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -28,23 +28,23 @@ from .. core.random_sampling    import NoiseSampler as SiPMsNoiseSampler
 
 from .. database import load_db
 
-from ..io                       import pmap_io          as pio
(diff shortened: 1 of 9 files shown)


jjgomezcadenas (approved):
Excellent. Reduction of technical debt long time due.

jjgomezcadenas:
@jmbenlloch please merge

PR #285, opened 2017-07-29
Corona improvements part 1: exceptions

paolafer:
@jacg I tried to do what you suggested, can you have a look and see if it's ok?


jjgomezcadenas (approved):
Improves the management of exceptions in the corona algorithms

PR #289, opened 2017-08-09
Fix event number in MCTrack table

### LINKED ISSUE #287: MC true information
I'm looking at a file of pmaps produced recently (last week) with irene, starting from a RWF file that @jmbenlloch gave me, and it looks like only the information of the first event is saved. I remember that it was an issue @jerenner and I were having a couple of months ago in Geneva, hasn't it been solved yet? It is quite important to have true info in order to understand the performance of corona and paolina.

jmbenlloch:
There was a bug in the mc_track_writer that was setting all event numbers to zero. It is now fixed and one test has been updated to check this condition.

This closes #287.


jjgomezcadenas (approved):
Bug fix with test

PR #290, opened 2017-08-11
Add function that returns the length of a track

paolafer:
@jacg I realized that in the paolina functions we miss a function that gives the length of the track. As far as I understand from the code, the shortest path is not available to the user (correct me if I'm wrong), therefore the user cannot calculate the length of a track by himself. On the contrary, the energy is accesible through the voxels. Is that true? Or am I missing the philosophy of it?


jjgomezcadenas:
Paola this is correct function was needed. Please add a test. 

Enviado desde mi iPhone

jjgomezcadenas (approved):
A useful function, really nothing but syntactic sugar.  A test is not essential (since shortest_paths has tests).

PR #291, opened 2017-08-18
Allow icusers to study s1s of individual pmts

abotas:
Before we could only look at the s2s of individual pmts, even though the code had all the functionality necessary to look at the s1s of individual pmts. 

Even though we will not typically use/save this data, it is useful for some analysis and can help us better understand our detector and make cool plots, and it costs nothing.

### DIFF
--- a/invisible_cities/core/exceptions.py
+++ b/invisible_cities/core/exceptions.py
@@ -45,5 +45,5 @@ class NoHits(ICException):
 class NoVoxels(ICException):
     pass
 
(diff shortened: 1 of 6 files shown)


jjgomezcadenas (approved):
This PR completes previous work, allowing the possibility of saving S1 for individual PMTs in addition to S2. The coding is straight foward and Ale-clean (good job) and is exercised by a decent number of tests.

PR #292, opened 2017-08-21
Paolina dev

jjgomezcadenas:
This PR:
1. Refactors s12_filter so that it uses the high-level interface.
2. Adds new classes representing Tracks and Blobs (as well as their containers)
3. Adds new functionality to paolina_functions
4. Add new helper functions (e.g, functions to draw voxels and tracks).
5. Adds type hints (to Paolina functions, the idea is to add type hints to all functions in IC).


jmbenlloch (approved):
Useful and needed changes including tests.

invisible_cities/cities/base_cities.py
```diff
@@ -763,15 +763,16 @@ def make_tracks(self, evt_number : float , evt_time: float,
...
         for trk in track_graphs:
-            distances = paf.shortest_paths(track_graph)
-            a,b       = paf.find_extrema(distances) # type: Sequence[Voxel]
-            Ea = energy_within_radius(distances[a], self.blob_radius) # type: float
-            Eb = energy_within_radius(distances[b], self.blob_radius)
-            blob_a = Blob(a, Ea) # type: Blob
-            blob_b = Blob(b, Eb)
-            blobs = (blob_a, blob_b) if Ea < Eb else (blob_b, blob_a) # type: Tuple[Blob, Blob]
+            distances = paf.shortest_paths(trk)
+            a,b       = paf.find_extrema(distances) # type: Tuple[Voxel, Voxel]
```
jacg: Why use comments to specify the type, rather than Python's annotation syntax, if 

+ you advocate the use of Python's built-in annotation syntax
+ it is available in all the versions of Python we support!

jjgomezcadenas:
I am replying from mail because Idon’t seem to be able to access the web here.
Comments are one of the ways suggested by type annotations to describe variables.

jacg:
I know. That's before they had a syntax for variable annotations. 

I really, *really*, **REALLY** hate semantically meaningful comments: the number one thing baby programmers learn about comments is that they are purely for the benefit of the humans, that the compiler completely ignores them. There should be a special circle of hell reserved for those who design language features that depend on semantics buried inside comments. I understand that they are sometimes a necessary evil (for some definition of necessary), but they are still evil.

Anyway, it seems that currently there is no syntax that allows you to annotate tuple unpackings (unless I've missed it), so we *don't* have any other option for now, for most of the cases in this snippet.

jjgomezcadenas:
I agree. Let's put all this bunch in the 10th circle of Hell.

invisible_cities/reco/paolina_functions.py
```diff
@@ -69,6 +69,14 @@ def neighbours(va : Voxel, vb : Voxel, scale : float = 1.5) ->bool:
 
     return tuple(nx.connected_component_subgraphs(voxel_graph))
 
+
+def voxels_from_track_graph(track: Graph) ->List[Voxel]:
+    """Create and return a list of voxels from a track graph"""
+
+    voxels = [Voxel(t) for t in t.nodes]
```
jacg: Firstly, this is just an extremely convoluted and inefficient way of writing `return t.nodes()`. (Actually, there's also a bug in it, the `()` are missing! I think that it does get fixed in a later commit.)

Secondly, I question the value of creating `voxels_from_track_graph(t)` as a synonym for `t.nodes()`. There is some point to it: it creates a name which has a higher level meaning in our context. But I'm far from convinced that it is worth the effort. Given that we use it exactly one place, my temptation is to conclude that we should remove it and just use `track.nodes()` instead.

jacg: Correction: given that we **don't use it ANYWHERE exept its own test**, my temptation is to conclude that we should remove it ...

jacg: Correction to the correction: yes, we were using it in one place. Sorry about the noise.

PR #293, opened 2017-08-22
Add more documentation to corona

abotas:
- corona is generally a bit more legible
- added a detailed description of corona and its arguments
- added an explanation of why this corona is very slightly more complex 
- added a usage example that shows how this corona can be configured to have the same functionality as the corona in art


jjgomezcadenas (approved):
Well here we have clear and extensively documented code. Good job.

PR #294, opened 2017-08-22
Imports SipmZeroCharge exception

jerenner:
Penthesilea does run properly when this exception is thrown because it was not imported


jjgomezcadenas (approved):
Good catch.

PR #295, opened 2017-08-23
add functions for increasing s2si thresholds + tests

abotas:
These are functions we can use to increase the cuts on the SiPMs: `thr_sipm` and `thr_sipm_s2`

The idea is that maybe we could initially run irene with cuts on the SiPMs on the low side, and have the ability to quickly increase them whenever we want without having to rerun all of irene. 

This is a work in progress -- I'm using these functions extensively now to debug them and to look more into where we should put our cuts. I still need to cythonize them and maybe add more tests. I am opening the PR now just so people can have a look if they want to!

### DIFF
--- a/invisible_cities/core/exceptions.py
+++ b/invisible_cities/core/exceptions.py
@@ -59,3 +59,6 @@ class NoVoxels(ICException):
 
 class InconsistentS12dPmtsd(ICException):
     pass
+
+class NegativeThresholdNotAllowed(ICException):
+    pass
--- a/invisible_cities/reco/pmaps_functions.py
+++ b/invisible_cities/reco/pmaps_functions.py
@@ -96,33 +96,6 @@ def rebin_s2si_peak(t, e, sipms, stride):
       {sipm: ccf.rebin_array(qs, stride, remainder=True) for sipm, qs in sipms.items()}
 
 
-def _impose_thr_sipm_destructive(s2si_dict: Dict[int, S2Si],
-                                 thr_sipm : float           ) -> Dict[int, S2Si]:
-    """imposes a thr_sipm on s2si_dict"""
-    for s2si in s2si_dict.values():                   # iter over events
-        for si_peak in s2si.s2sid.values():           # iter over peaks
-            for sipm in list(si_peak.keys()):         # iter over sipms ** avoid mod while iter
-                for i, q in enumerate(si_peak[sipm]): # iter over timebins
-                    if q < thr_sipm:                  # impose threshold
-                        si_peak[sipm][i] = 0
-                if si_peak[sipm].sum() == 0:          # Delete SiPMs with integral
-                    del si_peak[sipm]                 # charge equal to 0
-    return s2si_dict
-
-
-def _impose_thr_sipm_s2_destructive(s2si_dict   : Dict[int, S2Si],
-                                    thr_sipm_s2 : float           ) -> Dict[int, S2Si]:
-    """imposes a thr_sipm_s2 on s2si_dict. deletes keys (sipms) from each s2sid peak if sipm
-       integral charge is less than thr_sipm_s2"""
-    for s2si in s2si_dict.values():
-        for si_peak in s2si.s2sid.values():
-            for sipm, qs in list(si_peak.items()): # ** avoid modifying while iterating
-                sipm_integral_charge = qs.sum()
-                if sipm_integral_charge < thr_sipm_s2:
-                    del si_peak[sipm]
-    return s2si_dict
-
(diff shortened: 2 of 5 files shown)


abotas:
@jacg @jjgomezcadenas 

From now we only keep a python or cython version of the code in the repo? So if i make cython versions of some of these functions I should delete the python functions, correct? 

This is my current plan unless someone tells me otherwise. I guess the alternative would be to run tests over both the cython and python versions.

A small pro for generally keeping python versions around that i don't think was discussed yesterday, is that python code is easier to debug/develop. If we discover some buggy performance in a function that has gotten by the tests, we can switch to the python version of the function debug / develop. But maybe this is not worth slowing down our automatic testing...

jjgomezcadenas:
Indeed your argument is correct, and by removing functions from the API (via _xxx) they effectively disappear from the code. I think there is no need to go hunting python functions superseded by cython functions right now (most of them are indeed involved in some sort of testing). 

On the other hand, the Master's argument still apply. Python functions (such as the one we have just killed) which are superseded by cython and do not contribute to a test are playing no role and should be eliminated (in particular the one we have kill was a one-liner).

jjgomezcadenas:
And BTWl  *this PR* is a good example of the matter under discussion. The functions that you have just written (in python) are far from trivial, and the cython counterpart always more obscure. You are following the natural way, e.g., write the python function, then pass a test, then write the cython function (then add a test for the cython function). When the cycle is completed, one could, in principle get rid of the python functions but:

1. They don't bother anyone (_xx notation makes them invisible).
2. They are involved in a test that should be deleted if we delete the function.
3. They are usually easier to read than cython thus they contribute to the documentation. 
4. They may be useful if at some stage we want to expand the function, modify it or find a bug. Work with python, test, then go to cython, test, is the correct workflow.

So, to conclude:
1. if the python function superseded by the cython function is not involved in a test, we should eliminate it (it means it was never necessary).
2. If the python function is involved in a test I would leave it in place (otherwise I have to eliminate the function and the test, which does not make me happy).
3. When writing new code, as in this PR I would keep the natural flow, python-->test--->cython-->test and only eliminate python (+ test) if code is so trivial that python function plays no role.

To insist, by keeping them out of the API (via _xx) no harm is done.

jjgomezcadenas (commented):
I think you don't need the list( in

for sipm in list(si_peak.keys()): 

and similar lines, si_peak.keys() should return you an iterable. 

Also, I suggest that you use type hints as I have done in paolina_functions. 

See:

http://mypy.readthedocs.io/en/latest/cheat_sheet_py3.html
http://mypy.readthedocs.io/en/latest/kinds_of_types.html
http://mypy-lang.org/examples.html

Thus, for example:

def rebin_s2si_peak(t: ndarray, e:ndarray, sipms:Dict[int, ndarray])->Dict[int, ndarray]

type hints allow us to document much better what the function does, and eventually we can use a static type checker (such as mypy) to ensure consistency. 

BTW: why do you want to  impose a threshold destructively? This is a "book example" of a function with side effects. Why not create a new object and void those side effects?

The whole bunch of functions that you have written are side-effects sinners. Is it worth considering to write versions without side effects (e.g, create new data rather than modify it?) 

perhaps @jacg wants to illuminate us here.

abotas:
1) I've made the main function, `raise_sipm_thresholds` non-destructive and added more tests. Functions called by raise_sipm_thresholds are still destructive so that we do not have to copy the entire s2si_dict for each individual helper function, but these helper functions are intended to be private. 
2)  cythonized the functions that do the heavy lifting and tested the cythonized functions. 
3) added type hints to the python functions.

At the moment the python versions of the cython functions are still in the branch. Should I keep them or delete them. Both python and cython versions pass the tests (they are identical) but at the moment only the cython functions are run by the tests. Is this approvable as is? Should I delete the python versions? Should I have the tests run both versions?

jjgomezcadenas:
If the master @jacg confirms, let's bravely get rid of the python

jacg:
If the Cython functions pass the same tests that the pure Python functions used to pass (modulo name change of the function) then you can go ahead and remove the pure Python versions.

If the commit message (preferably its title) can highlight that a pure python version of a function was replaced by a Cython implementation, that would be useful if we ever doubt their equivalence in the future and want to see the old version.

Git, of course, has tools that help track this sort of stuff down, but how many of you

1. know about them?
2. know how to use them?

So, let's make it as easy as possible for the naive Git user to track down such changes.

abotas:
Done. python versions are deleted. I think (hopefully) the commit messages are clear as to where they were deleted.

PR #297, opened 2017-08-23
Typo in return np.array(qs_slice)

Aretno:


jacg:
Looks like this should have crashed if this line was ever executed! The fact that no test found this is telling us one of two things:

1. Tests for this function need to be added.

1. This function is not needed and should be removed.

jjgomezcadenas:
Actually, what happens is that we use the cython version of this function (which is tested and exercized). We have a bit of an issue with the duplication of python and cython functions, in general I am relecutant to remove too quickly the python versions, but here is an example of something that had become a fossil. Indeed the bug was not cathed because the function was not used (the cython version is).

Aretno:
Ok then I'll close the Pull request. Thanks.

jacg:
Well, whatever happens, it can't be left as it is! My preference would be to remove the function altogether. The other alternative is to add a test, and fix the bug.

But we can't just leave dead and buggy code lying around once it has been identified.

jjgomezcadenas:
Indeed. I believe that the best option in this case is to remove the function.

PR #298, opened 2017-08-23
Fix sipm ids and charges in slice

Aretno:
Deleted the function as suggested in https://github.com/nextic/IC/pull/297 

(I don't know if the previous PR could be edited with the new commit, sorry)

### DIFF
--- a/invisible_cities/reco/pmaps_functions.py
+++ b/invisible_cities/reco/pmaps_functions.py
@@ -52,24 +52,6 @@ def _integrate_S2Si_charge(s2sid):
     return { peak_no : _integrate_sipm_charges_in_peak_as_dict(peak)
              for (peak_no, peak) in s2sid.items() }
 
-
-def _sipm_ids_and_charges_in_slice(s2sid_peak, slice_no):
-    """Given s2sid_peak = {nsipm : [ q1, q2, ...qn]} and a slice_no
-    (running from 1, 2..n) returns:
-    Returns (np.array[nsipm_1 , nsipm_2, ...],
-             np.array[q_k from nsipm_1, q_k from nsipm_2, ...]]) when slice_no=k
-     """
-    number_of_sipms = len(s2sid_peak.keys())
-    ids      = []
-    qs_slice = []
-    for i, (nsipm, qs) in enumerate(s2sid_peak.items()):
-        if qs[slice_no] > 0:
-            ids.append(nsipm)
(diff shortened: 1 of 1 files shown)


jacg:
It can be edited, but the simplest thing to do is to add a comment in the old PR, saying that the function is being deleted here.

An even cleaner thing would have been to re-open the previous PR (rather than starting a new one) and remove the function there.

jjgomezcadenas:
Actually, I am a bit uneasy about deleting the function, rather than fixing it. Notice that we were stating explicitely that the funciton was not in the API (as the name _xxx indicates). So far we have been consistently keeping private versions of the python function whenever we had a cython public function. The python function have often been used to cross check the cython counterpart. 

In this case the function is so simple that it is really not worth, but I am not 100 % sure that I want to break the habit of having a python function for each cython function.

Although perhaps the best thing would be to delete *all* the python functions that are superseded by cython counterparts and not explicitly used in tests...

jacg:
The fundamental point is, that we should not have untested functions in our code base. And we *definitely* do not want broken functions in our code base. Now, whether you fix this problem by adding tests, or by removing functions is a second order issue.

Remember: having a function which does not get tested automatically and regularly, is equivalent to saying "I do not care if this function is broken". If you don't care if a function is broken, what is it doing in your code base?

I think that duplication of functions is downright counterproductive. There are arguments for keeping simple and clean pure-python versions of functions which have been superseded by uglier, optimized Cython versions. But the job of 'keeping' them is up to the version control system: the functions are still there in our VCS, should we ever need them, but they shouldn't be polluting the active code base.

My vote is to remove all superseded python versions

> The python function have often been used to cross check the cython counterpart.

The function has to pass the tests. You develop a slow first version of the function, you write a thorough test suite for it, then you write an optimized version of the function, and ensure that it still passes the same tests. Once it does, what is the point of keeping the old version? How does it help to cross check? (Is this perhaps some instinct that you developed in the bad old days when you didn't have automatic test suites?)

Given that this is not the first time that one of these obsolete Python functions causes confusion, supports my point. Let's get rid of them all.

jjgomezcadenas (approved):
Following the advise of the Master I hereby approve removing this obsolete function and vouch to keep removing all the others. Amen.

jjgomezcadenas:
@aretno please rebase your branch to main (ask @gonzaponte or @abotas if trouble), then one of them, please merge!

PR #300, opened 2017-08-24
Xy algos

jjgomezcadenas:
This PR refactors the XY reconstruction algorithms.

1.- The code no longer calls two algorithms (barycenter or corona)
which had to be selected by config, but a single algorithm (corona) with
a number of parameters. In particular, lm_radius can be set to a
negative number in which case corona will return an overall
barycenter (using all SiPms above Qthr threshold).

2. The condition imposing a single cluster when filling a kr_event has
been removed. That condition assumed that an overall barycenter
returning a single cluster *had to be used* in kr_city (e.g, to
compute KDSTs and maps). However, this is not a must and in fact,
better (less biased) results can in principle be obtained using a
corona algorithm which uses a limited number of SiPms around the local
maximum. On the other hand, the use of a corona may result in more
than one cluster in the tracking plane, even in the case of
single-particles (e.g, X-rays or Krypton), due, for example, to noisiy
SiPMs. In this case the physical cluster is expected to have more energy than the satellite and this is the criterium used to choose.

3. In the case of hit_city, if there is a reconstruction failure a hit
is still formed using a "NN cluster" (a Non-Numeric cluster). This
allows to keep track of the energy of the slice that failed to produce
a cluster. 

The new code passes all the tests. In addition, we have run many events using Penthesilea to check that there are no running failures. Since there is no new functionality (no new algorithms) but a refactor of existing ones, new tests do not appear to be a must (in fact, existing tests caught a bug introduced by accident when refactoring and allowed for immediate correction).

### DIFF
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -50,7 +50,7 @@ from .. reco.sensor_functions   import convert_channel_id_to_IC_id
 from .. reco.corrections        import Correction
 from .. reco.corrections        import Fcorrection
 from .. reco.xy_algorithms      import corona
-from .. reco.xy_algorithms      import barycenter
+#from .. reco.xy_algorithms      import barycenter
 
 from .. evm.ic_containers       import S12Params
 from .. evm.ic_containers       import S12Sum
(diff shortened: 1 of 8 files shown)


gonzaponte (approved):
Several improvements to the reconstruction algorithms. It also improves the readability of the code, which is very much appreciated.

PR #301, opened 2017-08-24
Dorothea small fixes

jjgomezcadenas:
The possibility of using corona instead of barycenter when processing "point like" events (e.g, Krypton events, implies that some times more than one cluster is produced by the algorithm. It is important to keep track of how many times this happens, in order to maximize the corona parameters for Krypton events. For example, if one runs corona ( 3 x 3) with a low SiPM threshold (e.g, less than 25 pes), about 10 % of the times 2 clusters are produced, while with a high SiPM threshold only 1% of the times a cluster is produced. Since the typical charge of a Krypton event in the anode is of the order of 300, a cut on 25 pes appears OK.

In order to keep tracks of the number of times that we have two hits a counter is introduced. In the same go, a few other counters are introduced to improve the book keeping of Dorothea. 

Finally, examples of configurations files to run Dorothea with barycenter or with corona are added to config.

### DIFF
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -655,14 +655,15 @@ class KrCity(PCity):
                 if len(clusters) == 1:
                     c = clusters[0]
                 else:
+                    cQ = [c.Q for c in clusters]
+                    self.cnt.increment_counter('n_events_more_than_1_cluster')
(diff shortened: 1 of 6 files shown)


gonzaponte (approved):
Some cosmetics for Dorothea and configuration files for common user-cases. Approved.

PR #302, opened 2017-08-24
Icaro stuff

gonzaponte:
Some cosmetics and new functions for icaro.

### DIFF
--- a/invisible_cities/icaro/hst_functions.py
+++ b/invisible_cities/icaro/hst_functions.py
@@ -5,8 +5,15 @@ import textwrap
 import numpy             as np
 import matplotlib.pyplot as plt
 
-from .. core              import fit_functions as     fitf
+from .. core              import fit_functions as fitf
 from .. evm.ic_containers import Measurement
+
+
+def create_new_figure(kwargs):
+    if kwargs.setdefault("new_figure", True):
+        plt.figure()
+    del kwargs["new_figure"]
+
 
 def labels(xlabel, ylabel, title=""):
     """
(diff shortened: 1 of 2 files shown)


jjgomezcadenas (commented):
Why do you need

del kwargs["new_figure"]?

jjgomezcadenas (commented):
Function
def histo(x, nbins, title="hsimple", xlabel="", ylabel="Frequency"):
    """histograms"""

    plt.hist(x, nbins, histtype="bar", alpha=0.75)
    plt.title(title)
    plt.xlabel(xlabel)
    plt.ylabel(ylabel)

seems to be older (I wrote it 100 years ago) and superseded by new hist. Should we get rid of it?

jjgomezcadenas (commented):
The file hst_functions contains many histogram related functions (essentially sugar on top of matplotlib) and then some functions that have nothing to do with histograms, such as "resolution", "gausstext", "savetofolder"... should those functions go to a different file? Also, some of them (e.g, "resolution") could use a test...

gonzaponte:
> Why do you need
> 
> del kwargs["new_figure"]?

The main functionality of that function is to create a figure unless it is otherwise specified in the kwargs parameter. The kwargs is the dictionary containing the set of parameters passed by name to any function making some kind of plot. The deletion in this case ensures the parameter "new_figure" is not propagated to other functions to avoid crash. It is kinda nasty, but it is also comfortable.

> seems to be older (I wrote it 100 years ago) and superseded by new hist. Should we get rid of it?

Sure

> The file hst_functions contains many histogram related functions (essentially sugar on top of matplotlib) and then some functions that have nothing to do with histograms, such as "resolution", "gausstext", "savetofolder"... should those functions go to a different file? 

I have been thinking about that, but I am not sure where the line is... All of them are in some way related to plotting (for instance, gausstext provides a string to be displayed using matplotlib). 

> Also, some of them (e.g, "resolution") could use a test...

Actually, resolution is tested in `hst_functions_test.py`

jjgomezcadenas (approved):
OK then. Not completely easy about the deleting but can't punch a hole on it.

PR #303, opened 2017-08-24
Hit readers

jjgomezcadenas:
So far we were capable to write hits to disk but we didn't have the capability to read the hits back. This PR adds this functionality and at the same time refactors (slightly) the io modules so that now:

**dst_io** only contains the generic functions load_dst and load_dsts (that will load any structure saved as a pandas dst). 

**hits_io** contains all the hits i/o functionality including the new function 

*load_hits(name_of_file) *

which returns a dictionary of hits keyed by event number. 

In addition, a new (small) file containing a few events with hits (from Talium MC) has been added. Using this file, tests have been added to check the correct reading of the hits. 

In particular, the test:

*test_load_hits_double_ratio_e_q_equals_one(TlMC_hits)*

checks a non-trivial condition, namely:

*The double ratio of energy deposition in the anode and in the cathode must be close to one.*

where the double ratio refers to the ratio between the average energies of the hits per event (without the hit with the maximum energy), and the average of the hits with the maximum energy. Physically those numbers must be close, an checking this condition shows that hit energies make sense.


gonzaponte (approved):
Important functions for reading back hits and exercised with tests. Approved.

PR #305, opened 2017-08-25
Add option to return std deviation instead of mean error

gonzaponte:


jjgomezcadenas:
Is it worth to make the functions more complex? Why not take always std??

gonzaponte:
Sometimes you need std, sometimes you need the mean error. The former one can be used to select data within some number of rms from the mean. The latter is the one we need for passing errors to fits.

If we hace to stick to one I would go for the mean error, because fits are far more common than rms-based selections, but I don’t think this is a huge complication to the functions.

jjgomezcadenas (approved):
OK, the code is clear enough and there is a test with std=True. Next time you could give a few more explanations in the PR text.

gonzaponte:
Sorry for that. I am very lazy!

PR #306, opened 2017-08-27
Mctracks

jjgomezcadenas:
A new module *mchits_io* containing functions
to read the MCTraks pytable and return high level event dictionaries is added.

Two high level classes are added to the event_model module. 

MCParticle defines a MC particle

MCHit defines a MC hit. 

functions in mchits_io module:

oad_mcparticles(file_name) --> returns event_dicts for particles : {event_no:MCParticle}
load_mchits(file_name) --> returns event_dicts for hits: {event_no:MCHit}

Two tests have been added. The test check that the
first particle and all its hits in disk is correctly loaded into the
high level class (for for MCParticle and for MCHit)

The  tests, while simple, check that the data format is
correct and the high level classes are correctly instantiated.


abotas (approved):
These are necessary additions to the repo that allow icusers to easily handle MC tracks and compare MC with data. 
The new high level classes MCParticle and MCHit are clear and concise. The classes inherit from their logical base classes. The load and read functions follow the format established in other IC load and read functions. The tests cover all of the functions added to repo. 
The PR is approved!

PR #312, opened 2017-08-30
Fix numpy warnings

### LINKED ISSUE #310: Many warnings while running tests
Starting at version 3.1 pytest displays warnings that were raised during the tests' excution.

Our test suite produces a few screenfuls of warnings. These need to be analyzed and

1. if the warning points at an error in our code, the code needs fixing

2. if the warning is a false positive, the code needs to be rewritten in a way that silences the warning.

gonzaponte:
This PR fixes a number of warnings issued when running tests. There were 3 types:

1. Operations on empty arrays. This has been fixed by introducing new functions to handle that particular case.
2. Overflow. Hypothesis floats were running up to 10**20, which caused overflow when calling np.std (which at some point squared the number).
3. Division by 0. This has been avoided by either removing those elements from the operation or treating the case separately.

Closes #310.


gonzaponte:
This PR comes from issue #310.

jacg:
Very nice PR.

Stylistically I'm very happy with it.

+ What is the reasoning behind the disappearance of the `-1` on line 108 of `fit_functions.py`?

+ Our word-of-mouth style guide prefers parenthesized expressions over `\` for line-breaks.

+ I don't really follow the changes in d89fdc33.

+ It might be a good idea to check quickly that there is no obvious performance degradation, as some extra copying of numpy arrays is introduced.

gonzaponte:
> Stylistically I'm very happy with it.

Wow!

> What is the reasoning behind the disappearance of the -1 on line 108 of fit_functions.py?

This was a reminder for me that I didn't notice before committing. It is something that should also be fixed but that is not in the line of this PR. I will remove it.

>Our word-of-mouth style guide prefers parenthesized expressions over \ for line-breaks.

Ok. I thought it was the other way around.

>I don't really follow the changes in d89fdc3.

1. `f_ref = self._fs[index] if self._fs[index] else 1` This ensures we don't have a 0 as reference value. Now I realize that this shouldn't pass silently, so I will add an assertion.
2. `valid = (self._fs > 0) & (self._us > 0)` This ensures the operations can be done safely. I will perform them when `valid=True` and use the default values provided when `valid=False`.
3. The copies `valid_fs = self._fs[valid].copy()` and  `valid_us = self._us[valid].copy()` are there for readability rather than because they are needed. These are the arrays I can do safe computations with.
4. The `self._us` calculations have been split into two lines for readability.

Let me know if there is something else unclear.

> It might be a good idea to check quickly that there is no obvious performance degradation, as some extra copying of numpy arrays is introduced.

Ok, but this is not a very demanded class and its method is called only once.

jacg:
> > Stylistically I'm very happy with it.

> Wow!

It's generally very satisfying to see how many of you have really got into the groove and are producing decent code with tests spontaneously appearing, and so on.

Anyway

1. Remove the orthogonal `-1` removal.
1. Add the missing assertion.

And I'll be happy to accept it.

Good stuff.

If you know that it's not in the critical path, then don't bother checking the performance.

jacg:
I'm getting a couple of test failures now. Travis agrees.

gonzaponte:
The assertion makes the test of zaira fail, probably because we don't have events in the nomalization bin. I have noticed that the test runs over 1 event only, which doesn't make sense...

gonzaponte:
I had to change the input file for the tests in Zaira. The file contained only one event, which is not enough for what Zaira does (a profile of energy in 2 dimensions). I will remove the file dst_3389_test.h5 if nobody objects.

jacg (approved):
Suppresses the pure-noise warnings as required. Tests have been added for the new warning-suppressing functions. Good work.

Issue #263, opened 2017-07-03
Failure in tests


paolafer:
One of tests is failing in my Mac, namely 

invisible_cities/reco/sensor_functions_test.py::test_cwf_blr[electrons_40keV_z250_MCRD.h5]

with the following error:

E           assert 1.7658670951263973 < 1
invisible_cities/reco/sensor_functions_test.py:47: AssertionError

Any idea of what the reason is, @jjgomezcadenas ?

jjgomezcadenas:
The test checks that the result of the convolution + deconvolution is equal (within tolerance) to the original waveform. The problem can be fixed increasing the tolerance, but I don't understand why the test passes everywhere except in your Mac.

Aretno:
I have updated my master to the latest repository and I'm having the same issue Paola has with this same test.

jjgomezcadenas:
Do you still have this problem, @paolafer , @Aretno ?? It is not showing up anywhere else

Aretno:
Yes, I have. I think @jahernando also had this issue. I tipically get "assert" values between 3-5.

jjgomezcadenas:
This is very funny. It never appears in my local or in TRAVIS. Not sure what is best here.

jerenner:
I am also getting a failed test on my Mac with invisible_cities/reco/sensor_functions_test.py:47: AssertionError with: 
E           assert 3.992222624122006 < 1

jjgomezcadenas:
I cannot reproduce this error locally. Are u running the last version of the code?

jerenner:
I got the error this morning when running the code on Alex's branch debug-cities

jmbenlloch:
Maybe is some old binary file (pyc or whatever), you could try running: `bash   manage.sh clean` and then `source manage.sh work_in_python_version 3.6` again.

Aretno:
No improvement on my side with the cleaning.

jerenner:
The cleaning does not seem to help for me either

jacg:
What is the status of this? Are these tests still failing for anyone?

Aretno:
It still does for me. I think @jmbenlloch also saw these tests failing in Canfranc computers.

jacg:
I suggest that you move your conda installation out of the way (`cp $HOME/miniconda $HOME/miniconda.moved-away`) and then do a full `source manage.sh install_and_check 3.6` in a fresh shell)

This should tell us whether its caused by some feature of your machine, or some feature of your existing conda state.

[...]

PR #314, opened 2017-08-31
ipmt pmap readers writers and loaders

abotas:
With these modifications we can write and load the s1 and s2 of individual pmts. The ipmt pmaps can be written and loaded separately or together with the traditional pmaps.

There are a few things we need to be aware of:

1) We've found we must flush the pytables after writing to avoid data losses at least when writing all the ipmt + pmaps simultaneously. However, the `file.flush()` is not included in this PR because it does not belong in the writers. The writers do not know how much they have written or when they are done writing. They are called for each event in the event loop. This is a little annoying since it would be nice to have commands necessary for writing inside the writers and not have to be aware in other places in the code that a `.flush()` is necessary. But also pytables documentation recommends flushing after any write operation so we should probably put flushes in all our run methods or file loops anyway. 

So, the .flush() command will be implemented in a separate PR (imminent) that allows irene or maybe pmap city to write ipmt pmaps.

2) One change implemented in this PR is not backwards compatible. s12pmt.pmtsd has been replaced by s12pmt.ipmtd because this naming is more descriptive, more consistent with the nomenclature necessary to distinguish between different pmap writers/loaders, and because I think no one is using the ipmt pmaps yet so hopefully there won't be many complaints.

3) All the write pmap functions (old and new) now accept `None` as an argument instead of an s1, s2, s2si s1pmt s2pmt or any combination of these. Before if there was an event without an s-whatever the writer had to be given an empty instance of that class, call the store method, and then do nothing. Now they store an si peak when there is an si peak for that event.  I am also eager to (in a separate PR) have the pmap constructors do something intelligent when they are given empty dictionaries in initialization. All of this will hopefully also speed irene up a little bit.


abotas:
I've rebased this over the most recent PR accepted to nextic/master

abotas:
rebased again over most recent PR merged into nextic/master

jjgomezcadenas:
Travelling review later today

jjgomezcadenas:
I am slightly unhappy with flush not being included in this PR. Why not?

jjgomezcadenas:
After we have processed all our data, we should flush the table’s I/O buffer if we want to write all this data to disk. We achieve that by calling the table.flush() method:

>>> table.flush()

Thus the recipe is pretty clear. Agree @abotas?

abotas:
Sure, we can put the flush in this PR. But I would flush the entire file `h5out.flush()` in `base_cities.city.run` just before we close the file.

This suggestion might sound weird but:
A `file.flush()` flushes all the pytables at once. 
And even though pytables in its documentation claims it flushes everything before closing a file, they are somehow failing to do this in our case, which i think is shown in this notebook:

https://github.com/abotas/ICARO/blob/city-monitors/icaro/reco/when-to-flush.ipynb

The notebook also shows:
1) Flushing after every event is 35x slower than never flushing
2) Never flushing when writing more than 4 pytables at once (at least on my laptop) always results in a lost pytable
2) flushing the write file once, just before closing it, is sufficient to make sure we never lose any events and it slows things down a only a couple ms. Since we don't lose anything I think we should always flush --> put the `file.flush` in `city.run()`

Should I add that line to the PR?

abotas:
I added `h5out.flush()` to `city.run()` just in case you like the solution. I can always change it

jacg:
You should avoid flushing frequently, as it defeats the whole point of buffering.

Try to restrict flushing to 

+ Just before you expect a crash
+ After having written everything you have to write. As @abotas says, this *should* happen automatically when using decent high-level tools, but I'm not entirely surprised that you can find cases in which it does not.

abotas:
Ok @jacg I think this implementation follows those guidelines exactly.

Also the PR failed `test_in_range_right_shape` with a `FailedHealthCheck` on Travis. I think this is what was being discussed in [#263](https://github.com/nextic/IC/issues/263). Everything passes on my local machine

abotas:
Also, the automatic travis build on my repo passed...

jacg:
Yes, it's clearly something flaky. It's just weird that, since it was committed in April, this is the first time we notice the problem, and it appears twice on the same day.

I've restarted the nextic job, so you should have Travis' approval in a few minutes.

gonzaponte:
There has been a new release of hypothesis today which, they say, shouldn't be visible. But it is being used by Travis. Any idea how to check this?

jjgomezcadenas (commented):
This is a major step forward, that allow us to write not only the PMAPS of the sum but also the PMAPS of the invididual PMTs. This reviewer has eaten peras with the main author and extensively checked the code and tests. Good work!

jjgomezcadenas (approved):

PR #316, opened 2017-08-31
Rename std to var

### LINKED ISSUE #288: Wrong definition of rms in barycenter
In the `barycenter` function coded in [invisible_cities/reco/xy_algorithms.py](https://github.com/nextic/IC/blob/f5aef7b8b58f3173e52ee17eb979122dd1a6fd07/invisible_cities/reco/xy_algorithms.py) we have:
```
std = np.average((pos - mu) ** 2, weights=qs, axis=0)
```
Which is the variance, not the standard deviation. I think this is not intented, but correct me if I am wrong.

gonzaponte:
There are several places in the code where a variable is named ***std*** (for std deviation) when it actually refers to variance. This PR renames every ***std*** to ***var***. Closes #288.

### DIFF
--- a/invisible_cities/evm/event_model.py
+++ b/invisible_cities/evm/event_model.py
@@ -163,7 +163,7 @@ class Voxel(BHit):
 
 class Cluster(BHit):
     """Represents a reconstructed cluster in the tracking plane"""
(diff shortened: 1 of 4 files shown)


jjgomezcadenas (approved):
This fixes a naming issue (introduced by the reviewer). T

abotas:
Merged

PR #320, opened 2017-09-01
fix find_peaks to avoid error when no pmt signal after tmin

abotas:
There was very dumb bug in `find_peaks` that could cause irene to crash for 1 out of a few hundred or few thousand events depending on trigger conditions and irene.conf. find_peaks would raise an error when there was no pmt signal after tmin. 

This PR gives a small fix in python and cython `find_peaks` and is tested in both. The fix has also been checked by running IC-dev, with the fix, in canfranc and verifying that irene no longer crashes.

### DIFF
--- a/invisible_cities/reco/peak_functions.py
+++ b/invisible_cities/reco/peak_functions.py
@@ -198,10 +198,11 @@ def _find_peaks(index, time=minmax(0, 1e+6), length=minmax(8, 1000000), stride=4
 
     peak_bounds = {}
     T = cpf._time_from_index(index)
-    # Start end end index of S12, [start i, end i)
-    i_min = int(time[0] / (25*units.ns))     # index in csum  corresponding to t.min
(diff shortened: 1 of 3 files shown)


gonzaponte:
The protection line you added
```where_after_tmin = np.where(index >= i_min)[0]   # where, in index, time is >= tmin```
will crash if `np.where` returns an empty array. Is it possible for `index` to have all values below `i_min`?

abotas:
It will not crash np.where returns a tuple:
` a = np.arange(10)`
 `np.where(a > 9)`
`(array([], dtype=int64),)`

gonzaponte:
Yes, but trying to do `np.min` on an empty array will also cause a crash.

gonzaponte:
Oh sorry, that is indeed the fix!

abotas:
HA!

gonzaponte (approved):
Important bug fix that was causing crashes when there was no signal in the searching region. Good catch.

PR #321, opened 2017-09-01
fix typo trheshold --> threshold

abotas:
This fixes a typo propagated from pmaps.peak to a few different places
in the repo. The method is tested. 

I think we should change this if it won't cause problems for anyone. @gonzaponte @jjgomezcadenas


abotas:
rebased over most recent PR of nextic / master

gonzaponte (approved):
Just a typo correction. No impact on the code whatsoever.

PR #322, opened 2017-09-04
pmap classes raise exception if initialized with empty dicts

abotas:
Pmap constructors: S1, S2, S2Si, S1Pmt, S2Pmt will now raise an exception when given empty dicts to initialize. Previously if we had an empty s2, we would call and initialize an empty S2Pmt --> S2 --> S12. This is now avoided. 

If the constructor is called to build a peaks out of empty dictionaries, the constructor now raises `InitializeEmptyPmapObject` which must be caught. 

The general flow of things has changed such that where previously if there were no s2 in an event, IC would pass around empty s2, s2si, and s2pmt, Now we just pass around `None`.

This PR merely creates this exception and catches it throughout the codebase where necessary.

### DIFF
--- a/invisible_cities/evm/pmaps_test.py
+++ b/invisible_cities/evm/pmaps_test.py
@@ -2,6 +2,7 @@ import numpy as np
 
 from numpy.testing import assert_equal
 from pytest        import mark
+from pytest        import raises
 
 from hypothesis             import given
 from hypothesis.strategies  import floats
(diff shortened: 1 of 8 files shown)


jjgomezcadenas:
Please add a test (raise the exception, then catch it!) :)!

jacg:
Don't catch it, use the high-level tools we have for the job:

    from pytest import raises

    with raises(InitializeEmptyMapObject):
        <create the situation which you expect to raise the exception>

abotas:
ok @jjgomezcadenas and ok @jacg

abotas:
Done and travis passed

jacg:
The last commit introduces 7 tests which are essentially identical. Please turn them into a single parametrized etst.

abotas:
ok- done

jacg:
Please use `pytest.mark.parametrize` with the callables and corresponding args-tuples as the parameters.

What you have works, but it's an ad-hoc low-level solution to a common problem for which pytest provides a high-level solution.

jacg:
@abotas Also, please check your gmail spam.

jjgomezcadenas (approved):
Home work done following Master's instructions. I hereby raise your level of Karma.

invisible_cities/evm/pmaps_test.py
```diff
@@ -293,3 +298,16 @@ def test_pmt_total_energy():
...
+     (({},  ), S1    ),
+     (({},  ), S2    ),
+     (({},  ), S12   ),
+     (({},{}), S2Si  ),
+     (({},{}), S1Pmt ),
+     (({},{}), S2Pmt ),
+     (({},{}), S12Pmt)))
+def test_pmap_constructors_raise_exception_when_initialized_with_empty_dicts(args, constructor):
+    with raises(InitializedEmptyPmapObject):
+        constructor(*args)
```
jacg: Interesting choice of order: given that in Python syntax the callable precedes its arguments, I would have put the callables *before* the arguments in the parametrization.

This way around works, but it looks a bit weird.

PR #324, opened 2017-09-05
Fix command line

### LINKED ISSUE #307: Command line options not working properly
Some command line options are not working properly. For example:

- -n, --nevents. It does not work because the variable used in cities is named nmax,
- --runall. This one is simply ignored.
- -s, --skip. This one is also ignored.
- -f, --first-event. This one is also ignored, and should be the same as -s.

I have only checked the -n option, but I didn't find anything about the others in the code.

### LINKED ISSUE #308: nmax option not stopping execution right away in Dorothea
When using a value of nmax inferior to the number of events, the program keeps looping over files. This happens *at least* in Dorothea, but I think it is something common to all cities.

jacg:
This work fixes numerous problems with CLI options relating to event selection.

+ `--skip` / `-s` has been removed
+ `--first-event` / `-f` which was a synonym for `--skip`
+ `--n-events-max` / `-n` has been removed
+ `--runall` has been removed

The above have been replaced with a single option `--event-range` with the abbreviation `-e`.

+ `-s 10` / `-f 10` --> `-e 10 last`
+ `-n 10` --> `-e 10`
+ `-s 10 -n 10` --> `-e 10 20`
+ `--runall` --> `-e all`

This should reduce the possibility of ambiguous conflicts between different options.

The config files in the repository have been updated to match these changes of parameter name.

Treatment of `--event-range` (or rather, its predecessors) has been fixed, in the sense that it now seems to work as it is supposed to. However, the way this is done **is in desperate need of abstraction**. As it stands, the implementor of any city `event_loop` has to copy-n-paste the all of

    what_next = self.event_loop_step()
    if what_next is EventLoop.skip_this_event: continue
    if what_next is EventLoop.terminate_loop : break
    self.cnt.increment_counter('n_events_tot')

in order to get it to work. This is completely unacceptable, but no obvious solution comes to mind within the constrains of the Python language in combination with the way in which the cities have been structured.

(If the cities were organized as streams, it would be trivial)

Another problem is testing of these issues. The new implementation comes with tests in the context of Penthesilea, but it is essentially untested in the other cities. As it's all copy-n-paste, there is no way of testing the functionality in isolation and reusing the tested functionality.

Testing it over and over again in the context of every city would not only be tedious and error-prone, but also costly in terms of test run time: testing this functionality requires a city to process actual events, in some cases all the events available in the data files, such tests would be very slow in some cities.

The above points close #307.

Further work in this PR closes #308.

The latter is untested because

+ it is not clear how to test how many files were opened, without introducing complications into the code
+ such tests would be very slow in general


jacg:
Moving this over from #307 ...

> The tests are still using `n_events_max`. Am I missing something?

I've removed the remaining vestiges of `n_events_max`. The problem mentioned in #307 goes away for `test_irene_runinfo_run_2983`, but still remains for `test_irene_electrons_40keV` and `test_isidora_electrons_40keV`.

It also fails to solve the problem for 2 new Penthesilea counter tests, which I didn't mention earlier.

gonzaponte:
The fact that the check is *at the end* of the loop implies that n+1 events are processed. In the failing test, for example, it requires to run over the interval [0,2), but it will only realize it is processing event number 2 (i.e. three in total) after the job is done. However, this is not the cause of the problem, because you are saying that `nactual == 1`, not 2, which is the required condition for the loop to stop. I will keep looking for the root of the problem.

gonzaponte:
I couldn't see the output of travis yesterday, but it seems the penthesilea tests are failing due to `n_events_tot` being 0.

jacg:
> I couldn't see the output of travis yesterday, 

You could always fetch the branch and run the tests locally.

> but it seems the penthesilea tests are failing due to n_events_tot being 0.

Yes, I alluded to this 3 comments up, right at the end. These are completely new Penthesilea tests, written just for the purpose of testing the event range.

In the meantime I've got to understand quite a bit more about this whole mess. As I said earlier, it's **really** horrible, and desperately needs abstracting.

First I will make it work based on the ad-hoc, error-prone boilerplate repetition style we have in the code right now. Once that is working, committed and available to everyone, I'll start thinking about how to do it properly.

gonzaponte:
> You could always fetch the branch and run the tests locally.

I know, but I was hoping that travis would eventually run.

> Yes, I alluded to this 2 comments up, right at the end. These are completely new Penthesilea tests, written just for the purpose of testing the event range

There is a chance that `n_events_tot` is 0 because the events in the test file are being filtered. I don't think `n_events_tot` is being used properly. I think it is actually the same as `n_events_out`, as it is not being updated when the events are filtered out.

gonzaponte:
I think I've found something. In `test_irene_electrons_40keV`, when doing `conf.update`, you use the keyword `events_range` instead of `event_range`. The same happens for `test_isidora_electrons_40keV `.

jacg:
> I don't think `n_events_tot` is being used properly.

Yes, I got the impression that these counters and their use are a complete mess. They've caused me a lot of confusion. Something needs to be done about this ...

jacg:
> when doing conf.update, you use the keyword `events_range` instead of `event_range`.

Well spotted. That makes the last failing tests pass.

Now *there's* an argument for static typing!

Actually, I've been mulling over ideas for spotting unexpected names in the configuration: lots of mistakes are bound to creep in with the large dynamic namespaces we have in the configuration, and performing more automatic sanity checks would probably save us a lot of time and pain down the years.

Edit: this is now addressed in #333: Detect unknown parameters.

gonzaponte:
And the failures in penthesilea are almost certainly because the events are filtered out. The events in `KrMC_pmaps.h5` are intentionally weird (missing S1 or missing S2). That was so we could test (ages ago) that we were running properly on this kind of events. Do we need a new PMaps file? I can take care of it.

jacg:
No, the failures in Penthesilea were because of the lack of a counter which keeps track of how many events have been read in (and maybe skipped). This is now fixed.

As far as I am concerned, this is a first-order fix for #307. #308 has not been addressed yet, but the groundwork has been laid and I think I have a fair idea how to proceed. What I don't know is how to test #308: I need multiple data files. Any ideas?

Should we merge this first with just a fix for #307, or inculde the fix for #308 here too?

jacg:
This PR is now up for review.

Please read the updated top comment in this PR, which is the currently proposed commit message for the merge.

Should #308 be addressed in a separate PR, or should the fix be included here?

jacg:
The fix for #308 has now been added.

+ It's pretty simple
+ It's untested
  + it would require costly, multiple-file tests in general
  + it seems difficult to test how many files were opened

gonzaponte:
I am taking a look, but it will take me some time. I will be posting questions or issues as I review more commits.

What is the point of adding a dummy element to `Configuration._file_stack` [here](https://github.com/nextic/IC/pull/324/commits/04d69f977b3deae1413200030b506239651d641e)? It doesn't seem obvious from the commit.

gonzaponte:
In [70680f8](https://github.com/nextic/IC/pull/324/commits/70680f87fdb4e7aa6813d69f742f702126f0d3e2#diff-bd3ab46d81ccf91e4b8a2a8ab1ada3e2R108), isn't it simpler and better `ns = ReadOnlyNamespace(**self._data)`?

jacg:
> What is the point of adding a dummy element to `Configuration._file_stack`?

To ensure that `_current_filename` works, even if no filenames have been opened yet. It enables testing `Configuration` without the need for any config files.

jacg:
> isn't it simpler and better `ns = ReadOnlyNamespace(**self._data)`?

No, as the comment between those two lines suggests, the namespace cannot be read-only until it has been initialized. It starts off in a writable state so that it can be populated, only *afterwards* can it be set into a read-only state.

Try it, and you will see the 41 tests fail.

gonzaponte:
Ok, I wasn't understanding it properly.

jacg:
Feel free to improve the comment, if you can make it clearer without writing a long essay in the middle of the code.

gonzaponte:
I will finish this tomorrow. But some tiny comments:
- Are we keeping the `EventLoop` class in `base_cities` for now or do you want to move it somewhere else?
- (pure cosmetics) Maybe [here](https://github.com/nextic/IC/pull/324/files#diff-1ce385d5350d35f26f1d70f787aed59aR139) we can sort the `len(er) == X` "ifs" in ascending order?
- Do you want to fix [this](https://github.com/nextic/IC/pull/324/files#diff-1ce385d5350d35f26f1d70f787aed59aR217) before merging?

jacg:
> Are we keeping the `EventLoop` class in `base_cities` for now or do you want to move it somewhere else?

I won't have time to do it. I'm happy for someone else to find a sensible place and move it. It could also do with a nicer `__str__` / `__repr__`. Remember that

+ it's really two symbols: `all` and `last`
+ they appear in the config report, so it would be nice for the representation to make sense in that context.

> (pure cosmetics) Maybe here we can sort the `len(er) == X` "ifs" in ascending order?

That's how I had it originally, but I decided to group them by result, rather than by input. I have no strong feelings either way. Feel free to change it if you think it's better.

> Do you want to fix [max event number reporting] before merging?

Again, I won't have time before I disappear. Feel free to do it, or to leave it until later. I think it's important that the fix to the broken command line go in ASAP.

Additionally, I have a couple of PRs in the pipeline which are waiting for this to me merged ...

jjgomezcadenas:
> Are we keeping the EventLoop class in base_cities for now or do you want to move it somewhere else?

Keep event loop in base_cities. Ain't broken, don't fix it. 


> > Do you want to fix [max event number reporting] before merging?

> Again, I won't have time before I disappear. Feel free to do it, or to leave it until later. I think it's important that the fix to the broken command line go in ASAP.

Agree

> Additionally, I have a couple of PRs in the pipeline which are waiting for this to me merged ...

So let's move this fast

jjgomezcadenas:
@gonzaponte we need to move this fast.

gonzaponte:
First thing in the morning.

gonzaponte:
I mentioned this a while back:
> The fact that the check is *at the end* of the loop implies that n+1 events are processed

This is still happening for `irene` and `isidora`. I can fix it if your time is limited.

gonzaponte:
That is my last comment, by the way. Apart from that, I am ready to approve.

jacg:
> I mentioned this a while back:
>
> > The fact that the check is at the end of the loop implies that `n+1` events are processed
>
> This is still happening for irene and isidora. I can fix it if your time is limited.

Elsewhere, I (@jacg) wrote

> Is there any good reason why `event_loop` in diomira, penthesilea, dorothea (and maybe others) performs the `max_events_reached` check and increments the counter *before* processing the event,
while in irene, isidora (and maybe others) this is done after processing the event?
>
> I suspect it's just an erroneous inconsistency introduced by requiring the implementor to recode this boilerplate yet again.

To which @jjgomezcadenas replied

> The reason why the max_event_reached is on top of the loop for IRENE and on the bottom for Doris and Patty is related with the fact that Irene does not filter data (so the position of the condition at the top or the bootoom of the loop is irrelevant) while Doris and Patty do (so they implement the condition at the bottom, since we intend max_event_reached on processed events, not read events). One could move the condition to the bottom of the loop in IRENE, but I don't think there is any inconsistency there.

Which, given that I was buried 10 deep in a stack of active tasks at the time, I simply took to mean "it's correct as it is".

Reading it a little more carefully now:

Of course, I **strongly** disagree with @jjgomezcadenas about there not being an inconsistency. The inconsistency might be inconsequential in terms of semantics of the code, but having half of the checks done at the top and half at the bottom most certainly **is** inconsistent, and it led to me wasting time on wondering about it when I first encountered it (and made a few important things drop out of my short-term memory) and it is now wasting our time once more on discussing it again.

Moral of the story: even if an inconsistency has no semantic consequences for the code itself: get rid of it anyway, in order to avoid humans wasting precious resources on wondering about whether it matters or not!

All of this just lends more weight to my point: this loop-event-range control is in desperate need of a reusable abstraction.

Now, as to what to do about it in this right now ... if @gonzaponte and @jjgomezcadenas confirm that they both think that moving the checks to a consistent position will not change the semantics of the code, then it should be done ASAP. If such confirmation arrives this morning, I can fix it. If not, it's up to you.

gonzaponte:
Sorry if I am not understanding it correctly, but my point is that it is not only an inconsistency but a bug! This is causing these cities to run a different number of events than the one requested. Which makes me wonder how are the tests even passing!

jacg:
> my point is that it is not only an inconsistency but a bug! 

And my point is that I infer from @jjgomezcadenas reply to my query, that he thinks it is *not* a bug. Which is why I want the two of you to come to agreement on this. You two have far more background information available on this than I do.

jacg:
> Which makes me wonder how are the tests even passing!

If the tests were written with buggy expectations, then it's perfectly possible that buggy code passes them. The tests that check expectations of number of processed events across the cities are ad-hoc, so don't expect guaranteed consistency across what these tests expect of the different cities!

jjgomezcadenas:
To cut a long story short: let's fix the incosistency or bug. I assume that this leads to lower my level of karma by 10 units. Thanks.

jacg:
Heheheh ... it's not that easy! First we have to understand whether it is a bug! Because moving the check, as @gonzaponte implies, will probably make some tests fail. And if it doesn't, we'll have to scratch our heads even more.

gonzaponte:
It does, I am working on it.

jacg:
And with that, gentlemen, I bid you farewell.

(I might have the odd minute here and there, but I'm essentially out of the picture for the next week.)

jjgomezcadenas:
Irene:

        for evt in range(NEVT):
            DO STUFF BUT DO NOT FILTER DATA
            if self.max_events_reached(self.cnt.counter_value('n_events_tot')):
                break
            else:
                self.cnt.increment_counter('n_events_tot')

IS EQUIVALENT TO:

        for evt in range(NEVT):
            if self.max_events_reached(self.cnt.counter_value('n_events_tot')):
                break
            else:
                self.cnt.increment_counter('n_events_tot')
            DO STUFF BUT DO NOT FILTER DATA
            
E.G, if you do not fileter data away, the orders of the factor does not alter the product.

if you filter data (e.g, Doris, Patty), then the order of factors is important. 

Thus, I believe the code in irene is not buggy (but I agree is inconsistent).

jjgomezcadenas:
Well, you go but you leave us with a monster (see what you have done of @gonzaponte, formerly a nice guy, now well on its way to become another iron fist...)

gonzaponte:
> E.G, if you do not fileter data away

Irene does filter data:
```
            if not self.check_s12(s12sum): # ocasional but rare empty events
                self.cnt.increment_counter('n_empty_events')
                continue
```

But independently of that, because the loop begins at 0, if you check the number of events processed *at the end of the loop* you will have processed 1 event, but the counter is still 0! Which causes the mismatch. Moreover, the counter `n_events_tot` was being used as `n_events_out` because it was only updated *after* the filters.

I am almost done with this, anyway.

gonzaponte:
I can't push to this PR. Is this a configuration issue, @jacg?

jacg:
> I can't push to this PR.

What are the symptoms?

I pushed a rebase on top of #329, just after it was merged into master.

So you might need to fetch the latest version of the branch, and rebase locally on top of that, before pushing.

gonzaponte:
Yup, that was it. I thought I did the fetch after the rebase.

gonzaponte:
Can someone review these 4 commits?

jjgomezcadenas:
Done, I am happy with the last 4 commits. But I guess it's you (not me) who needs to approve this PR

gonzaponte:
> But I guess it's you (not me) who needs to approve this PR

Yes, but I guess those four commits needed some extra pair of eyes.

gonzaponte (approved):
This PR addresses multiple issues:

- Reduce CLI options concerning event selection to a unique option.
- Improve event loop control
- Fix the file loop control when asked to loop over a limited number of events.

Multiple tests have been added, and some others fixed. The issues that arose have been heavily discussed and reviewed. Great work by the master. Approved.

gonzaponte:
Closes #336.

PR #325, opened 2017-09-06
allow irene to find ipmt pmaps

abotas:
This pr will allow irene to make use of the ipmt pmap finders and writers in order to save ipmt pmaps. I've run this irene in my notebook a few times and everything looks good so far.

1)  In order to tell irene to write ipmt pmaps (or not) I had to put a flag `compute_ipmt_pmaps` in pmap_city.conf. This means that previous irene config files will not work without first adding the flag. This may be kind of annoying for people running productions @jmbenlloch @jerenner @gonzaponte. If there is a better way of doing this, please let me know, and I will implement it. The only alternative I could think of would be to create another city, but that sounds potentially even more annoying for you guys.

2) I did some time/memory profiling for on Kr data. From run 4446 with `thr_sipm = 1pes` and 'thr_sipm_s2 = 2pes' and found after running 1k events:
speed
1.92 evt / sec with IPMT
2.06 evt / sec without 
file size
2.37 x larger than without IPMT
So, time does not appear to be an issue, but the file size does grow substantially...  

3) @jjgomezcadenas I don't remember if we decided we wanted to put the ipmt pmaps in a different group, called `IPMT` in the hdf5 file... did we?

### DIFF
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -55,6 +55,7 @@ from .. reco.xy_algorithms      import corona
 from .. evm.ic_containers       import S12Params
 from .. evm.ic_containers       import S12Sum
 from .. evm.ic_containers       import CSum
+from .. evm.ic_containers       import CCWf
 from .. evm.ic_containers       import DataVectors
 from .. evm.ic_containers       import PmapVectors
 from .. evm.ic_containers       import TriggerParams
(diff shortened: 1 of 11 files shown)


jjgomezcadenas:
1. Sad as it is, I think we need to introduce the flag compute_ipmt_maps. Definitively NOT a new city (see separated discussion with the master @jacg on that).

2. Something has gone really wrong with our new version of the cities, it seems. IRENE use to run in 0.5 seconds per event.

3. Size is acceptable, I think. PMAP files are rather small, so that the increase is tolerable. What do you think, @jmbenlloch ?

4. If we have not put the IPMTs in a different group probably we don't need to bother.

abotas:
@jjgomezcadenas the speed is the same! .5 sec/event. I gave event / sec

jjgomezcadenas:
OK, then all is fine! 
But why is Travis working again??

abotas:
... I have no idea .. my last push was more than an hr ago

abotas:
it looks like it's stalled: 
https://travis-ci.org/nextic/IC

abotas:
I think this is done. I've looked through things in the notebook and it looks good.

I run over a few k events and nothing crashes, things look as expected.

jjgomezcadenas:
OK, Travis still stale...

abotas:
but travis does not want to work...

abotas:
Travis builds once, passes, then without new commits decides to build again, and stalls... 
https://travis-ci.org/nextic/IC/pull_requests
Build #940 passed and then it started to build again for some reason and stalled???

abotas:
Now it has passed

jjgomezcadenas (approved):
Major piece of work, with many tests and extensive data running to back it up. Again, I rise your level of Karma.

PR #326, opened 2017-09-06
Simplify counters

jacg:
This is an issue which should eventually be upgraded to a PR. As GitHub has removed the ability to turn issues into PRs, I am starting it as a PR, even before any code has been written.

Elsewhere, @jjgomezcadenas says

> As for the counter class, the idea was to have a generic counter and that was the best I could come up with. I am sure that the master can come with much prettier implementations, but old dog does not learn new tricks... (or it takes it a long wile).

As far as I can tell, it's just a namespace, which carries around a bunch of things which happen to be counters.

It has one useful feature: the ability to increment a whole bunch of counters in one go.

It has a lot of demerits, mostly to do with unnecessarily complicating things and introducing a new syntax for doing stuff that is simply done in standard Python. Which of the following 2 lines do you prefer?

    self.cnt.increment_counter('name')  # Counter
    self.cnt.name += 1                  # Normal way

How about implementing it as a subclass of `Namespace`, whith an added method for incrementing multiple counters. Though I'm not sure that it's worth it, over a plain namespace.

### DIFF
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -75,7 +75,7 @@ from .. sierpe                  import blr
 from .. sierpe                  import fee as FE
 
 from .. types.ic_types          import minmax
-from .. types.ic_types          import Counter
+from .. types.ic_types          import Counters
 from .. types.ic_types          import NN
 from .. types.ic_types          import xy
 
@@ -111,8 +111,8 @@ class City:
         4. provides access to the data base.
         """
 
-        self.cnt = Counter()
-        self.cnt.init_counter('n_events_for_range')
+        self.cnt = Counters(n_events_for_range = 0)
+
         conf = Namespace(**kwds)
         self.conf = conf
 
@@ -177,7 +177,7 @@ class City:
         self.run()
         t1 = time()
         dt = t1 - t0
-        n_events = self.cnt.counter_value('n_events_tot')
+        n_events = self.cnt.n_events_tot
         print("run {} evts in {} s, time/event = {}".format(n_events,
                                                             dt,
                                                             dt / n_events))
@@ -276,8 +276,8 @@ class City:
                   .format(evt, n_events_tot))
 
     def event_range_step(self):
-        N = self.cnt.counter_value('n_events_for_range')
-        self.cnt.increment_counter('n_events_for_range')
+        N = self.cnt.n_events_for_range
+        self.cnt.n_events_for_range += 1
 
         if isinstance(self.last_event, int) and N >= self.last_event:
             # TODO: this side-effect should not be here
@@ -289,7 +289,7 @@ class City:
             return EventLoop.skip_this_event
 
     def event_range_finished(self):
-        N = self.cnt.counter_value('n_events_for_range')
+        N = self.cnt.n_events_for_range
         return isinstance(self.last_event, int) and N >= self.last_event
 
 
@@ -708,7 +708,7 @@ class KrCity(PCity):
                     c = clusters[0]
                 else:
                     cQ = [c.Q for c in clusters]
-                    self.cnt.increment_counter('n_events_more_than_1_cluster')
+                    self.cnt.n_events_more_than_1_cluster += 1
                     print('found case with more than one cluster')
                     print('clusters charge = {}'.format(cQ))
 
--- a/invisible_cities/cities/diomira.py
+++ b/invisible_cities/cities/diomira.py
@@ -57,8 +57,8 @@ class Diomira(MonteCarloCity):
         super().__init__(**kwds)
         conf = self.conf
 
-        self.cnt.set_name('diomira')
-        self.cnt.init_counters(('n_events_tot', 'nevt_out'))
+        self.cnt.init(n_events_tot = 0,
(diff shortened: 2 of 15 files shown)


jjgomezcadenas:
Clearly the implementation using a namespace is better. 

NN is used (instead of nan) because is hard to write and readback nans in pytables (or at least I had issues with this).

jacg:
> hard to write and readback nans in pytables (or at least I had issues with this).

How about `float('-inf')`?

Is its purpose just to be stored, or will it ever be ordered WRT any other numbers? (You realize, if you say, 'no, never!' then it's almost guaranteed that next month someone will write some code that uses it in an order :-)

jjgomezcadenas:
as usual, the master solution shames that of oldDog. 
I take that we can simply write NN = float('-inf') and keep NN notation (is already in a number of places).

NN is always used when we cannot compute a number, e.g, when we cannot find a value for a hit, but we still want to fill the hit. I believe we won't ever order WRT other numbers (and also I agree with you that somebody will come up tomorrow with some code trying to order it, but to the best of my knowledge this is just let me write/read, keep track of NNs...)

jacg:
> I take that we can simply write `NN = float('-inf')` and keep NN notation

Probably.

Have you checked whether pytables is happy with `float('-inf')`?

Do I have your approval to merge the change to master without messing around with PRs, if it passes our existing test suite?

jjgomezcadenas:
Not sure if pytables is happy with float(-inf') need to check, and need a test to ensure this.

jjgomezcadenas:
@gonzaponte, @abotas  this needs review, and in turn, we need to check that we can write and read
float('-inf') from/to pytables (e.g., we need a test). Can you guys handle it? I am afraid that I will be quite useless for development for the next few days (maybe weeks).

jacg:
>  this needs review

This doesn't need review, as there's nothing here yet!

jacg:
> I am afraid that I will be quite useless for development for the next few days (maybe weeks).

Around tomorrow midday I'm mostly offline for at least 3 weeks (with some availability in the middle week).

jjgomezcadenas:
Sorry. What I meant was that we need to check if float('-inf') works well with pytables. I also thought @jacg was going to implement the change from class to namespace. 

To summarize, I can try to do a few things:

1) Change to namespace.
2) Implement the change of NN if pytables happy.
3) abstract common code in Penthesilea/Dorothea.
4) formalize better the use of counters across the cities.

All this can be carried out in this PR. Agree?

jacg:
> 1. Change to namespace.
> 2. Implement the change of NN if pytables happy.
> 3. abstract common code in Penthesilea/Dorothea.
> 4. formalize better the use of counters across the cities.
>
> All this can be carried out in this PR. Agree?

The `NN` issue has nothing to do with the counters. (I just put the comment there, because `NN` was in my face as I made the dummy commit.) Not sure what no. 3 is doing here.

I can have a go at no. 1, including propagating the consequences to the rest of the code. Can't guarantee that I'll finish it before going offline.

No. 2 should be a separate PR. I'll won't go anywhere near it, because others are better placed to check what pytables thinks of it.

No. 4 sounds like it needs some calm and careful thought, and I am (we are?) in panic mode at the moment.

jacg:
Oh, this work also needs to be done on top of #324.

jacg:
> Oh, this work also needs to be done on top of #324.

And if I could start directly on top of #333, that would be very helpful. So that's what I will do.

jacg:
This is ready for a first look.

I have dropped the ability to name the counter namespace. It didn't seem to be very useful: the name only ever appeared on stdout when running a city, and the counter's name always matched the city name.

The only differences between the subclass and the original `Namespace` are

+ It's called `Counters` rather than `Namespace` and that name appears when it is printed out.
+ A `set` method allows setting a bunch of values at once. This saves repeating `self.cnt` at the cost of adding a `.set` and a bunch of commas.
+ An `init` method, which is just like `set` but checks that the names being set are new additions to the namespace, rather than overwriting an existing value.

I am not convinced that what `Counters` gives us is worth it: we could just use plain old `Namespace` instead.

Because the implementation is utterly trivial, and I'm not sure that it's worth keeping, I have not written any tests. That is left as an exercise for anyone who cares.

jjgomezcadenas:
I am getting confused.

This PR is getting huge and you are disappearing soon. Are we reviewing/approving this any time soon? There are substantial changes and a lot of data production is gonna happen in the next month. We need to decide if we move this into main now or wait...

jacg:
> This PR is getting huge

It's big, but that's because it's tedious and touches a lot of different places. But it's not complicated.

Perhaps you are looking at the 400 commits that are listed here, and are forgetting that most of those do not belong to this PR, but to #333 and #324 on which it is based.

#324 is the most complicated of the lot, has the biggest impact, and is the most important. And there are 3 other PRs based on it!

> Are we reviewing/approving this any time soon? 

I suggest you hurry up and review #324.

jacg:
Please review this.

jjgomezcadenas (approved):
As extensively discussed, subclassing Namespace is a more elegant and concise way of producing counters than the original Counter class.

PR #327, opened 2017-09-06
Use event numbers from input files for MC

jmbenlloch:
We were computing a new event number in Diomira because in the
previous MC production, each file restarted the event counter to
zero. This is not the case anymore, each file has a unique event
number. Since this was a patch intended to use that old production, it
could be removed now.


jjgomezcadenas (approved):
Fixes an essential bug that has ruined the previous production. Need to be urgently approved in order to re-launch production.

PR #330, opened 2017-09-07
prevent crash in search for s2si when s2 is None

abotas:
This morning I was running the ipmt-irene on data that triggered on s1 and found a small bug in the ipmt irene code from yesterday.. 

Before irene would catch the exception when it found no s2, but then would still ask for the s2.s2d to compute the s2si. This has been fixed and corresponding tests have been implemented. 

I also checked against the that irene no longer crashes when running on this same s1 data.


gonzaponte (approved):
Obvious fix with test included. Approved.

PR #332, opened 2017-09-07
Fix reporting of config overrides on CLI

jacg:
The configulation printer sometimes erroneously reported that some option has been set in a config file, when in reality it was set on the CLI. This has been fixed.


jacg:
Please review this.

jjgomezcadenas (approved):
Needed technical fix

PR #333, opened 2017-09-07
Detect unknown parameters

jacg:
This is based on top of the as-yet-unmerged #324. The first significant commit not present in #324 is the one titled 'Introduce detection of unexpected parameters', so you should ignore an commits before that one when looking at this PR.

Given that we now treat the parameter namespaces completly dynamically with `**kwds`; typos, obsolete parameter names and other mistakes pass silently. This leads to lots of annoying and difficult to track bugs and confusion.

This PR introduces detection of unexpected parameters:

+ Each city may provide a class attribute named `parameters`, being a sequence of parameter names that are recognized at this level, but not recognized by the superclass.
+ At construction time, the construction arguments are checked for the presence of parameters which have not been declared at the city's level or its superclasses. If an unexpected parameter is found, an `AssertionError` is raised. (This should probably be replaced by a new IC exception.)

Please check through the all the `parameters` in each class to see whether they have been listed at the appropriate level.

### DIFF
--- a/invisible_cities/cities/city_test.py
+++ b/invisible_cities/cities/city_test.py
@@ -8,7 +8,7 @@ package: invisible cities. See release notes and licence
 last changed: 01-12-2017
 """
 
-from   pytest import raises
+from pytest import raises
 
 from .. core.exceptions import NoInputFiles
 from .. core.exceptions import NoOutputFile


jacg:
Please review this.

jjgomezcadenas:
@jacg I have stated looking at this branch, but I guess you (or any one you delegate) needs to resolve parameters first, right?

I have a doubt/comment. The technique proposed here is to specify in the code of each city the list of parameters that are the city defines. In a sense we are "statically tipying" our parameters. The point is that we win in safety but we "loose" in the sense that we loose the ability to define a new parameter of a city by simply defining in in the config file. Now we need to change the config file **and the code**. Are we sure that this is what we want?

jacg:
> In a sense we are "statically tipying" our parameters.

Indeed.

> The point is that we win in safety but we "loose" in the sense that we loose the ability to define a new parameter of a city by simply defining in in the config file. 

Exactly.

> Now we need to change the config file **and the code**.

Only when introducing a parameter into the config file that did not exist before. Introducing such a parameter imples changing the code *anyway*, because if you don't change the code, then there is nothing in the code that could possibly make use of the parameter!

> Are we sure that this is what we want?

Having seen the hundreds of mistakes and bugs that were silently passing in our code, because of the lack of such checks, while doing the work on the last few PRs, my conclusion is an unequivocal: "Yes, we absolutely **do** want this".

There were lots of misspelt parameters, outdated parameters, obsolete parameters, all over our config files and our tests. All of these mistakes passed silently for months! Reading the code or the config files gave the impression that something was being said, while in practice it was being silently ignored. I am certainly convinced that we do want this.

> but I guess you (or any one you delegate) needs to resolve parameters first, right?

I don't understand what you mean by 'resolve parameters'. You guys certainly need to go through the parameter declarations, and verify that I put them in the right places and at the right levels.

jjgomezcadenas:
I meant rebase and solve the test failure...

Enviado desde mi iPhone

jjgomezcadenas (approved):
Great work. See the discussion.

jjgomezcadenas:
@jacg there are conflict to be rebased in this branch. Will you have the time to fix them or shall we try? Trhanks!

jacg:
> there are conflict to be rebased in this branch.

This is probably some artefact of the Git-identity-rebase. I have already merged this branch, and it's already in `nextic/master`. Somehow GitHub doesn't recognize that as having happened.

PR #334, opened 2017-09-08
Index pytables after populating them

abotas:
This PR addresses the `create_index` issue. The Pytables node manager is apparently buggy. Francesc Alted suggested we: 
1) Populate all our pytables and close the write file. 
2) reopen the file and create indices. 

This is a little strange since after we have closed the writefile we have to reopen the file and remember which columns in which tables to index. To do this: 

- I've modified the writers -- with the exception of `xy_writer` whose pytable @gonzaponte says does not need to be indexed -- so that the writers mark a column or columns in their various tables to be indexed, instead of actually indexing them.
- Then a new method `City.index_tables` walks through the nodes in the write file that are pytables tables,  and checks the tables for columns that need to be indexed and indexes them.

There is one question we need to answer:
Where do we call `City.index_tables` ? I think we should call it in `City.end()` **as long as all cities use the `City.end()` method**. The other option I can think of is to put it at the end of `City.run()`. For now it is implemented in `City.end`

### DIFF
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -195,23 +195,36 @@ class City:
 
         with tb.open_file(self.output_file, "w",
                           filters = tbl.filters(self.compression)) as h5out:
-
             self.write_parameters(h5out)
             self.writers = self.get_writers(h5out)
             self.file_loop()
-            h5out.flush()
 
     def end(self):
         """Postoprocessing after city execution:
         1. calls the end method of the daemons if they have been invoked.
         2. prints the counter dictionary
         """
+        self.index_tables() # index cols in tables marked for indexing by city's writers
+
         if hasattr(self, 'daemons'):
             for deamon in self.daemons:
                 deamon.end()
 
         print(self.cnt)
         return self.cnt
+
+    def index_tables(self):
+        """
+        -finds all tables in self.output_file
+        -checks if any columns in the tables have been marked to be indexed by writers
+        -indexes those columns
+        """
+        with tb.open_file(self.output_file, 'r+') as h5out:
+            for table in h5out.walk_nodes(classname='Table'):        # Walk over all tables in h5out
+                if 'columns_to_index' not in table.attrs:  continue  # Check for columns to index
+                for colname in table.attrs.columns_to_index:         # Index those columns
+                    table.colinstances[colname].create_index()
+
 
     def display_IO_info(self):
         n_events_max = 'TODO: FIXME'
--- a/invisible_cities/io/hits_io.py
+++ b/invisible_cities/io/hits_io.py
@@ -81,8 +81,8 @@ def hits_writer(hdf5_file, *, compression='ZLIB4'):
                              fformat     = HitsTable,
                              description = 'Hits',
                              compression = compression)
-
-    hits_table.cols.event.create_index()
+    # Mark column to index after populating table
+    hits_table.set_attr('columns_to_index', ['event'])
 
     def write_hits(hits_event):
         hits_event.store(hits_table)
--- a/invisible_cities/io/indexation_test.py
+++ b/invisible_cities/io/indexation_test.py
@@ -14,7 +14,7 @@ from .  pmap_io            import     pmap_writer
 
 
 
-@fixture
+@fixture(scope="session")
 def init_city(ICDIR, config_tmpdir):
         conf = configure(('city ' + ICDIR +  'config/city.conf').split())
         file_out = os.path.join(config_tmpdir, "empty_file.h5")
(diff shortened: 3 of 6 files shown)


jjgomezcadenas:
I like the solutions proposed. I will try to review this later today (tonight most likely)

abotas:
Ok perfect! 
I've run irene with ipmt / dorothea / penthesilea over 5k Kr events and it all looks good. `index_tables()` is tested.

If this looks good to everyone I think it is ready to be merged.

gonzaponte:
Mmm, I don't like setting an attribute to the pytable object.

What about this:
- Writer-builders to return both the writer and the columns to be indexed (i.e. return 2 values).
- This is caught in the `run` method of the cities, which is common to all of them.
- Call `index_tables` after `file_loop` in `run`, while the output file is still open.

Moreover, I don't like the fact that you loop over all tables in the file. I am sure we can find a way of iterating over the tables we are actually using in the city. For instance, the writer-builders can return the actual column, so you only need to call the `create_index` method (assuming we use something like I propose above).

abotas:
First, I should make clear, we currently close the write-file after populating it and then reopen it to index the tables because of a bug in pytables that makes this solution more stable than indexing simply after populating the write-file before closing it. This we can and should reverse if pytables fixes its bug.

abotas:
I don't think iterating over all the tables in the file is a problem, because there are like 4-6 tables and most or all of them need to be indexed. And we have to find the tables somehow. 

We could avoid setting an attribute to the table marking which column or columns to index. But I think we would need to:
- have each writer return 1) the write function, 2) the table name, 3 the column(s) to index
- accumulate these in a dictionary `{table name : [cols]}` (across all writers)
- either put `index_tables` in run() and pass it the dictionary, returned by `get_writers()` or make the dictionary an attribute of the city, and have `index_tables` make use of it that way.

I just don't think this is better. Not more useful and i don't think more elegant.

I agree that setting an attribute to the table to mark which columns to index is strange and maybe shady. But it's because there's a bug in pytables.  I just don't see the need to avoid setting an attribute because if nothing else, it serves as a log of what col was supposed to be indexed. Also, this is already implemented!

@gonzaponte I encourage you to implement something that avoids setting an attribute if you want! If not, then if others agree that this is to be avoided, I will do it

jjgomezcadenas:
I think the solution is good enough as it is. By the way, an statement

I don't like setting an attribute to the pytable object.

Should be backed up with a reason (what is wrong about setting arguments to pytables?)

In turn, the proposed solution seems somewhat complicated and perhaps error-prone. 

Most of all, I think that the "if ain't broken don't fix it" principle applies here. The proposed solution works and has been tested and it solves a but that is not ours.

Thus, from my point of view, is OK.

I propose that @abotas rebases and then I will approve.

gonzaponte:
> what is wrong about setting arguments to pytables?

It is very obscure. You are implicitly returning something of importance. For me, it should be explicitly given.

> I just don't think this is better. Not more useful and i don't think more elegant.

It doesn't seem that tedious to be, but:

> @gonzaponte I encourage you to implement something that avoids setting an attribute if you want

I don't have the time and @jjgomezcadenas seems to be ok with it, so let's move on.

jjgomezcadenas:
guys, tests are failing... what is going on, @abotas ?

gonzaponte:
Wrong call to `Configuration.as_dict`. You need to rebase and, since that class changed in the last PR, I think you only need to remove the `.as_dict`.

abotas:
Ok, rebased. Tests pass locally

gonzaponte:
The City in `test_hits_table_is_indexed` could be a fixture. What do you think, @abotas?

abotas:
You're right. One second

abotas:
made it a fixture. tests pass locally. I'm going to sleep. We can approve/merge tonight and then @jmbenlloch can start MC irene with ipmt productions whenever he's ready! If we are in a hurry @gonzaponte or whoever can push whatever other changes!

gonzaponte:
This is up for review, @jjgomezcadenas. Let me know if you are unhappy about anything.

jjgomezcadenas:
@abotas @gonzaponte I would like to see the tests passing Travis, but it seems stale. As soon as I see this I will approve this.

gonzaponte:
You can already approve it. It can't be merged without Travis giving the ok
and the tests pass in my machine.

jjgomezcadenas (approved):
Extensive and thorough work. Good job.

jjgomezcadenas:
@abotas @gonzaponte tests still failing

gonzaponte:
forgot last push. My bad.

jjgomezcadenas:
@abotas @gonzaponte @jacg @jmbenlloch any one to merge approved PRs?

gonzaponte:
They are already merged.

PR #336, opened 2017-09-08
Fix event number for MCTracks in irene

jmbenlloch:
Irene was passing its own event counter to mc_writer instead of using
the original event number. Due to this, the MC production is not
working as irene was not writing MCTracks. This commit fixes the problem.

### DIFF
--- a/invisible_cities/cities/irene.py
+++ b/invisible_cities/cities/irene.py
@@ -87,7 +87,7 @@ class Irene(PmapCity):
             write.pmap         (event, *pmap)
             write.run_and_event(self.run_number, event, timestamp)
             if self.monte_carlo:
-                write.mc(mc_tracks, evt)
+                write.mc(mc_tracks, event)
 
     def check_s12(self, s12sum):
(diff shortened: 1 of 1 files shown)


gonzaponte:
This is fixed in #324, which will be approved shortly

jmbenlloch:
> This is fixed in #324, which will be approved shortly

Ok, then I will close this PR when #324 is merged.

jmbenlloch:
This is still unfixed in #324. I have changed the code again to use the proper variable.

gonzaponte:
Correct. My mistake.

gonzaponte (approved):
This PR fixes the event number associated with the MC tracks. Good catch. Approved.

jjgomezcadenas:
@jmbenlloch, @gonzaponte, @abotas Travis seems stale. How can we fix it?

gonzaponte:
Travis seems to take very long lately to queue our jobs. Not sure if we can
actually do something.

PR #340, opened 2017-09-18
penthesilea test

abotas:
This adds a penthesilea test @jjgomezcadenas wanted (and we didn't have). It makes sure penthesilea runs properly over 9 events


gonzaponte:
- Shouldn't be the `event_range` parameter more meaningful (you are running less than 50 evts, I think)? And shouldn't you check that the actual number of events run is the correct one?

- Do you need to override all those parameters?

Looks good to me.

abotas:
You're about `event_range` -- I'll change it now

I check that the correct number of events is run in assert_dataframes with the last line.

I override the parameters just in case `penthesilea.conf` (or confs on which `penthesilea.conf` is dependent) is ever changed. To check that the parameters are what they should be I need params to be stable. So overriding all these params is a safety precaution, to prevent the test from failing when it shouldn't.

gonzaponte:
> I check that the correct number of events is run in assert_dataframes with the last line.

I know, but it would be an extra check for the internal counters. It is not a big deal and it is probably checked in some other test, but it may be helpful.

> I override the parameters just in case penthesilea.conf (or confs on which penthesilea.conf is dependent) is ever changed. To check that the parameters are what they should be I need params to be stable. So overriding all these params is a safety precaution, to prevent the test from failing when it shouldn't.

That is what I thought. Default config files are there as an example and for documentation, but they are also used for tests, I think. I am not convinced that *not overriding* is better, but I wanted to bring it up for discussion. I am happy with this either way.

jacg:
+1 for explicitly setting the parameters in the test itself.

abotas:
@gonzaponte -- ok it now checks that the correct number of events selected are recorded in the counter. I included this assert to prevent the test from going through the trouble of loading and comparing dataframes if the test is going to fail anyway because no events were found. More thorough counter tests should go somewhere else because we want this test to fail only if Penthesilea specific code is broken, no?

`event_rang` is fixed

gonzaponte:
> More thorough counter tests should go somewhere else because we want this test to fail only if Penthesilea specific code is broken, no?

Each city (as far as I know) has a test ensuring the number of events processed is the correct one, if that is what you mean.

invisible_cities/cities/base_cities.py
```diff
@@ -668,6 +668,8 @@ def event_loop(self, pmapVectors):
 
             # Count events in and break if necessary before filtering
             what_next = self.event_range_step()
+            print(evt_number)
```
gonzaponte: What is this print doing here? You didn't mean to push it, did you?

jjgomezcadenas:
guys, please, speed up. This is a simple PR and we don't need to spend a week here.

jjgomezcadenas:
which means, for example that is faster if @gonzaponte pushed a commit removing the unwanted print than if he comments, waits for @abotas reaction and so on. As my sadistic watch reminds me every day. Move!

abotas:
The print is gone -- I should have caught that, thank you. I check a couple counters. If these cannot be checked somewhere that checks the counters they should still not be checked by penthesilea because penthesilea does not implement them. They should be checked in base cities. I think we can approve this PR, no?

jacg:
> they should still not be checked by penthesilea because penthesilea does not implement them. They should be checked in base cities.

Hmm ... just because something is correctly implemented in a base class and tested, in no way guarantees that a subclass can't override the correct implementation with a buggy one. So, from the pure logic perspective, your argument seems to be flawed.

jjgomezcadenas:
But then, again, we cannot pre-empt all the situations, right? The test does what it does correctly, more tests can be added later when needed.

jacg:
We cannot achieve 100% certainty. I am merely raising awareness of a way in which this paradigm introduces greater uncertainty.

jjgomezcadenas:
Agreed. On the other hand the test is still OK, isn't it?

jacg:
Sorry, no time to look at any actual code at the moment.

abotas:
ok, I understand what you mean. A city can overwrite sound code from its base with buggy garbage. The logic is flawed. I still have the idea that testing the counters here violates principle that tests should be orthogonal.

If we really want to test the counters in penthesilea we should probably have a test that implements and runs a pcity 'for pcity in pcity, Dorothea, and penthesilea', and checks the counters. 

We could make this a separate PR.

What do we think?

abotas:
oh I missed a few comments while I was writing.

abotas:
I don't think my logic is flawed this time. If this is not a separate test --> we duplicate code. Can we approve this?

gonzaponte:
> guys, please, speed up. This is a simple PR and we don't need to spend a week here.

It is simple, but we are not in a hurry. We are currently validating the latest tag, which is far more important. In general, PRs **do not** need to be approved the day after they are pushed, particularly if they are not fixing some wrong behaviour. IMHO, that makes very easy for bugs to appear.

> I check a couple counters. If these cannot be checked somewhere that checks the counters they should still not be checked by penthesilea because penthesilea does not implement them. They should be checked in base cities.

As the master said, the counters must be checked in each implementation. As I was saying, Irene, Diomira and Dorothea do check the counter returns the correct value, but they do so as part of some test. Maybe we should make a parametrized test for all cities with that assertion.

> I think we can approve this PR, no?

Yes

gonzaponte (approved):
Useful test for Penthesilea ensuring correct execution over 5 evts with individual pmts pmaps.

gonzaponte:
I just realized that the name of the test doesn't make sense (it says 5evt, but runs over 9 and produces 6). Sorry.

jacg:
> In general, PRs do not need to be approved the day after they are pushed, 

Indeed.

> particularly if they are not fixing some wrong behaviour.

Exactly.

> IMHO, that makes very easy for bugs to appear.

Words of wisdom, right there.

Less haste, more speed.

The list of nonsense that has got into our codebase because of too hasty action, is getting quite long. Fixing the nonsense post-hoc, usually costs more time and effort than getting it right in the first place, not to mention the cost of the consequences of running code that contains nonsense.

jacg:
As for the business of testing the counters over and over again in all the cities, this is something that has been bothering me for a long time, but I do not have the resources at present to dedicate to finding a good solution.

(And, of course, I remain convinced deep down, that if the code were structured as a dataflow network, or written in a functional style, many of these problems would be vastly mitigated ... but I am even farther away from having the resources to dedicate to this.)

abotas:
I will have to open another -- more careful -- PR to address the incoherent name... Part of the slowness with this PR was avoidable and my fault. 

I think both @gonzaponte and I suggested the same solution to testing the counters that hopefully works reasonably well given the framework we currently have.

PR #341, opened 2017-09-18
Cythonize xy algos

abotas:
This PR cythonizes corona and barycenter along with their helper functions. It also fixes a few small inefficiencies in the cluster finding algos. It makes dorothea and penthesilea a little faster.

### DIFF
--- a/invisible_cities/reco/xy_algorithms_c.pyx
+++ b/invisible_cities/reco/xy_algorithms_c.pyx
@@ -20,14 +20,15 @@ cpdef barycenter(np.ndarray[double, ndim=2] pos, np.ndarray[double, ndim=1] qs):
        qs = vector (q1, q2...qs) --> (1xn)
 
         """
-    if len(pos) == 0: raise SipmEmptyList
-    if  sum(qs) == 0: raise SipmZeroCharge
+    cdef double Q = np.sum(qs)
+    if   len(pos) == 0: raise SipmEmptyList
+    if          Q == 0: raise SipmZeroCharge
     cdef np.ndarray[double, ndim=1] mu  = np.average( pos           , weights=qs, axis=0)
     cdef np.ndarray[double, ndim=1] var = np.average((pos - mu) ** 2, weights=qs, axis=0)
     # For uniformity of interface, all xy algorithms should return a
     # list of clusters. barycenter always returns a single clusters,
     # but we still want it in a list.
-    return [Cluster(sum(qs), xy(*mu), xy(*var), len(qs))]
+    return [Cluster(Q, xy(*mu), xy(*var), len(qs))]
 
 
 cpdef corona(np.ndarray[double, ndim=2] pos, np.ndarray[double, ndim=1] qs,
@@ -105,15 +106,15 @@ cpdef corona(np.ndarray[double, ndim=2] pos, np.ndarray[double, ndim=1] qs,
            msipm          =  K3)
     """
 
-    if len(pos) == 0: raise SipmEmptyList
-    if  sum(qs) == 0: raise SipmZeroCharge
+    if len(pos)   == 0: raise SipmEmptyList
+    if np.sum(qs) == 0: raise SipmZeroCharge
 
     cdef np.ndarray[long, ndim=1] above_threshold
     above_threshold = np.where(qs >= Qthr)[0]            # Find SiPMs with qs at least Qthr
     pos, qs = pos[above_threshold], qs[above_threshold]  # Discard SiPMs with qs less than Qthr
 
-    if len(pos) == 0: raise SipmEmptyListAboveQthr
-    if  sum(qs) == 0: raise SipmZeroChargeAboveQthr
+    if len(pos)   == 0: raise SipmEmptyListAboveQthr
+    if np.sum(qs) == 0: raise SipmZeroChargeAboveQthr
 
     # if lm_radius or new_lm_radius is negative, just call overall barycenter
     if lm_radius < 0 or new_lm_radius < 0:
--- a/invisible_cities/reco/xy_algorithms_c.pyx
+++ b/invisible_cities/reco/xy_algorithms_c.pyx
@@ -1,4 +1,4 @@
-import sys
+cimport numpy as np
 import numpy as np
 
 from .. core.system_of_units_c import units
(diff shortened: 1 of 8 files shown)


jacg:
Can you quartify how much this gives us in terms of speed?

abotas:
I time profiled these changes together running over 1k Kr events and saw a 10% speed increase

abotas:
The plan is to cythonize a loop that looks like it might be slow, currently in `HitCity`. -- that may make a larger difference. If this PR is not approved soon I may add to this branch since these changes are closely related.

jacg:
The statements

> I time profiled these changes

and

> The plan is to cythonize a loop that looks like it might be slow

seem a bit contradictory. If you have profiled it, then you **know** whether the loop is significantly slow.

I wonder whether you mean 'timed' when you say 'profiled'.

abotas:
Hmm - ok I understand. I timed, I didn't profile! The new plan is to profile.

jacg:
Good. Now, please take to heart the golden rule of optimization: **Don't even think of optimizing anything, until profiling has demonstrated that the thing you are going to optimize uses a SIGNIFICANT proportion of the resource whose use you want to reduce with the optimization.**

Or, in a less verbose form: **DO NOT optimize without the profiler's permission**.

invisible_cities/reco/xy_algorithms_c.pyx
```diff
@@ -0,0 +1,162 @@
...
+
+    cdef np.ndarray[long, ndim=1] above_threshold
+    above_threshold = np.where(qs >= Qthr)[0]            # Find SiPMs with qs at least Qthr
+    pos, qs = pos[above_threshold], qs[above_threshold]  # Discard SiPMs with qs less than Qthr
+
+    if len(pos) == 0: raise SipmEmptyListAboveQthr
+    if  sum(qs) == 0: raise SipmZeroChargeAboveQthr
+
+    # if lm_radius or new_lm_radius is negative, just call overall barycenter
+    if lm_radius < 0 or new_lm_radius < 0:
```
gonzaponte: Can I suggest the use of `<=` here or are we interested in using these variables with value `0`?

invisible_cities/reco/xy_algorithms_c.pyx
```diff
@@ -0,0 +1,162 @@
...
+    return c
+
+
+cpdef discard_sipms(np.ndarray[  long, ndim=1] sis,
+                    np.ndarray[double, ndim=2] pos,
+                    np.ndarray[double, ndim=1] qs):
+    return np.delete(pos, sis, axis=0), np.delete(qs, sis)
+
+
+cpdef get_nearby_sipm_inds(tuple cs, float d,
```
gonzaponte, line 159: why do we use `cs` as tuple here? I see that in the first call to this function we explicitly cast some value to tuple.

abotas:
It seems unnecessary to me that we sort the peaks of the s2si.s2d in line 924 of base_cities.py ... am I wrong? since these are dictionaries the keys order should not be meaningful by design I don't think. @gonzaponte

gonzaponte:
> It seems unnecessary to me that we sort the peaks of the s2si.s2d in line 924 of base_cities.py ... am I wrong?

It may be unnecessary, but it is more pleasant for us, humans, that prefer everything in order. I think our code is sensitive to that (I have checked that hits are loaded and associated reading the peak number, so no problem there), but I am still not convinced. Does it cost us in terms of performance?

jjgomezcadenas:
It goes the other way around, no? Do we win anything by sorting?

As @abotas has pointed out, in a dictionary you organise things by key, so what is the point of sorting? I don't get it either.

jacg:
@abotas  dictionaries do not have any meaningful order, but the data that you get out of them (for example via `.items()` will have a meaningful order if sorted ... as long as you then don't put them back in into a dictionary.

I just want to make user that you understand that `sorted(some_dict.items())` does not sort `some_dict` or modify it in any way whatsoever. It does guarantee that the client of `sorted(...)` receives the items in a meaningful order.

It seems that the loop body appends these items into a list, therefore the list (which *does* have an order) will contain the items in sorted order.

@gonzaponte  The correct way to answer your question about performance is ... to profile the code. Anything else is pointless speculation which will almost certainly turn out to be wrong.

Given that some of the items in this loop are filtered, you will get some performance gain by sorting the final (in general shorter, filtered) list than all the items coming out if the dictionary. Whether this has any noticeable or worthwhile impact on the performance of the code ... you'll have to ask a profiler (or, at least, time the whole code).

jacg:
@jjgomezcadenas He is not sorting the dictionary, he is sorting data that are being appended to a list. Lists *do* have an order. What we gain by this, is that the items in the list are in a sensible rather than random order, which, if nothing else, makes life easier for some human who gets to see the list's contents.

jjgomezcadenas:
All clear! Thanks, master.

abotas:
The only thing I have to add to the conversation about the `sorted` keys is that in our case it does not seem to be costing us much at all.

In fact nothing seems to be particularly slow here. 

Time profiling shows that the most time consuming part of penthesilea is `create_hits_event`. `create_hits_event` has some fancy python loops that I cythonized [here](https://github.com/abotas/IC-1/commit/c1b80734db030e7c0f102c47854adfb51b5ccb0d#diff-237ebaacfbe9a2ed97f1ca5dc841564f), but this did not speed things up.

The most time consuming part of `create_hits_event` is actually `compute_xy_position`. Recall, virtually the only thing `compute_xy_position` does, is call corona/barycenter. But these functions are not slow, they are just called many times.

In fact corona and barycenter are so fast that it is difficult to see just how long they take with real profilers since a profiler's precision is limited by the fact my computer's  "[underlying clock is only ticking at a rate of .001 seconds](https://docs.python.org/3/library/profile.html#module-cProfile)"

To see how long these functions are actually taking in python vs cython i have to use [timeit](https://docs.python.org/3/library/timeit.html). (timit executes the functions many times and divides by the # of executions to see how long they take). According to timit, cythonized corona and barycenter are very, very similar in speed to their python versions, with the cython versions usually performing very, very slightly faster. For all practical purposes they are the same speed.

Time profiling also indicates that the various classes in event model are not slow. I was stubborn (and a little stupid) so I cythonized them anyway -- I also tried stripping `Hit` down to its bare fundamentals --  and saw no improvement in speed.

---
---
---
If people agree, I think I should resign, close this PR, and conclude this was mostly an embarrassing case of premature optimization. I think we were also worried at some over the summer that Dorothea was slow and later realized she was not...? 

Alternatively, if our paradigm for IC is that we want as much code in Cython as possible I will clean up my cythonized `event_model` classes, push them to this PR and be done with it.

Either way, there were a few obvious small inefficiencies I found during the course of this PR that I should quickly fix in a clean, separate PR. I can also celebrate that I think I now know how to use a couple profilers

jacg:
> I think I should resign,

That's taking it a bit far :-)

> this was mostly an embarrassing case of premature optimization.

As long as you learned from the experience, and remember it for the future, it was worthwhile.

I am acutely aware that people often don't believe the advice I give, or believe it only intellectually but not viscerally, until they suffer as a consequence of not following my advice ... only then does the true meaning and value of what I said sink in. It's a fairly standard part of the learning process for most people (including me).

Your last comment shows that you (eventually) did sterling work on the profiling. Documenting it here for all IC-ists to see, might just help others to learn from your painful experience and thus save them the need to have the painful experience themselves, before getting the point.

Well done!

> I can also celebrate that I think I now know how to use a couple profilers

Indeed!

Spread the knowledge!

BTW, did you try any of the flamegraph tools? If so, which ones did you try, and which ones did you find useful?

> I think we were also worried at some over the summer that Dorothea was slow and later realized she was not...?

It's not yet clear whether that was real or not. There is the possibility that for a few commits in `master`  she is drastically slower. Waiting for @gonzaponte to find some time to confirm.

jacg:
> if our paradigm for IC is that we want as much code in Cython as possible

That would be very silly, wouldn't it!

We want the flexibility, dynamism and ease of development that pure Python offers, wherever reasonably possible. We are prepared to forgo it and drop to the more rigid, more complex, more difficult to modify and experiment with Cython, **only** in those cases where the benefits of a **significant** run-time performance increase outweigh the concomitant increase in cost of development.

Remember, once you've Cythonized a function, if I want to fiddle with it, try alternatives, I have to **recomplie** it every time. If I'm doing this in an interactive session, I have to restart the whole session every time. (Even though IPython does heroic things to mitigate this!)

gonzaponte:
Bravo @abotas, good effort!

> It's a fairly standard part of the learning process for most people (including me).

Not only standard, it is fundamental. Without feeling the pain, we never learn the lesson. At least that is my experience.

> Waiting for @gonzaponte to find some time to confirm.

This will be confirmed at some unforeseeable point in the future, when we actually compare current `Dorothea` versus old `Dorothea`.

jacg:
> Without feeling the pain, we never learn the lesson.

It's not *strictly* necessary: Some people are able to learn from the experience of others in some situations. Unfortunately, these occurances are rather rare.

abotas:
Unfortunately it seems I prefer to learn by first sticking my hand in the fire. But I did learn! 

@jacg I did not use flamegraph tools -- i mostly used cprofile. But I've been looking at the flamegraph stuff these past couple days and it looks very interesting! I will try using it in the future and keep you updated. 

Ok well with that I am closing the PR

PR #342, opened 2017-09-21
Corrections update

gonzaponte:
This PR:
- Adds the normalization by a constant to the class `Correction`.
- Generalizes the class so this new normalization fits better.
- Adds safety checks to the class.
- Provides tests for all of the above.
- Changes the philosophy of correction factors normalization. The normalization now occurs when loading from file, rather than when storing them. The function that reads back the corrections has been modified, accordingly.

### DIFF
--- a/invisible_cities/cities/zaira.py
+++ b/invisible_cities/cities/zaira.py
@@ -9,11 +9,9 @@ last revised: JJGC, 12-July-2017
 import numpy  as np
 import tables as tb
 import pandas as pd
-from operator import attrgetter
 
 from .  base_cities         import DstCity
 from .. core.core_functions import in_range
-from .. core.configure      import configure
 from .. io.kdst_io          import xy_correction_writer
 from ..reco.corrections     import Correction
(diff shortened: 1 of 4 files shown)


jahernando (approved):
The commit extends the functionality of Correction class to normalize now to a given value. The normalization strategies and options are now passed by key-arguments.  
Test are modified accordingly. New tests are included to check the normalization to a given value works and to raise exceptions when options of a given normalization strategy are not passed in the key-arguments. 
The PR is approved.
Nevertheless, it seems that the functionality of Corrections when using an interpolation other than 'nearest' is not properly tested. Isn't it?
If not, as it is nor crucial for this release, it is approved, but these tests should be included in a future PR.

jahernando (commented):
I approve it again!

jahernando (approved):

jahernando (approved):
Approved

gonzaponte:
This PR has been approved by @jahernando, but requires the review of someone with write access.

PR #343, opened 2017-09-21
Add machinery to perform lifetime corrections depending on XY

gonzaponte:
This PR adds an interface to perform XY-dependent lifetime corrections.


abotas:
This looks good to me. You've added a function that allows a lifetime correction dependent on XY. Are you done adding commits @gonzaponte ? If so it looks like it needs to be rebased and then I will approve it.

gonzaponte:
I will push one more and rebase.

gonzaponte:
I think this is finished now.

abotas:
Travis failed but I can't seem to see the log, do tests pass locally?

gonzaponte:
Yes. The failure occurs during the execution of `source manage.sh install ${IC_PYTHON_VERSION}` and before running the tests. The error is:
```
unable to execute 'x86_64-conda_cos6-linux-gnu-gcc': No such file or directory
error: command 'x86_64-conda_cos6-linux-gnu-gcc' failed with exit status 1
```

Can we force a rebuild?

gonzaponte:
Seems to be some problem with Travis. Other PR builds crash at the same point.

abotas:
Yes, I restarted the build and it failed again

gonzaponte:
Continuum released a new version of anaconda/miniconda on Sep 26th. In that release they set the gcc compiler to a non-existing executable... We can either wait for this to be fixed, or manually pick the previous release. I would wait a couple of days before doing something like that.

abotas:
Something strange is going on because I think travis is supposed to use the latest release of miniconda3 for linux which on the website is listed as 4.3.21. 

Travis seems to download this version but then update to an unstable version 4.3.27

abotas:
Maybe we should try prevent the update to 4.3.27 -- it is still not working.

gonzaponte:
Yes, we have a `conda update conda` instruction in the `manage.sh` script I didn't notice before. I don't know if @jacg can say something here.

Summary:
- Latest version of miniconda contains a bug and doesn't find C compiler
- The version we download is not buggy, but we update to the buggy one in our code.

The question is: should we wait for a fix by continuum or remove the `conda update conda` line from the code?

jmbenlloch:
I think we should fix the Conda version we are using, my last PR #346 does that. In the near future I will prepare another PR with fixed versions for all our dependencies.

Now I think it is clear we need #346 to be able to work and to avoid having similar issues again.

jjgomezcadenas:
Please, approve this, unless @jacg says otherwise. We need to unblock this nasty situation. Use privileges if needed to bypass Travis, then we can unblock the situation.

abotas (approved):
This pull request looks good! You've added a function that allows a lifetime correction dependent on XY, a couple writer functions, loaders, and corresponding tests. 

The writers and loaders follow the format established by other writers and loaders. The tests are clear and succinct.

abotas:
merged

PR #345, opened 2017-09-26
S12selector full output

gonzaponte:
So far, the rejected peaks of an event passing the S12Selector were being propagated to the rest of the code, without knowing which ones they were. This had two side effects:
- We always picked the first S1, but in the case of multiple S1s, we didn't have the certainty that it would be the right one.
- When multiple S2s are present, all of them were being written to file, independently of whether they were selected or not. Thus, on top of the extra disk usage, we were having a wrong number of S2s  associated to the event.

This PR incorporates the result of the selection over each peak in the event to the output of the S12Selector. Other functions and the cities involved are now aware of it and use the complete output to ignore the rejected peaks.

Furthermore, a bunch of tests have been added to the S12Selector class, which was untested as well as the functions defined to use it. The test suite for the cities `Dorothea` and `Penthesilea` have been also extended with some new tests to ensure the selection is performed correctly.


abotas:
It looks like a terrible bug that's been fixed! A few first requests:
1) Could you put some comments in`S12SelectorOutput` and explain what is going on -- maybe this should be obvious but it's not immediately obvious to me. The class seems necessary and makes some code in base cities easily legible but needs some documentation!

2) I think you pushed a `__init__.py` in filters that should be gotten rid of, no?

And a couple questions specific surrounding a more general one -- have you looked at data with these changes?
1) Do you have any idea what fraction of events this was bug affecting? 
I imagine it depends on a lot of stuff: `.confs` and the trigger, isotope etc, but any estimate would be, though maybe not necessary, super nice.

2) For the purposes of our analysis, what was this bug hurting? The two things you mention in your description i think imply:
- we lost efficiency 
- for some fraction of events, our reconstruction of the Z coordinate of hits was avoidably wrong.

The 2nd is very bad since it totally screws up our correction for lifetime. If this bug was affecting many events it will be worth reprocessing data with these changes to try to improve energy resolution.

gonzaponte:
> Could you put some comments inS12SelectorOutput and explain what is going on

Sure!

> I think you pushed a __init__.py in filters that should be gotten rid of, no?

I pushed it because I couldn't import the modules in `filters` from an interactive session. Maybe this is something wrong with my configuration and, surprisingly, it doesn't complain when they are imported in the cities! Anyway, all the other packages have it, so it should be there, shouldn't it?

> have you looked at data with these changes?

The new tests use actual data (not MC).

> Do you have any idea what fraction of events this was bug affecting?

That is difficult to say... But the number is not huge (otherwise we wouldn't get good results). I would say less than 10%, but we were not paying too much attention to multiple-S2 events, so it may be bigger.

> For the purposes of our analysis, what was this bug hurting? 

- As you say, we were potentially choosing the wrong S1, which affects the Z coordinate.
- For the S2, the column `nS2` in the DST was being interpreted as "number of S2 peaks in the event, regardless of the selection". Now, the selection has an impact on that number as well as in the number of rows produced in the DST.

abotas (approved):
My comments have been addressed, you are right about the `__init__.py`, pull request approved!

abotas:
Merged!

PR #346, opened 2017-09-29
Fix conda version to one known to work

jmbenlloch:
Anaconda introduced a bug in version 4.3.27 and in general we should keep our software versions under control, this is a first step towards that. The previous version 4.3.21 works well.

In the future we will need to decide when to upgrade our dependencies, but we need control over them or errors like #343, will keep appearing both in Travis, Canfranc, users, etc.


gonzaponte (approved):
Agreed.

PR #348, opened 2017-10-02
Kr city init counter

### LINKED ISSUE #347: counters bug running dorothea with corona
There is a counter `n_events_more_than_1_cluster` in `Counter` that is never initialized (i think it is only in one place in the code and we are trying to add 1 to it). 

When I try to run dorothea with corona it crashes immediately. 
I think this is a simple fix, but I don't know if this is a counter we care about or if we should just get rid of it? @gonzaponte @jjgomezcadenas 

```
event in file = 0, total = 0
Traceback (most recent call last):
  File "/software/IC-dev//bin/city", line 19, in <module>
    city_class.drive(args)
  File "/software/IC-dev/invisible_cities/cities/base_cities.py", line 187, in drive
    instance.go()
  File "/software/IC-dev/invisible_cities/cities/base_cities.py", line 196, in go
    self.run()
  File "/software/IC-dev/invisible_cities/cities/base_cities.py", line 219, in run
    self.file_loop()
  File "/software/IC-dev/invisible_cities/cities/base_cities.py", line 716, in file_loop
    self.event_loop(pmapVectors)
  File "/software/IC-dev/invisible_cities/cities/base_cities.py", line 690, in event_loop
    evt = self.create_dst_event(pmapVectors)
  File "/software/IC-dev/invisible_cities/cities/dorothea.py", line 24, in create_dst_event
    return  self.create_kr_event(pmapVectors)
  File "/software/IC-dev/invisible_cities/cities/base_cities.py", line 826, in create_kr_event
    self.cnt.n_events_more_than_1_cluster += 1
AttributeError: 'Counters' object has no attribute 'n_events_more_than_1_cluster'
```

jahernando:
Solve issue 347:  Dorothea configured with corona crashed due to a non initialized counter that counts the number of events with multiple clusters. This pull request inits the counter in KrCity and adds a test. 
Closes #347

### DIFF
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -743,6 +743,7 @@ class KrCity(PCity):
 
     def __init__(self, **kwds):
         super().__init__(**kwds)
+        self.cnt.init(n_events_more_than_1_cluster = 0)
         #self.reco_algorithm = find_algorithm(self.conf.reco_algorithm)
 
 
--- a/invisible_cities/cities/dorothea_test.py
+++ b/invisible_cities/cities/dorothea_test.py
@@ -140,3 +140,17 @@ def test_dorothea_filter_events(config_tmpdir, Kr_pmaps_run4628):
 
     assert np.all(dst.event.values == events_pass)
     assert np.all(dst.peak.values  ==   peak_pass)
+
+def test_dorothea_issue_347(KrMC_pmaps, config_tmpdir):
+    PATH_IN =  KrMC_pmaps[0]
+    PATH_OUT = os.path.join(config_tmpdir, 'KrDST.h5')
+    conf = configure('dummy invisible_cities/config/dorothea_with_corona.conf'.split())
+    # with this parameters Corona will find several clusters
+    conf.update(dict(files_in   = PATH_IN,
+                     file_out   = PATH_OUT,
+                     lm_radius     = 0.1,
+                     new_lm_radius = 0.2,
+                     msipm         = 1))
+    dorothea = Dorothea(**conf)
+    dorothea.run()
+    dorothea.end()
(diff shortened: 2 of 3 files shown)


abotas:
Thanks @jahernando! 
I think the fix i think is perfect but in the test can you add an assert statement checking that the counter has the correct value after the `end` method?

Right now the test ensures dorothea does not crash, but I don't think it checks that the counter counted the correct number of events with more than one cluster. If we are going to keep the counter we should make sure it isn't lying to us.

abotas:
Ahh.. wait I see maybe checking that this counter has the correct number is actually very difficult because we would have to look at the geometry of the individual events and the corona...

abotas:
Maybe this PR is actually perfect as is, as it does ensure the counter is initialized and closes the issue as you say! 

We just need another PR in the future that goes checks this counter and probably various others...

abotas:
Actually, I think we can just check for how many events the list `clusters = self.compute_xy_position(s2si, peak_no)` in line 822 of `base_cities` has a length > 1. 
This should equal the final number of the counter i think. 

@jahernando do you agree? If you agree but are busy let me know and I can add it to the PR!

jacg:
@jahernando Please configure your git to use the email that you have registered with GitHub. Then we'll need to modify your commits to use that address.

jahernando:
@jacg I just changed my email address in my GitHub to be the same one used in git for the commits. Is that enough?

jacg:
That seems to be an excellent solution, which should avoid the need to rebase your commits.

jahernando:
Hi Alex,
I just added into dorothea_test.py an assert to check that the counter counts the number of events with more than 1 clusters in a given data file and with a given configuration of corona algorithm!

abotas:
Perfect! But I think you didn't push the changes to github. The commits here don't have the change!

jahernando:
Yes, I did not :) But now It should be there!

gonzaponte (approved):
Fix to a uninitialized variable that was stopping us from using `Dorothea` when multiple clusters are found (i.e. with `corona`). A test has also been provided to ensure the counter is working properly.

Good job.

Issue #134, opened 2017-02-21
Record conda and pip package versions for reproducibility


jacg:
In order to be able to reproduce old runs, it is not enough to check out the particular version of IC with which the run was performed, but one must also ensure that exactly the same versions of packages installed with `conda` and `pip` (listed in the `make_environment` function of `manage.sh`) are used.

This means that we need an easy way to record and store the versions of such packages.

gonzaponte:
Closed via #349.

Issue #299, opened 2017-08-23
Failure in two diomira tests.


Aretno:
I updated today to the current status of nextic/master and now two diomira tests are failing (in addition to the blr one mentioned in https://github.com/nextic/IC/issues/263). Using python 3.6. I tried to do a clean also just to make sure nothing was messing with it.

```
_____________________________________________________________________________________ test_command_line_run[diomira] _____________________________________________________________________________________

city = 'diomira', tmpdir_factory = <_pytest.tmpdir.TempdirFactory object at 0x1055f7358>

    @mark.slow
    # @mark.parametrize('city',
    #                   'diomira isidora irene dorothea zaira penthesilea'.split())
    # TODO understand what's wrong with isidora (in Travis)
    @mark.parametrize('city',
                      'diomira irene dorothea zaira penthesilea'.split())
    
    def test_command_line_run(city, tmpdir_factory):
        ICTDIR = getenv('ICTDIR')
        # Use the example config file included in the repository
        config_file_name = join(ICTDIR, 'invisible_cities/config', city+'.conf')
        # Ensure that output does not pollute: send it to a temporary dir
        temp_dir = tmpdir_factory.mktemp('output_files')
        out_file_name = join(temp_dir, city+'.out')
        # The actual command that we want to test
        command = ('city {city} {config_file_name} -o {out_file_name}'
                   .format(**locals()))
        try:
>           check_output(command, shell = True, stderr=STDOUT)

invisible_cities/cities/command_line_execution_test.py:28: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../anaconda/envs/IC3.6/lib/python3.6/subprocess.py:336: in check_output
    **kwargs).stdout
[...]

PR #349, opened 2017-10-03
Fix dependencies versions

jmbenlloch:
This PR fixes the versions of IC dependencies (both from conda & pip). We need control over this versions if we want reproducibility and avoiding possible bugs due to different versions used in different machines (it has already happened in the past in travis, canfranc and between some IC users).

Since we want to analyze data for the first NEW papers, we should keep this under control as soon as possible. Later on we should decide how and when we want to update this packages.

The versions I have hardcoded in manage.sh are the most up-to-date now. I have tested this setup in several machines and in all of them all tests passed.

### DIFF
--- a/manage.sh
+++ b/manage.sh
@@ -65,25 +65,25 @@ function make_environment {
     echo creating ${YML_FILENAME}
 
     cat <<EOF > ${YML_FILENAME}
-name: IC${PYTHON_VERSION}
+name: IC3.6
 dependencies:
-- python=${PYTHON_VERSION}
-- cython
-- numpy
-- pandas
-- pymysql
-- pytables
-- pytest
-- scipy
-- matplotlib
-- jupyter
-- notebook
-- sphinx
-- networkx
+- cython=0.26=py36_0
+- jupyter=1.0.0=py36_3
+- matplotlib=2.0.2=np113py36_0
+- networkx=1.11=py36_0
+- notebook=5.0.0=py36_0
+- numpy=1.13.1=py36_0
+- pandas=0.20.3=py36_0
+- pymysql=0.7.9=py36_0
+- pytables=3.4.2=np113py36_0
+- pytest=3.2.1=py36_0
+- python=3.6.2=0
+- scipy=0.19.1=np113py36_0
+- sphinx=1.6.3=py36_0
 - pip:
-  - hypothesis
-  - flaky
-  - pytest-xdist
+  - flaky==3.4.0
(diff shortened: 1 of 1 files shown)


gonzaponte:
Great!

@jacg mentioned in #299 the need for an automatic test of versions. Maybe this is a good PR to introduce them.

jjgomezcadenas (approved):
Good job.

jmbenlloch:
> @jacg mentioned in #299 the need for an automatic test of versions. Maybe this is a good PR to introduce them.

If anybody has a clear idea of how to that and volunteers to implement it now, I agree. If not, in mi opinion we should prioritize this and worry later about the possible updates.

gonzaponte:
> If anybody has a clear idea of how to that and volunteers to implement it now, I agree. If not, in mi opinion we should prioritize this and worry later about the possible updates.

Agreed. I am opening a new PR so we do not forget about it. This PR is very important.

jmbenlloch:
> Agreed. I am opening a new PR so we do not forget about it. This PR is very important.

Ok, then I'll merge this one when travis give us green light. This closes #134.

jacg:
Hmm, I see that you have hard-wired the versions in `manage.sh`. It doesn't belong there. It should have gone in a `.yaml` file, one per currently supported python version (which today happens to be just 3.6, but in general is more than one: specifically, when 3.7 is released, there will be a period of transition while we wait for all our dependencies to become available, during which we will need to support 3.6 and 3.7 ... and so on for future python releases).

gonzaponte:
Then we shouldn't create the `.yaml` file in the `manage.sh`, right? It should be a proper file in the repo. Is this a bad practice?

jacg:
> Then we shouldn't create the .yaml file in the manage.sh, right? It should be a proper file in the repo.

Exactly.

> Is this a bad practice?

The versions of the dependencies we use are part of the state of our code at any particular commit. So having these version numbers committed seems reasonable. We already went down that route, by hard-wiring them into `manage.sh` which is tracked.

It seems perfectly reasonable to me.

Committing the versions does raise questions about the process of upgrading to newer versions. We need to give this some thought. (Which I cannot afford to do at this very moment: note to self: TODO).

jjgomezcadenas:
So, since we have a fix that works now, should we defer further fixes for a while? We have our platter full...

jacg:
It's not urgent, as it works. But it is a little bit of technical debt. It should probably be recorded in an issue.

PR #350, opened 2017-10-03
fix typo in TrackCity

gonzaponte:


jacg:
The absence of a test failure pointing out this bug indicates that a test is missing.

gonzaponte:
Yes, I am aware, but this section of the code is still work in progress and it is not used yet. I think the developers @paolafer and @jerenner will develop and properly test it at some point.

jacg:
> this section of the code is still work in progress and it is not used yet.

Then it shouldn't be in `master`.

gonzaponte:
I know... but it is already there. Do you want me (or someone else) to remove it and keep it in a personal branch?

jacg:
Yes, I think it should be removed from `master` until it is thought out, developed, tested and approved.

It could be done in a branch visible to only one person, or on a branch which the relevant developers fetch from eachothers' repos, or in an ongoing PR ... but not in `master`.

gonzaponte:
Ok. @jjgomezcadenas, do you agree?

jjgomezcadenas:
Actually I don't agree. If I recall right, I wrote that code and was part of a PR that was approved. I don't see a reason now to "unapproved it". It's not the first time that we find code that has not been tested. It's all part of the perpetual tradeoff between moving the code forward and reducing technical debt. Let's simply add a test, since we have spotted the bug, rather than removing the code.

jacg:
Earlier @gonzaponte said

> this section of the code is still work in progress and it is not used yet.

Now @jjgomezcadenas says

> [If I recall right,] I wrote that code and was part of a PR that was approved.

It seems to me that one of these statements must be mistaken, or that the PR approval was a mistake.

+ If the former is true, the code should not be in `master`. In which case it should be *removed* from `master`.

+ If the latter is true, then a test needs to be added, ASAP.

jjgomezcadenas:
@jacg is this PR obsolete?

PR #351, opened 2017-10-03
Save hits of all particles

paolafer:
The compute_mchits_dict() function was selecting only hits coming from electrons, while the correct behaviour is that all hits are returned. This PR fixes this bug.


gonzaponte:
Hide that merge commit before it is noticed by the master!!!

jacg:
> Hide that merge commit before ...

Too late! I've just seen it!

jacg:
What disturbes me even more is that this significant change to the behaviour of the function is not detected by any test in our test suite. There is a test, but the nonsense that the function did before, was perfectly fine as far as the test was concerned.

This stuff should never have made it into `master`. It was the reviewers' job to catch it!

paolafer:
Ooops, I'm not sure I know what I did with the merge commit...

jjgomezcadenas:
I don't think the function was doing a non-sense, it was simply returning the hits of the electrons, rather than the hits of all charge particles. There was no test simply because I didn't realise that there were more particles (such as positrons) to return. I think the process of catching an incomplete function and expanding it is just normal. We can now add a test. And again, I disagree about the stuff not making it into master. It was not trivial to catch that error unless you happened to be the person who works at all times with the MC (e.g, Paola, who has eventually caught it).

jacg:
> I don't think the function was doing a non-sense

Its docstring is: """Returns all hits in the event"""; its behaviour was to ignore any hits that did not come from electrons.

It was clearly the wrong behaivour, and thereby qualifies as nonsense.

> It was not trivial to catch that error unless you happened to 

... unless you happened to write a test ?

Ah, no ...

> be the person who works at all times with the MC (e.g, Paola, who has eventually caught it).

If that is the case, then I would suggest that said person should, perhaps, have been involved in the review.

I get the impression that you're worried about your karma again. If it's any consolation, the same person has, today, also found a missing test and corresponding nonsense behaviour in some code that I have written. 

Please note that rather than defending or justifying the code or the absence of the test, I wrote the test, started a PR and am in the process of fixing the nonsense.

jjgomezcadenas:
Not worried about my karma. Worried about converging. We are approaching again red lines and need to come up with a complete release soon (now) to launch productions, etc. The release needs to include the MC hit business. I am happy enough to be hanged by my thumbs for writing incomplete code and an incomplete test, but let's face realities. Paola did not review the code when I wrote it because she was not available at that moment (we have limited resources). She has reviewed it when she has had the opportunity, caught a bug, proposed a solution. She should probably also add a test. This is business as usual, no? No need to speculate whether this should have been in the main or not. It is now, and what we need to do is fix it.

There is a continuous tension between what we all want to do and what we need to do. Not all reviewers are the same, our resources are limited, and our schedules are tight. All what I am trying to say is that we need to our best with what we have. Stating that such piece of code should not be in the repository does not help. Since it is, let's fix it.

gonzaponte (changes requested):
Please:
- Add a test to ensure *all* hits are actually taken.
- Remove merge commit.
- Rebase.

gonzaponte:
Ok. Good. Can you squash those 4 commits so we don't have the large file in the history?

gonzaponte (approved):

PR #352, opened 2017-10-03
Voxel size

jacg:
+ `voxelize_hits` was not respecting the requested `voxel_dimensions`.

+ There was no test of this behaviour.

This PR

+ adds a test demonstrating the problem
+ fixes `voxelize_hits` so that it passes the test
+ improves `__str__` in BHit and subclasses

This last point makes it much easer to understand and debug issues such as the main one in this PR.

### DIFF
--- a/invisible_cities/evm/event_model.py
+++ b/invisible_cities/evm/event_model.py
@@ -115,7 +115,8 @@ class BHit:
     def E   (self): return self.energy
 
     def __str__(self):
-        return '<{} {}>'.format(self.pos.tolist(), self.E)
+        return '{}({.X}, {.Y}, {.Z}, E={.E})'.format(
+            self.__class__.__name__, self, self, self, self)
 
     __repr__ =     __str__
 
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -46,6 +46,12 @@ def voxelize_hits(hits             : Sequence[BHit],
     voxel_edges_lo = bounding_box_centre - number_of_voxels * voxel_dimensions / 2
     voxel_edges_hi = bounding_box_centre + number_of_voxels * voxel_dimensions / 2
 
+    # Expand the voxels a tiny bit, in order to include hits which
+    # fall within the margin of error of the voxel bounding box.
+    eps = 3e-12 # geometric mean of range that seems to work
+    voxel_edges_lo -= eps
+    voxel_edges_hi += eps
+
     hit_positions = np.array([h.pos for h in hits])
     hit_energies  =          [h.E   for h in hits]
     E, edges = np.histogramdd(hit_positions,
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -39,12 +39,19 @@ def voxelize_hits(hits             : Sequence[BHit],
     if not hits:
         raise NoHits
     hlo, hhi = bounding_box(hits)
-    hranges = hhi - hlo
-    bins = np.ceil(hranges / voxel_dimensions).astype(int)
-    nonzero_bins = np.clip(bins, a_min=1, a_max=None)
+    bounding_box_centre = (hhi + hlo) / 2
+    bounding_box_size   =  hhi - hlo
+    number_of_voxels = np.ceil(bounding_box_size / voxel_dimensions).astype(int)
+    number_of_voxels = np.clip(number_of_voxels, a_min=1, a_max=None)
+    voxel_edges_lo = bounding_box_centre - number_of_voxels * voxel_dimensions / 2
+    voxel_edges_hi = bounding_box_centre + number_of_voxels * voxel_dimensions / 2
+
     hit_positions = np.array([h.pos for h in hits])
     hit_energies  =          [h.E   for h in hits]
-    E, edges = np.histogramdd(hit_positions, bins=nonzero_bins, weights=hit_energies)
+    E, edges = np.histogramdd(hit_positions,
+                              bins    = number_of_voxels,
+                              range   = tuple(zip(voxel_edges_lo, voxel_edges_hi)),
+                              weights = hit_energies)
 
     def centres(a : np.ndarray) -> np.ndarray:
         return (a[1:] + a[:-1]) / 2
--- a/invisible_cities/reco/paolina_functions_test.py
+++ b/invisible_cities/reco/paolina_functions_test.py
@@ -39,7 +39,7 @@ def big_enough(hits):
 
 posn = floats(min_value=10, max_value=100)
 ener = posn
-bunch_of_hits = lists(builds(Voxel, posn, posn, posn, ener),
+bunch_of_hits = lists(builds(BHit, posn, posn, posn, ener),
                       min_size =  1,
                       max_size = 30).filter(big_enough)
 
--- a/invisible_cities/reco/paolina_functions_test.py
+++ b/invisible_cities/reco/paolina_functions_test.py
@@ -129,9 +129,6 @@ def test_voxelize_hits_respects_voxel_dimensions(hits, requested_voxel_dimension
                 np.isclose(off_by, unit)).all()
 
 
-
-
-
 @given(bunch_of_hits, box_sizes)
 def test_make_voxel_graph_keeps_all_voxels(hits, voxel_dimensions):
     voxels = voxelize_hits    (hits  , voxel_dimensions)
--- a/invisible_cities/reco/paolina_functions_test.py
+++ b/invisible_cities/reco/paolina_functions_test.py
@@ -9,6 +9,7 @@ from numpy.testing import assert_almost_equal
 
 from pytest import fixture
 from pytest import mark
+from pytest import approx
 from pytest import raises
 parametrize = mark.parametrize
 
@@ -92,7 +93,7 @@ def test_voxelize_hits_does_not_lose_energy(hits, voxel_dimensions):
     def sum_energy(seq):
         return sum(e.E for e in seq)
 
-    assert_almost_equal(sum_energy(voxels), sum_energy(hits))
+    assert sum_energy(voxels) == approx(sum_energy(hits))
 
 
 random_graph = builds(partial(fast_gnp_random_graph, p=0.5),
--- a/invisible_cities/reco/paolina_functions_test.py
+++ b/invisible_cities/reco/paolina_functions_test.py
@@ -55,6 +55,7 @@ def test_voxelize_hits_should_detect_no_hits():
     with raises(NoHits):
         voxelize_hits([], None)
 
+
 @given(bunch_of_hits)
 def test_bounding_box(hits):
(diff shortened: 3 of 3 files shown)


jacg:
The test proposed here is not correct: it fails to correctly express the idea "I want to make sure that this `float` is (approximately) a multiple of some other `float`".

I'm struggling to come up with a clean, concise, obviously correct (etc.) way of expressing this idea. On top of that, I keep getting the impression that I'm re-inventing some wheel.

Any suggestions how to do this well?

gonzaponte:
> The test proposed here is not correct: it fails to correctly express the idea "I want to make sure that this float is (approximately) a multiple of some other float".

What is the `D` loop for? The `D` variable does nothing. If, as I feel, it is skippable, why do you say it does not express the idea correctly? It does *exactly* that.

jacg:
Indeed, the `D`-loop is just leftover rubbish.

The non-rubbish does not express the idea correctly, because `5.9999999 % 1` is approximately `0.999999` which is nowhere near to `0`, even though `5.999999` is damn close to `6` which is a multiple of `1`.

In other words, if it is just *below* a multiple of whatever, then the test fails.

gonzaponte:
I see. Why not:

```
np.allclose(distance_between_voxels, requested_voxel_dimensions)
```
?

gonzaponte:
Sorry, didn't think too much about that. I see what you mean.

gonzaponte:
The simplest ad-hoc way would be, I think:
```
discrete_distances = distance_between_voxels % requested_voxel_dimensions
np.allclose(distance_distances - np.round(discrete_distances), 0, rtol=eps)
```

for some `eps` much smaller than `requested_voxel_dimensions`. I guess the question is: Is there a simpler, more obvious way, right?

paolafer:
Why don't you use the absolute value of the difference between the coordinates of two voxels minus voxel_size < epsilon?

jacg:
Because this is guaranteed to fail for any pair of voxels that are not neighbours.

Is there any guarantee that successive elements of the voxel sequence returned by `voxelize_hits`, will be neighbours?

Even then, some neighbours will have identical centres in *some* dimensions, and then the scheme also fails.

jacg:
I've pushed a failed attempt at fixing the test and the implementation. The test fails, so at least one of the two must be incorrect. I'll be almost completely unavailable for the next 40 hours. Please feel free to improve on my pathetic effort in the meantime.

paolafer:
I can take this, Jacek!

jjgomezcadenas:
You mean @jacg (aka Master) :)!

paolafer:
I've tried to work on it, @jacg, and I found a solution that seems sensible to me and passes the tests (you can find it [here](https://github.com/paolafer/IC/blob/vox-size/invisible_cities/reco/paolina_functions.py)). The only problem is that it gives nonsense results, so there must be a flaw somewhere. I'll keep on working on this, I just wanted to push it in my branch, in case you want to have a look when you're back.

jacg:
@paolafer In the version of the test that I see in your branch, the only `assert` in the test is commented out. In other words, the test will *always* pass. When I uncomment the loop with the assert, I get exactly the same failure as I see in my proposed solution.

jacg:
There was a bug in the test I proposed earlier. I have fixed that now, and the proposed fixed implementation now passes the fixed test.

@paolafer Initial investigation suggests that the solution you propose in your branch does *not* pass the new test. So please check whether the solution proposed in **this** branch also gives 'nonsense results'.

I'm still not entirely happy with the test or the implementation: they seem too long-winded; but first let's check whether they make sense.

paolafer:
@jacg I'm not sure I understand what you're telling me to do... In which branch did you do your changes? Your voxel-size branch seems to me the same as two days ago. What should I test with your branch?

jacg:
> Your voxel-size branch seems to me the same as two days ago.

No, the branch has changed since then. Most importantly, `1` has been replaced with the voxel dimension in the second `isclose` in the test.

> What should I test with your branch?

Earier you said that 

> I found a solution that seems sensible to me and passes the tests (you can find it here). The only problem is that it gives nonsense results,

The solution you propose does not pass the test which is currently in `jacg/voxel-size`. Please fetch the latest version of `jacg/voxel-size` and use it to see whether it too gives you 'nonsense results' for whatever you meant when you said 'nonsense results'.

paolafer:
Ok, now I understand.

jacg:
@paolafer I've just pushed a change. See if you find this version of `voxelize_hits` easier to follow.

paolafer:
Yes, this simplifies things!

jacg:
As far as I am concerned, this one is now ready for review.

I would like the code to be a bit cleaner, but

+ it's already a bit better than the initial attempts
+ if it works, we should probably just take it and get on with other things
+ I suspect we might not be able to find a significantly crearer approach

paolafer:
I would like to test it and compare the results with the previous version, @jerenner, maybe you want to try your code on this version, too.

jacg:
> I would like to test it and compare the results with the previous version

Of course! I consider that to be a **necessary** part of the review.

jacg:
You gotta love hypothesis!

In the `hard` profile hypothesis found the following counterexample for `test_voxelize_hits_does_not_lose_energy`:

    hits = [BHit(96.11764705882354, 10.0, 10.0, E=10.0), BHit(12.470588235294109, 10.100000000000016, 10.100000000000016, E=10.0)], voxel_dimensions = array([ 2.78823529,  1.        ,  1.        ])
    
        @given(bunch_of_hits, box_sizes)
        def test_voxelize_hits_does_not_lose_energy(hits, voxel_dimensions):
            voxels = voxelize_hits(hits, voxel_dimensions)
        
            if not hits:
                assert voxels == []
        
            def sum_energy(seq):
                return sum(e.E for e in seq)
        
    >       assert_almost_equal(sum_energy(hits), sum_energy(voxels))
    E       AssertionError: 
    E       Arrays are not almost equal to 7 decimals
    E        ACTUAL: 20.0
    E        DESIRED: 10.0


[Aside: with the way pytest reports *ACTUAL* and *DESIRED*, those arguments to `assert_almost_equal` should be swapped.]

So, what is going on? 

1. There are two hits, separated by an almost exact multiple of voxel lengths, in some direction.
2. The algorithm places the outer edges of the extreme voxels on the hit positions (as accurately as floating point operations allow).
3. One of the hits that sits on a voxel boundary is considered to lie outside the voxel, and it is lost, along with its energy!


Floating point numbers are such fun!

Any suggestion how to deal with this sensibly and painlessly?

jacg:
Could you please help me to preserve my sanity? Try the following on your machine, show me the results and tell me if you think they make sense.

    from numpy import nextafter
    nextafter( 0.1, -1)
    nextafter( 0.1, +1)
    nextafter( 1  , -1)
    nextafter( 1  , +1)
    nextafter(10  , -1)
    nextafter(10  , +1)

gonzaponte:
```
>>> from numpy import nextafter
>>> nextafter( 0.1, -1)
0.099999999999999992
>>> nextafter( 0.1, +1)
0.10000000000000002
>>> nextafter( 1  , -1)
0.99999999999999989
>>> nextafter( 1  , +1)
1.0
>>> nextafter(10  , -1)
9.9999999999999982
>>> nextafter(10  , +1)
9.9999999999999982
```
Ha! It is not only assymetric, but inconsistent!

jacg:
Bug or feature?

gonzaponte:
They do have a [test](https://github.com/numpy/numpy/blob/master/numpy/core/tests/test_half.py#L249) about it, but I don't quite understand it. The array they create doesn't make sense to me.

jacg:
OK, here's the relevant part of the docs: 

> Return the next floating-point value after x1 towards x2, element-wise.

the key part being **towards x2**. (Unfortunately this is written right at the top of the docs, so looking for the meaning of `x2` in the documentation for the `x2` parameter didn't give me any joy.)

So, instead of `+1` it requires something like `+np.inf`

jacg:
Well,

    eps = 1e-12
    voxel_edges_lo -= eps
    voxel_edges_hi += eps

seems to do the trick, for values of `eps` ranging from about `1e-9` to `1e-14`.

But it's not a solution that I'm particularly happy with. It seems like a hack.

I should really try to understand the exact circumstances under which the problem arises, but my time available for IC is going to be very limited until the end of the week, and essentially zero for the week-and-a-half after that. So I'm probably dropping off the radar for a while.

jacg:
I have pushed the aforementioned solution.

I weep for `voxelize_hits` who was fairly slim in its youth, but is now becoming rather obese.

paolafer:
Just to confirm, the current version or your voxel-size branch should have this problem fixed, right, @jacg? I will proceed to test it.

jacg:
Yes, the fix is there. How general the fix is, I do not know, as I haven't analysed the problem to fully understand it.

Also, it shouldn't make much difference in real life, because the frequency with which this problem arises should be absolutely miniscule: how likely is it that two extreme voxels are separated by 'exactly' a multiple of the voxel width?

paolafer:
Yes, it should probably make no difference.

jjgomezcadenas:
@paolafer do not let this PR to grow a beard

paolafer:
I'm on it!

paolafer (commented):
Could you please clarify to me what the assertion of test_voxelize_hits_respect_voxel_dimensions means? I understand that the remainder of the division should be 0, but why unit?

jacg:
`unit` is just a short synonym for `requested_voxel_dimensions`. As discussed above in this thread, we want the dimensions of the voxels that were returned to be (very close to) a multiple of this unit size. The test does this by looking at the remainder when dividing by the unit size.

The problem is that both `1.000000001` times the unit size and `5.999999999` times the unit size **should** succeed, but the latter evaluates to `0.9999999999` and will fail if you only admit `off_by % 0`: you need `off_by % unit` for that one.

In summary, the second case is needed for those values which are approximately close from below, because the first case only spots those which are approximately close from above.

paolafer (approved):
This PR fixes a bug in the dimension of the voxels. Now voxels have all the same dimensions. In my opinion it can be approved and merged.

jacg:
The voxels all had the same dimensions before ... it's just that those dimensions were usually slightly smaller than the requested ones.

The code in this PR leaves a margin in the outer voxels (in order to match the requested voxel size exactly). Consequently the blobs will have slightly lower energy than before.

paolafer:
As far as I recall, the first changes in this PR have been done to address the fact that the positions of the voxels were not separated by a multiple of the voxel dimension. Maybe I'm confused and that problem was solved by another PR?

jacg:
Yes, this PR addresses the fact that the returned voxels were not of the specified required dimension, but slightly smaller. You earlier statement (emphasis mine)

> **Now** voxels have all the same dimensions.

is misleading, because it suggests that before this PR the voxels did *not* have the same dimensions. This is not true: they *did* have the same dimensions *as eachother* but those dimensions were slightly smaller than the requested dimensions.

paolafer:
Ok, it's all clear! I have used the code and it produces voxels with the requested dimensions. Green light from my side!

PR #354, opened 2017-10-04
Percentual sipm threshold

gonzaponte:
This PR introduces the possibility of using two types of thresholds for the SiPMs:
- Common: a single value in pes for all the SiPMs.
- Individual: each SiPM has its own threshold based on a common percentage of noise reduction. Since each SiPM has a different PDF, they also have different thresholds.

The code to compute the thresholds based on the percentage was already in IC, but it has been cleaned up a bit. The cython function to extract the signal from the sipms has been adapted to take both types of thresholds and a flag has been introduced in the cities to switch between both modes. The tests included in this PR are, perhaps, still not very exhaustive. I have introduced three:
- A test for `irene` to check that can run in both modes (this is actually a parametrization of a previous test to run with the two values of the flag). However, it does not check the output, as this is more complicated, but it is somehow complemented with:
- Two tests for the function `signal_sipm` in `peak_functions`, one for each threshold mode. These tests ensure the output of the function is the one we expect.

The configuration files from now on must include a new option `thr_sipm_type` that must be equal to `"common"` or to `"individual"` (case insensitive).

### DIFF
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -516,7 +516,7 @@ class CalibratedCity(DeconvolutionCity):
         elif conf.thr_sipm_type.lower() == "individual":
             # In this case, the threshold is a percentual value
             noise_sampler = SiPMsNoiseSampler(self.run_number)
-            self.thr_sipm = noise_sampler.ComputeThresholds(conf.thr_sipm)
+            self.thr_sipm = noise_sampler.compute_thresholds(conf.thr_sipm)
         else:
             raise ValueError(("Wrong value in thr_sipm_type. It must"
                               "be either 'Common' or 'Individual'"))
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -488,6 +488,12 @@ class CalibratedCity(DeconvolutionCity):
           a) equalize to pes;
           b) compute a MAU that follows baseline and keep samples above
              MAU + threshold.
+             The threshold can be used in two ways:
+             - A single value for all sipms (thr_sipm_type = "common"). In this
+               case, thr_sipm is the threshold value in pes.
+             - A value for each SiPM, based on a common the percentage
+               of noise reduction (thr_sipm_type = "individual"). In this case,
+               thr_sipm is the percentage.
        """
 
     parameters = tuple("""n_mau thr_mau thr_csum_s1 thr_csum_s2 n_mau_sipm thr_sipm thr_sipm_type""".split())
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -490,7 +490,7 @@ class CalibratedCity(DeconvolutionCity):
              MAU + threshold.
        """
 
-    parameters = tuple("""n_mau thr_mau thr_csum_s1 thr_csum_s2 n_mau_sipm thr_sipm""".split())
+    parameters = tuple("""n_mau thr_mau thr_csum_s1 thr_csum_s2 n_mau_sipm thr_sipm thr_sipm_type""".split())
 
     def __init__(self, **kwds):
 
@@ -503,9 +503,17 @@ class CalibratedCity(DeconvolutionCity):
         self.thr_csum_s2 = conf.thr_csum_s2
 
         # Parameters of the SiPM signal
-        self.n_MAU_sipm = conf.n_mau_sipm
-        self.  thr_sipm = conf.  thr_sipm
-
+        self.n_MAU_sipm  = conf.n_mau_sipm
+        if   conf.thr_sipm_type.lower() == "common":
+            # In this case, the threshold is a value of threshold in pes
+            self.thr_sipm = conf.thr_sipm
+        elif conf.thr_sipm_type.lower() == "individual":
+            # In this case, the threshold is a percentual value
+            noise_sampler = SiPMsNoiseSampler(self.run_number)
+            self.thr_sipm = noise_sampler.ComputeThresholds(conf.thr_sipm)
+        else:
+            raise ValueError(("Wrong value in thr_sipm_type. It must"
+                              "be either 'Common' or 'Individual'"))
 
     def calibrated_pmt_mau(self, CWF):
         """Return the csum and csum_mau calibrated sums."""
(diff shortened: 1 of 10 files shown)


gonzaponte:
I have had doubts with regard to the removal of the type of the argument `thr` in `signal_sipm`. Initially, I thought it would be simpler to remove it rather than changing it to an array so the tests would remain unchanged and therefore everything is more robust. I didn't realize at the time that there were no tests for that function. Now it is not clear to me whether it is better to accept an untyped value (which is later converted into an array anyway) or to force the user the call it with an array. The latter makes the function simpler, maybe, but the former is more comfortable. There is probably not a big difference, but it is open for discussion.

jahernando (commented):
I do not understand why you replace the typed parameter 'trh' in invisible_cities/reco/peak_functions_c.pyx.

This is not related with the PR but in the line 450, I assume threshold is in pes, but to apply a filter is multiplied by adc_to_pes, it should not be divided?

gonzaponte:
> I do not understand why you replace the typed parameter 'trh' in invisible_cities/reco/peak_functions_c.pyx.

So it can accept both a `float` (common threshold) and an `array` (individual thresholds).

> In the line 450, I assume threshold is in pes, but to apply a filter is multiplied by adc_to_pes, it should not be divided?

I assume is line 450 of `peak_functions_c.pyx`. This is actually a wrong name that has been there since the beginning. It is actually `pes_to_adc`.
Actually, this is everywhere, even in our database... Since this has nothing to do with this PR, I will open an issue.

jahernando (changes requested):
Please add a comment in base_cities to indicate the possible values of the thr_sipm_type and thr_sipm

jahernando:
> This is actually a wrong name that has been there since the beginning.
> 
> 

I see this is the usual JJ’s tactic of confusing the enemies and the friends around.

gonzaponte:
> Please add a comment in base_cities to indicate the possible values of the thr_sipm_type and thr_sipm

I think it is obvious. There are two comments indicating what `thr_sipm` means in each case.

```
if   conf.thr_sipm_type.lower() == "common":
    # In this case, the threshold is a value of threshold in pes
    [...]
elif conf.thr_sipm_type.lower() == "individual":
    # In this case, the threshold is a percentual value
    [...]
else:
    raise ValueError(("Wrong value in thr_sipm_type. It must"
                      "be either 'Common' or 'Individual'"))
```

jahernando:
Yes, when you read the code, but not in the header comments of the class, the first place one look to understand the behavior.

gonzaponte:
Ok. I agree.

jahernando (dismissed):
The PR includes the possibility of do Zero-Supression of SiPM wave-forms, using a threshold on the cdfs of the SiPMs noise, and accepting only time slices in the wf with a signal with a p-probability greater than a certain value. 

The PR extends the interface of signal_sipm function, that now it includes a parameter to select the ZS method (thr_sipm_type) and a parameter of the actual threshold (thr_sipm) value(s).

There are two test to verify that the signal_sipm is properly doing the ZS for the two method 'common' and 'individual'.

Not related with this PR, it seems to me that an extra test is needed to verify the operation of ComputeThreshold class. Simply to check that the returned thresholds for the individual SiPMs correspond in fact to a given value of their cdfs.

I approve the PR.

jahernando (commented):
In fact, I think is better that you include the test of the ComputeThreshold class that I was commenting before, in this PR, to have a more consistent picture.

gonzaponte:
I don't know why the test `test_uniform_smearing_stats` fails in travis. I can't even get a failure in my computer. My version of hypothesis is quite old (3.12.0), but this is something purely statistical. It should only fail 0.3% of the time, and I allow 1 failure in 5 runs.

jacg:
Two things:

1. `pip install hypothesis 3.28`
2. `HYPOTHESIS_PROFILE=hard pytest -k uniform_smearing_stats`

gonzaponte:
Oh, I thought we had switched back to `dev`. The `hard` profile is causing of the failure. Are we changing the profile to `dev` back again? Should I be more permissive in the assertion or use the context manager to reduce the number of examples for that particular test?

jacg:
> I thought we had switched back to `dev`

Oh nonononnonoooooo! Why would we?

The whole point of `dev` is that we use it during development, so that we get quick turnaround in our local repos as we're hacking away. The whole point of `hard` is that Travis can run it, because given that Travis already has to download and install our whole system, the relative overhead of running more thorough tests isn't that significant and, usually, there is no human sitting there doing nothing but waiting for it to complete.

> Are we changing the profile to `dev` back again?

For Travis? No way! (Travis *never* had `dev`, `dev` is an emasculated version of the default profile.)

`dev` skimps on the testing. We're prepared to live with it in our local repos, as long as someone, somewhere, runs the hard tests regularly for us. That someone is Travis.

> Should I be more permissive in the assertion or use the context manager to reduce the number of examples for that particular test?

Sorry, can't help you there, as that would require me to dig into details for which I have no time at the moment.

gonzaponte:
Ok, I agree. Thank you.

jjgomezcadenas:
@jahernando this is stale. I think this is an important PR to include in our next release.

abotas:
The PR looks great and very thoroughly tested to this reviewer! 

I think it's unfortunate that we add a flag "common" or "individual for sipm thr type. But if we want to keep the common threshold in some cases, which i think we do, the flag seems necessary. 

Just to reiterate, this PR allows individual `thr_sipm`, (the time bin by time bin cut)  but not individual `thr_sipm_s2` (the integral s2 cut), correct? 

---
One thing that bothers me a little, is that if we decide to use individual `thr_sipm`s, i think then using a common `thr_sipm_s2` on the integral s2 gets a little messy. I am imagining that we might introduce a bias where our software will incorrectly show noisy SiPMs seeing considerably less light than well behaved SiPMs. To some extent maybe this is unavoidable.  _However particularly for long s2s, SiPMs that are slightly more noisy might have to record considerably more light to pass the `thr_sipm_s2` when we use individual `thr_sipm`._ Does this make sense? It would be nice for a few people to consider this for a second, even if it is just to tell me I am imagining an issue. @gonzaponte @jjgomezcadenas whoever!

All that said, I don't think that^^ is a problem that must be addressed in this PR, just something we want to be aware of if it is a potential issue!

abotas:
There are quite a few commits here so I will refrain from merging until tomorrow, just to give other reviewers one last (quick) chance to complain if they want! But again, the PR looks good and ready to this reviewer!

abotas (approved):
Approved.

abotas:
Rebased.

PR #355, opened 2017-10-05
mc_track_writer does not work for non-consecutive events

jmbenlloch:
In the last MC production we found a bug: when processing files with 10k events, the number of events in `MCTracks` table usually was much smaller (around ~1k).

The problem is that `mc_track_writer` does not work properly if any event is skipped, so if irene filters out some event, since that moment the rest of MCTracks will not be copied to the output file.

This is shown in a test included in the first commit in this PR.


jmbenlloch:
The second commit includes the fix.

gonzaponte:
Ok. It took me a while to understand it. What you do is:
- Read from the last row. If an event (or more than one) has been skipped, the event number will be lower than the one you requested so you continue reading, but doing nothing with it.
- When the event number is the requested one it begins dumping to the table.
- When the event number changes again, stop reading.

Is this correct?

jmbenlloch:
Yes, that is what the code does.

gonzaponte (approved):
Important fix concerning the propagation (copy) of true information to higher-level data structures. Because of event selection, there was a mismatch between the true events and the reconstructed ones.

On the other hand, the `mc_track_writer` was untested, which made this error go silently for a long time. Although the test introduced here is a good one, we need to be more exhaustive. I suggest we move on with this PR now, but we pay the technical debt as soon as we can.

Good job.

PR #358, opened 2017-10-05
Cleanup

jacg:
This PR contains a miscellaneous collection of fixes and cleanups that emerged out of my post-hoc review of a number of PRs that were accepted over the summer. This has been sitting in my repo over the last weeks as I have been otherwise occupied. As I don't expect to have time in the near future to pursue these any further, I submit it as it is.

### DIFF
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -955,24 +955,6 @@ class TrackCity(HitCity):
          """
         return paf.voxelize_hits(hits, self.voxel_dimensions)
 
-    def make_tracks(self, evt_number: float, evt_time: float, voxels: List[Voxel]) -> TrackCollection:
-        """Makes a track collection. """
-
-        tc = TrackCollection(evt_number, evt_time) # type: TrackCollection
-
-        track_graphs = paf.make_track_graphs(voxels, self.voxel_dimensions) # type: Sequence[Graph]
-
-        for trk in track_graphs:
-            distances = paf.shortest_paths(trk)
-            a,b       = paf.find_extrema(distances) # type: Tuple[Voxel, Voxel]
-            ba, bb = paf.blobs(trk, blob_radius) # type: Tuple[List[Voxel], List[Voxel]]
-            blob_a = Blob(ba) # type: Blob
-            blob_b = Blob(bb)
-            blobs = (blob_a, blob_b) if blob_a.E < blob_b.E else (blob_b, blob_a) # type: Tuple[Blob, Blob]
-            track = Track(paf.voxels_from_track_graph(trk), blobs)
-            tc.tracks.append(track)
-        return tc
-
 
 class TriggerEmulationCity(PmapCity):
     """Emulates the trigger in the FPGA.
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -140,7 +140,7 @@ class City:
         self.print_mod    = conf.print_mod  # default print frequency
         self.first_event, self.last_event = self._event_range()
         self.set_up_database()
-        self.daemons = list(map(invoke_daemon, kwds.get('daemons', [])))
+        self.daemons = tuple(map(invoke_daemon, kwds.get('daemons', [])))
 
     def detect_unknown_parameters(self, kwds):
         known = self.allowed_parameters()
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -947,7 +947,7 @@ class TrackCity(HitCity):
         self.voxel_dimensions  = self.conf.voxel_dimensions # type: np.ndarray
         self.blob_radius       = self.blob_radius # type: float
 
-    def voxelize_hits(self, hits : Sequence[Hit]) ->List[Voxel]:
+    def voxelize_hits(self, hits : Sequence[Hit]) -> List[Voxel]:
         """1. Hits are enclosed by a bounding box.
            2. Boundix box is discretized (via a hitogramdd).
            3. The energy of all the hits insidex each discreet "voxel" is added.
@@ -955,8 +955,7 @@ class TrackCity(HitCity):
          """
         return paf.voxelize_hits(hits, self.voxel_dimensions)
 
-    def make_tracks(self, evt_number : float , evt_time: float,
-                    voxels : List[Voxel]) ->TrackCollection:
+    def make_tracks(self, evt_number: float, evt_time: float, voxels: List[Voxel]) -> TrackCollection:
         """Makes a track collection. """
 
         tc = TrackCollection(evt_number, evt_time) # type: TrackCollection
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -222,9 +222,8 @@ class City:
         """
         self.index_tables() # index cols in tables marked for indexing by city's writers
 
-        if hasattr(self, 'daemons'):
-            for deamon in self.daemons:
-                deamon.end()
+        for daemon in self.daemons:
+            daemon.end()
 
         print(self.cnt)
         return self.cnt
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -84,7 +84,7 @@ from .. filters.s1s2_filter     import s2si_filter
 from .. filters.s1s2_filter     import S12Selector
 from .. filters.s1s2_filter     import S12SelectorOutput
 
-from .. daemons.idaemon         import invoke_daemon
+from .. daemons.idaemon         import summon_daemon
 
 from typing import Sequence
 from typing import List
@@ -140,7 +140,7 @@ class City:
         self.print_mod    = conf.print_mod  # default print frequency
         self.first_event, self.last_event = self._event_range()
         self.set_up_database()
-        self.daemons = tuple(map(invoke_daemon, kwds.get('daemons', [])))
+        self.daemons = tuple(map(summon_daemon, kwds.get('daemons', [])))
 
     def detect_unknown_parameters(self, kwds):
         known = self.allowed_parameters()
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -332,7 +332,6 @@ class City:
            obtained by divinding the RD PMT vector and the sample
            time of the electronics (25 ns). """
         with tb.open_file(filename, "r") as h5in:
-            #pmtrd, sipmrd = self._get_rd(h5in)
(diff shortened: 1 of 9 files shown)


jjgomezcadenas:
Yes, I will review this, since it is mostly cleaning after my mess. Thanks, @jacg and apologies!

gonzaponte:
THANK YOU! All these tiny stuff has been bothering me for a long time... and there is still a lot to do.

Please, @paolafer @jerenner @jjgomezcadenas, take a look at the changes in the whole paolina/tracks/blobs stuff. I know very little about that section of the code.

Some comments:
- In the second commit, notice the misspelling of "daemon":
```
for deamon in self.daemons:
    deamon.end()
```
- You changed the output of `compute_blobs` and the tests do not complain, how is so?
- In your last commit you mention that the `voxels_from_track_graph` is used only in one place in the code, but I don't see any change apart from the removal of a test.

jacg:
>     In the second commit, notice the misspelling of "daemon":

Bloody hell, it took me over 3 minutes of staring at it to spot where it was!

paolafer:
As far as the paolina functions are concerned, I think the PR is fine. A function which returned voxels from the graph has been eliminated, because it was redundant and a function which promises to return blobs returns effectively blobs and nothing else.

jacg:
> You changed the output of compute_blobs and the tests do not complain, how is so?

Memory suggests that it's because these values were not being used at the only call site (which is why I removed them). But looking through it now, it looks like I didn't change the call site, which occurs in `make_tracks` which ... isn't used or tested anywhere!

@jjgomezcadenas Any idea why `make_tracks` is not used or tested anywhere?

If I completely remove `compute_blobs` and/or `make_tracks`, all the tests still pass. Even in the current `nextic/master`!

jacg:
> In your last commit you mention that the voxels_from_track_graph is used only in one place in the code, but I don't see any change apart from the removal of a test.

I can see the removal of 

+ `voxels_from_track_graph`
+ a test for the latter

So, where was this one place that `voxels_from_track_graph` was used? you ask! It's in our friend `make_tracks` which (as I pointed out in my last comment)

+ is not tested
+ is not used anywhere

which would explain why I didn't spot that I forgot to remove it the call to `voxels_from_track_graph`, because, as you all well know, I use failing tests to tell me what remains to be done.

jacg:
> In the second commit, notice the misspelling of "daemon":

I have fixed this one locally, but as its buried at the bottom of a tall rebase, I won't push it just yet.

jacg:
> ... and there is still a lot to do.

Absolutely! But it's a start.

jjgomezcadenas (commented):
This PR includes three different actions in the code.
1) A fix in the way daemons where invoked. The new version is much better than the original.
2) A general cleanup of code. Needed and much appreciated.
3) Removal of at least two functions that were being used in analysis by at least one developer (me). I don't agree with removal of those functions. The arguments that functions were not being used in IC is not correct. IC hay many functions that are not being used by IC code (reconstruction) but by client IC code (analysis, often in notebooks). 



>Sanitize bizarre setting of daemons attribute  …

The right (and simpler) way to do it

>Remove now-superfluous hasattr before daemons loop  

Consistent with previous

>Turn daemons list into a tuple 

If you insist

> Fix misleading function name invoke_daemon -> find_daemon

Imposing your aesthetics over mine. That's OK

>Stop suppressing ModuleNotFoundError for unknown daemons
 
Of course

>Remove commented-out code
>Add missing whitespace in type annotations
>Remove pointless (and sometimes incorrect!) docstrings
>Cosmetics
>Indent incorrectly positioned method

Thanks
>Remove unused and duplicated make_tracks function  

make_tracks is actually being used in my notebooks. In general a function in IC does not need to be exercised in IC itself, because ICARO (e.g, the analysis code) is a client of IC. If the function is duplicated I agree with removing the duplication, but I hope that make_tracks has not completely disappeared from the code, can you confirm?

>Fix incorrect return type annotation in compute_blobs

>Stop returning extrema voxels from compute_blobs
 
Why? For analysis purposes I found that useful. 

>Remove pointless compexity in voxels_from_track_graph  …

Why pointeless? For analysis I wanted to use objects of type Voxel, not of type track.node

> Remove voxels_from_track_graph  
Why? I was using that function in analysis.

jacg:
>> Turn daemons list into a tuple

> If you insist

It is never meant to be modified after it has been created. It is the equivalent of declaring it const in C++. It is The Right Thing To Do.

>> Fix misleading function name invoke_daemon -> find_daemon

> Imposing your aesthetics over mine. That's OK

It is not aesthetics, it is semantics: the function does not invoke any daemons, it merely finds them. Giving the function a name which implies that it does something it does not, is The Wrong Thing To Do.

(Hmm ... I wonder whether you like *invoke* because *invoking daemons* has a nice poetic sound to it. If that is the case, then I *will* impose my aesthetics: this is a programming project, not a literary one; *invoke* has a specific meaning in the context of programming, a meaning which would mislead about the function's behaviour.)

>>    Remove unused and duplicated make_tracks function

> If the function is duplicated I agree with removing the duplication, but I hope that make_tracks has not completely disappeared from the code, can you confirm?

The same function (though formatted slightly differently and with a few superfluous lines commented out) can be found at the bottom of `paolina_functions.py`

> make_tracks is actually being used in my notebooks. In general a function in IC does not need to be exercised in IC itself, because ICARO (e.g, the analysis code) is a client of IC.

A function in IC *does* need to be exercised in IC itself: there must be a test which would notice the disappearance of serious breakage (ideally any breakage) of this function. Guess what happens in the tests when the second copy of `make_track` is completely removed.

The tests all still pass!

Please appreciate that having functions in the code whose removal triggers no tests, is a huge obstacle in the way of any developer who needs to fix something in such a function. It is a place where a high rate of interest is paid on technical debt.

>> Stop returning extrema voxels from compute_blobs

> Why?

Because it looked completely superflous.

> For analysis purposes I found that useful.

Then back it goes.

>> Remove pointless compexity in voxels_from_track_graph …

> Why pointeless? For analysis I wanted to use objects of type Voxel, not of type track.node

The nodes of a graph are of whatever type you put into the graph. If you put in `Voxel`s (as we do) then `track.nodes()` returns `Voxel`s. There is no type `track.nodes`.

Your code

1. obtained a list of `Voxel`s from `track.nodes()`
2. unpacked each `Voxel` into its component positions and energy
3. used the components to construct a new, identical `Voxel`
4. created a list containing the new `Voxel`s: one identical to the list originally obtained from `track.nodes()`.

As such it is a waste of 

1. CPU cycles
2. Memory
3. Lines of code, and therefore ...
4. ... readers' (developers') brain cycles.

>>Remove voxels_from_track_graph

> Why? I was using that function in analysis.

After removing the unnecessary complexity I have just described, It's obviously just a trivial wrapper around `track.nodes()`. Therefore it's value is questionable. In an exchange with you on the topic elsewhere in GitHub, I pointed this out and acknoweledged that, in spite of its triviality, it's name, which sends a higher-level message in IC, does provide some value. But I questioned whether it was worth the associated maintenance cost. (The fact that we are discussing it here, for at least the second time now, is a cost.)

I proposed removing it and solicited your opinion. I got no reply. I concluded that if you do not choose to defend it when I threaten it with removal, then you do not care sufficiently for it to be kept.

Actually, I thought you probably got lost in the many exchanges that were taking place, and concluded that the best way to get your attention would be to remove it in a PR ... and here we are!

So, let me repeat: Does the benefit of being able to write `voxels_from_track_graph(track_graph)` instead of `track_graph.nodes()` justify the cost of maintaining the function and its test?

I remind you that the results are **absolutely** identical.

jacg:
I have rebased to

1. Fix the spelling of `daemons`
2. Remove the commit which stops returning extrema voxels from `compute_blobs`
3. Catch up with `nextic/master`

jjgomezcadenas:
Previous discussion:

>Hmm ... I wonder whether you like invoke because invoking daemons has a nice poetic sound to it. 

You got it

>I will impose my aesthetics: this is a programming project, not a literary one;

so be it.


>I proposed removing it and solicited your opinion. I got no reply. I concluded that if you do not choose to defend it when I threaten it with removal, then you do not care sufficiently for it to be kept.
Actually, I thought you probably got lost in the many exchanges that were taking place, and concluded that the best way to get your attention would be to remove it in a PR ... and here we are!

Yes, I got lost in the cross fire. These days is difficult for me to keep track of the activity. 

>So, let me repeat: Does the benefit of being able to write voxels_from_track_graph(track_graph) instead of track_graph.nodes() justify the cost of maintaining the function and its test?

I believe so. It saves cycles of the end-user, who does not need to remember that track.nodes() actually returns the voxels. Not a big issue, anyway, but I would keep it. At least my limited CPU finds the syntactic sugar useful.

jacg:
>> Hmm ... I wonder whether you like *invoke* because *invoking daemons* has a nice poetic sound to it.

> You got it

I would let you `summon_daemons` if you want. That's a name which does not imply that the daemon will be called in the programming sense.

>> Does the benefit of being able to write `voxels_from_track_graph(track_graph)` instead of `track_graph.nodes()` justify the cost of maintaining the function and its test?

> I believe so. 

Fine by me. But the test needs rewriting. The function is meant to be a dead ringer for `x.nodes()` so the test should check exactly that, rather than the complex stuff it does right now, including calling `voxelize_hits` and `make_track_graphs` which are completely orthogonal to its behaviour.

I might find the time to fix it today.

jjgomezcadenas:
>I would let you summon_daemons if you want. That's a name which does not imply that the daemon will be called in the programming sense.

Who's the poet now? summon_daemons is as poetic as it gets (in fact more than invoke_daemons) and syntactically equivalent to find_daemons. Would love to see that name, if you wish to humour an old nuts... but totally up to you, find_daemons is perfect too.

>Fine by me. But the test needs rewriting. The function is meant to be a dead ringer for x.nodes() so the test should check exactly that, rather than the complex stuff it does right now, including calling voxelize_hits and make_track_graphs which are completely orthogonal to its behaviour.

Agreed!

>I might find the time to fix it today.

Thanks again!

jacg:
> syntactically equivalent to find_daemons

Methinks you mean *semantically* equivalent.

> Would love to see that name

Your wish is my command. Consider it done.

jjgomezcadenas:
> Methinks you mean semantically equivalent.

Indeed.

>Your wish is my command. Consider it done.

OK, I guess you are gonna fix the test too, right? After that I guess this can be approved with my thanks!

jacg:
1. Spelling of daemons has been fixed.
2. `compute_blobs` returns extreme voxels once more.
3.  Daemons are now summoned rather than found.
4. `voxels_from_track_graph` is back, and ...
5. ... its test has been cleaned up.
6. We're sitting pretty on top of `nextic/master`.

Anything else?

jjgomezcadenas (approved):
With hearty thanks!

PR #359, opened 2017-10-06
Fix chi2 and errors in fit function

bpalmeiro:
- Chi2
Implemented chi2 computation, based on scipy.stat.chisquare is only valid when data smearing is Poisson distributed. So, a general chi2 function is added, with formula extracted from [1].

- Errors
Fit errors, given by scipy.optimize.curve_fit were not correct as flag absolute_sigma was set to false. When true, "[data] sigma describes one standard deviation errors of the input data points", so covariance matrix estimation is based on these values. Otherwise, data sigma is only taken as "relative weights of the data points" and covariance matrix is not affected by the overall magnitude of the values in sigma, just by relative magnitudes.

Both improvements are shown in https://github.com/bpalmeiro/ICARO/blob/Chi2_demostration/icaro/Chi2_demonstration/Chi2_demonstration.ipynb

[1]  Numerical Recipes. The Art of Scientific Computing, 3rd Edition, 2007, ISBN 0-521-88068-8. Pages 778-780

### DIFF
--- a/invisible_cities/core/fit_functions.py
+++ b/invisible_cities/core/fit_functions.py
@@ -128,19 +128,30 @@ def fit(func, x, y, seed=(), fit_range=None, **kwargs):
         if "sigma" in kwargs:
             kwargs["sigma"] = kwargs["sigma"][sel]
 
+    abs_sigma = "sigma" in kwargs
+
     vals, cov = scipy.optimize.curve_fit(func,
                                          x, y,
                                          seed,
+                                         absolute_sigma = abs_sigma,
                                          **kwargs)
 
-    fitf = lambda x: func(x, *vals)
-    fitx = fitf(x)
-    chi2, pval = scipy.stats.chisquare(y[fitx != 0], fitx[fitx != 0], len(vals))
+    fitf    = lambda x: func(x, *vals)
+    fitx    = fitf(x)
+    nonzero = fitx != 0
+
+    sigma_r = kwargs.get("sigma", np.ones_like(y))
+    sigma_r = sigma_r[nonzero]
+
+    chi2, pval = get_chi2_and_pvalue(y    [nonzero],
+                                     fitx [nonzero],
+                                     len(y[nonzero]) - len(vals),
+                                     sigma_r)
 
     return FitFunction(fitf,
                        vals,
                        get_errors(cov),
-                       chi2 / (len(x) - len(vals)),
+                       chi2,
                        pval)
 
 
--- a/invisible_cities/core/fit_functions.py
+++ b/invisible_cities/core/fit_functions.py
@@ -28,6 +28,41 @@ def get_errors(cov):
     return np.sqrt(np.diag(cov))
 
 
+def get_chi2_and_pvalue(ydata, yfit, ndf, sigma=None):
+    """
+    Gets reduced chi2 and p-value
+
+    Parameters
+    ----------
+    ydata : np.ndarray
+        Data points.
+    yfit : np.ndarray
+        Fit values corresponding to ydata array.
+    sigma : np.ndarray
+        Data errors. If sigma is not given, it takes the poisson case:
+            sigma = sqrt(ydata)
+    ndf : int
+        Number of degrees of freedom
+        (number of data points - number of parameters).
+
+    Returns
+    -------
+    chi2 : float
+        Reduced chi2 computed as:
+            chi2 = [sum(ydata - yfit)**2 / sigma**2] / ndf
+    pvalue : float
+        Fit p-value.
+    """
+
+    if sigma is None:
+        sigma = ydata**0.5
+
+    chi2   = np.sum(((ydata - yfit) / sigma)**2)
+    pvalue = scipy.stats.chi2.sf(chi2, ndf)
+
+    return chi2 / ndf, pvalue
+
+
 # ###########################################################
 # Functions
 def gauss(x, amp, mu, sigma):
--- a/invisible_cities/reco/corrections_test.py
+++ b/invisible_cities/reco/corrections_test.py
@@ -2,16 +2,18 @@ from collections import namedtuple
 
 import numpy as np
 
-from ..core             import fit_functions as fitf
-from ..core.exceptions  import ParameterNotSet
-from ..reco.corrections import Correction
-from ..reco.corrections import Fcorrection
-from ..reco.corrections import LifetimeCorrection
-from ..reco.corrections import LifetimeRCorrection
-from ..reco.corrections import LifetimeXYCorrection
-from ..reco.corrections import opt_nearest
-from ..reco.corrections import opt_linear
-from ..reco.corrections import opt_cubic
+from ..core                import fit_functions as fitf
+from ..core .exceptions    import ParameterNotSet
+from ..reco .corrections   import Correction
+from ..reco .corrections   import Fcorrection
+from ..reco .corrections   import LifetimeCorrection
+from ..reco .corrections   import LifetimeRCorrection
+from ..reco .corrections   import LifetimeXYCorrection
+from ..reco .corrections   import opt_nearest
+from ..reco .corrections   import opt_linear
+from ..reco .corrections   import opt_cubic
+from ..icaro.hst_functions import poisson_sigma
+from ..icaro.hst_functions import shift_to_bin_centers
 
 from numpy.testing import assert_allclose
 from pytest        import fixture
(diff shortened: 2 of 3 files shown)


invisible_cities/core/fit_functions.py
```diff
@@ -27,6 +27,35 @@ def get_errors(cov):
...
+
+    Parameters
+    ----------
+    ydata : np.ndarray
+        Data array
+    ydata : np.ndarray
+        Fit funtion values corresponding to ydata array
+    sigma : np.ndarray
+        Data errors array. If no sigma is given, it takes the poisson case.
+
```
gonzaponte: ndf is not documented

invisible_cities/core/fit_functions.py
```diff
@@ -96,16 +125,26 @@ def fit(func, x, y, seed=(), fit_range=None, **kwargs):
     vals, cov = scipy.optimize.curve_fit(func,
                                          x, y,
                                          seed,
+                                         absolute_sigma=True,
                                          **kwargs)
 
     fitf = lambda x: func(x, *vals)
     fitx = fitf(x)
-    chi2, pval = scipy.stats.chisquare(y[fitx != 0], fitx[fitx != 0], len(vals))
+
```
gonzaponte: use a variable to hold the result of `fitx != 0`. It is used in 4 different places.

invisible_cities/core/fit_functions.py
```diff
@@ -96,16 +125,26 @@ def fit(func, x, y, seed=(), fit_range=None, **kwargs):
...
+    if not "sigma" in kwargs:
+        sigma_r   = None
```
gonzaponte: You can use `sigma_r = kwargs.get("sigma")` which returns `None` if the key is not found.  Or `sigma_r = kwargs.get("sigma", None)`, which is more explicit.

gonzaponte (changes requested):
Great work! Very well documented and demonstrated.

jjgomezcadenas:
Really good work. The Galician gang (GG's) are increasing even more their level of Karma.
I suggest to turn the examples in the notebooks into tests! Simply take the results that you obtain for chi2 and P-value with both methods and assert them. Like that the test (with a few comments) also documents the work!

Incidentally, the result:
chi2     : 0.9911404817192085
P-value  : 1.328398e-21

Which you obtain for Poissonian errors with old method does not make sense for P-value. 

Good job (I will live @gonzaponte to approve it, after you add the little changes he suggests and perhaps the docs-test. :)!

invisible_cities/core/fit_functions.py
gonzaponte: I don't want to be pedantic, but can we replace `len(data)` by `number of data points`?

gonzaponte (dismissed):

gonzaponte:
I like this, but a second view made me realize that the `ndf` argument in `get_chi2_and_pvalue` is optional and set to 1. This shouldn't be like that, as the number of degrees of freedom is essential to the output. Also, at the risk to be pedantic again, can you mark the optional arguments in the docstring? Finally, we must add tests to ensure this new function is working properly and review those of the `fit` function which are probably very soft on the `chi2` output.

jjgomezcadenas:
This review is stale, @gonzaponte, why? No advantage in letting open PRs which appear good enough to grow old. We are not running a winery here.

gonzaponte:
> why?

Because we are still working on making it right once and for all. There are different interpretations of the parameters for scipy's `curve_fit` function. We want to handle all of them correctly. It seems stale, but we are pera-discussing it behind the scenes.

We have a working version which yields incorrect results only for the chi2 value and the errors of the parameters (which we are not paying too much attention to). We need to fix this, but it will not have a big impact on our lives.

invisible_cities/reco/corrections_test.py
```diff
@@ -1,13 +1,14 @@
...
-from ..reco.corrections import LifetimeRCorrection
-from ..reco.corrections import LifetimeXYCorrection
+from ..core                import fit_functions as fitf
+from ..core.exceptions     import ParameterNotSet
+from ..reco.corrections    import Correction
+from ..reco.corrections    import Fcorrection
+from ..reco.corrections    import LifetimeCorrection
+from ..reco.corrections    import LifetimeRCorrection
+from ..reco.corrections    import LifetimeXYCorrection
+from ..icaro.hst_functions import poisson_sigma
```
gonzaponte: Please move `poisson_sigma` to `core_functions.py`. The `icaro` package will (probably) be removed in the near future.

invisible_cities/core/fit_functions.py
```diff
@@ -137,8 +137,8 @@ def fit(func, x, y, seed=(), fit_range=None, **kwargs):
     fitx = fitf(x)
     nonzero = fitx != 0
 
-    sigma_r = kwargs.get("sigma", None)
-    if sigma_r is not None:
+    sigma_r = kwargs.get("sigma", 1)
+    if np.size(sigma_r) is not 1:
```
gonzaponte: Mmm, I don't like this `size` check... How about setting the default to `np.ones` of the corresponding size and applying the `nonzero` selection over it?

invisible_cities/reco/corrections_test.py
```diff
@@ -360,9 +361,10 @@ def test_corrections_1d(gauss_data_1d):
     mean = np.mean(Eevt)
     std  = np.std (Eevt)
 
-    y, x = np.histogram(Eevt, np.linspace(mean - 3 * std, mean + 3 * std, 100))
-    x    = x[:-1] + np.diff(x) * 0.5
-    f    = fitf.fit(fitf.gauss, x, y, (1e5, mean, std))
+    y, x  = np.histogram(Eevt, np.linspace(mean - 3 * std, mean + 3 * std, 100))
+    x     = x[:-1] + np.diff(x) * 0.5
```
gonzaponte: This is the same as the function `shift_to_bin_centers` in the `icaro` package. Can we move it also to `core` and use it here?

invisible_cities/reco/corrections_test.py
```diff
@@ -380,5 +382,6 @@ def test_corrections_2d(gauss_data_2d):
 
     y, x = np.histogram(Eevt, np.linspace(mean - 3 * std, mean + 3 * std, 100))
     x    = x[:-1] + np.diff(x) * 0.5
-    f    = fitf.fit(fitf.gauss, x, y, (1e5, mean, std))
+    sigma = poisson_sigma(y)
+    f    = fitf.fit(fitf.gauss, x, y, (1e5, mean, std), sigma=sigma)
     assert f.chi2 < 3
```
gonzaponte: This `f.chi2 < 3` is maybe a bit soft. I think we can `flaky` these two tests.

gonzaponte (commented):
Good work! This looks nice!

jjgomezcadenas:
@gonzaponte I don't think that ICARO will be removed from IC!!!!

What we are going to do is to prototype in ICARO (repo), and the promote to IC/ICARO when we are happy enough, that means that PRS to IC/ICARO will continue but with the standard level of strictness (tests and such). The package will not disappear.

This said, poisson_sigma it's probably better in core

gonzaponte:
@jjgomezcadenas Ok, then it is not obvious to me where these functions (those in `icaro/hst_functions.py` that are not for plotting) should go. I think we can postpone that discussion for another PR, do you agree?

jjgomezcadenas:
Why not stat_tools.py also in IC/ICARO? Either now or in a different PR, your pick

gonzaponte:
> Why not stat_tools.py also in IC/ICARO?

Agree. I was thinking about something like that too.

> Either now or in a different PR, your pick

I like to maintain PRs as self-contained as possible, so I would make a separate PR. But it can be performed by the same author as this one :)

jjgomezcadenas:
What on earth happened here, @gonzaponte, @bpalmeiro? This is huger amount of work, we need to get it out of the way

gonzaponte:
This is almost ready (I still need to review it a bit), however I am going to sit with @bpalmeiro to show him how to clean up the PR (it contains far too much noise). The time difference and the busy last weeks didn't help.

jjgomezcadenas:
This has been opened for ever, @bpalmeiro, @gonzaponte . Are you serious or what?

gonzaponte:
1. This is not urgent. There is nothing depending on this PR.
2. This has no conflict with TBR.
3. We will use this PR to practice some of the git-related operations and leave a clean history while reviewing with peras. However, @bpalmeiro is on vacation. I was going to do it by myself and redo it with him when he is back, but:
4. I don't have time AT ALL.
5. @bpalmeiro is back next week. We will begin the pera-review + cleanup then.
6. Again, there is nothing urgent here. Nothing in the data-processing chain depends on this. Please, keep calm, I am doing my best.

gonzaponte (approved):
This PR fixes a long-term issue with the definition of the chi2 and the p-value. The calculations have been encapsulated within a function, which is tested. The previous test suite for the rest of the code has also been fixed, improved and extended.

Great job!

PR #362, opened 2017-10-06
Fixes to manage.sh

gonzaponte:
This PR:
- Implements a function that was missing
- Adds a command already available to help string


jacg (approved):
Simple addition of things that were clearly missing. I haven't checked thoroughly whether anything else is missing, but that is no reasons to delay the admission of these fixes.

jacg:
Sorry, this has in fact been merged, but a little oversight in the process makes it look like it was closed instead. There doesn't seem to be a way to fix that which is not more trouble than it's worth.

PR #365, opened 2017-10-08
Redefine threshold and interval

gonzaponte:
This PR address the inconsistency of the definition of threshold and interval. This affects the value of some parameters in the code.
The work performed here is summarized in these points:
- Replace `>` by `>=` when dealing with thresholds.
- Fix inconsistency between the python and cython implementation of `minmax` (use right-open intervals in both).
- Exhaustively test python and cython `minmax` implementations.
- Cosmetics.

Making all intervals right-open creates, nonetheless, an uncomfortable situation when dealing with integer values. We have in our configuration:
```
some_integer_variable_min = 1
some_integer_variable_max = 2
```
Which means that only `some_integer_variable = 1` is accepted, but one may think it accepts both values looking at the configuration. On the other hand, setting both to 1 wouldn't work. I guess one just has to remember this is equivalent to python's `range` function, but it is not obvious to me. Opinions on this topic are very much appreciated.


invisible_cities/types/ic_types_test.py
```diff
@@ -1,55 +1,69 @@
...
 
-def make_cminmax(a,b):
+def make_minmax_c(a,b):
     # Ensure that arguments respect the required order
     if a > b: a, b = b, a
-    return cminmax(a, b)
+    return minmax_c(a, b)
+
+
+def make_xy(a,b):
```
jacg: Why do we want `make_xy` and `make_xy_c` as synonyms for `xy` and `xy_c`?

invisible_cities/types/ic_types_test.py
```diff
@@ -47,6 +47,31 @@ def test_minmax_does_not_accept_min_greater_than_max(a,b):
...
+
+@given(minmaxes)
+def test_minmax_center(mm):
+    center = (mm.max + mm.min)*0.5
+    assert mm.center == center
+
+
+@given(minmaxes)
+def test_minmax_eq(mm):
+    assert mm == mm
```
jacg: Eh? `assert mm == mm`? You are testing that any minmax equals itself. OK, but that's a far cry from actually testing the correct behaviour of `==`!

jacg:
How urgent is this? I'd like to go over it calmly, as I see many little things that could be improved.

invisible_cities/types/ic_types_test.py
gonzaponte: We don't. I didn't realize it was purely an alias.

invisible_cities/types/ic_types_test.py
gonzaponte: Yes, that's quite embarrassing... Actual test coming up.

gonzaponte:
> How urgent is this? I'd like to go over it calmly, as I see many little things that could be improved.

Take your time. I am still not sure we want to include this in our next tag.

invisible_cities/types/ic_types_test.py
jacg: Well, strictly speaking, it's *not* an alias, because `make_xy` is a function, while `xy` is a type. Consequently, even though

    make_xy(a,b) == xy(a,b)

and

    type(make_xy(a,b)) == type(xy(a,b))

unfortunately

    type(make_xy(1,2)) != make_xy

whereas, quite sensibly

    type(xy(1,2)) == xy

gonzaponte: I am not very good with the nomenclature, but you get the point. The correct word would be... a wrapper?

jacg: My point isn't about nomenclature. Yes, it's a wrapper, and it's (almost) a synonym, and (almost) an alias. But that's *not* the point.

The point is that it's utterly useless. It only wastes a couple of lines of code, and breaks some functionality.

I'm guessing that you were motivated to write it because you wanted to pass a callable to hypothesis' `builds`. The `xy` types themselves (like all types) are already perfectly good callables themselves (otherwise you wouldn't be able to call them inside your `make_xy` functions!). Like everything else in Python, they are first-class objects (just data), so you can pass them or store them or do whatever you want with them, without the need to wrap them.

In general, when you find yourself writing something of the form 

    def wrapper(*args, **kwds):
        return implementation(*args, **kwds)

or

    lambda *args, **kwds : implementation(*args, **kwds)

then **don't bother** and just use `implementation` itself, directly.

In both the above samples you are creating a function which takes some inputs, passes them on (without looking at them or modifying them at all) to `implementation` and then returns (without looking at or modifying it) whatever `implementation` returns. It matters not one bit whether `implementation` is a type or not.

That's a pointless waste of lines of code, CPU cycles, key presses, developer brain cycles and reader brain cycles.

gonzaponte: I agree 100%, but I didn't write that. I simply didn't realize that function was *just* a wrapper and used it as it was. I will remove it.

jacg: You mean you didn't write the functions `make_xy` and `make_xy_c`?

gonzaponte: Exactly. I did rename the latter, though.

jacg: Ah, OK, at first glance it looked like they appeared in your last commit, but looking more closely I see that they were there already before.

Looks like @jjgomezcadenas made them in mid-June.

@jjgomezcadenas Please do not write pointless wrappers around types.

jjgomezcadenas:
>@jjgomezcadenas Please do not write pointless wrappers around types.

Following Joaquim Sabina, I deny everything:
https://www.youtube.com/watch?v=ToSbLQKqkPA

gonzaponte:
I have renamed the PR. Suggestions are welcomed.

jjgomezcadenas:
what is the status of this, @gonzaponte ?

gonzaponte:
I will redo it after TBR, I have to remember the changes I've made, but it will be easier with the new interface. Not urgent, anyway.

PR #366, opened 2017-10-08
Refactor pmaps

### LINKED ISSUE #360: Weird behavior in S2Si
I have found a weird behavior in this method of the S2Si class:
```
cpdef sipm_total_energy(self, int peak_number, int sipm_number):
        """For peak and and sipm_number return Q, where Q is the SiPM total energy."""
        cdef double et
        if self.number_of_sipms_in_peak(peak_number) == 0:
            return 0
        try:
            et = np.sum(self.s2sid[peak_number][sipm_number])
            return et
        except KeyError:
            raise SipmNotFound
```

This means that
- If the peak does not contain any SiPM, the function will say it is 0
- If the peak does contain at least 1 SiPM, but not the one we are looking for, the function will raise an exception.

I think we should raise `SipmNotFound` in both cases. Do you agree @jjgomezcadenas? I am preparing a PR with some small changes in that module.

gonzaponte:
This is mainly a cleanup and refactoring of the `pmaps` module, but it also contains a couple of changes to the external user:
- So far, functions searching for some SiPM raised `SipmEmptyList` when the `Peak` contained no SiPMs, but `SipmNotFound` when there are some SiPMs, but not the one we are looking for. It seems the `SipmEmptyList` exception doesn't belong there because the function doesn't care about how many SiPMs are there, but whether the one we asked for is there or not.
- Some methods that didn't handle exceptions now do, because the handling of exceptions has been contained in dedicated methods.

Closes #360.

### DIFF
--- a/invisible_cities/evm/pmaps.pyx
+++ b/invisible_cities/evm/pmaps.pyx
@@ -48,17 +48,17 @@ cdef class Peak:
         def __get__(self):  return not np.any(np.isnan(self.t) | np.isnan(self.E))
 
     def total_energy_above_threshold(self, thr):
-        sat = self.E > thr
-        return np.sum(self.E[sat])
+        over_thr = self.E > thr
+        return np.sum(self.E[over_thr])
 
     def width_above_threshold(self, thr):
-        sat = self.E > thr
-        t   = self.t[sat]
-        return t[-1] - t[0] if sat.any() else 0
+        over_thr = self.E > thr
+        t        = self.t[over_thr]
+        return t[-1] - t[0] if over_thr.any() else 0
 
     def height_above_threshold(self, thr):
-        sat = self.E > thr
-        return np.max(self.E[sat]) if sat.any() else 0
+        over_thr = self.E > thr
+        return np.max(self.E[over_thr]) if over_thr.any() else 0
 
     def __str__(self):
         if self.width < units.mus:


jjgomezcadenas:
> Avoid reinvention of the wheel

Agreed. Better to use argmax here. 

>Simplify good_waveform property

so you change: 

    return (False
            if np.any(np.isnan(self.t))  or
               np.any(np.isnan(self.E))
            else True)

by 

    return not np.any(np.isnan(self.t) | np.isnan(self.E))

Are you sure this is a "simplification"? The code was concise enough already and I personally need more time to parse your solution.

>Replace common code with a function

Why on Earth you prefer

    eth = self.E[self.signal_above_threshold(thr)]

to

     eth = self.E[self.E > thr]

when 

    def signal_above_threshold(self, thr):
         return self.E > thr

You are achieving absolutely nothing here. The condition *E > thr*  is **clearer** than the 
condition *signal_above_threshold(self, thr)* (in one case we see directly the variable we are cutting on, in the other we need to go and find out what the function does), but in any case your function is, at best, syntactic sugar, you just change a name by other. There is no simplification in the code for changing one line by another (longer and more obscure) line.

jjgomezcadenas:
> Remove useless if

good catch

jjgomezcadenas:
>Use signal_above_threshold function to simplify other methods

I like the simplification but it works the same (in my opinion better) if you write:
stat = self.E > thr

jjgomezcadenas:
>Make the str representation of Peak more readable

Nice!

jjgomezcadenas:
>Remove commented out lines

That's an honest man job

jjgomezcadenas:
>Gather all try/except structures in class methods

This is nice. Indeed it simplifies the code and makes it more readable and structured. I hereby raise your level of Karma by 5 jaceks.

jjgomezcadenas:
>Make the output of S12 more readable

I raise my cup to a perfectionist!

jjgomezcadenas:
> Cosmetics

The mark of the master!

jjgomezcadenas:
> Combine str method of S1 and S2

more honest man work

jjgomezcadenas:
> Remove misleading checks

Yes, building on the previous simplification of exceptions. Bravo.

jjgomezcadenas:
>Remove useless if

Useless indeed. Good!

jacg:
> so you change:

    return (False
            if np.any(np.isnan(self.t))  or
               np.any(np.isnan(self.E))
            else True)

> by

    return not np.any(np.isnan(self.t) | np.isnan(self.E))

FWIW, I prefer the latter. Sticking a bool-valued expression in an `if` to arrive at a pain `False` or a plain `True` is an anti-pattern. If the expression evaluates to a boolean, just use its value directly, or stick a `not` in front of it if it's the wrong way around.

If you prefer the `or` to the `|` then 

1. That's the way it's done with numpy arrays: if you're gonna use numpy, you'd better get used to it
2. You could `return not (np.any(np.isnan(self.t)) or np.any(np.isnan(self.E)))`

jjgomezcadenas (approved):
Since this reviewer was terrified by the new policy of setting the reviewer's head on the stake, he has tried hard, and commented the commits one by one. Overall, this PR improves the code readability and in the case of the grouping of exception I think it also reduces notably technical debt. This reviewer has found a couple of minor points. One of them has already been solved by the master in favor of the defendant. The second point is minor. This reviewer feels that he can keep his head on his shoulder approving this PR.

gonzaponte:
> Why on Earth you prefer
>
> eth = self.E[self.signal_above_threshold(thr)]
> to
>
> eth = self.E[self.E > thr]

I actually don't have a strong opinion here. I did this with the commit "Use signal_above_threshold function to simplify other methods" in mind. I felt like gathering the same selection in a method could be more robust, but I also agree with
> The condition E > thr is clearer than the condition signal_above_threshold(self, thr)

I do not oppose reverting these changes, but I wanted to give it a try.

jacg:
@gonzaponte @jjgomezcadenas I'm planning to merge this sometime this afternoon, in some spare scrap of time. Please come to an agreement on the `signal_above_threshold` point, so that I can get on with the mechanics of merging.

I prefer `eth = self.E[self.E > thr]`: it's so obvious and concise that I think hiding it behind an abstraction doesn't really help.

gonzaponte:
I think we all agree that it is not worth the abstraction. I will push a commit reverting it and rebase.

jacg:
@gonzaponte Are you familiar with `git revert` and Magit`s `V`?

gonzaponte:
>Are you familiar with `git revert` and Magit's `V`?

A bit. I can try to do it that way, if you prefer.

jacg:
Don't worry, I'm about to merge in the next few seconds. It's just that you could have saved yourself a bit of effort, and made it a bit clearer in the log all in one go. Try to remember it for the next time it comes up.

gonzaponte:
Ok. I think I will need some practice before, though.

jacg:
You can practice all you want in a local branch.

In the meantime, I think I've repeated the same mistake I made a few days ago when merging ... unless something has changed in the way GitHub keeps track of things.

Oh, and it looks like you slipped in a commit just after I took the branch. Hmm, OK, I'll do a quick rebase of master, if nobody's looking.

jacg:
OK, I think it was you extra commit that threw me off track. All fixed now, I think.

PR #369, opened 2017-10-09
Temporary workaround for #357: use older hypothesis

jacg:
As discussed in #357, taking the hypothesis version back to 3.28 should speed up the Travis builds significantly.

This is not a long-term solution, but it should let us work much more efficiently in the short term.

#357 needs to remain open, until a detailed analysis of the situation is performed and a good long-term solution is found.

Issue #357, opened 2017-10-05
Travis taking too long


gonzaponte:
Lately, Travis is taking too long to make the build. The last one is over 20 minutes and seems to be stuck in the tests rather than in the downloads or in the setup. I don't know if the parallel testing is the issue here, but I don't see any other reason. Also, the output of Travis contains stuff I haven't seen before, but I did never pay too much attention to it anyway.

jacg:
This seems to be caused by a new version of hypothesis: lots of tests are now hitting the default timeout limit of 60s. We don't see this in our local repos, because

1. Most of us are running older versions of hypothesis, locally
2. Travis runs the `hard` profile, which tries 1000 cases, while our `dev` profile, which we use by default locally, tries only 10 cases.

I have run the tests on 8 processors with both the `hard` and `dev` profiles on my machine in different versions of hypothesis:

[edit: Added `hard_nocov` column this uses the `hand` profile with `use_coverage = False`. `use_coverage` was not available in hypothesis 3.28]

| Hypothesis version | `dev` run time | `hard` run time | `hard_nocov` |
|--------------------------|----------------:|-----------------:|-----:|
| 3.28.0             | 31s | 100s |
| 3.29.0             | 34s | 250s | 84s |
| 3.30.0             | 33s | 250s | 76s |
| 3.31.0             | 39s | 600s | 83s |
| 3.31.2 (ours at time of writing)   | 38s | 550s | 85s |
| 3.31.4 (latest at time of writing) | 37s | 500s | 71s |
| 3.32.0 (added in an edit)           | 33s | 170s | 76s |

So the sort-term solution would be to specify `hypothesis=3.28` in `manage.sh`. In the long-term we need do understand why these times went up so drastically. [edit: hypothesis `3.32` has been released, and much of the performance problem disappears!]

Incidentally, we are installing hypothesis with `pip` but it seems to be available in `conda` now.

jjgomezcadenas:
Is this connected with the (more or less solved) issue of upgrading hypothesis?

jacg:
I think that this had to stay open, as long as the solution was the **temporary** workaround of **holding back** the version of hypothesis.

Now that we are solving it by using the **most recent** hypothesis, we can probably assume that the problem was caused by hypothesis rather than by us.

If that is the case, the solution can be considered **permanent**, and the issue can be closed.

PR #370, opened 2017-10-11
Reduce maximum length of test array

gonzaponte:
This avoids occasional health check failures.

### DIFF
--- a/invisible_cities/core/core_functions_test.py
+++ b/invisible_cities/core/core_functions_test.py
@@ -60,7 +60,7 @@ def test_in_range_positives():
     assert np.count_nonzero(core.in_range(data, 0, 10)) == 500
 
 
-@given(random_length_float_arrays(max_length = 1000))
+@given(random_length_float_arrays(max_length = 100))
 def test_in_range_right_shape(data):
     assert core.in_range(data, -1., 1.).shape == data.shape
 


jacg:
I find it hard to believe that this has any effect, given that `random_length_float_arrays` already has a default value of `max_length = 100`!

jacg:
Ignore me. Brain elsewhere. (That's why I should ignore IC stuff while doing other things!)

gonzaponte:
> I find it hard to believe that this has any effect, given that random_length_float_arrays already has a default value of max_length = 100!

It was set to `max_length = 1000` before.

Anyway, I couldn't reproduce the error on my computer after running the test 15 times, but I did see this error on Travis a couple of times already.

jacg:
Could you reproduce the error locally *before* the change?

gonzaponte:
> Could you reproduce the error locally *before* the change?

That was what I meant. I couldn't reproduce the error *before* the change.

jacg:
Hmm, so it's just a random stab in the dark. We could merge it and just hope that it works. If Travis builds are failing because of it, it might be worth a try. It *might* be a quick solution to an existing problem.

Another option is to get a Travis docker image and try to investigate what is going on in that, and try to make **sure** that we merge something that actually solves the problem.

gonzaponte:
I agree, this is just a patch. Probably this particular test is not worth the effort (it only checks the output shape is correct, no need for gigantic arrays to check that), but we did see when we first introduced these tests that hypothesis couldn't generate numpy arrays of length greater than 1024. For some reason, the generation of these arrays is very slow, which raises the HealthCheck failure.

jacg:
Recall that we have recently started holding back the version of hypothesis as a *temporary* fix to another, quite possibly related, problem. (#357, #369)

We can't continue doing random things. Sometime we're going to have to invest time and effort into getting a proper understanding of what is going on and what the proper solution is.

gonzaponte:
I know, but we definitely don't have the time now. I tried to dig a bit in hypothesis' changelog. They do mention a `deadline` setting, which turns slow tests into errors, but it is off by default.

jacg:
> I couldn't reproduce the error *before* the change.

Just to be sure: did you do this using the `hard` profile?

gonzaponte:
> Just to be sure: did you do this using the hard profile?

Yes, I did.

gonzaponte:
Update. I've run the test again 20 times with `max_length=1000` and `--hypothesis-profile hard`. I've got this ***warning*** twice.
```
invisible_cities/core/core_functions_test.py::test_in_range_right_shape
  /Applications/anaconda3/envs/IC3.6/lib/python3.6/site-packages/hypothesis/core.py:805:
HypothesisDeprecationWarning: Your tests are hitting the settings timeout (60.00s). This
functionality will go away in a future release and you should not rely on it. Instead,
try setting max_examples to be some value lower than 975 (the number of examples
your test successfully ran here). Or, if you would prefer your tests to run to completion,
regardless of how long they take, you can set the timeout value to hypothesis.unlimited.
    state.run()
```

I am running the same hypothesis version as Travis (3.28.0).

jacg:
With hypothesis 3.28.0 and the `hard` profile, this test typically takes roughly 40s on my machine.

Switching to hypothesis 3.32 (released on Monday), it takes about 2.6s.

Hypothesis 3.32 recovers much (though not all) of the performance degradation we saw since 3.28. See the table of timings in #357, which I have updated to include 3.32 as well as timings with `use_coverage=False`.

Please let me know if you observe timing differences similar to mine on this test.

If you do, I propose that we bump hypothesis to 3.32.

jacg:
@gonzaponte Although you haven't confirmed it specifically, I'm assuming that #372 has made this problem go away.

We have two options:

1. Close this PR
2. Reduce the size from 1000 to 100 anyway, and save around 2s on the test. This is reasonable if, as you say earlier, large array sizes are unlikely to be important in this test: there's no point in having a test take so much time if it is not being spent usefully.

gonzaponte:
> Although you haven't confirmed it specifically, I'm assuming that #372 has made this problem go away.

Sorry, I forgot about this. I do see the same performance difference (66 s -> 8 s). It has passed the test 50 times with `max_length=1000` and the hard profile without any failure or warning. 

>We have two options:
>
>1. Close this PR
>2. Reduce the size from 1000 to 100 anyway, 

I think we agree that the latter makes sense.

jacg (approved):
We seem confident that limiting the array length will speed up the test, without degrading the quality of the test.

Issue #363, opened 2017-10-07
Assign merge commits to reviewer who approved PR


jacg:
We currently have a policy to assign merge commits to the main author of the commits in a PR. I propose that we switch to assigning the merge commit to the reviewer who approved the PR, with the approval comment included in the merge commit comment, along with the main comment introducing the PR.

[edit: We should also include the PR number in the commit message. Better still, in the title of the commit message. There may be plenty of background information on the code stored in the conversation of the PR, but, currently, finding the corresponding PR is difficult.]

In the case of single-commit PRs we do not have a separate merge commit, so there only the author would appear.

gonzaponte:
I don't disagree, but can you elaborate on the logic behind it?

jacg:
Review and approval of a PR are a significant part of the PR's history. Currently we note authorship in the commits themselves, but we have no record of the identity of the reviewer, except trawling through the PRs on GitHub, but there is no link between the commits that make up the PR in a local repo and the GitHub PR in which they were accepted.

The author is probably the primary source of wisdom about a piece of code. The reviewer is likely to be the second most informed person on that particular topic. It is useful information to have.

I can get a fair idea of the author of a piece of code with `magit-blame` in seconds. Identifying the reviewer currently takes a too much effort.

I would propose that the PR number should be included in the merge commit message too.

gonzaponte:
Sounds good to me.

> I would propose that the PR number should be included in the merge commit message too.

Yes, I find that particularly useful.

jjgomezcadenas:
I smell blood soon to be spilled here...

jacg:
>> the PR number should be included in the merge commit message

> Yes, I find that particularly useful.

Yup, I think it's the most important part of the idea. Adding the URL of the PR in the message body could be good.

> I smell blood soon to be spilled here...

Oh yessss, the hounds shall be unleashed.

jacg:
Do we agree to the following merge policy?

Mergers should use the PR number and title as the title of the merge commit. The merge commit comment should contain the most useful and informative combination of the PR's opening comment and the reviewer's acceptance comment.

If we stick to this policy then I have no strong opinion whether the merge commit should be assigned to the author or the reviewer who approved the PR.

gonzaponte:
I do agree.

> Mergers should use the PR number and title as the title of the merge commit.

I want to emphasize that we (me particularly) should use more descriptive titles in our pull requests. Reviewers must demand a better title if it describes the PR poorly.

[...]

PR #372, opened 2017-10-11
Bump hypothesis to 3.32

jacg:
The serious performance degradation that invaded Travis with the advent of hypothesis 3.29, has been mostly fixed in hypothesis 3.32. On the other hand, 3.32 makes `test_in_range_right_shape` run an order of magnitude faster.

This PR contains a commit, which advances our hypothesis version no 3.32.0

Additionally, two new hypothesis profiles are added: versions of your existing `dev` and `hard` which switch off `use_coverage`. I think that we want to keep `use_coverage` on, but these profiles will make it easier to investigate performance degradations in the future.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -4,10 +4,12 @@ from hypothesis import settings
 from hypothesis import Verbosity
 
 # In addition to the 'default' profile, we provide
-settings.register_profile("hard"     , settings(max_examples = 1000))
-settings.register_profile("dev"      , settings(max_examples =   10))
-settings.register_profile("noshrink" , settings(max_examples =   10,
-                                                max_shrinks  =    0))
-settings.register_profile("debug"    , settings(max_examples =   10,
+settings.register_profile("hard"      , settings(max_examples = 1000))
+settings.register_profile("dev"       , settings(max_examples =   10))
+settings.register_profile("hard_nocov", settings(max_examples = 1000, use_coverage=False))
+settings.register_profile("dev_nocov" , settings(max_examples =   10, use_coverage=False))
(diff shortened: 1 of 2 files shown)


jjgomezcadenas (approved):
As the previous discussion reveals, this PR appears to solve a part of our problems.

jjgomezcadenas:
I have just approved this PR. According to new policy I must merge it, but I am afraid to mess it up. Since approving this PR contains little intelligence from the side of the reviewer would appreciate if somebody else merges and saves me the trouble (I swear a I will learn the mechanics as soon as I am out of the woods with proposal-writing)

jacg:
No, according to our new (not yet confirmed) policy, whoever merges it, must put your name on the merge commit. Don't worry about it. I'll take care of it.

jjgomezcadenas:
Actually, with great regret  I *do vote in favour* of the proposed policy. I know I will be sorry for this, and severely wiped for my frequent mistakes, but I think is a great idea that must be implemented asap.

jacg:
I think that the important part of what is being discussed in #363 - that the PR number be included in the merge commit title - has already been agreed.

This on its own will be enough to trace all your mistakes back to you :-)

The second most important part is that the approval message should be included in the body of the merge commit comment. That's nice, that's helpful, but once the PR number is in there, all that information is relatively easily retrieved.

The idea with which I kicked off the whole discussion - that the approver be assigned as the author of the merge commit - is probably the least interesting part of the policy. It's helpful to have that information in front of your eyes when you are looking at the log, but that too is information that can be retrieved from the PR number.

The bit I forgot to do in the merge commit for this PR is to inculde a link to the PR.

Issue #374, opened 2017-10-12
Missing MC true information in hits created by Penthesilea


paolafer:
I've realized that MC true info is not written in the Penthesilea step. I think we should have MC true info in MC output files, in general.

gonzaponte:
I am working on it, but we need #376 first.

PR #376, opened 2017-10-16
Replace KrMC_pmaps file

gonzaponte:
This PR replaces the previous `KrMC_pmaps` file with a new one. We need it to contain MC tracks (#374) and we've (I've) lost track of the original dataset it has been taken from. Besides the replacement:
- a file containing the output of `dorothea` (with a particular configuration) has been added
- a file containing the output of `penthesilea` (with a particular configuration) has been added
- the tests have been adapted to this new file and improved a bit in some cases
- the hard-coded dependencies of some tests have been fixed
- the fixture that manages the relevant file info has been extended to contain more data
- some "duplicated" tests have been combined

On top of that, I have cleaned up a bit the modules I've touched.


jjgomezcadenas:
The checks are failing. Why?

gonzaponte:
I don't know, they passed locally. I will take a look tomorrow, I can't
right now.

abotas:
`City.go` fails when no events are actually run since it tries to divide by `n_events` when `n_events` is 0.
```
n_events = self.cnt.n_events_tot
print("run {} evts in {} s, time/event = {}".format(n_events,
                                                            dt,
                                                            dt / n_events))
```

gonzaponte:
I've found the error. It came with the deletion of the previous file, which
I did the last and forgot to run the tests. I think I can push the
necessary changes in a few hours.

gonzaponte:
This is ready for review.

On a completely different subject, I have reorganized my commits so they look more natural and to facilitate the review but, for some reason, in the PR they appear in a completely different order (and I am not sure it is chronologically either). Do you know how to avoid this, @jacg?

jacg:
@gonzaponte Have you been re-ordering commits with interactive rebase?

gonzaponte:
> Have you been re-ordering commits with interactive rebase?

Yes.

jacg:
A commit has two different timestamps on it: the author date and the commit date. When you move commits around with rebase, the author date does not change, but the commit date does. This explains why the commits are not in chronological order: you are looking at the dates that the commits were authored (the first time you committed them), but you shuffled them around with the rebase.

Nothing wrong with this. Keep doing it.

As to the order in which they appear in the PR ... I'm not sure ... they sort of seem to follow the commit time order, some of the time.

Sorry, no time to look further into it at the moment.

gonzaponte:
FWIW, an annoying workaround to this is to reword every commit when doing an interactive rebase, as I've just done. It is not fancy nor handy, but it serves my purpose.

jacg:
There's an Emacs package called `magit-rockstar` or something like it (by the same author as Magit IIRC), which automatically does this sort of thing for you. I think it was designed for precisely for your use-case, actually.

gonzaponte:
I forgot to actually add the MCtracks to the new file... I have also renamed the new files (they said `_10evt` when they only contain 5 events) and rearranged a bit some changes that were included in the wrong commit.

gonzaponte:
> There's an Emacs package called magit-rockstar or something like it

Beautiful


> (by the same author as Magit IIRC)

Yup, same guy.

Thank you for the info 🙂

gonzaponte:
Sorry, the MC tracks are not correctly written. My test to check the file is ok is buggy. Do not merge yet.

gonzaponte:
Should be ok now.

We can include `ipmt` pmaps as well, so we can base all the tests in the same pmaps file. Opinions? @abotas, in particular.

invisible_cities/reco/pmaps_functions_test.py
```diff
@@ -70,73 +58,58 @@ def timebin_size_must_be_equal_to_stride_times_25_ns_shows_bug_in_old_data(KrMC_
...
+        assert sorted(s1_dict.keys()) == S1_evts[:i]
+
+    for i, max_events in enumerate(S2_evts):
+        s2_dict = df_to_s2_dict(s2t, max_events)
+        assert sorted(s2_dict.keys()) == S2_evts[:i]
+
+    for i, max_events in enumerate(Si_evts):
+        si_dict = df_to_s2si_dict(s2t, sit, max_events)
+        assert sorted(si_dict.keys()) == Si_evts[:i]
+
```
abotas: Maybe this could be parameterized?

invisible_cities/io/dst_io_test.py
```diff
@@ -1,37 +1,26 @@
...
-from . dst_io                 import load_dsts
+from ..core.testing_utils import assert_dataframes_close
+from .      dst_io        import load_dsts
 
 
-def test_load_dst(Kr_dst_data):
-    (filename, group, node), df = Kr_dst_data
-    dst = load_dst(filename, group, node)
+def test_load_dst(KrMC_kdst):
+    # The function load_dst is actually used in the fixture
```
abotas: Maybe `load_dst` should be loaded (again) here even if it slows down the tests a couple hundred microseconds just in case someone changes the fixture?

invisible_cities/reco/pmaps_functions_test.py
gonzaponte: Not obvious to me how to do it properly, in a way it is obvious what I am doing. Because it uses particular items in the fixture, it is not something I can set externally.

I can return namedtuples in the fixture, though. That would make the test parametrizable in a readable way.

invisible_cities/io/dst_io_test.py
gonzaponte: Indeed

invisible_cities/reco/pmaps_functions_test.py
abotas: Ok it just seemed like repeated code. Not a big deal obviously but ideally i think we would make it a little more concise

gonzaponte: Yes, I didn't like it either, but at the time I decided it was better to repeat the code than to make it obscure. If you like the idea of namedtuples I can implement it tomorrow.

abotas:
About the `ipmt` pmaps I don't have any special opinion... it might be marginally preferable to only have one kr pmaps file that has everything, yes. But to do that we would need the corresponding RWF and to run irene over them with the ipmts. If we have all that available and it is easy then sure if not, from my point of view it's not a big deal

gonzaponte:
> If we have all that available and it is easy then sure

I think we do.

abotas:
Ok sounds good then

invisible_cities/reco/pmaps_functions_test.py
abotas: That sounds good to me

abotas:
Once these changes are made I am satisfied. Looks good!

gonzaponte:
I am not going to include the `ipmt` pmaps in this PR so we don't delay the approval any further. Rearding the parametrized test, it is not worth it. Because `df_to_s2si_dict`'s signature is different from the others, it would have to go in a different test, and parametrizing it only for `df_to_s1_dict` and `df_to_s2_dict` doesn't make much of a difference. In this case, I'd repeat the code for the sake of readability. I will push a few commits with the `namedtuple` stuff, though.

abotas:
That's fine!

gonzaponte:
Ok, I have
- Introduced namedtuples in `conftest.py` and used them to make tests more readable
- Used the function  `load_dst` in the test that is supposed to test it.
- Made some cosmetics here and there
- Renamed the original files back to `_10evt` because I mistakenly thought they contained only 5 events.
- Reformulated a test that I found unreadable.

I have gone through everything again and I think it is ok. Please review.

abotas (approved):
Looks good! And you fixed an ugly old test of mine I think. Wonderful.

PR #379, opened 2017-10-19
fix minor penthesilea and dorothea mistakes

abotas:
This tiny PR fixes some (minor) slowness in HitCity and the xy-algorithms and fixes a typo in a penthesilea test


invisible_cities/cities/base_cities.py
```diff
@@ -756,8 +756,8 @@ def compute_xy_position(self, s2si, peak_no):
         """
         IDs, Qs = cpmp.integrate_sipm_charges_in_peak(s2si, peak_no)
         xs, ys   = self.xs[IDs], self.ys[IDs]
-        #return self.reco_algorithm(np.stack((xs, ys)).T, Qs)
-        return corona(np.stack((xs, ys)).T, Qs,
+        #return self.reco_algorithm(np.stack((xs, ys), axis=1), Qs)
```
jacg: Is there any good reason for keeping this commented-out line?

jacg:
> fixes some (minor) slowness

Can you quantify this?

invisible_cities/cities/penthesilea_test.py
```diff
@@ -9,7 +9,7 @@
 from .. io                   import dst_io as dio
 
 
-def test_penthesilea_run_on_Kr_MC_ipmt_pmaps_5evt(ICDIR, config_tmpdir):
+def test_penthesilea_run_on_Kr_MC_ipmt_pmaps_6evt(ICDIR, config_tmpdir):
```
gonzaponte: This is already being changed in another PR, where the "ipmt" name is removed, as the test has nothing to do with "ipmts"

invisible_cities/cities/base_cities.py
gonzaponte: There is a similar one in the `__init__` method than can be removed as well.

invisible_cities/cities/penthesilea_test.py
abotas: The "ipmt" is included because the name of the test is identifying the name of the input file in `test_data` that the test is testing.

gonzaponte:
- Please remove the commented-out lines mentioned in the review.
- Have you checked that `np.sum` is indeed faster than python's `sum`? Sometimes it is not.

abotas:
On my machine for sums of very short qs (len ~5) python's `sum` is about a microsecond faster / execution, while for large (len ~1k) qs,`np.sum` is about a little less than 10x faster. I changed it mostly thinking that in the future (for next100 simulations or something) this would become a problem. 

`np.stack`ing on the first axis instead of the 0th and then doing the transpose is always faster (though there is a lot of variance in how much faster) and it also just seemed more correct.

invisible_cities/cities/base_cities.py
abotas: Ok I have removed the commented code

abotas:
If the reviewers are satisfied i think this is ready..?

gonzaponte:
I thought I left a comment here, but somehow that has not happened. I wanted to emphasize that the number of SiPMs with signal should remain more or less constant regardless of the detector because the light emission is very focused (is this word correct?). Have you timed both functions for lengths in the range 10-50? This is the range we are getting from MC and data. I don't think there is a big difference, though.

abotas:
On my computer at length 10 python is a tiny bit faster, at length 50 numpy is  a tiny bit faster. My reasoning was they are about the same, and numpy seems safer.

gonzaponte:
Ok. Approved then.

abotas:
And the speeds are all on the order of 10 microseconds

gonzaponte (approved):
This PR introduces tiny performance improvements to `PCity`-derived cities and some cosmetics.

abotas:
Rebased

gonzaponte:
Ups, forgot to merge. I'll do it in a moment.

gonzaponte:
You pushed the conflicts in the first commit and fixed it in the last one. Wanna fix it?

abotas:
I didn't see that, thank you -- I squashed the commits and I think that has solved it

gonzaponte:
I hope I have performed the merge properly, I hold my digits in high esteem...

PR #380, opened 2017-10-21
Write mc tracks in PCity

### LINKED ISSUE #374: Missing MC true information in hits created by Penthesilea

gonzaponte:
This PR introduces the ability to propagate the MC tracks in the input file to the output file for cities deriving from `PCity`. This affects mainly to `Penthesilea`, as we haven't had such need in `Dorothea`.

Moreover, it introduces minor fixes with regard to the organization of parameters, fixes a small detail in a previous test and applies some cosmetics to the code. A couple of tests have also been introduced to ensure this new functionality works properly.

Closes #374


paolafer:
I've tested the code and it works: MC true track info is written for all the events, even when more than one file is processed with Penthesilea. From the point of view of functionality, this PR can be approved.

jmbenlloch (approved):
This is an important functionality, I have read the new code and everything seems good. New tests has been added for the new functionality and it is working correctly, so I'll approve and merge this.

PR #381, opened 2017-10-25
Add Zrms column to Dorothea's output

gonzaponte:
This PR includes a new column in the output DST produced by `Dorothea`.  This column is the standard deviation of the peak in time, providing another useful variable for the analysis. It is correlated with the `S2width` column, but not the same. Change summary:

- The `kdst` produced by `Dorothea` now contains another column (`Zrms`).
- The `Peak` class now contains a new attribute (`rms`) and a new method (`rms_above_threshold`).
- Cosmetics.


invisible_cities/evm/pmaps_test.py
```diff
@@ -43,6 +44,13 @@ def s2si_input(draw, min_size=1, max_size=10):
     return size, t, E, qs1, qs2
 
 
+def rms(x, y):
+    if len(x) == len(y) < 2: return 0
+    mean = np.average(x, weights=y)
+    var  = np.average((x-mean)**2, weights=y)
+    return var**0.5
+
```
abotas: Can we not use `np.var`? (also in `xy-algorithms`?)

Also, since we do this in a couple places, maybe we can put an rms function in `invisible_cities.core`

invisible_cities/evm/pmaps.pyx
```diff
@@ -60,6 +61,17 @@ cdef class Peak:
...
+        over_thr   = self.E > thr
+        n_over_thr = np.count_nonzero(over_thr)
+        if n_over_thr < 2: return 0
+
+        t    = self.t[over_thr]
+        E    = self.E[over_thr]
+        mean = np.average( t         , weights=E)
+        var  = np.average((t-mean)**2, weights=E)
+        return var**0.5
+
```
abotas: It seems like we are repeating some code -- i want to reiterate that maybe we can put some mean+var function in core (if we can't just use np.var)

abotas (changes requested):

invisible_cities/evm/pmaps_test.py
gonzaponte: `np.var` does not accept weights, that is why we use this `np.average` formulation.

invisible_cities/evm/pmaps.pyx
gonzaponte: Sounds good.

jjgomezcadenas:
@abotas @gonzaponte, this is stale. Please do not leave PRs to fester for...

gonzaponte:
Yes, this was reviewed just after the cm days. I will make the changes today or tomorrow.

abotas (approved):

abotas:
The changes have been addressed. PR is approved

PR #382, opened 2017-10-26
Track length

paolafer:
I've realized that the length function in paolina_functions returned the number of voxels instead of the actual length of the track. I have modified that function and added a test to make sure the returned length is the correct one. I have first used the test on the old code, which failed, and then on the new one, which passes.


invisible_cities/reco/paolina_functions.py
```diff
@@ -100,8 +100,19 @@ def find_extrema(distance : Dict[Voxel, Dict[Voxel, float]]) -> Tuple[Voxel, Vox
...
 
-def length(track_graph):
-    return len(shortest_paths(track_graph))
+def length(distance : Dict[Voxel, Dict[Voxel, float]]) -> float:
+    """Calculate the length of a track."""
+    if not distance:
+        raise NoVoxels
+    if len(distance) == 1:
+        return 0.
+    max_distance = 0.
```
jacg: This is essentially copy-pasted from `find_extrema`.

We do not do copy-n-paste code reuse in IC: it is an evil practice which leads to confusion and to wasting lots of time and effort.

If you need to reuse some code, wrap it into a convenient abstraction, such as a function, and use that in multiple places.

In this case you should extract the loop from `find_extrema` into a function, and call that function in both `find_extrema` and `length`.

invisible_cities/reco/paolina_functions_test.py
```diff
@@ -209,3 +209,27 @@ def test_voxelize_single_hit():
     hits = [BHit(1, 1, 1, 100)]
     vox_size = np.array([10,10,10], dtype=np.int16)
     assert len(voxelize_hits(hits, vox_size)) == 1
+
+def test_length():
+    voxel_spec = ((10,10,10,  1000),
+                  (10,10,11,     1),
+                  (10,10,12,     2),
+                  (10,10,13,     4),
+                  (10,10,14,     8),
```
jacg: The 'interesting' energies in `voxel_spec` give the impression that the energies somehow matter in this test. AFAICT, the energies are irrelevant, so don't make them look interesting. It would be best if you could get away with `None` or `'dummy'`. But I suspect that you will need the energies to be

+ numeric
+ greater than zero

invisible_cities/reco/paolina_functions.py
paolafer: Ok.

invisible_cities/reco/paolina_functions_test.py
```diff
@@ -209,3 +209,27 @@ def test_voxelize_single_hit():
...
+    voxels = [Voxel(x,y,z, E) for (x,y,z,E) in voxel_spec]
+    tracks  = make_track_graphs(voxels, np.array([1,1,1]), contiguity=1.85)
+
+    assert len(tracks) == 1
+    distances = shortest_paths(tracks[0])
+    track_length = length(distances)
+
+    expected_length = 8 + np.sqrt(2)
+
+    assert_almost_equal(track_length, expected_length)
```
jacg: As a general point for @everybody-in-IC, I recommend using [`pytest.approx`](https://docs.pytest.org/en/latest/builtin.html#comparing-floating-point-numbers) for doing float comparisons. It now supports numpy arrays (and other containers)!

invisible_cities/reco/paolina_functions_test.py
paolafer: Yes, they must be positive numbers. What do you suggest?

invisible_cities/reco/paolina_functions.py
paolafer: I am also a bit uncomfortable with the argument of length and find_extrema. From an external user's point of view, I would expect to look for the length or the extremes of a track, not of a dictionary of distances. I understand that calling all_pairs_dijkstra_path_length() every time I ask for a length is maybe unnecessarily time-consuming? Is this the reason why the find_extrema function has the overmentioned dictionary as an argument?

invisible_cities/reco/paolina_functions_test.py
jacg: 1?

paolafer: Ah, ok, I was thinking more about random numbers, but maybe it's an overkill...

invisible_cities/reco/paolina_functions.py
jacg: The primary reason is probably decomposition and testability. Optimization might well be a secondary reason.

In `paolina_functions.py` at least, `find_extrema` is *always* called just after `shortest_paths`.

I think it would be reasonable to add a function which calculates the extrema directly from a track: it would simply call `shortest_paths` and `find_extrema` in succession.

You might want to rename `find_extrema` to something more clunky and use the original name for the new one which accepts tracks. And then it would be reasonable to do something similar for `length`.

invisible_cities/reco/paolina_functions_test.py
jacg: The energy values are completely irrelevant for the test, beyond being nonzero. So I would follow the principle of making them as dummy as possible. But nonzero random numbers would be fine too.

invisible_cities/reco/paolina_functions.py
```diff
@@ -97,35 +97,31 @@ def shortest_paths(track_graph : Graph) -> Dict[Voxel, Dict[Voxel, float]]:
...
         if d > max_distance:
             first, last, max_distance = source, target, d
-    return (first, last)
+    return (first, last, max_distance)
 
+def find_extrema(track: Graph) -> Tuple[Voxel, Voxel]:
+    """Find the extrema of a track."""
+    distances = shortest_paths(track)
+    extrema_and_length = find_extrema_and_length(distances)
+    return (extrema_and_length[0], extrema_and_length[1])
```
jacg: The last two lines look like a very convoluted way of saying `return find_extrema_and_length(distances)`.

invisible_cities/reco/paolina_functions.py
```diff
@@ -97,35 +97,31 @@ def shortest_paths(track_graph : Graph) -> Dict[Voxel, Dict[Voxel, float]]:
...
-    max_distance = 0.
-    for source, target in combinations(distance, 2):
-        d = distance[source][target]
-        if d > max_distance:
-            max_distance = d
-
-    return  max_distance
+    distances = shortest_paths(track)
+    extrema_and_length = find_extrema_and_length(distances)
+    return extrema_and_length[2]
```
jacg: The *magic number* `2`is really unhelpful. I would prefer something like

    _, _, length = find_extrema_and_length(distances)
    return length

jacg:
We follow PEP8 when it comes to leaving two blank lines between top-level functions.

invisible_cities/reco/paolina_functions.py
jacg: Actually, no, what I said isn't correct. It's more like `return extrema_and_length[:-1]`.

The fact that I misinterpreted it shows that it needs to be rewritten in a clearer way.

jacg (commented):
This PR introduces two main changes

1. `length` calculates the geometric rather than the topological distance between the extrema of a track. 

2. `length` and `find_extrema` accept track graphs, rather than the dictionary of distances between.

... hmm, but before approving, I just want to add a test that ensures that we calculate the length along the track, rather than any shortcut that might be available.

jacg (approved):
OK, I'm happy with this, apart from the general brokenness of `contiguity` but fixing that does not belong to this PR.

paolafer:
> I just want to add a test that ensures that we calculate the length along the track, rather than any shortcut that might be available.

I'm a bit confused about the test_length_around_bend function. What does it test, exactly? When there's a bend, my understanding is that the shortest path *should* take the shortcut, because it is shorter than following the corner. Maybe I've completely missed the point, because it's not clear to me what the test does...

jacg:
Given the voxels marked by `X` distributed in 2D like this:

    XX
     X
    XX

do you want `length` to return the length of this path

    ++
     +
    ++

or this one

    +
    +
    +

?

The former takes the long way around, only passing through space occupied by voxels in the track; the latter takes a shortut by passing through **space where there is no voxel** belonging to the track at all.

All this assumes face-contiguity. You seem to be talking about cutting corners when edge-continuity is used to define neighbours, so your preferred distance would be along a path like this.

    +
     +
    +

I remind you that `contiguity` is currently completely broken.

paolafer:
Yes, I was talking about cutting corners. The path 

+
+
+

is incorrect.

We need to fix the contiguity definition to face/edge/corner, which is the definition that we have tested so far. I propose to leave for future developments the study of different definitions of contiguity.

jacg:
I don't understand how you want to fix the contiguity definition while at the same time leaving it until later.

I have started #383, where I will fix this damn mess once and for all.

jacg:
> I was talking about cutting corners.

I think that `test_length_cuts_corners` from #383 demonstrates that it works the way you want it to work.

paolafer:
> I don't understand how you want to fix the contiguity definition while at the same time leaving it until later.

No, no, what I meant is that I want to fix it! If in the future anyone comes up with a better definition of contiguity, then we'll go back to it.

> I think that test_length_cuts_corners from #383 demonstrates that it works the way you want it to work.

Yes, it does!

jacg:
I guess that by 'fix' you mean 'remove flexibility', rather than 'correct'. In any case, I think it's all taken care of in #383.

PR #383, opened 2017-10-27
Fix `contiguity`

jacg:
Once upon a time there was an IC contributor whose handle starts with **j** (and ends with **jgomezcadenas**) who decided to improve some existing IC code by adding extra functionality to it. The functionality in question consisted of a parameter called `contiguity` to `make_track_graphs` whose purpose was to allow some degree of control over the criterion by which two voxels are considered to be neighbours.

For some strange reason, the developer gave the parameter a default value of `1`, meaning that (by default) no voxels would ever be considered neighbours, unless floating point arithmetic error brought the centres of two adjacent voxels to *less than* the required voxel separation.

What is even stranger is that the developer did not write any tests for this functionality, thus committing one of the greater IC sins.

The deveoper probably thought that, in such a simple case, it is not worth wasting time on writing tests: the solution is obviously correct.

As it turns out, the strange choice of default value was not the only problem with the implementation. The implementation created by the original developer of the code, considered voxels that share *entire* faces to be neighbours, and the logic embedded in the code was devised for that specific case. Merely changing the length used in that logic, was not enough to implement the new feature. But, **without an automated test**, how was the poor developer supposed to know that he had failed to achieve what he wanted?

Many moons later, another IC developer (whose handle starts with **p**) kept getting confused by the behaviour of `make_track_graphs` and related functions. She kept asking the *original* developer (the one who did **not** add `contiguity`) to explain how it was supposed to work.

Unfortunately, she chose to raise these questions is private emails, rather than in GitHub issues, so the *actual* author of `contiguity` remained oblivious of the fact that his untested and buggy addition to the code was repeatedly wasting time of two IC contributors. 

The moral of the story is, **always wirte tests for new functionality that you add to IC**.

This PR kicks off by adding tests that prove that `contiguity` is broken.

Further commits that fix it will follow ...

### DIFF
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -77,7 +77,7 @@ def make_track_graphs(voxels           : Voxel,
     """
 
     def neighbours(va : Voxel, vb : Voxel, scale : float = 1.0) -> bool:
-        return ((abs(va.pos - vb.pos) / voxel_dimensions) < contiguity).all()
+        return np.linalg.norm((va.pos - vb.pos) / voxel_dimensions) < contiguity
 
     voxel_graph = nx.Graph()
     voxel_graph.add_nodes_from(voxels)
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -1,5 +1,6 @@
 from functools   import reduce
 from itertools   import combinations
+from enum        import Enum
 
 import numpy    as np
 import networkx as nx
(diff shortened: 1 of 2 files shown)


gonzaponte:
What a great tale. You should consider writing novels...

jacg:
OK, I think this is ready for review.

jjgomezcadenas:
Indeed, in particular because, if memory serves, the real story is not exactly as told. I believed that once upon the time, a developer whose name starts by j (and ends as you all know), found a hardwire number in a call to a function (the value of the hardwire number was either 1.5 or 2, I don't remember). The above mentioned subject (let's call IT J) then decided to NAME the parameter with something that sounded pretty (e.g, contiguity), and set such parameter to the value of one for reasons only known to him (possibly he though it was the step between voxels). He didn't add a test because he didn't do anything, really, only invent a name to avoid a hard-wired number (of course he also changed the hard wired number, otherwise there would be no plot to our story).

That is, at least, as I recall the tale, but it could be that I have smoked too much pot today...

jacg:
Magic numbers are evil, and I applaud initiatives which give them names. The thing that went wrong in this case, is that the name somehow became a user-settable parameter of the function. It's at this point that things broke, because the logic that used the number was not designed to work with anything outside the open interval `(1, sqrt(2))`, while the presence of the parameter implied that its purpose was to allow picking values in any of the open intervals

+ `(1, sqrt(2))`
+ `(sqrt(2), sqrt(3))`
+ `(sqrt(3), 2)`

Anyway, this PR now provides

+ tests
+ fixes
+ a more self-explanatory and self-validating interface

So, let's

+ review it, and ...
+ **all** remember to test, *test*, TEST, **test**, everything we write.

In this particular case it's interesting to note how 'I didn't really do anything' turned out to be much more. We live and learn. And we test.

jacg:
Ladies and gentlemen,

There's really not much here. Could someone please review it?

jjgomezcadenas (approved):
Brilliantly implemented, totally clarifies an obscure issues that has crated some trouble. A real lesson in good implementation practice including the use of symbols rather than strings. Shows very high levels of Karma.

jacg:
Travis is [aware](https://www.traviscistatus.com/incidents/xgpn2tq6mx7k) that public builds are failing.

PR #384, opened 2017-10-27
Let voxels know their size

jacg:
A number of situations are popping up where it would be useful to query a voxel for its size, rather than having to hund backwards in the code for the size that was requested, or having to infer the size from the spacing. The latter is not only tedious, but also nontrivial, error-prone and iiable to be done on the basis of unfounded assumptions.

There are essentially no commits in this PR yet: it looks otherwise, because it is based on #383 which has not been approved and merged yet.

The purpose of opening the PR now is to solicit input on the question of what the interface should be.

We should discourage changing a voxel's size after construction, so I propose a read-only interface. This essentially excludes returning an `np.array`. Options include

+ 3 separate properties, one for each dimension
+ 1 property, returning a tuple
+ 1 property returning an np.array (dangerous but convenient)

Any thoughts?

### DIFF
--- a/invisible_cities/evm/event_model.py
+++ b/invisible_cities/evm/event_model.py
@@ -139,7 +139,6 @@ class MCHit(BHit):
     @property
     def T   (self): return self.time
 
-
     def __str__(self):
         return '<pos = {} E = {} time = {}>'.format(
                 self.pos.tolist(), self.E, self.time)
@@ -158,8 +157,12 @@ class MCHit(BHit):
 
 class Voxel(BHit):
     """Represents a Voxel"""
-    def __init__(self, x,y,z, E):
+    def __init__(self, x,y,z, E, size):
         super().__init__(x,y,z, E)
+        self._size = size
+
+    @property
+    def size(self): return self._size
 
 
 class Cluster(BHit):
--- a/invisible_cities/evm/event_model_test.py
+++ b/invisible_cities/evm/event_model_test.py
@@ -40,11 +40,12 @@ def event_input(draw):
 
 @composite
 def voxel_input(draw, min_value=0, max_value=100):
-    x     = draw(floats  (  1,   5))
-    y     = draw(floats  (-10,  10))
-    z     = draw(floats  (.01,  .5))
-    E     = draw(floats  ( 50, 100))
-    return x, y, z, E
+    x     = draw(floats(  1,   5))
+    y     = draw(floats(-10,  10))
+    z     = draw(floats(.01,  .5))
+    E     = draw(floats( 50, 100))
+    size  = np.array([draw(floats(1,2)), draw(floats(1,2)), draw(floats(1,2))])
+    return x, y, z, E, size
 
 @composite
 def cluster_input(draw, min_value=0, max_value=100):
@@ -137,9 +138,9 @@ def test_hit(ci, hi):
 
 @given(voxel_input(1))
 def test_voxel(vi):
-    x, y, z, E = vi
+    x, y, z, E, size = vi
     xyz = x, y, z
-    v = Voxel(x, y, z, E)
+    v = Voxel(x, y, z, E, size)
 
     np.allclose(v.XYZ  , xyz,                rtol=1e-4)
     np.allclose(v.pos , np.array(xyz),       rtol=1e-4)
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -65,7 +65,7 @@ def voxelize_hits(hits             : Sequence[BHit],
 
     cx, cy, cz = map(centres, edges)
     nz = np.nonzero(E)
-    return [Voxel(cx[x], cy[y], cz[z], E[x,y,z]) for (x,y,z) in np.stack(nz).T]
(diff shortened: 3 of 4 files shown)


jjgomezcadenas:
A tuple seems to be very appropriate here (almost by definition), no? We could even return a named tuple

jacg:
Yes, I'd like to go with a namedtuple. I'm just wondering whether an `np.array` wouldn't be more convenient in certain use cases. I suggest that we go with the namedtuple until such use cases present themselves.

Do we have and xyz namedtuple lying around somewhere? ... memory is hazy.

jjgomezcadenas:
Actually, I think we have an xyz objet in ic_types...

jacg:
As the only client of the voxel size, for now is the upcoming change in `make_track_graphs`, and that code naturally wants to use `np.array`s, we'll go with that for now. If a use-case comes up in the future that requires something else, we'll provide it when it becomes necessary.

paolafer (approved):
As discussed in the meeting, Voxel has the attribute of its size.

jacg:
Well done @paolafer.

Adding the URL of the PR in the commit message would have been good.

gonzaponte:
That is my fault, actually. I thought this was done automatically, because in the web interface and in the mail that says "Merged #384.", the "#384" appears as a hyperlink to the PR.

How should we do it? With the brackets + parenthesis syntax that is used, for example, in markdown or put the raw url in the commit body?

jacg:
> I thought this was done automatically

GitHub does all sorts of automagic in that respect, so if you are viewing it on GitHub, or reading an email sent by GitHub the automagic is there. If you are reading the commit message anywhere else (such as the Git log on the command line or in Magit), the magic isn't there.

> How should we do it?

Probably just the raw URL, as the brackets are markdown syntax, and you might be looking at this somewhere where markdown is not recognized. On the other hand, the markdown syntax is quite unobtrusive, so putting the brackets in would be OK too.

gonzaponte:
Ok. Loud and clear!

jacg:
One more thing: Please wrap the commit messages at around 72 characters, rather than writing one long line per paragraph.

gonzaponte:
I used to do that, but then I realized that emacs splits the lines when you are writing the commit body. It seems this is just an emacs trick for visualization, but the actual text doesn't contain line breaks. Noted!

paolafer:
> Probably just the raw URL, as the brackets are markdown syntax, and you might be looking at this somewhere where markdown is not recognized. On the other hand, the markdown syntax is quite unobtrusive, so putting the brackets in would be OK too.

So, should use the brackets or shouldn't we? Is this choice left to personal decision?

> One more thing: Please wrap the commit messages at around 72 characters, rather than writing one long line per paragraph.

Do you mean that the total length of the message must be 72 characters? Or each line of the message, ending with a full stop?

jacg:
@paolafer

> should use the brackets or shouldn't we?

Personal choice.

@gonzaponte 

> emacs splits the lines when you are writing the commit body

Only if you have something like `visual-line-mode` switched on. Or do you mean something else?

gonzaponte:
> Or do you mean something else?

Mmm, I've just checked and it seems to break the line regardless of the state of `visual-line-mode`. Besides, when it breaks the line it doesn't show the usual arrow on the left margin (I don't know if this is a configuration option or it is always there) that indicates the line has not been broken.

jacg:
> It seems to break the line regardless of the state of `visual-line-mode`

I see what you mean. But as far as I can see (that is to say, `whitespace-mode` shows it) there are genuine newlines in there.

> when it breaks the line it doesn't show the usual arrow on the left margin [...] that indicates the line has not been broken.

Exactly, so I would expect the lines to be broken in the commit message ... and a quick test suggests that that is exactly what happens, so I'm confused as to how you managed to get unbroken lines in there, if you were using Emacs!

paolafer:
> so I'm confused as to how you managed to get unbroken lines in there, if you were using Emacs!

I'm confused about the whole discussion :-D. I remember I wrote the whole message without newlines (but probably I've missed the whole point of the discussion...)

gonzaponte:
Ok, so it is something with @paolafer's emacs. Did you (@paolafer) write the commit message in emacs or did you paste it from another editor (I don't remember).

jacg:
Yes, it's starting to sound like @paolafer's emacs (or lack thereof) is the key.

Please remember that `M-q` (`fill-paragraph`) is a handy way of getting the lines to be the right length, in just one keystroke.

paolafer:
In my emacs, the commit message appears correctly, with the lines broken, and also in the browser... Where do you read just one long line?

gonzaponte:
> In my emacs, the commit message appears correctly, with the lines broken, and also in the browser... Where do you read just one long line?

Mmm, I just realized that the commit I was referring to was [3e2cd77](https://github.com/nextic/IC/commit/3e2cd77f11d0fd7e94f6ccf111758788aab7d8f3) from @abotas. Did we both make the same mistake?

jacg:
Yes, probably. I recall looking at a whole bunch of commits, many of them by @abotas, and those certainly had long lines. I suspect that I painted @paolafer with the same brush.

Slander!

Apologies.

Issue #80, opened 2017-01-16
describe JCK-1


jjgomezcadenas:
We are effectively moving towards JCK-1 syntax, and the last PR tries hard to follow the JCK notions. However, if we are going to ask other developers to follow it, we need a description/discussion (with examples, that can be taken from the Irene.py code) to illustrate the most important ideas.

jacg:
> We are effectively moving towards JCK-1 syntax

It's not syntax. It's a style, or a guideline, but it's not syntax.

PR #385, opened 2017-10-29
Add new normalization option "center"

gonzaponte:
It is a shortcut for
```
norm_strategy = "index"
norm_opts = {"value": central_bin}
```
where `central_bin` is the arrays' shape tuple halved.

The idea is that the user doesn't need to know the indices corresponding to the central bin.

### DIFF
--- a/invisible_cities/reco/corrections.py
+++ b/invisible_cities/reco/corrections.py
@@ -2,9 +2,9 @@ import numpy as np
 import scipy as sc
 
 
-from ..core                     import fit_functions as fitf
-from ..core.exceptions          import ParameterNotSet
-from .. evm.ic_containers       import Measurement
+from ..core               import fit_functions as fitf
+from ..core.exceptions    import ParameterNotSet
+from .. evm.ic_containers import Measurement
 
 
 class Correction:
--- a/invisible_cities/reco/corrections.py
+++ b/invisible_cities/reco/corrections.py
@@ -84,6 +84,11 @@ class Correction:
             f_ref = self._fs[mult_index]
             u_ref = self._us[mult_index]
 
+        elif   strategy == "center":
+            index = tuple(i//2 for i in self._fs.shape)
+            f_ref = self._fs[index]
+            u_ref = self._us[index]
+
         elif   strategy == "index":
             if "index" not in opts:
                 raise ParameterNotSet(("Normalization stratery 'index' requires"
--- a/invisible_cities/reco/corrections.py
+++ b/invisible_cities/reco/corrections.py
@@ -34,9 +34,11 @@ class Correction:
 
     def __init__(self,
                  xs, fs, us,
-                   norm_strategy = False, norm_opts = None,
+                   norm_strategy = False,
+                   norm_opts     = None,
                  interp_strategy = "nearest",
-                 default_f = 0, default_u = 0):
+                 default_f       = 0,
+                 default_u       = 0):
 
         self._xs = [np.array( x, dtype=float) for x in xs]
         self._fs =  np.array(fs, dtype=float)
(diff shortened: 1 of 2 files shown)


jacg:
Te-he ... if you are going to stick faithfully to the alignment strategy that you picked, then you'll have to move the colons to the next column, in the other branches, as "center" is one character longer than "index" and "const", which determined your colon-alignment before.

I think you've maybe taken it a bit too far, in this case, but I'm certainly not complaning, and I *do* like the `return` at the beginning standing out (... but I am perhaps just a little worried about the monster I have created :-).

jacg:
Please *do* stick whitespace around mathematical operators (this is mainly in the tests). I'm ambivalent about `**`, but we really do want them around `/` and `//`.

jacg:
Convince me that `norm_index` in the 1d test really should be based on `E` rather than `X`.

jacg:
This needs to be reviewed by someone who uses corrections on a daily basis: I would have to visit dark and forgotten corners of my mind in before getting up to speed on what this means exctaly, which I would love to do ... if I had the time.

Please, don't let this simple PR fester here for an unreasonable amount of time.

gonzaponte:
Thank you for the corrections.

> we really do want them around `/` and `//`.

I guess `*`, `-`, `+` and any other operator follow the same rule, right? If so, I will fix the spacing for the files modified in this PR.

> Convince me that `norm_index` in the 1d test really should be based on `E` rather than `X`.

I will not. `X` should be the one used.

jacg:
> I guess `*`, `-`, `+` and any other operator follow the same rule, right?

Yes. All infix operators. Even `**`, but, somehow, that one seems ... different.

gonzaponte:
What is your opinion in this case? `size  = draw(integers(min_value=2  , max_value=10))`

Should the `=` signs used in the keyword arguments be surrounded by spaces? I personally put spaces when there are both positional and keyword arguments **in the same line**, so it is more obvious which of those are the keyword ones. If the keyword arguments are in different lines I put the spaces back in place. I do not have a rule for this one, though.

Let me know if we have agreed on a different style so I can, at least, avoid introducing more of mine.

jacg:
[PEP8](https://www.python.org/dev/peps/pep-0008/#id28) says that there should be no spaces around the `=` in keyword arguments. I rationalize this by observing that the separation between different arguments should be more obvious than the separations between components of the same argument.

Which is why we override PEP8 on this one when we spread the arguments over multiple lines: when keyword arguments appear on separate lines, we *do* put the whitespace around the `=`.

It seems that you are following our rule without being aware that we have one, and that you are doing it for exactly the reason that we made the rule. Carry on.

Somewhere we have an issue (ah, yes, #80) telling us that we should document our rules. I'm not going to have time to do it any time soon.

jjgomezcadenas:
@gonzaponte what is the status of this PR?

gonzaponte:
> what is the status of this PR?

This needs a review and a rebase (with conflict management). I will take care of the latter. As pointed out by @jacg:

> This needs to be reviewed by someone who uses corrections on a daily basis

And I think we failed, here:
> Please, don't let this simple PR fester here for an unreasonable amount of time.

jjgomezcadenas (commented):
Actually, the meat of the PR is relatively straight forward and clearly written, it's simply adding an option (center) which makes sense.  However, I am not sure that the option is tested

jjgomezcadenas (approved):
Oh, yes, test is there, clear and sharp. Why on Earth have we lingered so long to approve this PR?

gonzaponte:
> Why on Earth have we lingered so long to approve this PR?

Because we have very few people reviewing PRs...

jjgomezcadenas:
Not really, see my comment on issue. But that's OK, rebase and let's get rid of this.

jacg:
I have a few minutes, so I could try to rebase and merge this ... which is what I shall do, unless I hear that someone else is already doing it.

gonzaponte:
I am doing it

gonzaponte:
Rebased.

jacg:
Naughty @gonzaponte: you went to the trouble of editing the commit message to include approval comment, but didn't add the PR number to the title! Tsk, tsk. Minus one karma point.

gonzaponte:
Shame on me!

However, you are turning into a softie. I've heard promises of "amputated digits"... ;)

jacg:
Karma points ... digits ... same thing. Anyway, as my most fervent disciple, I need you to be able to type efficiently.

jjgomezcadenas:
Oh, my. @jahernando @gonzaponte has officially become @jacg little grass hoper.

jacg:
Well, *someone* will have to keep you guys from making it all go pear-shaped when I'm not around. And I remind you, he has demonstraded that he understands the perils of the IO monad.

PR #386, opened 2017-10-30
Merge tracks

paolafer:
This PR adds a track merging functionality to paolina functions. 
When finding two voxels in two different tracks separated by at most one voxel, the algorithm adds that voxel to the graph. 
The energy is redistributed among the two old voxels and the new one in the following way: the old voxels keep 75% of their energies and the new one acquire the energy subtracted. This is something that could be optimized, I just thought about a simple criterium to start with. 
The min_nodes parameter require the tracks have more than min_nodes voxels to undergo merging, in order to merge only long tracks, and not merge X ray depositions with long tracks.


jacg:
Why is this useful?

paolafer:
Because with our current reconstruction, sometimes we don't reconstruct enough hits to have connected tracks with a reasonable size of voxels and it is common to have a long tracks just divided in two halves. Ideally one wants smaller voxels to reconstruct better the features of the tracks, so one way to keep connectivity is "merging" long tracks, provided that they are close enough and long enough (to be sure that they are not a long track plus an X ray, for instance).

jacg:
I've been doing a binary search for the limiting voxel size which keeps the track connected, for this sort of purpose. Of course, I don't have many data!

Can you provide data for some events to which you would want to apply this?

What is your criterion for deciding whether you have reasonable voxel size?

Also, slight aside, can you think of any situation in which it would be useful to have a non-cubic voxel? We have this ability, but it seems completely unused.

jjgomezcadenas:
I believe the natural size of the voxel is a parallelepiped, given that our system of references is rectangular, and our errors decouple nicely in X-Y then Z. But we have never tried spherical voxels, which could be better to walk from bob to blob. An interesting idea to be tried, I think.

jacg:
> I believe the natural size of the voxel is a parallelepiped,

I think you mean 'shape' when you say 'size'.

I was actually going the other way, to *less* flexibility. Currently our code admids cub**oid** voxels, but we only ever seem to use cub**ic** ones. I'm wondering whether to remove this unused flexibility.

Not in any hurry to do so, but I'm starting to hard-wire cubes in experiments I do (to save time), and I want to know whether I'm missing out on something important by doing so.

jjgomezcadenas:
Actually, I think that cuboid is fine. We have used in the past voxels of 10 x 10 x 5 and I would like to try them again. The point is that there is less diffusion in z than in XY, so the voxel is naturally smaller in Z than in XY

jacg:
This is exacty the sort of anisotropy that I was expecting. Good.

But it does complicate my idea of binary-search-for-optimal-voxel-size.

jacg:
@paolafer I have rebased on top of `master` which now contains #383. Please fetch and check: I've added a commit which removes the magic-number contiguities you were using.

paolafer:
Ok, I've rebased my local merge-track branch on the origin one. After solving a few conflicts, everything's ok. I'm afraid that I have pushed a useless commit without being aware, also...

paolafer:
Regarding your implemented tests, I don't understand exactly what they do, probably because I'm not familiar with parametrize...

jacg:
> I'm afraid that I have pushed a useless commit without being aware

Then please get rid of it (see below)

> I'm not familiar with parametrize

IC developers *should* be familiar with `parametrize` as writing (and reading!) good tests is fundamental to developing in IC, and `parametrize` is fundamental to writing good tests.

Both getting rid of accidentally pushed commits, and learing to understand `parametrize` are things that are best explained face-to-face. Is there anyone currently in your geographical vicinity who could help you with this?

jacg:
Which test do you mean, BTW? (Where you do not understand the use of `parametrize`.)

paolafer:
In test_contiguity and test_length_cuts_corners.

jacg:
The first line of `parametrize` for `test_contiguity` states that we have 3 parameters:

+ `contiguity`
+ `proximity`
+ `are_neighbours`

You see the same parameters declared as the formal parameters of `test_contiguity` itself.

The rest of the `parametrize` content gives sets of values for these parameters. The test is executed for each of these sets, of which there are 15, grouped in 3 groups for the human reader's convenience.

So, the first time `test_contiguity` is executed with `contiguity=FACE`, `proximity='share_face'` and `are_neighbours=True`; because when we are using face contiguity and two voxels share a face, then it should be true that they are neighbours.

Then the same test is repeated once again, but this time `proximity` has become `'share_edge'`, which means that the voxels should no longer be considered as neighbours, so `are_neighbours` is specified as `False` this time.

And so on, for all 15 sets of arguments.

With that you should be able to figure it out for yourself, for `test_cuts_corners`.

paolafer:
Yes, that's clear! Beautiful tests :-). I've fixed a typo, by the way.

jjgomezcadenas:
@paolafer what is the status of this PR?

paolafer:
I think I have to address @jacg's comment "Can you provide data for some events to which you would want to apply this?"...

jacg:
I think that would be good, yes. But please be aware that I will have very little time for IC things until around December 6th.

jjgomezcadenas:
what is the status of this, @paolafer, @jacg ?

jacg:
It's in my stack, somewhere below #396.

PR #388, opened 2017-11-02
Add "scale" option when loading lifetime map

gonzaponte:
Add a scale factor to the function that loads the lifetime map. This allows the use of the lifetime map in runs that are not right after/before the one used to extract the map.


gonzaponte:
Tests are failing because the database has been modified recently. Fix coming up shortly.

abotas (approved):
Adds a feature, documentation, and a test. Looks good. Approved.

PR #391, opened 2017-11-07
Refactor interpolation in `Correction` class

gonzaponte:
This PR revisits the interpolation section of the `Correction` class. The main changes are:
- Major refactoring: A chunk of code that performed interpolation has been replaced by a handy function from `scipy`
- As a result, the linear and cubic interpolations are directly available.
- The output of the `__call__` method is consistent with the types of the input. Namely, if the inputs are arrays (scalars), the output contains arrays (scalars).
- Tests for all of the above
- Cosmetics, as usual

### DIFF
--- a/invisible_cities/reco/corrections.py
+++ b/invisible_cities/reco/corrections.py
@@ -37,6 +37,7 @@ class Correction:
                    norm_strategy = False,
                    norm_opts     = None,
                  interp_strategy = "nearest",
+                 interp_opts     = None,
                  default_f       = 0,
                  default_u       = 0):
 
@@ -47,11 +48,12 @@ class Correction:
         self.norm_strategy   =   norm_strategy
         self.norm_opts       =   norm_opts
         self.interp_strategy = interp_strategy
+        self.interp_opts     = interp_opts
         self.default_f       = default_f
         self.default_u       = default_u
 
-        self._normalize          (  norm_strategy, norm_opts)
-        self._define_interpolator(interp_strategy)
+        self._normalize          (  norm_strategy,   norm_opts)
+        self._define_interpolator(interp_strategy, interp_opts)
 
     def __call__(self, *x):
         """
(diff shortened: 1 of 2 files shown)


jjgomezcadenas:
@abotas I have approved the easy ones, so that I could throw this one at you.

invisible_cities/reco/corrections.py
```diff
@@ -79,7 +79,7 @@ def _normalize(self, strategy, opts):
 
         elif   strategy == "const":
             if "value" not in opts:
-                raise ParameterNotSet(("Normalization stratery 'const' requires"
+                raise ParameterNotSet(("Normalization strategy 'const' requires"
```
abotas: The double parenthesis are not necessary here I don't think.

invisible_cities/reco/corrections.py
```diff
@@ -91,14 +91,14 @@ def _normalize(self, strategy, opts):
 
         elif   strategy == "index":
             if "index" not in opts:
-                raise ParameterNotSet(("Normalization stratery 'index' requires"
+                raise ParameterNotSet(("Normalization strategy 'index' requires"
```
abotas: and neither here

invisible_cities/reco/corrections.py
```diff
@@ -9,6 +9,15 @@
 from .. evm.ic_containers       import Measurement
 
 
+opt_nearest = {"interp_method": "nearest"}
```
abotas: Since the code below does not depend on these global variables and they serve mostly as documentation I guess this is ok. I'll tag @jacg just in case he has serious complaints. Otherwise I guess we can let this be.

jacg: Does removing these variables cause any tests to fail? If not, then get rid of them: even if they do serve a purpose of documentation (which I can't see at first glance), without tests depending on it, it will diverge from reality sooner or later, and documentation which does not reflect reality is evil.

But do note that just because it'not used in the code below ... well this is a module's global scope so its purpose could well be for someone to import it and use it elsewhere.

Run the tests, and then apply the principle **if it's not tested, then it doesn't exist**.

If removing it doesn't break any test in the test suite, then get rid of it.

The same applies to your earier comment about the nested brackets.

gonzaponte: Those variables are there as a shortcut for "frequently used options", as described in the commit. The tests depend on them as they are used to specify which normalization option is used within the tests.

> its purpose could well be for someone to import it and use it elsewhere.

That is its other functionality, yes. To be used in the same way as described above.

jjgomezcadenas:
what is the status of this, @abotas, @gonzaponte ?

gonzaponte:
Awaiting review by @abotas, which will probably do after TBR.

jjgomezcadenas:
That could be a mistake, big rebase-suffering. I would advise to try to look at it and approve before.

gonzaponte:
This is orthogonal to TBR, I don't think it is a problem.

abotas:
I can do either. It's no problem for me either way

abotas (approved):
Requested changes have been addressed. PR is approved!

abotas:
rebased

PR #392, opened 2017-11-07
Fix `test_noise_sampler_compute_thresholds`

gonzaponte:
This test was failing because of an unexpected change in the
database. The test relies ultimately on the data provided by the
database, which was, in a way, hard-coded in a fixture.


abotas (approved):
Great! This is a straightforward fix.

PR #393, opened 2017-11-16
Remove comma from config

jacg:
The `event_range` parameter, when set in the config file, required an ugly trailing comma, in those cases where only one value was given, for example

    event_range = 10,
    event_range = all,

This PR removes the need for that comma, so the above examples may now be written as

    event_range = 10
    event_range = all

Such trailing commas have also been removed from the sample config files present in the repository.

Additionally, the default behaviour for `event_range` was to run *all* events in the file, although the standard config files overrode this, effectively defaulting to running a single event by default. Running a single event by default has now been built-in to the configuration management code.


jjgomezcadenas (approved):
This exercise has been done con Peras, and the reviewer has learned a couple of useful tricks. It's a clear improvement for the clarity of our conf files.

jacg:
@andLaing This removes the need for the ugly comma, as we discussed.

As for running all events by default, @jjgomezcadenas wanted the most harmless thing possible to happen by default, that is to say, run just one event. If you want to run them all, there are two options:

1. specify `event_range = all` in the config file
2. specify `--event-range all` on the command line.

An alternative spelling of the latter is `-e all`.

gonzaponte:
I thought this PR has already merged, but since it is not I will use it to show @paolafer how to merge a multiple-commit PR.

jacg:
> I will use it to show @paolafer how to merge a multiple-commit PR.

... which is exactly why it has been left unmerged ... as discussed on the morning after it was approved.

gonzaponte:
Ok, I didn't remember that conversation. Good!

gonzaponte:
Congrats to @paolafer!!!

PR #394, opened 2017-11-19
Proposal for new PMaps interface

### LINKED ISSUE #177: PMAPs interface
The PMAPs interface we currently have is an accidental by-product of the implementation choices we happen to be going with right now: no effort has been made to design a decent interface for PMAPs.

The presence or absence of zeros (whether it be in the HDF5 file or in the datastructure used in the transient representation) in the S2Sis is a matter of implementation: the client doesn't care how and where the information is stored, as long as it is accessible conveniently and quickly.

Similarly, whether times or sample numbers are stored in the energy plane components of the pmaps and/or the tracking plane components, is purely a detail of implementation: the client just needs to access information as quickly and conveniently as possible.

Our choice to store times and energies in the S12s, but only energies in the S2Sis, is an implementation detail which we currently throw into the client's face.This choice of implementation *leaks out through the interface*: Compare and contrast how to access energies and their corresponding times in the different planes, when you are holding on to some peak `p`:

+ S12s:
    - energy: `p.E[n]` OK.
    - time  `p.t[n]` Nice symmetry.
+ S2Sis:
    - energy: `p[n]` Eh? Where did the `.E` go?
    - time:  erm ... `a_related_but_different_p[event_no].t[n]`? WTF!

This is an *abominable*, inconsistent interface with a horrible abstraction leak. It is like it is because the implementors (that's us!) haven't bothered to think about the *client* (that's also us!) interface. It needs to be killed with fire!

What needs to be done:

+ [ ] Collect use cases, showing the sorts of things we expect clients to do with PMAPs.
+ [ ] Design interface, one which allows the client to use PMAPs cleanly, efficiently and consistently.
+ [ ] Hide implementation behind interface.

If y'all provide me with use cases, I can tackle the rest in April. Of course, this is best done *con peras*, but the collection of use cases can start right now. The collection of use cases is also the part where I can probably contribute least.

There is lots more to be said on this topic, but this should be enough for a start.

jacg:
Closes #177.

### DIFF
--- a/invisible_cities/evm/new_pmaps.py
+++ b/invisible_cities/evm/new_pmaps.py
@@ -37,6 +37,9 @@ class _Peak:
     @property
     def width(self): pass
 
+    @property
+    def rms(self): pass
+
     def energy_above_threshold(self, thr): pass
 
     def  width_above_threshold(self, thr): pass
--- a/invisible_cities/evm/new_pmaps.py
+++ b/invisible_cities/evm/new_pmaps.py
@@ -25,6 +25,24 @@ class _Peak:
     @property
     def times(self): pass
 
+    @property
+    def time_at_max_energy(self): pass
+
+    @property
+    def total_energy(self): pass
+
+    @property
+    def height(self): pass
+
+    @property
+    def width(self): pass
+
+    def energy_above_threshold(self, thr): pass
+
+    def  width_above_threshold(self, thr): pass
+
+    def    rms_above_threshold(self, thr): pass
+
 
 class S1(_Peak): pass
 class S2(_Peak): pass
--- a/invisible_cities/evm/new_pmaps.py
+++ b/invisible_cities/evm/new_pmaps.py
@@ -1,12 +1,17 @@
 class PMap:
 
-    def peaks(self, kind = object):
-        return [ peak for peak in self._peaks if isinstance(peak, kind) ]
+    @property
+    def s1s(self): pass
 
     @property
-    def n_peaks(self): pass
+    def s2s(self): pass
 
-    def get_peak(self): pass
+    # Optionally:
+    @property
+    def number_of_s1s(self): pass
+
+    @property
+    def number_of_s2s(self): pass
 
 
 class _Peak:
--- a/invisible_cities/evm/new_pmaps.py
+++ b/invisible_cities/evm/new_pmaps.py
@@ -1,0 +1,44 @@
+class PMap:
+
+    def peaks(self, kind = object):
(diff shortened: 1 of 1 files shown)


jjgomezcadenas (commented):
I agree with this interface.

gonzaponte:
@abotas and I are not comfortable with the interface for `PMap`.
1. `n_peaks` and `get_peak` don't make sense, because:
*a)* They are meaningless and not practical. We need the n-th S1 or the n-th S2, but not the n-th peak. Likewise, we don't care about the sum of the number of S1s and the number of S2s, but rather the number of each type separately.
*b)* You could fix *a)* by making them methods, rather than properties and follow the same logic as in `peaks`, but then it becomes a monstrous interface for what is a couple of sequences with different physical interpretations.
2. We think `PMap` should be a provider of two sequences: s1s and s2s. They would contain objects of the type `S1` and `S2` respectively. If we find ourselves in the need of a third type, we can always ***extend*** the interface, without breaking it. The sequences we are trying to provide can be given as:
*a)* the `peaks` method currently implemented, which returns one thing or the other depending on the argument. This makes it easier for a 3rd peak type to be included.
*b)* with specific `s1_peaks` and `s2_peaks` properties. In this case, the third peak would imply the ***extension*** of the interface.

Can discuss it tomorrow?

jacg:
Please push a commit with your proposed new interface on top of the commit that is already here, leaving the original one intact.

Input from me will be, at best, sporadic, during the next week.

jacg:
You appear to have removed the ability to get the nth peak altogether, and your comment above suggests that you still wanted to be able get the nth peak of your desired type.

gonzaponte:
> You appear to have removed the ability to get the nth peak altogether, and your comment above suggests that you still wanted to be able get the nth peak of your desired type.

Yes, but `s1s` and `s2s` are intended to return indexable sequences. Therefore we can access the n-th S1 peak as `pmap.s1s[index]`. We (@abotas, @jahernando and I) prefer this rather than `pmap.get_s1_peak(index)` because it is equally informative and simpler.

jacg:
> it is equally informative and simpler.

... and goes against the Python 3 standard of returning lazy iterables.

But that's probably OK in this case.

gonzaponte:
Ok. We are going to start working on the implementation. Unless someone complains (soon), this will be approved later today.

abotas:
I've rebased this over master

abotas:
We have slightly extended `Peak`'s interface. We've added some of the most used properties that exist in the old `Peak` class.

abotas (approved):
It looks the IC community has given its blessing -- this new pmaps interface is hereby approved.

PR #395, opened 2017-11-20
Add assert to test_s1_s2_si_from_pmaps

abotas:
test_s1_s2_si_from_pmaps was missing `assert`s before its `np.allclose()` checks so the `np.allclose()` could return false and the test would still pass. This has been fixed.


jjgomezcadenas (approved):
Some one whose name better not be mentioned, forgot to assert what had to be asserted, and this PR mercifully fixes the slip. Thanks Ale

Issue #389, opened 2017-11-06
Possible bug in voxelize_hits


paolafer:
After a careful comparison of our codes and results using paolina functions, @jerenner and myself realized that, using different versions of those functions, the results change substantially.
In particular, while I'm using the current master, Josh is using a master at commit 07e9f9a21bc97bdedfd90d66a89d713a301ab99a.
In particular, the energy distribution of the blobs is very different, the selection of events being the same.
As an example, I attach one event of data, where the new master reconstructs two blobs with really small energy and the older version reconstructs two blobs with similar and quite high energy. I attach a file with the reconstructed hits and two files with the list of voxels and blobs given by the old (bug_josh) and the current (bug_new) versions of the master.
[possible_bug.zip](https://github.com/nextic/IC/files/1446656/possible_bug.zip)

Whether the correct result is the one with the new or with the old version, is not clear to me. The only difference is in the way we define the boundaries of the voxels, which shouldn't introduce such a big difference. If it does, we should study why the old (and incorrect, in the sense that it didn't do what we wanted) version gives better results, since it may open a window to improving the code.

paolafer:
After a bit more of playing with voxelization, I may have a possible explanation why the two versions of the algorithm perform differently, actually why the "wrong" version (let's call it 'old') performs better than the "right" ('new') one. @jacg, please, shout out loud if I misunderstood some part of your code.

Let me remind you the difference between the two versions:

old: the size of the voxels, although given to the algorithm by the user as a fixed parameter, are slightly different event by event. This happens because the axis ranges of the 3D histogram used for the voxelization are not calculated explicitely, but they're set automatically by the np.histogramdd() function, given the number of bins. Therefore, depending on the positions of the most outer hits, the ranges vary, thus the bin size varies as well.

new: the size of the voxels are fixed by parameter. The ranges of the histogram axes are calculated accordingly. Therefore, although the ranges depend also on the positions of the hits, they do it in a more rigid way. A consequence of this fact is that the most external hits may fall inside a voxel whose border is far from the hit, something like this:

           |       x |          | ...
       

instead of:
            
           | x       |          | ...


The second voxelization allows to make a mor accurate estimation of the extremes of a track, and I think that this is the reason why the 'old' algorithm works better in blob reconstruction.

This shouldn't make a big difference, if the voxel size is small, but since we're using 15 mm cubic voxels, I suspect that it does.

[...]

PR #396, opened 2017-11-22
Let voxels have variable size event by event.

paolafer:
This PR addresses issue #389. A fixed voxelization for all the events is not optimal and, in fact, it turns out to have a big impact in the blob definition. Therefore, in this PR the method voxelize_hits has been changed in order to allow voxels to have a variable size that is at most as large as the size used as an input. Contiguity is calculated according to this event-by-event variable voxel size. @jerenner, once this PR will be approved, we need to use this code for the analysis, since the one we were using until now didn't calculate contiguity correctly.

### DIFF
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -40,32 +40,24 @@ def voxelize_hits(hits             : Sequence[BHit],
     if not hits:
         raise NoHits
     hlo, hhi = bounding_box(hits)
-    bounding_box_centre = (hhi + hlo) / 2
-    bounding_box_size   =  hhi - hlo
-    number_of_voxels = np.ceil(bounding_box_size / voxel_dimensions).astype(int)
-    number_of_voxels = np.clip(number_of_voxels, a_min=1, a_max=None)
-    voxel_edges_lo = bounding_box_centre - number_of_voxels * voxel_dimensions / 2
-    voxel_edges_hi = bounding_box_centre + number_of_voxels * voxel_dimensions / 2
-
-    # Expand the voxels a tiny bit, in order to include hits which
-    # fall within the margin of error of the voxel bounding box.
-    eps = 3e-12 # geometric mean of range that seems to work
-    voxel_edges_lo -= eps
-    voxel_edges_hi += eps
-
+    hranges = hhi - hlo
+    bins = np.ceil(hranges / voxel_dimensions).astype(int)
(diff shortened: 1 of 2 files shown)


jacg:
There are a number of things that bother me about this PR, and the
reasons why it exists:

+ It is largely a reversal of the the meat of #352.

+ It exists to counteract exactly the (apparently undesirable) effects
  introduced by #352 that I described in #352 itself, aswell as in
  #389 and in my answers to private emails. (Given that these emails
  discussed technical details of code that was being actively PRed,
  these discussions should have been in the PR and the related issue,
  so that all concerned could remain informed!)

+ I kept on repeating those points, precisely so that those who are
  actually going to use the code to analyse real data, could decide
  whether the change was *really* desirable, **before** accepting
  it. So I was surprised to find that #352 was accepted in the middle
  of what I perceived to be a discussion about its unsuitablity.

All this comes back to advice that I have been making repeatedly:
**less haste, more speed**: we make faster progress in the long term
by taking it slowly and being more careful in the short term!

The diff between 19d8298e (where the work being undone was done in the
first place) and this PR, is posing more questions than my currently
available time permits me to consider **carefully**, so I cannot
review this PR at present.

paolafer:
I asked for the change, because we were using the code assuming that voxels had fixed size, while they hadn't. Probably we should have fixed this assumption and make a comparison between the behaviour of fixed and variable size voxels, but we didn't. Personally, I didn't realize at that time that variable size voxels, used properly, could have been better than fixed-size. I had to use the code for a while to be able to understand that. 
I could try to show the different behaviour with some plots in an analysis meeting to decide together which solution is better, if you think it is worthwhile.

jacg:
In https://github.com/nextic/IC/issues/389#issuecomment-346811581 you present data supporting the hypothesis that variable-sized voxels are better.

I have looked over the code in more detail and am now happy with it.

However, before accepting, merging and removing the fixed-size voxels code, why don't we consider whether we don't want to keep *both* possibilities, giving `voxelize_hits` a parameter for switching between the two (with variable-size being the default)?

jjgomezcadenas:
@paolafer are you on this? I would like to get this PR out of the way asap

paolafer:
Yes, I'm on this, see conversation with @jacg in #389.

jjgomezcadenas:
What is the status of this, @paolafer? It would be good to approve it before TBR

jacg:
It's waiting for me to find some time to make progress on it.

PR #397, opened 2017-11-22
Fix normalization option "max" for N-dimensional corrections

gonzaponte:
The normalization option "max" was failing when used in corrections with more than 1 dimension. This was not detected before because we didn't have a test for that case (shame on me!). This PR adds that test, which also demonstrates the bug and fixes it.


jjgomezcadenas (approved):
This is good stuff. It adds a test to demonstrate bug and then a solution that fixes the bug. Concise and efficient coding, which starts to be a trademark of the house.

PR #398, opened 2017-11-22
Fix overflow warning

gonzaponte:
A test was raising an overflow warning continuously due to the use of
a small value within an exponential function. This commit makes
that value a bit bigger so the warning is not raised, but the test
remains the same.


jjgomezcadenas (approved):
This is an easy one. Since @abotas is procrastinating I take the liberty to approve it

PR #400, opened 2017-11-24
Fix outdated instructions in contribution guide

jacg:
The guide still contained some information that directly contradicts our current practices!

Much work remains before the guide is complete, but this was urgent: as we expect new contributors over the coming weeks, we want to make sure that the contribution guide doesn't instruct them to do the things that we want to avoid!


jjgomezcadenas (approved):
Urgent update of doc

PR #403, opened 2017-11-27
Add SiPM spectra city

andLaing:
Addition of city and related functions for the generation of SiPM
photoelectron spectra (binned pdfs).

Additions include:

1) invisible_cities/cities/sipmPDF.py (and corresponding test) which
   reads raw data and generates two types of spectra for each channel
   once converted into pe (zeroed and in pe and zeroed, in pe and
   difference from mau);

2) invisible_cities/io/hist_io.py, a histogram writer to save on
   output data size.

Changes to existing functions include:

+ change to invisible_cities/peak_functions_c.pyx.signal_sipm to
  include calibration option so that all bin data is saved rather than
  just selected data

  + cal=0 (default) is equivalent to unchanged function,
  + cal=1 saves sipmrawf-mean,
  + cal=2 saves sipmrwf-mean-mau

Undid accidental include of some changes from an unrelated
branch.


jjgomezcadenas (commented):
I have started to review this PR, probably best to talk in person (on Wednesday) to make faster progress. Meanwhile, please, find a name you like for the city 
https://en.wikipedia.org/wiki/Invisible_Cities
and change the name of your script (naming cities following Calvino's is mandatory in IC :)!

paolafer:
Could I join the discussion on Wednesday? I would like to understand better the way a city works, technically.

jacg:
> I would like to understand better the way a city works, technically.

... talkin 'bout a revolution sounds like a *whisper* ...

*whisper, whisper* ... a city just transforms streams of data by piping the data through a bunch of components that perform the transforms, with some added bookkeeping info added in to the mix. Unfortunately the current code structure makes this behavioural structure very difficult to see. If only cities were expressed as dataflow networks, all this would be much clearer, easier to understand, easier to test, easier to fix, easier to debug, easier to reuse, easier to explain, easier to parallelize, easier to ... ssshhhhh ... there's someone coming ... quiet ... *whisper* ...

jjgomezcadenas:
Boys and Ladies, indeed, something is coming.... :)!
Meanwhile, yes, let's meet tomorrow circa 3:30 for a look at @andLaing city...

bin/city
```diff
@@ -5,15 +5,20 @@ from importlib import import_module
...
-    module_name = 'invisible_cities.cities.' + city_name
-    city_class  = getattr(import_module(module_name), city_name.capitalize())
-except ModuleNotFoundError:
-    print('available cities: <TODO look at contents of config directory>')
-    print('usage TODO: <get this from our config parser>')
-    traceback.print_exc()
-    exit(1)
-else:
-    city_class.drive(args)
+for i, fldr in enumerate(city_dirs):
```
jacg, line 12: We use the name 'directory' to refer to directories in IC, so please avoid introducing a different term for it: 'folder'. Please spell it `for dir_ in city_dirs`. The trailing underscore ensures that we avoid shadowing a built-in. Alternatively, `directory` would be fine too.

But, more importantly, is there any good reason why this city can't live with all the other ones? Do we really need to introduce an extra place where cities might live?

This touches on something else that has been bothering me with this city: it seems to introduce modifications to parts of IC where I wouldn't normally expect authors of new cities to have to go. (This is just the impression I get after a cursory look: I haven't had the time to look at it in any detail.)

*If* (and it's a big *if*) the need to add this extra directory turns out to be genuine, then we can do without the `enumerate` and the `if i == len(city_dirs)-1`. The correct spelling of that idea is to cut the old body of the `NoModuleFoundError` handler, replacing it with `pass` or `continue`, and to move it to the `else` block of the `for loop` (yes, the `for loop`!) The addition of `break` to the `else` block of the `try` will still be necessary.

invisible_cities/calib/sipm_pdf.py
```diff
@@ -22,7 +22,7 @@
 from ..  cities.base_cities import EventLoop
 
 
-class SipmPdf(CalibratedCity):
+class Sipm_pdf(CalibratedCity):
```
jacg: We conform to PEP8 when it comes to class names (unless there is some overriding reason not to), so this change would have to be reverted, were it not for the fact that it's not a Calvino name, so @jjgomezcadenas is not going to accept the old name either.

andLaing:
Juanjo suggested the change and José María came up with a patch to make it work correctly but it’s definitely something that should be discussed and, if still wanted, improved.

As to the more fundamental changes (I assume you mean the changes to reco/peak_functions_c.pxd <https://github.com/nextic/IC#diff-58ef3e9a0e2c13c279e6c874859e4e6a> and base_cities.py), Juanjo still has to review them. However, to have access to the base level information necessary for sensor calibration (not only in the existing case but for other modules that will have to be written) this or another probably more fundamental change needs to be made. The existing functions set to zero anything not passing a selection cut which is of no use if we want to calibrate but the code should be the same up to the first point of deviation in my opinion. I’m sur ethat you experts can come up with a better change but, as mentioned before, a change is necessary.

A

andLaing:
JJ approved (actually did the commit I think) with SipmPdf but it is unacceptable due to the .capitalize() function in bin/city so José María suggested the change.

A

jacg:
'.capitalize()` will cease to be an issue as soon as the name has been Calvino-ized, unless the book contains a name with a capital in the middle!

The point about peak functions needing to change seems reasonable. I was under the impression that there were changes elsewhere, but perhaps I just imagined those.

But why does this city need to live in a directory of it own? Can't it play nicely with the others?

jacg:
Sorry, accidentally hit wrong button.

jjgomezcadenas:
>But why does this city need to live in a directory of it own? Can't it play nicely with the others?

The logic behind it is as follows. The set of scripts that we call cities (all of them baptised with pretty girlish names), are, in a general sense, data-transformation scripts. What the different cities do is to manipulate data producing new data which typically contains higher level of abstraction. We start for example with raw waveforms and end up with tracks.

But then, again, one could try to organise things in a slightly more specialised way. For example, one could distinguish between reconstruction cities, analysis cities and calibration cities. They are all quite similar structurally, but slightly different functionally. One could define the role of a reconstruction city a to produce progressively more sophisticated DSTs, while the role of a calibration city is to produce files with calibration constants (to be used typically by reconstruction cities). Analysis cities would be fed with the DSTs produced by reconstruction cities and output even higher level objets, such as fitted histograms or even fully written Papers! And perhaps Simulation cities would produce simulation data to be used by the others...

The attempt to create a calib directory goes into this direction. Calibration has its own dynamics and I was under the impression that calibration scripts would be better off under their own directory. This would not make any sense if we were to write only one calibration script but presumably there will be many.

This also opens the door to place "neural networks cities" in their own directory. In this way, also, one type of city does not interfere with the other.

As a side-effect of calibration cities moving to a different directory I was ready to accept that any city not leaving in city directory could have a useful-descriptive-and-ugly-name (the sort of stuff @andLaing loves and I hate). 

One could push the idea further, and rename cities to recoCities for example, but I would be happy to leave cities as the implicit "reconstruction city directory", and then as needed add more directories for analysis (if required), neural networks and calibration.

Tha was the rationale, open to debate!

jacg:
Don't have time to give this deep thought right now (it seems perfectly reasonable at first blush), but one thing seems obvious: if you are going to admit descriptive names for cities, and use those cities as command line scritps, please make the city names **absolutely identical** to the names of the executable scritps.

This would mean

1. removing the `capitalize` from `bin/city`

2. renaming the existing calvinistic cities to all lowercase (or start using the capitalized names as the names of the scritps ... I would vote for the former).

jacg:
Oh, and I hate the term 'script' more and more every day.

jjgomezcadenas:
OK, opinions, any one? (and BTW, what would be the kosher name for "script")?

jacg:
> what would be the kosher name for "script"

Program(me).

andLaing:
Hi

Can I maybe get an approve or deny some time today?

From the comments it seems to me that maybe reverting the changes of folders etc and letting JJ do those changes and submit them as a pull request might be the most popular approach but correct me if I'm wrong. In any case the reviewers still need to check the changes to the reco/peak_functions_c.pxd functions etc...

jacg:
No need for a new PR, whatever changes need to be done, can be pushed to this branch.

I can do the stuff I mentioned in `bin/city` tonight, but I'm kinda waiting on others to decide whether

1. we want multiple city directories
2. we want to rename the Calvinistic cities to lowercase

The peak functions changes I'll leave for others to review and fix if necessary.

jjgomezcadenas:
Here my proposal:

1. I believe @jacg does not oppose to adding more directories in order to organise better the hierarchies in IC. I think calibration is a good user case and I can see another one coming, e.g, Neural Network cities, that I believe can be separated from both reconstruction cities (aka cities) and calibration cities (aka calib). Therefore, unless I get a no-go, I take that we can proceed forward with putting the new calibration city in directory calib.

2) I propose that we keep the name cities as an alias for reconstruction cities and if you indulge me, let's keep the convention of calvinized names for cities leaving there. Cities (program) leaving somewhere else do not need to be calvinized. 

3) In order to skip the problem of capitalise vs names of program.py for the time being I propose that we keep for now the convention of name.py and class Name (so the new file would be called sipmpdf.py and the class Sipmpdf as originally proposed by Andrew. This does not violate PEP, I think. We can always open another PR to change naming conventions. 

Is this agreeable?

jjgomezcadenas:
To wrap-up.

1. Since @jacg gives us green light to use multiple city directories I take the liberty (and the responsibility) to make of that a fait accompli.  
2. Unless @jacg says the contrary we carry on with name.py Name.py convention, then we change to whatever he decides. I believe that city.py class City is OK, because it allows, longname.py  class LongName.

andLaing:
OK, I'll revert the names of class+files for the city now...

jacg:
> I believe that city.py class City is OK, because it allows, longname.py class LongName.

This is precisely an example of something that does *not* work! When trying to run a city with the name `longname` (regardless of whether and where there are any capital letters in it) our system will look for a class called `Longname` **note the absence of the capital N in the middle**.

So, if you choose not to match executable and class name exactly, then you will have to restrict yourself to just a single capital letter at the beginning of the class name.

Alternatively you could have a mapping like this: `long_name_example` <==> `LongNameExample`.

jjgomezcadenas:
>Alternatively you could have a mapping like this: long_name_example <==> LongNameExample.

I think this one is nice, but let's worry about it later. I am now reviewing the city and the associated functions.

jjgomezcadenas:
At this point I have reviewed and fully revisited the function used by Andrew's new city. The city itself is also OK. From this point of view the PR can be approved. 

Unless @jacg wants to add his own review I will approve it tomorrow.

jjgomezcadenas:
Also, see:

https://github.com/jjgomezcadenas/ICDEV/blob/master/reco/cython_vs_python.ipynb

jacg:
This study should also consider [numba](https://numba.pydata.org/) which today surpasses Cython in speed for some purely numerical codes!

jjgomezcadenas:
>This study should also consider numba which today surpasses Cython in speed for some purely numerical codes!

Done, see notebook. Numba indeed performs better (although not by a large factor) than CYTHON, with the same caveats (e.g, neither CYTHON, nor Numba give optimal performance if using NUMPY rather than coding by loops). 

Apparently using NUMBA is easier than CYTHON, given the fact that you just need to add a decorator. BUT (big BUT). In the study I get an error (see code near the end of notebook) which I am unable to trace. For some unknown reasons the code doesn't like a perfectly kosher instruction (checked that al data OK) and crashes with a very verbose (and at least for me) very unhelpful message. 

Which bring the issue that NUMBA is more a black box than CYTHON.

jacg:
A few quick points about the notebook:

+ Don't copy the timings by hand, use varaiables! For timeits you need to use the `-o` option to `%timeit` like this:
    timings_abc = %timeit -o fn(a,b,c)
    timings_abc.best

+ The notebook would benefit immensely from having a summary table which collects all the results in a single easy overview. If you use variables rather than copy-by-hand, the table will be trivial to maintain and get right.

+ Stop using `print` all over the place. The REPL prints out the last value in each cell for you.

+ **JYTHON** !????!!!

... oh ... I also recommend having a look at Jupyterlab.

jacg:
Besides the horrible, *horrible*, **horrible** crime of siticking Numba work under the heading Jython (for which you lose many karma points), the excessive copy-pasting is not helpful. You have (I suspect, but I'm too lazy to check it precisely) copy-pasted the entire definition of a long function just so that you can stick the `jit` decorator on top of it. If you understand what the decorator syntax means, then you should realize that you can simply do 

    numba_version = jit(pure_python_version)

which has the added benefits that

+ You cannot possibly make any copy-paste errors
 
+ `numba_version` automatically changes (upon next evaluation of its cell) if you change `pure_python_version`

+ People like me, who try to read it, don't have to wonder what it is that you have copy-pasted and whether you have copy-pasted it correctly and remembered to keep it in sync with the original.

gonzaponte:
I haven't read the whole discussion, so I apologize if something I put here has already been discussed:

1. Numpy versions can still be optimized. You are using python for loops in them. Using `np.some_function(arguments, axis=0,1,2)` is equally readable and much faster (factor 4 in my computer).

2. You can do the same trick with `scipy.signal.lfilter`. It accepts the `axis` argument and makes your life easier, reducing the number of python loops.

3. The tests should also verify performance for large arrays (shape (1000, 1000) or so) as this our use case.

While working on TBR we have encountered that Cython is (probably) not necessary at all, but we haven't documented it yet.

jacg:
I haven't had the time yet to look at the code in any detail, so thank you @gonzaponte for spotting this one.

Let me repeat, once again (**everyone please pay attention**), the basic rule of numpy: **If you are looping over numpy arrays with a for-loop, then you are doing it wrong!**

Numpy is designed to work efficiently by using *internal* loops, that is to say loops which are embedded in low-level argorithms that come built-in to numpy: using pure-python loops on arrays guarantees you sub-optimal performance.

To be absolutely clear, if you care about speed, **do not write python for loops overy numpy arrays**.

jjgomezcadenas:
Thanks for the various tips, I will implement them in the NB. However, take notice:

1) Numba crashes for no good reason in one of the cases in the NB. I have not been able to trace why, and this only is enough for me to eliminate Numba as an option (no black magic please). 

> While working on TBR we have encountered that Cython is (probably) not necessary at all, but we haven't documented it yet.

I don't believe you.

jacg:
I don't believe anyone who claims that Cython **is** necessary without showing me profiler results to back it up. So far, this default strategy has had a remarkably high success rate. I look forward to the first counterexample.

jjgomezcadenas:
Oh, this has been lost in the night of times, but the BLR deconvolution function was accelerated by a factor 100 when passed to CYTHON. Also the NB shows that (so far) using CYTHON is better than using Python. I will add the fixes suggested by Gonzalo and we will see.

jacg:
I don't care if Cython can make it 1000 times faster, or a million times faster, if it's not in the bottleneck of your code, it's not worth Cythonizing, and people need to learn to **profile** to figure out where the bottlenecks are **before** optimizing.

jjgomezcadenas:
Actually, in the case of the BLR that was *the* bottleneck, and a very narrow one, slowing the overall performance of Irene and Diomira by at least one order of magnitude. But yes, agreed, profile first, optimise later. Getting into the habit of this was the purpose of this NB. 

Incidentally, where should the NB go (I have asked this question as an issue, no answer)? Right now is in a private repo.

jacg:
> But yes, agreed, profile first, optimise later. Getting into the habit of this was the purpose of this NB.

If the purpose of the NB was to get in the habit of profiling before optimizing, why does the NB contain optimizations but no profiler output?

The comparisons you make in the notebook are both useful and interesting. But there is not even the slightest acknowledgement of the existence of the concept of profiling in that whole notebook (as far as I could see in the quick look I had). You seem to confuse profiling with benchmarking. I really want people to understand that **profiling** is a vital **prerequisite** for optimization.

+ Profile **before** optimization.
+ Benchmark **during** optimization. 

 Deciding where the NB should go, is in my TODO bag, without much hope of being fished out this week.

jjgomezcadenas:
Yes, sorry,  all what I have done so far is benchmarking. Profiling will come. Meanwhile we are polluting the space of a (largely unrelated) PR with a conversation that should move somewhere else.

jjgomezcadenas (approved):
As promised, since no body else has complained, I consider that Andrew's first city can be approved. This PR has triggered a lot of discussion, both on performance and on city structure! Bravo!

@gonzaponte, @abotas or @jmbenlloch, please merge

gonzaponte:
I am sorry to open a discussion here so late, but I haven't been able to follow this PR. There are some things I have noticed:
- The "private" function `_signal_sipm` in `calib_sensor_functions_c` is called in `base_cities.py`, therefore it shouldn't be "private".
- The *new* functions defined in `calib_sensor_functions_c` are used only to compare the output of some tests. They shouldn’t be there unless used for something else.
- I think the `Cal` flag in `signal_sipm` would be better implemented as a symbol, or at least as something more verbose.
- The `hist_io` module is not tested.
- Some tests for the `calib_sensor_functions_c` are not reliable. In fact, they pass because:
- We (@abotas and I) have observed that the MAU is not being used at all (at least not in my laptop). When we print out the vector used for the mau coefficients, it is full of zeros, which is a result of filling the array with int(1) / int(100). It seems that cython performs integer division when both arguments are integers(*). I will open a separate issue for this topic, as it is a(n important) bug.


From the code maintenance point of view, there are lots of:
- Leftover docstrings.
- Commented-out lines.
- Unused imports (new ones, I mean).


Most of these issues are fixed in the PR adapting the code to the new pmap interface, which will be submitted shortly.


(*) Surprisingly, this doesn’t happen in a jit-compiled jupyter notebook cell… Maybe @jacg has some input here (no hurries).

andLaing:
Hi

That the 'new' functions aren't used is something that I noticed after the merge, something went wrong with one of JJ's commits and some last changes weren't integrated. I had made changes for the next calibration related PR but since I knew you were reviewing all these things I didn't rush to do another PR.
Cal sun signal_sipm: I agree, in the first PR it was just the minimal changes required so that we could compare exactly what was happening in the low level distributions using the code as it was. In the branch I sent you the other day I make some suggested changes.
hist_io: completely my fault but since the io modules are quite inconsistently tested (rwf_io which I used as a base doesn't have a test either, for example) I wasn't sure what the politics was. I'll look into it.
Rest are questions I can't really answer...

jacg:
Sorry guys, I'm dealing with the fallout of a hardware failure, which I am combining with a long-overdure review and upgrade of IT infrastructure, all while preparing for an imminent and heavy month on the road, with a funeral thrown in for good measure. In brief, from the IC perspective I effectively don't exist until the workshop.

PR #405, opened 2017-11-28
Fix `test_weighted_mean_and_var_unbiased_frequentist`

gonzaponte:
This test was failing frequently due to the lack of compatibility
between hypothesis and flaky modules. This has ben fixed by replacing
the hypothesis dependence with a random number.

### DIFF
--- a/invisible_cities/core/core_functions_test.py
+++ b/invisible_cities/core/core_functions_test.py
@@ -101,9 +101,8 @@ def test_weighted_mean_and_var_gaussian_function(mu, sigma, ndata):
 
 @flaky(max_runs   = 4,
        min_passes = 3)
-@given(integers(min_value=5, max_value=20))
-def test_weighted_mean_and_var_unbiased_frequentist(ndata):
-    mu, sigma = 100, 1
+def test_weighted_mean_and_var_unbiased_frequentist():
+    mu, sigma, ndata = 100, 1, np.random.randint(5, 20)
     data = np.random.normal(mu, sigma, size=ndata)
     values, freqs = np.unique(data, return_counts=True)
 


gonzaponte:
> the lack of compatibility between hypothesis and flaky modules.

Maybe @jacg can answer this question (when he finds some time, not urgent): is there a way to use hypothesis in flaky tests? I have been searching the documentation and there seems to be some sort of tricks to deal with them but doesn't seem obvious. I am constantly "forced" to remove hypothesis and replace it with a random number generator.

jacg:
Very briefly ...

1. I don't know about compatibility, but it seems difficult, as hypothesis is all about reproducability of results, for the ability of shrinking towards simplest failing cases, while flaky is all about tests which give different results for identical inputs.

2. I don't like it. By it, I mean replacing hypothesis with random numbers. We have to ask ourselves why the tests give different results for the same inputs. Ideally we would have deterministic behaviour in our tests. If a test can be made deterministic, it should use hypothesis. If it cannot be made deterministic, then it probably cannot use hypothesis.

gonzaponte:
Thank you for the clear response.

In this particular case, I used a hypothesis-provided integer `ndat` to decide the number of random samples to be taken from a particular distribution. Because of its randomness, the function will not have a perfectly defined result, but it will rather be distributed in a certain manner. Given that distribution, I know the function I am testing should yield some value between `a` and `b` some fraction of the time.

From your comment, I understand that `ndat` should not be given by hypothesis, but still want to make it variable in some range so the test is a bit more complete. Is it ok to use a random number in this case? Should I just parametrize it and use a reduced number of values?

jacg:
> I used a hypothesis-provided integer ndat to decide the number of random samples to be taken from a particular distribution.

So here is a source of randomness which is beyond hypothesis' control. If this randomness cannot be placed under hypothesis' control, then I don't see how hypothesis can work sensibly with this test.

More generally, you are testing something which has only statistical likelyhood of success, which isn't going to work with hypothesis, unless someone somewhere has done something clever to address this sort of issue. I am not aware of that having been done, and do not have time to search for it at the moment.

In short (without having looked at the details myself), this looks like a genuine case of a test where we don't want hypothesis to be involved.

gonzaponte:
All clear. Thank you :)

abotas (approved):
This PR makes a small change to increase test stability. The comments have converged to agreement -- When the calculations made in the test are not deterministic we should avoid using hypothesis. PR is approved!

PR #406, opened 2017-11-28
Implementation of new PMAPs interface

abotas:
This PR consists of a full parallel python implementation with a corresponding test suite for the new PMAPs following their interface defined in PR #394, and is meant to fulfill step 2 in @jacg's comment at the end of issue #177 [edit] ... and, more importantly, step 2 of #407.

### DIFF
--- a/invisible_cities/core/testing_utils.py
+++ b/invisible_cities/core/testing_utils.py
@@ -31,6 +31,20 @@ def exactly(value, **kwargs):
     return approx(value, rel=0, abs=0, **kwargs)
 
 
+def previous_float(x):
+    """
+    Return the next float towards -inf.
+    """
+    return np.nextafter(x, -np.inf)
+
+
+def next_float(x):
+    """
+    Return the next float towards +inf.
+    """
+    return np.nextafter(x, +np.inf)
+
+
 def float_arrays(size       =   100,
                  min_value  = -1e20,
                  max_value  = +1e20,
@@ -78,3 +92,28 @@ def assert_dataframes_equal(df1, df2, check_types=True, **kwargs):
 
 def assert_dataframes_close(df1, df2, check_types=True, **kwargs):
     _compare_dataframes(assert_allclose, df1, df2, check_types, **kwargs)
+
+
+def assert_SensorResponses_equality(sr0, sr1):
+    # This is sufficient to assert equality since all of SensorResponses other
+    # properties depend solely on .all_waveforms, and all of those properties
+    # are tested.
+    assert sr0.ids           == exactly(sr1.ids)
+    assert sr0.all_waveforms == approx(sr1.all_waveforms)
+
+
+def assert_Peak_equality(pk0, pk1):
+    assert pk0.times == approx(pk1.times)
+    assert_SensorResponses_equality(pk0.pmts , pk1.pmts )
+    assert_SensorResponses_equality(pk0.sipms, pk1.sipms)
+
+
+def assert_PMap_equality(pmp0, pmp1):
+    assert len(pmp0.s1s) == len(pmp1.s1s)
+    assert len(pmp0.s2s) == len(pmp1.s2s)
+
+    for s1_0, s1_1 in zip(pmp0.s1s, pmp1.s1s):
+        assert_Peak_equality(s1_0, s1_1)
+
+    for s2_0, s2_1 in zip(pmp0.s2s, pmp1.s2s):
+        assert_Peak_equality(s2_0, s2_1)
--- a/invisible_cities/evm/new_pmaps.py
+++ b/invisible_cities/evm/new_pmaps.py
@@ -1,3 +1,7 @@
+from textwrap import dedent
+
+import numpy as np
+
 class PMap:
 
     @property
@@ -51,20 +55,66 @@ class S1(_Peak): pass
 class S2(_Peak): pass
 
 
-class SensorResponses:
+class _SensorResponses:
+    def __init__(self, ids, wfs):
+        self._check_valid_input(ids, wfs)
+
+        self._ids              = np.array(ids, copy=False, ndmin=1)
+        self._all_waveforms    = np.array(wfs, copy=False, ndmin=2)
+        self._wfs_dict         = dict(zip(self._ids, self._all_waveforms))
+        self._sum_over_sensors = np.sum(self._all_waveforms, axis=0)
+        self._sum_over_times   = np.sum(self._all_waveforms, axis=1)
 
     @property
-    def all_waveforms(self): pass
+    def all_waveforms(self):
+        return self._all_waveforms
 
-    def waveform(self, sensor_id): pass
+    def waveform(self, sensor_id):
+        return self._wfs_dict[sensor_id]
 
-    def time_slice(self, slice_number): pass
-
-    @property
-    def ids(self): pass
-
-    @property
-    def sum_over_times(self): pass
+    def time_slice(self, slice_number):
+        return self._all_waveforms[:, slice_number]
 
     @property
-    def sum_over_sensors(self): pass
+    def ids(self):
+        return self._ids
+
+    @property
+    def sum_over_times(self):
+        return self._sum_over_times
+
+    @property
+    def sum_over_sensors(self):
+        return self._sum_over_sensors
+
+    def where_above_threshold(self, thr):
+        return np.where(self.sum_over_sensors > thr)[0]
+
+    @classmethod
+    def build_empty_instance(cls):
+        return cls([], [])
+
+    def __repr__(self):
+        n_sensors = len(self.ids)
+        header = dedent(f"""
+            ------------------------
+            {self.__class__.__name__} instance
+            ------------------------
+            Number of sensors: {n_sensors}""")
+
+        sensors = [f"""
+            | ID: {ID}
+            | WF: {wf}
+            """ for ID, wf in self._wfs_dict.items()]
+        return header + "".join(map(dedent, sensors))
+
+    def _check_valid_input(self, ids, wfs):
+        if len(ids) != len(wfs):
+            msg  =  "Shapes do not match\n"
+            msg += f"ids has length {len(ids)}\n"
+            msg += f"wfs has length {len(wfs)}\n"
+            raise ValueError(msg)
+
+
+class PMTResponses (_SensorResponses): pass
+class SiPMResponses(_SensorResponses): pass
--- a/invisible_cities/evm/new_pmaps.py
+++ b/invisible_cities/evm/new_pmaps.py
@@ -8,33 +8,16 @@ from .. core.core_functions    import weighted_mean_and_std
 
 class PMap:
     def __init__(self, s1s, s2s):
-        self._s1s = tuple(s1s)
-        self._s2s = tuple(s2s)
-
-    @property
-    def s1s(self):
-        return self._s1s
-
-    @property
-    def s2s(self):
-        return self._s2s
-
-    # Optionally:
-    @property
-    def number_of_s1s(self):
-        return len(self.s1s)
-
-    @property
-    def number_of_s2s(self):
-        return len(self.s2s)
+        self.s1s = tuple(s1s)
+        self.s2s = tuple(s2s)
 
     def __repr__(self):
         s  = f"""
         ---------------------
         PMap instance
         ---------------------
-        Number of S1s: {self.number_of_s1s}
-        Number of S2s: {self.number_of_s2s}
+        Number of S1s: {len(self.s1s)}
+        Number of S2s: {len(self.s2s)}
         """
         return dedent(s)
 
@@ -43,45 +26,17 @@ class _Peak:
     def __init__(self, times, pmts, sipms):
         self._check_valid_input(times, pmts, sipms)
 
-        self._times = np.asarray(times)
-        self._pmts  = pmts
-        self._sipms = sipms
+        self.times = np.asarray(times)
+        self.pmts  = pmts
+        self.sipms = sipms
 
-    @property
-    def times(self):
-        return self._times
-
-    @property
-    def pmts(self):
-        return self._pmts
-
-    @property
-    def sipms(self):
-        return self._sipms
-
-    @property
-    def time_at_max_energy(self):
-        return self.times[np.argmax(self.pmts.sum_over_sensors)]
-
-    @property
-    def total_energy(self):
-        return self.energy_above_threshold(0)
-
-    @property
-    def total_charge(self):
-        return self.charge_above_threshold(0)
-
-    @property
-    def height(self):
-        return np.max(self.pmts.sum_over_sensors)
-
-    @property
-    def width(self):
-        return self.width_above_threshold(0)
-
-    @property
-    def rms(self):
-        return self.rms_above_threshold(0)
+        i_max                   = np.argmax(self.pmts.sum_over_sensors)
+        self.time_at_max_energy = self.times[i_max]
+        self.height             = np.max(self.pmts.sum_over_sensors)
+        self.total_energy       = self.energy_above_threshold(0)
+        self.total_charge       = self.charge_above_threshold(0)
+        self.width              = self. width_above_threshold(0)
+        self.rms                = self.   rms_above_threshold(0)
 
     def energy_above_threshold(self, thr):
         i_above_thr  = self.pmts.where_above_threshold(thr)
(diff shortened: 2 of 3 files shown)


gonzaponte:
First 4 commits belong to #394. We have rebased this branch on top of the other for obvious reasons.

jacg:
Erm, #394 should have been approved and merged before work on this started.

abotas:
@gonzaponte and I are also wondering if in the cases in which all the property does is return a private attribute of the same name preceded by an `_`, we should just use public attributes.

 `_Peak.pmts`, for example, is currently a property that returns attribute`_Peak._pmts` and nothing more. We understand the point of these properties might be that users cannot set `Peak.pmts` if it is a property and in that sense, these properties make the code slightly more robust. Users can however, always set `_Peak._pmts` which will of course change the behavior of the property `_Peak.pmts`, so it's unclear how much safer the properties actually make the code. 

Our main motivation for considering public attributes over properties + private attributes, is that properties are bulky; using public attributes would make the code much more concise and more readable.

jacg:
Public data attributes and properties are indistingushable, from the interface perspective ... as long as the property has both a setter and a getter.

But these properties are read-only, so turning them into public data attributes, grants the client extra powers. We have two options:

1. Go ahead and make it a data attribute---after all, this is the norm in Python---and trust that users' good sense and our excellent testing culture will prevent any serious problems.

2. Decide that the benefit of the increased safety and semantic clarity of read-only properties outweighs their bulkier implementation.

The good news is that it really doesn't matter: the interface remains (almost) exactly the same, so you can make one choice today, and change your mind about it tomorrow, or next week, or next year, without negatively affecting anybody ... except for those who write code which directly sets these attributes ... but such people deserve what they get :-)

abotas:
How do we review this PR? Should someone other than @gonzaponte and I do it since we are both authors? Or should one of us just review it anyway?

This is time sensitive since we are now going to begin a new PR to adapt the existing IC code to the new pmaps.

jahernando:
I can try to do it...

abotas:
@jahernando Sounds good to me!

jjgomezcadenas:
Another possibility is that @gonzaponte and @abotas give a full presentation, discussing the interface, implementation and tests to a "committee", including @jahernando, @jerenner, @paolafer, @jmbenlloch and yours truly (I assume the master is off for a while). Like that many of the usual  suspects get acquainted. We could organise the review on Thursday afternoon, say 3:30 pm. How does that sound?

gonzaponte:
That would be good, but if we (were we includes you) decide that something is wrong, we have no time to fix it.

jjgomezcadenas:
What about tomorrow 4:30  pm then (I have a rendez-vous already at 3:30) ?

jacg:
I'll try to have a look this evening, and give some feedback, if appropriate.

jacg:
Why does the word 'parallel' feature in the title of this PR?

abotas:
I wrote in parallel to express that we have not deleted the old pmaps implementation, but maybe that's not an important piece of information.

jerenner:
I can make tomorrow at 4:30pm

jahernando:
fine for me, Wednesday 29 at 16:30.

gonzaponte:
Actually, the implementation is just 4 commits long (those starting with "Implement" + [Replace...](https://github.com/nextic/IC/pull/406/commits/2e41b8b0dfada853cbeb63883508660323eaa3c0)). The other 33 commits are pure tests (and a couple of functions used just by them). We have opted to commit each test separately, but they can be squashed if that is considered to be best. Do you still prefer to have the presentation? No problem with that, just making sure you know is not a huge PR.

jjgomezcadenas:
The reason why I want the presentation is more to get every one in the team acquainted with an important piece of work and as an excuse to have everybody thinking proactively. It is extending the PERAS concept to Big Brothers Band. Just a bit of fund for everybody. If agreed I will send an e-mail around. 

(interested to hear the opinion of the master on BBB reviewing)

jacg:
I think that as many of you as possible getting together to discuss the new interface and its implementation, is a good idea. This *is* an important and central piece of work.

I would also invite @andLaing to the party. And Carmen (sorry, still don't know the GitHub id).

I regret that I will not be able to join you.

jacg:
+ I appreciate where you're coming from, but I find the double underscore in `test__Peak_*` a bit much.

+ The first three commits have titles which claim to add generators, but generator has a quite specific technical meaning in Python, and it's not the things being added in those commits. The things being added also have a very specific technical name: *fixtures*. Please reword these.

+ By the end of *The Big Refactor*, `new_pmaps` will have to be renamed to `pmaps`.

+ Isn't `one_of(just(S1), just(S2))` a long-winded way of writing `sampled_from((S1, S2))` ?

+ Prefer `pytest.approx` to `nupmy.allclose` etc.

+ Please squash out commits which fix mistakes in prevous commits in this PR, such as 'Demand absolute equality in new pmaps tests (not allclose)'.

jjgomezcadenas:
>Prefer pytest.approx to nupmy.allclose etc.

Why? I thought numpy was as kosher as it gets.

gonzaponte:
> I appreciate where you're coming from, but I find the double underscore in test__Peak_* a bit much.

I agree.

> The first three commits have titles which claim to add generators, but generator has a quite specific technical meaning in Python, and it's not the things being added in those commits. The things being added also have a very specific technical name: fixtures.

Indeed.

> By the end of The Big Refactor, new_pmaps will have to be renamed to pmaps.

That is the plan, yes.

> Isn't one_of(just(S1), just(S2)) a long-winded way of writing sampled_from((S1, S2)) ?

Probably, I wasn't aware of the function.

> Prefer pytest.approx to nupmy.allclose etc.

Ok.

> Please squash out commits which fix mistakes in prevous commits in this PR, such as 'Demand absolute equality in new pmaps tests (not allclose)'.

We were leaving that as "we have noticed that this is the right way to do it" and leaving the trace of our decision. Unless you say otherwise, we will do it tomorrow.

jacg:
> I thought numpy was as kosher as it gets.

Numpy is as kosher as it gets for doing efficient array operations. Pytest is as kosher as it gets for testing in pytest.

`pytest.approx` gives us a single function which covers most use cases and is best integrated into pytest, and is going to give you the best error reporting.

jacg:
> We were leaving that as "we have noticed that this is the right way to do it" and leaving the trace of our decision. Unless you say otherwise, we will do it tomorrow.

If you think that this change of mind is an important part of the narrative, then leave it in. The other side of the argument is that we want the commits to record, as cleanly as possible, what there *is*, rather than the meandering path we took to get there.

I would vote for getting rid of the fix commits, but I would accept keeping them in.

jacg:
>> Isn't `one_of(just(S1), just(S2))` a long-winded way of writing `sampled_from((S1, S2))` ?

> Probably, I wasn't aware of the function.

I'll let you into a little secret: neither was I ... but the original was so ugly that I thought *there **must** be a better way!*

jacg:
After sleeping on it, I am less convinced by the multiple micro-commits. I think I would prefer one commit each for

+ the fixtures
+ the tests (this one could be split into one commit per class being tested)
+ the implementations (again, maybe split into one per class)

Maybe the last two should be ordered thus:

1. `SensorResponse` tests
2. `SensorResponse` implementation
3. `_Peak` tests
4. `_Peak` implementations
5. `PMap` tests
6. `PMap` implementations

That way the building blocks can be considered in isolation and before the higher-level components built upon them.

jacg:
The s1s and s2s should be stored as tuples, not lists, in `PMap`.

jacg:
I would argue that the argument presented here

>> You appear to have removed the ability to get the nth peak altogether, and your comment above suggests that you still wanted to be able get the nth peak of your desired type.

> Yes, but s1s and s2s are intended to return indexable sequences. Therefore we can access the n-th S1 peak as `pmap.s1s[index]`. We (@abotas, @jahernando and I) prefer this rather than `pmap.get_s1_peak(index)` because it is equally informative and simpler.

applies equally to `pmap.number_of_s1s` being replaced by `len(pmap.s1s)`

gonzaponte:
> After sleeping on it, I am less convinced by the multiple micro-commits

Me too. It has been quite tedious to reorganize commits and do that sort of stuff. I agree with the scheme you propose.

> The s1s and s2s should be stored as tuples, not lists, in PMap.

Indeed.

> applies equally to pmap.number_of_s1s being replaced by len(pmap.s1s)

Yes, we were discussing whether to leave the property or not. We thought it might be better for some cases, just to be more verbose. We don't have a strong opinion here.

jahernando:
In agree with @jacg that if we do not grant access via explicit method to the S1 and S2 peaks, `map.get_s1_peak(index)`, it it is not coherent to have a method to provide the number of S1s or S1s. I will simply go for `len(pmap.s1s)`

gonzaponte:
> Prefer pytest.approx to nupmy.allclose etc.

Should we define `exactly = partial(approx, rel=0, abs=0)` for testing absolute equality (particularly for numpy arrays)?

jacg:
What is the status of this? How did the meeting go? Is this ready to be approved and merged?

jacg:
Please get rid of all the `_type = "<name of the surrounding class>" lines. Use `self.\_\_class__.\_\_name__` instead (with a property defined in the base class to make reading more convenient, if you really must), but please don't use strings to represent types: types are objects in Python, and are therefore perfectly capable of representing themselves. And they know their own names, so you don't need to tell them what their name is!

jacg:
`_Peak.__str__` looks very wrong!

First of all, repeated `string +=` is a classic Python anti-pattern. Secondly, Python's triple-quoted strings exist to make this sort of thing easy (all those manual `\n`s are awful). Thirdly, use the power of `str.format` to avoid repetition of `self` etc.

I suggest something like this


    def __str__(self):
        n_samples = len(self.times)
        return """---------------------
    {__class__.__name__} instance
    ---------------------
    Number of samples: {n_samples}
    Times: {times / mus} µs
    Time @ max energy: {time_at_max_energy / mus}
    Width: {width / mus} µs
    Height: {height} pes
    Energy: {total_energy} pes
    Charge: {total_charge} pes
    RMS: {rms / mus} µs
    """.format(**{**vars(self), **vars(units), locals()})


I **do** like the unicode mu.

gonzaponte:
In the meeting we agreed on that we should have the equivalent of `total_energy` for the tracking plane, which is `total_charge`. For symmetry, we have also added the `total_charge_above_threshold` method. We (@abotas and I) later decided that this repetition could be better implemented if we added a `where_avode_threshold` method to the `SensorResponses` class.

Also, we have created the descriptive classes: `PMTResponses` and `SiPMResponses` that derive from `_SensorResponses`.

On top of that, we added some safety checks to the `_SensorResponses` and `_Peak` classes so they cannot be initialized with wrong inputs.

> Please get rid of all the _type [...] lines. [...] And they know their own names, so you don't need to tell them what their name is!

Yes. Thanks.

> First of all, repeated string += is a classic Python anti-pattern. Secondly, Python's triple-quoted strings exist to make this sort of thing easy (all those manual \ns are awful).

Actually I don't like the 's +=' pattern I used either, but I do find the code display more pleaseant. The reason I did this is because I dislike how the dedented string looks in the code and I think the `dedent` function in `textwrap` doesn't work for all cases, but I'll check.

> Thirdly, use the power of str.format to avoid repetition of self etc.

That is ok, but I think your proposal doesn't work as it doesn't execute the code within the curly brackets.

jacg:
I appreciate that it's difficult to write automated tests for `__str__`, but is there something that can quickly generate interesting instances of `_Peak` etc, so that I can at least *look* at some output without having to create instances by hand?

gonzaponte:
> is there something that can quickly generate interesting instances of _Peak etc, so that I can at least look at some output without having to create instances by hand?

Isn't the fixture `peaks` in `new_pmaps_test.py` what you are looking for?

jacg:
Yes, that's exactly what I'm looknig for. Thanks.

jacg:
> Should we define `exactly = partial(approx, rel=0, abs=0)` for testing absolute equality (particularly for numpy arrays)?

An interesting idea. I like it ... but it feels a bit weird, and I'm not sure why. I suggest we try it and see if it is useful and if it causes any trouble.

jacg:
The `sensor_responses` strategy generates waveforms in which the same float value is repeated over and over again, with a very small number of entries being different. It is not clear to me

+ whether this is intended and desirable
+ exactly what in the code makes it happen

gonzaponte:
> whether this is intended and desirable

Not intended **by us**. It is as desirable as any other, I think.

> exactly what in the code makes it happen

That is hypothesis heuristics, isn't it?

jacg:
If hypothesis documents this behaviour, then I'm not seeing where.

jacg:
It has come to my attention that a class which has an explicit `__repr__` but no explicit `__str__` has the `__str__` inherit the behaviour of `__repr__` by default.

That means that we can just implement the `__repr__`s and not bother with the `__repr__ = __str__` trick. I'll make this change.

But don't we want a distinct, more succinct implementation of `__str__`. Something that maybe just reports the type, maybe including one or two key numbers, such as total energy and or charge?

gonzaponte:
> If hypothesis documents this behaviour, then I'm not seeing where.

This is the numpy part, so it should be [here](http://hypothesis.readthedocs.io/en/latest/numpy.html?highlight=arrays):

> Array values are generated in two parts:
>Some subset of the coordinates of the array are populated with a value drawn from the elements strategy (or its inferred form).
> If any coordinates were not assigned in the previous step, a single value is drawn from the fill strategy and is assigned to all remaining places.

gonzaponte:
> But don't we want a distinct, more succinct implementation of __str__. Something that maybe just reports the type, maybe including one or two key numbers, such as total energy and or charge?

I don't know. I think we will just use this string representations for debugging and therefore they will need to be as informative as possible. I think @abotas wanted to do something like yours as well.

jacg:
OK, the hypothesis array generator makes sense.

I don't see why you say that you will use str representations for debugging. The motto is 'repr is for the developer, str is for the client'.

You want some succinct and not completely overwhelming means of viewing information, and that is what `__str__` is for.

gonzaponte:
> I don't see why you say that you will use str representations for debugging.

I mean that *in general*, we are not interested in printing Peak objects unless we are debugging something. But I may be wrong. Anyway, I would only hide the time array for the "short version".

jacg:
In that case, I suggest that we go with implementing just `__repr__` and letting `__str__` pick up that implementation ... then we can add a more succinct `__str__` if it turns out to be useful.

As for me doing it ... as there are no tests on which I can rely, I'll pass: it's best that the people who wrote it and know what they expect to find in the output should check it.

It would be better if the `s += ...` lines were replaced with a `'\n'.join(<list of collected strings>)` and *even* better if we used a triple quoted string. I won't believe that `textrwap.dedent` doesn't work, without seeing it with my own eyes.

I can live with the `s += ...` stuff for a little longer, but the `type_ = ...` must go ASAP.

gonzaponte:
I agree with everything, but this will be postponed for a few hours. I have a lot to do at the moment.

gonzaponte:
> > Should we define exactly = partial(approx, rel=0, abs=0) for testing absolute equality (particularly for numpy arrays)?

> An interesting idea. I like it ... but it feels a bit weird, and I'm not sure why. I suggest we try it and see if it is useful and if it causes any trouble.

It feels a bit weird to include it in this PR. Should I open a micro-PR adding this definition to `core/testing_utils.py` or simply include it here?

jacg:
If you feel you have the time and strength for a micro-PR, then do that. If not, sneak it in here, and I'll forgive you.

gonzaponte:
>  I won't believe that textrwap.dedent doesn't work, without seeing it with my own eyes.

Well, I am afraid it doesn't work because of numpy arrays' string representation:

```
import numpy as np
from textwrap import dedent
array = np.random.rand(50)
print(dedent("""
    4 spaces to be dedented. This should be on the left margin.
    However, because I put this array: {}
    it doesn't work as expected
    """.format(array)))
```

gonzaponte:
> If you feel you have the time and strength for a micro-PR, then do that. If not, sneak it in here, and I'll forgive you.

Done in #413.

gonzaponte:
> Well, I am afraid it doesn't work because of numpy arrays' string representation:

It works when converted to lists, but then we have 16-digit-numbers which are quite difficult to read.

jacg (approved):
Clear, well-tested and desperately needed.

gonzaponte:
Sorry for the noise. I have rebased on top of master and reworded some commits.

PR #413, opened 2017-12-05
Add function for comparing exact values

gonzaponte:
This function is intended to have similar behavior to `pytest`'s
approx function, which compares floating point numbers, but for
integers. This is particularly interesting when comparing numpy
arrays, as the expression is a bit tedious:

```
np.all(array1 == array2)
```

to be compared with

```
array1 == exactly(array2)
```

which is more pleasant to the human eye and keeps the symmetry with
floating point comparisons.

### DIFF
--- a/invisible_cities/core/testing_utils.py
+++ b/invisible_cities/core/testing_utils.py
@@ -1,8 +1,34 @@
 import numpy  as np
 
+from pytest import approx
+
 from numpy.testing          import assert_array_equal, assert_allclose
 from hypothesis.strategies  import integers, floats
 from hypothesis.extra.numpy import arrays
+
+
+def exactly(value, **kwargs):
+    """
+    Assert that two numbers (or two sets of numbers) are equal to each
+    other with zero tolerance. This function is intended to have a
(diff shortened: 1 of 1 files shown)


jacg:
I would be partial (excuse the awful pun) to **not** using `partial` here: `partial` is great for quickly adapting something to a specific situation in-line. In this case, I think it would be better to write a function with a helpful docstring, given that we are creating a tool that we expect to be reused.

gonzaponte:
Ok. The docstring makes sense, but I think implementing the function with `partial` still makes sense, doesn't it?

jacg:
Without `partial` it would look something like this

    def exactly(value):
        "docstring"
        return approx(value, abs=0, rel=0)

and with `partial` it would look something like this

    def exactly(value):
        "docstring"
        return partial(approx, abs=0, rel=0)(value)


I don't see how `partial` helps. What am I missing?

gonzaponte:
Nothing, you are right! I got confused.

abotas (approved):
Straightforward and useful. Approved.

PR #415, opened 2017-12-07
Add mean XY of the peak to hDST

Aretno:
Added the Xpeak and Ypeak columns to Penthesilea hits (hDST). These are the XY mean position of the peak and is added to the Hit class as an XY object.

For calculating the position a function called `compute_xy_peak_position` has been written. This function is almost equal to `compute_xy_position` from the kDSTs with the exception that the configuration for corona is hardcoded to be the general configuration (to avoid using the corona config used for the hit-building). In case it fails, the mean position is set to be 0., 0., as with NN hits.

To make the code backwards compatible the getattr built-in function is used, in case the hDST does not have the XYpeak fields, the hits will be loaded using a default value of -1000 for those fields.


gonzaponte (approved):
This PR extends Penthesilea to include peak magnitudes to the output. The test suite has been accommodated to check the values of the new variables. Reviewed with peras.

Good job!

PR #418, opened 2018-01-09
Event Display

coreyjadams:
Here is the first draft of the pull request.  Let's not let perfection stop progress, so I submit this knowing you will find things to change.

I describe here the basics of the visualization tool:

bin/view-ic.py is a python executable that runs and launches the interactive viewer.

invisible_cities/viewer/gui is a package that contains the classes that interface with the screen and user directly: there is a widget for the 3D open GL view, a manager class for that, and a GUI for the interaction buttons

invisible_cities/viewer/manager is a package that handles interface with the invisible_cities data extraction and the user's actions, as well as the detector geometry information

invisible_cities/viewer/datatypes is a package that handles the details of drawing objects in the view.  In principle, this can be trivially extended to handle more objects, but I am still learning the ropes about the reconstruction event model and have only implemented MCHits, MCTracks, and PMaps


coreyjadams:
Additionally, there is not yet any incorporation with travis CI.  Since there are no changes to IC except additional python scripts for the viewer, I anticipate no change in other tests.  The viewer does not yet have tests of it's own.

jacg:
@coreyjadams Please submit PRs from feature branches, rather than `master`.

jacg:
@coreyjadams Please train your editor to eliminate trailing whitespace from files when it saves them, or ensure in some other way that trailing whitespace is removed before you commit files.

I'll take care of cleaning it all up in this PR.

jacg:
@coreyjadams The program crashes until the `pyqtgraph` and `pyopengl` packages are added to the conda environment. I will add these to the `.yml` specification in `manage.sh`. (Unfortunately this will require me to modify your `master` branch ... which is why you should submit PRs on feature branches)

When I execute `bin/view-ic.py` as suggested above, I get

    Traceback (most recent call last):
      File "bin/view-ic.py", line 55, in <module>
        main()
      File "bin/view-ic.py", line 36, in main
        manager = EvdManager3D(args.file)
      File "/home/jacek/src/IC-worktree/invisible_cities/viewer/manager/EvdManager3D.py", line 28, in __init__
        self.init_manager(_file[0])
    IndexError: list index out of range

Please improve this error report. I have no idea what file is required.

Once this code runs reliably out of the box, we can probably approve it quickly and worry about the technical debt later.

(Do you think that you might have some tests before the end of the week? If so, we'll wait until they are ready. If not, they can follow in a separate PR later.)

coreyjadams:
These are useful comments, thanks.

Pull requests from feature branches - sure.  Remove trailing whitespace - sure.  I can handle both of those in the future, thanks for kind reminders.

I know exactly what this error means, and it's no issue to make this more informative.  For your curiosity, it's expecting an input file on the command line.  I thought there was a check for this, but clearly not.  I am open to discussing configuration of the viewer in this sense.  In other experiments, I have made "change-event-on-the-fly-viewer"s where the reconstruction path can run *before* the viewer renders anything, to allow quick visualization of reconstruction algorithms.  This would require some changes here, but the framework interface is sufficiently divorced from the user interface to make this relatively easy.

Regarding pyqtgraph, PyQt (and implicitly Qt itself), and pyopengl - yes, they are required.  Are there tests on dependency fulfillment?  Additionally, before approving this pull request we should verify it works on Mac OSX.  Qt support has varied significantly in the past between linux and mac.  The viewer does support both PyQt4 and PyQt5 but if we are focused on just one (PyQt5 likely) I will adjust some code to catch differences.

And regarding tests: I hope to implement a few tomorrow, if just on the pieces that directly read pmaps.  I know the pmap model is changing, so I think implementing those tests and waiting for the updated model (and likely revising this code to fit) can be prerequisites for this approval.

(ps I am happy with the names "viewer" for the package and "view-ic.py" for the executable, but if there is a need to change it let me know.)

coreyjadams:
A comment on the dependencies:  I tried a fresh install of IC to see if it would work "out of the box" and noticed  that the version of pyqt installed by conda doesn't work with open gl.  There is a workaround that I found here that solved the problem:
https://github.com/ContinuumIO/anaconda-issues/issues/1267

Which involves installing PyQt5 and sip from pip instead of conda.

jacg:
Could you give me an example of a *specific* input file that can be used as the command line argument to `view-ic.py`, preferably one already present in our repository; failing that, one that I can download from somewhere else.

At the moment I'm getting crashes which have nothing to do with pyqt/opengl etc but which seem to be related to parsing either the filename or the contents of the file. As I have no time to spare, I'd like to make sure that I am using an adequate input file, before anything else.

PR #421, opened 2018-01-10
Small fixes in fit functions

bpalmeiro:
These are just small changes that remained from PR #359:
    - Some Cosmetics.
    - Removing a non-zero selection in fit function, no longer necessary.
    - Adding a test that demonstrates that invalid values for sigma were accepted.
    - Correcting correspondingly the code for checking invalid values in sigma.

### DIFF
--- a/invisible_cities/core/fit_functions.py
+++ b/invisible_cities/core/fit_functions.py
@@ -130,28 +130,22 @@ def fit(func, x, y, seed=(), fit_range=None, **kwargs):
             kwargs["sigma"] = kwargs["sigma"][sel]
 
     sigma_r = kwargs.get("sigma", np.ones_like(y))
+    if np.any(sigma_r <= 0):
+        raise ValueError("Zero or negative value found in argument sigma. "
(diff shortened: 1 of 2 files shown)


jahernando:
The PR fix an issue with the fit functions related of how to handle the fit when the client provides null errores. If this case happens, now the function throws an exception. The PR incorporates a test. 

If will be nice to incorporate a comment in the doc-string suggesting the user to provide Poisson's errors when using histogram contents for the fit.

jahernando (approved):
The PR solves an issue related with the treatment of the data when the user sends null errors to the fit function. Now it raises an exception.

The PR includes a test.

bpalmeiro:
We should clean the commit history and some cosmetics before merging

bpalmeiro:
Done! It's ready to be "approved " and merged.

gonzaponte:
@jahernando, please check that you're still happy with this PR so it can be merged.

jahernando:
Yes, please merge, I understood that you do the merge!

PR #423, opened 2018-01-11
Switchable voxel size strictness

paolafer:
Originally, `voxelize_hits` created voxels slightly smaller than the required voxel size.
We have added the option of using voxels of strict size. It seems that this worsen the results of 2-blob discrimination for the cases we have tried, but we want to keep the flexibility for future investigation.

Fixes #396.


jacg:
`test_voxelize_hits_respects_voxel_dimensions` is failing ... and I don't have time to look into that right now.

paolafer:
It was failing because we were using variable voxel dimensions (remember, we change the default value for the strictness parameter). The last commit should fix the problem.

jjgomezcadenas (approved):
Simple and clear. Good job.

PR #424, opened 2018-01-12
Eliminate misleading dependence on voxel size from functions

paolafer:
I've realized that we kept the explicit dependence on voxel size in make_track_graph function, when we added the size property to Voxel. This PR fixes this.

### DIFF
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -174,7 +174,8 @@ def make_tracks(evt_number       : float,
                 blob_radius      : float = 30 * units.mm) -> TrackCollection:
     """Make a track collection."""
     tc = TrackCollection(evt_number, evt_time) # type: TrackCollection
-    track_graphs = make_track_graphs(voxels, voxel_dimensions) # type: Sequence[Graph]
+    track_graphs = make_track_graphs(voxels) # type: Sequence[Graph]
+#    track_graphs = make_track_graphs(voxels, voxel_dimensions) # type: Sequence[Graph]
     for trk in track_graphs:
         a, b, voxels_a, voxels_b    = compute_blobs(trk, blob_radius)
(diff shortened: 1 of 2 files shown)


jacg:
This would have been a nice exercise for @carmenromo ...

paolafer:
She did the same for the merge_tracks function.

jjgomezcadenas (approved):
Simple and clear.

gonzaponte:
@paolafer After merging #423 this branch has conflicts. Solve them, please.

jacg:
> this branch has conflicts

They look pretty simple. If you could work through the process with @carmenromo, that would be good.

paolafer:
I’m traveling right now and I won’t be back until Sunday. I don’t know if a Skype-peras-conflict-solving is worth the effort in this case...

jacg:
No, no point in doing it over Skype, this needs genuine peras (I could have done it myself, if that were an option). If it's urgent to get this done then just do it yourself. If it can wait until you get back, then I would recommend waiting. As the conflict is tiny, it's easy to understand in its entirity. As @carmenromo is familiar with this code, it should make some sort of sense, rather than being an abstract exercise, and it would serve as a good introduction to both conflict resolution and rebasing code within a PR.

In short, it's a good didactic opportunity which we should take advantage of, if we can afford to.

paolafer:
I understood that @jjgomezcadenas wanted the more PR's approved as soon as possible, correct me if I'm wrong. @carmenromo and myself are working on the merging algorithm, which will offer opportunities of learning about rebasing and (maybe) conflict resolution, too.

jacg:
Repetition and good guidance are vital to learning. I doubt that this PR is holding anything else back. So I would recommend using it as a training opportunity. However, if you or @jjgomezcadenas think otherwise, then just go ahead and merge it.

PR #425, opened 2018-01-13
Add tests for the BLR algorithm

gonzaponte:
Somehow the module `sierpe/blr.pyx` has managed to escape the master's radar until now. This PR introduces some tests to this code, which belongs to the core of our processing chain. There are two types of tests:

- Property checking: by using a toy signal, we check that the integral is positive and that the baseline is recovered.

- Stability: by using some real waveforms we store the current value of the blr waveform in order to have something to compare with in the future. This assumes we trust the current algorithm.

### DIFF
--- a/invisible_cities/cities/base_cities.py
+++ b/invisible_cities/cities/base_cities.py
@@ -428,16 +428,16 @@ class DeconvolutionCity(RawCity):
        Since a Deconvolution city reads RWF, it is also a RawCity.
     """
 
-    parameters = tuple("""raw_data_type n_baseline thr_trigger acum_discharge_length""".split())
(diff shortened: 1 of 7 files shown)


gonzaponte:
I would like to read opinions about everything but in particular about the comments in the tests named `*ad_hoc_signals*`.

jjgomezcadenas (approved):
Very good. There was plenty of tests and studies for BLR but they never made it to IC. Good work!

gonzaponte:
Don't merge, this is still work in progress.

jjgomezcadenas:
Actually, I think no more tests are needed for this function. Please wrap up and let's move to something else.

gonzaponte:
There is one more test to be included which should verify the filtering behavior of `deconv_pmt`. Tomorrow.

PR #426, opened 2018-01-15
Event model interface and implementation

gonzaponte:
This PR implements the new event model discussed in the sprint week.


gonzaponte:
Hypothesis complains about the `tracks` composite taking too long to generate data. However, it doesn't seem like a huge amount of data to generate. Ideas? If we don't find any other solution, we will have to manually tell the test that data generation is slow.

jjgomezcadenas:
Tests are failing here. Why?

gonzaponte:
> Tests are failing here. Why?

Because
> Hypothesis complains about the `tracks` composite taking too long to generate data

jjgomezcadenas:
Or, skip hypothesis for this case. One way or the other we need to fix it quickly.

gonzaponte:
Event model agreed at the sprint week meeting:
event_model_diagram_v0

gonzaponte:
I propose the following one:
event_model_change

The main changes are:
- Store methods out of the event model (something I think we already agreed on).
- Few extra classes: `xy` and `xyz`, which are handy.
- `Cluster` is now a specialization of `xy`.
- The attribute `seed` in `blob` is an instance of `xyz`.
- `Hit` is now a specialization of both `xyzE` and `Cluster` rather than something that contains a `Cluster`. This is the main thing, as it removes the duplicity of the values `x` and `y` and it is the only difference I haven't implemented yet.

The last point is an important one. Do we want to keep this duplicity? That means adding 2 extra columns to the table we store in disk. I am not sure we need/want that.

jjgomezcadenas:
I like the proposal. Here some comments:

1.  Class xy already exists (look in ic_types.py, in event_model).
2. Since all classes are transient, you don't need that Hit collection inherits from Event (nor TrackCollection). You will need to **save** event number and time stamp when you write hit collections/track collections, so that you can handle the transformation from tables to per-event-objects, but since you are decoupling the writers from the transient classes, you can let the writers/readers figure that thing out. This means that the writer will have access to Event and Hit Collection and store a HitCOllectionTable that has event number and time stamp as columns, but this HitCollectionTrack is not (or not yet) in the event model and is only used to write/read.
3) Actually, I think we need to specify also the PERSISTENT classes in the event model. For example, given a HitCollection, you are going to need a HitCollectionTable (that includes event number and timestamp). HitCollectionTable is actually an aggregate of HitCollection plus some store() method. Some for TrackCollectionTable. 
4) Notice that behind HitCollectionTable we will still have a pure nh5 object, but I would add the persistent class an extra decoupling layer. If we ever change from nh5 to something else (say Spark objects) the nh5 imp will change but this will be achieved by changing the store method and the writers.

gonzaponte:
1. Yes, thanks.
2. Ok, I don't have a problem with that.
3. I don't think we should specify the persistent classes, because the user doesn't care about persistent objects as long as you provide them with de relevant IO functions. If we decide at some point to change the persistent representation, the user doesn't need to know.
4. In line with the previous point, I would actually decouple the event model from the IO functions completely. I would define the event model, certify that this is what we want and then code the IO operations ***as separate functions***. This way, if you change your persistent representation, that has no impact at all in the event model. I think the only difference between what you say and what I say is just the highlighted part, right?

jjgomezcadenas:
>I don't think we should specify the persistent classes, because the user doesn't care about persistent objects as long as you provide them with de relevant IO functions. If we decide at some point to change the persistent representation, the user doesn't need to know.

Actually, if we use persistent classes I think we must. The event model does not concern only "the user", it specifies everything related with the event model. If we are going to write a class for persistency, then we must "declare it" in the event model.

By persistent classes I mean classes that know how to write themselves (via a store()) method. 

We may choose not to use such classes, though. One could simply write functions that "consume" transient classes and "produce" pytable classes which are those that we actually write (we still need pytable classes which constitute a sort of "persistent model" anyway, but they are of no interest to the user). I am OK also with this approach, it removes one layer and puts all the "interface" in the writer/reader classes. 

So yes, I think we can converge in the approach of eliminating altogether persistent classes and write functions that consume transient classes and produce pytable classes (then we need functions that actually write/read those pytable classes, not clear to me if we want to do all the operations in the same function). 

In this approach, though, HitCollection (TrackCollection) do not need to inherit from Event, the "persistent functions" can consume a HitCOllection AND an Event transient classes and produce a pytable object. Agree?

gonzaponte:
> Agree?

Yes! 🙂

jjgomezcadenas:
OK, if no one complains loud (speak now or keep your peace), plane go ahed, *fix* the hypothesis problem (hack if you must) and implement the EVM. I suggest we review and approve first the EVM, *without* the writers, then open a new PR for the writers. Agreed?

gonzaponte:
> I suggest we review and approve first the EVM, without the writers, then open a new PR for the writers. Agreed?

Absolutely. That was my plan, actually.

paolafer:
> `Hit` is now a specialization of both `xyzE` and `Cluster` rather than something that contains a `Cluster`.

I'm a bit confused about this: where does the Hit object takes x and y from, then? Doesn't it have two different values anyway?

gonzaponte:
> I'm a bit confused about this: where does the Hit object takes x and y from, then? Doesn't it have two different values anyway?

From both, but there is no ambiguity: it is the same, and even if it is not, one overwrites the other and the order of inheritance or the `__init__` implementation determines which one is actually kept.

gonzaponte:
Here is the new proposal without multiple inheritance:
event_model_gonzalo_mod_v2

Issue #387, opened 2017-11-02
Masked channels


paolafer:
During today's S&A meeting I've realized that we are not taking into account masked channels in the corona algorithm. They matter, because we ask the sensors within corona be more than a fixed number. If one of these sensors is masked, this number should decrease by 1 (this is how we had  corona implemented in art). I can take care of this, but I wanted to raise the issue, first, because this change implies using database information on xy_algorithm functions, and I don't know if IC ethics allows that...

paolafer:
Maybe we could keep this information outside xy_algorithm, just changing the corona parameter nsipm in the place where the algorithm is called (HitCity, maybe?).

gonzaponte:
>  this change implies using database information on xy_algorithm functions

The functions can take an extra argument with the masked sensors or something like that. I think we can and should avoid introducing the `database` dependence in `xy_algorithms`

gonzaponte:
> Maybe we could keep this information outside xy_algorithm, just changing the corona parameter nsipm in the place where the algorithm is called (HitCity, maybe?).

I have just taken a look at the code and this seems a reasonable option. However, I don't see yet how are going to take this into account, because we don't know if a masked channel is around the signal region or not. And even if we know, we can't know whether it would have had signal or not.

paolafer:
Right, you need to know which one is the hottest SiPM, to know its neighbours...

gonzaponte:
Ok, I have seen the C++ implementation as well. I like your proposal. Are you going implement it?

paolafer:
Yes, but I think that we *need* to make xy_algorithm dependent on database. Is that ok?

gonzaponte:
I don't see the need for that. As you mentioned earlier...

> we could keep this information outside xy_algorithm, just changing the corona parameter nsipm in the place where the algorithm is called

jjgomezcadenas:
This is not a problem, we use the data base all the time and have a local copy of the data base at our disposal.

gonzaponte:
*Conceptually*, `corona` does not need to know about the database. The function is designed to be "abstract" in the way that it takes x-y positions and charges and produces an invariant result. This encapsulation makes our life easy in many ways (testing is a good example of that). If you need to handle special cases, do it separately and pass "pure" information to your algorithm. If we begin to introduce special cases within the function, it loses its generality. On top of that, the database is a mutable structure and should not be used in something as "pure" as a reconstruction algorithm, I think.

paolafer:
[...]

PR #430, opened 2018-01-17
Take into account masked channels in corona

paolafer:
This PR addresses issue #387, making possible to take into account the presence of masked SiPMs in the corona algorithm, when calculating the minimum number of SiPMs within the radius. 
Specifically, if N masked SiPMs are found, the minimum number of SiPMs required becomes msipm - N.

### DIFF
--- a/invisible_cities/reco/xy_algorithm_test.py
+++ b/invisible_cities/reco/xy_algorithm_test.py
@@ -20,7 +20,10 @@ from .       xy_algorithms     import corona
 from .       xy_algorithms     import barycenter
 from .       xy_algorithms     import discard_sipms
 from .       xy_algorithms     import get_nearby_sipm_inds
-from .. core.exceptions        import SipmEmptyList
+from .       xy_algorithms     import get_neighbours
+from .       xy_algorithms     import have_same_position_in_space
+from .       xy_algorithms     import is_masked
+
 
 @composite
 def positions_and_qs(draw, min_value=0, max_value=100):
@@ -38,13 +41,16 @@ def test_barycenter(p_q):
     assert np. isclose(B.Q  , qs.sum())
     assert B.nsipm == len(qs)
 
+
 def test_barycenter_raises_sipm_empty_list():
     with raises(SipmEmptyList):
         barycenter(np.array([]), None)
 
+
 def test_barycenter_raises_sipm_zero_charge():
     with raises(SipmZeroCharge):
         barycenter(np.array([1,2]), np.array([0,0]))
+
 
 @fixture
 def toy_sipm_signal():
@@ -53,6 +59,7 @@ def toy_sipm_signal():
     qs = np.array([ 6,   5]) * units.pes
     pos = np.stack((xs, ys), axis=1)
     return pos, qs
+
 
 def test_corona_barycenter_are_same_with_one_cluster(toy_sipm_signal):
     pos, qs = toy_sipm_signal
@@ -79,6 +86,7 @@ def test_corona_barycenter_are_same_with_one_cluster(toy_sipm_signal):
     assert c_cluster.R          == b_cluster.R
     assert c_cluster.Phi        == b_cluster.Phi
 
+
 def test_corona_multiple_clusters(toy_sipm_signal):
     """notice: cluster.xy =(x,y)
                cluster.posxy = ([x],
@@ -90,6 +98,7 @@ def test_corona_multiple_clusters(toy_sipm_signal):
     for i in range(len(pos)):
         assert np.array_equal(clusters[i].XY, pos[i])
         assert clusters[i].Q == qs[i]
+
 
 def test_corona_min_threshold_Qthr():
     """notice: cluster.XY =(x,y)
@@ -113,6 +122,7 @@ def test_corona_min_threshold_Qthr():
     assert clusters[0].Q ==  99
     assert clusters[0].XY == (990, 0)
 
+
 def test_corona_msipm(toy_sipm_signal):
     pos, qs = toy_sipm_signal
     # this is gonna raise and error:
@@ -123,6 +133,7 @@ def test_corona_msipm(toy_sipm_signal):
         pass
     except ClusterEmptyList:
         pass
+
 
 @parametrize(' Qlm,    rmax, nclusters',
              ((6.1,      15, 0),
@@ -139,6 +150,8 @@ def test_corona_simple_examples(toy_sipm_signal, Qlm, rmax, nclusters):
         pass
     except ClusterEmptyList:
         pass
+
+
 @fixture
 def toy_sipm_signal_and_inds():
     k = 10000
@@ -149,6 +162,7 @@ def toy_sipm_signal_and_inds():
     i  = np.array([0, 5, 1000, 9999])
     return k, i, pos, qs
 
+
 def test_discard_sipms(toy_sipm_signal_and_inds):
     k, i, pos, qs = toy_sipm_signal_and_inds
     xysel, qsel = discard_sipms(i, pos, qs)
@@ -157,6 +171,7 @@ def test_discard_sipms(toy_sipm_signal_and_inds):
         assert ind         not in xsel
         assert ind +     k not in ysel
         assert ind + 2 * k not in qsel
+
 
 def test_get_nearby_sipm_inds():
     xs  = np.array([0,1,2,3,4,0,1,2,3,4,0,1,2,3,4])
@@ -172,6 +187,34 @@ def test_get_nearby_sipm_inds():
             assert np.sqrt((xs[i] - xc)**2 + (ys[i] - yc)**2) <= d
         else:
             assert np.sqrt((xs[i] - xc)**2 + (ys[i] - yc)**2) >  d
+
+
+def test_get_neighbours():
+    pos = np.array([(35, 55)])
+
+    exp_xs = np.array([35, 35, 35, 25, 25, 25, 45, 45, 45])
+    exp_ys = np.array([55, 65, 45, 55, 65, 45, 55, 65, 45])
+    expected_neighbours = np.stack((exp_xs, exp_ys), axis=1)
+
+    found_neighbours = get_neighbours(pos, pitch = 10. * units.mm)
+
+    number_of_sipm_found_correctly = 0
+    for found in found_neighbours:
+        assert any(have_same_position_in_space(found, expected) for expected in expected_neighbours)
+        number_of_sipm_found_correctly += 1
+
+    assert number_of_sipm_found_correctly == 9
+
+
+def test_is_masked():
+    pos_masked = np.array([(0, 2),
+                           (2, 1)])
+
+    sipm_masked = [(0, 2)]
+    sipm_alive  = [(3, 2)]
+
+    assert     is_masked(sipm_masked, pos_masked)
+    assert not is_masked(sipm_alive , pos_masked)
 
 
 def test_masked_channels():
--- a/invisible_cities/reco/xy_algorithm_test.py
+++ b/invisible_cities/reco/xy_algorithm_test.py
@@ -172,3 +172,46 @@ def test_get_nearby_sipm_inds():
             assert np.sqrt((xs[i] - xc)**2 + (ys[i] - yc)**2) <= d
         else:
             assert np.sqrt((xs[i] - xc)**2 + (ys[i] - yc)**2) >  d
+
+
+def test_masked_channels():
+    """
+    Scheme of SiPM positions (the numbers are the SiPM charges)
+    1 1 1
+    1 6 1
+    1 1 1
+    1 5 0
+    1 1 1
+    This test is meant to fail if either
+    1) in the case of an empty masked channel list,
+       the actual threshold in the number of SiPMs
+       around the hottest one turns out to be different from msipm
+    2) the masked channel is not taken properly into account
+       by the code
+    """
+    xs         = np.array([0, 0, 0, 1, 1, 1, 2, 2, 0, 0, 1, 1, 2, 2])
+    ys         = np.array([0, 1, 2, 0, 1, 2, 0, 2, 3, 4, 3, 4, 3, 4])
+    qs         = np.array([1, 1, 1, 1, 5, 1, 1, 1, 1, 1, 6, 1, 1, 1])
+    pos        = np.stack((xs, ys), axis=1)
+    masked_pos = np.array([(2, 1)])
+
+    # Corona should return 1 cluster if the masked sipm is taken into account...
+    expected_nclusters = 1
+    found_clusters = corona(pos, qs,
+                            msipm         = 6              ,
+                            Qlm           = 4   * units.pes,
+                            new_lm_radius = 1.5 * units.mm ,
+                            pitch         = 1   * units.mm )
+
+    assert len(found_clusters) == expected_nclusters
+
+    # ... and two when ignored.
+    expected_nclusters = 2
+    found_clusters = corona(pos, qs,
+                            msipm          = 6              ,
+                            Qlm            = 4   * units.pes,
+                            new_lm_radius  = 1.5 * units.mm ,
+                            pitch          = 1   * units.mm ,
+                            masked_sipm    = masked_pos     )
+
+    assert len(found_clusters) == expected_nclusters
--- a/invisible_cities/reco/xy_algorithms.py
+++ b/invisible_cities/reco/xy_algorithms.py
@@ -12,12 +12,17 @@ from .. core.exceptions        import ClusterEmptyList
 from .. types.ic_types         import xy
 from .. evm.event_model        import Cluster
 
+from .. core.system_of_units_c import units
+
+from itertools import product
+
 
 def find_algorithm(algoname):
     if algoname in sys.modules[__name__].__dict__:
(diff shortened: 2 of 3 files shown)


paolafer:
I would like @andLaing (among others) to review this PR, but it seems I cannot suggest him as a reviewer.

invisible_cities/reco/xy_algorithm_test.py
```diff
@@ -172,3 +172,52 @@ def test_get_nearby_sipm_inds():
...
+    pos = np.stack((xs, ys), axis=1)
+
+    nclusters = 2
+
+    xs_m = np.array([2])
+    ys_m = np.array([1])
+
+    masked_pos = np.stack((xs_m, ys_m), axis=1)
+
+    try:
```
gonzaponte: These `try`/`except` clauses introduce a lot of noise. The output of corona is determined unambiguously by its input and it will either run or raise an exception, but the result will be the same every time because the input doesn't change. Since you are comparing the result I assume it will always run so you can jsut simply leave the code within the `try` sections.

I would also rename `nclusters` to `expected_nclusters` (or something similar) and assign the result of `corona` to a verbose name so the comparison `assert len(something) == expected_nclusters` is more obvious.

invisible_cities/reco/xy_algorithms.py
```diff
@@ -149,11 +150,15 @@ def corona(pos, qs,
...
+        masked_neighbours = [x for x in pos[within_new_lm_radius] if x in masked]
 
-        # if there are at least msipms within_new_lm_radius, get the barycenter
-        if len(within_new_lm_radius) >= msipm:
+        # if there are at least msipms within_new_lm_radius, taking
+        # into account any masked channel, get the barycenter
+        if len(within_new_lm_radius) >= msipm - len(masked_neighbours):
             c.extend(barycenter(pos[within_new_lm_radius], qs[within_new_lm_radius]))
 
+        print(len(c))
```
gonzaponte: Remember to remove this print

gonzaponte:
Ideally, the first commit would be split in two: one for the test and another for the fix in corona.

invisible_cities/reco/xy_algorithms.py
paolafer: Thanks for spotting it!

invisible_cities/reco/xy_algorithm_test.py
paolafer: I've used try/except following the other examples of tests, but I can remove them.

andLaing:
Hi

As far as I see algorithmically everything makes sense and is what I remember from the early iterations of the corona algorithm that I worked on with Paola.

2 small and maybe irrelevant comments:
1) in the algorithm with the changes (so from line ~153-158 in xy_algorithms.py) you don't ever actually use the list 'masked_neighbours' only its length. Maybe it'd be more efficient to use a generator to count directly and not bother making the list explicitly.
2) Maybe it's worthwhile to add a 'dumdumb' protection in case someone tries to be clever and accidentally gets a situation where msipm - len(masked_neighbours) <= 0 which would cause odd behaviour in the list of points (unless it's already implemented in barycenter)

A

paolafer:
Thanks! 
1. I agree and I've modified the list, although I don't think it will make a big difference in performance, since we're dealing with few masked SiPMs.
2. I don't see the danger: if msipm - len(masked_neighbours) <= 0, the condition will be always fulfilled (which may not be what one desires, but it doesn't cause an odd behaviour, right?)

andLaing:
Would it not mean you're sending empty arrays to the barycenter algorithm that could cause odd behaviour? It's probably protected inside the algorithm, no matter.

paolafer:
Can we approve this PR? Or is there anything else to be discussed?

jjgomezcadenas:
I see no reason to hold this any longer. Usual suspects, speak out or hold your peace forever.

gonzaponte:
I have found a magic behavior in this code. In the line
```
n_masked_neighbours = sum(1 for x in pos[within_new_lm_radius] if x in masked_sipm)
```
everything works because the values in `masked_sipm` and in `pos` are somehow the same **object**. If they were to have the same value, but with different objects, this would crash. This is demonstrated here:
```
import numpy as np
a = np.array([1, 2])
b = np.array([1, 2])
c = [a]
assert a in c # OK!
assert b in c # ValueError!!
```
Even though the values are the same, the objects are not and numpy does not let you compare two arrays like that. You'd have to define a function that does the equivalent of `if x in masked_sipm`, but that compares the arrays properly (with `np.allclose`).

paolafer:
I see the problem. If I have understood correctly, in this case the two values **are** the same object, right? So, the function behaves as it is meant to, why should we change it? It is not a general function, it is used only in that case. Am I missing something here?

gonzaponte:
> So, the function behaves as it is meant to, why should we change it?

It works because something magical happens elsewhere. If at any point that magic is removed it will crash. This function should not know anything about what happens elsewhere.

jacg:
`masked_sipm` is also a parameter with a mutable default value. This is going to cause pain and confusion. **Do not use mutable objects as default arguments in Python**. If no neater solution is available, then this is the standard pattern to apply in these cases:

    def fn(parameter_with_mutable_default = None):
        if parameter_with_mutable_default is None:
            parameter_with_mutable_default = <a new instance of the mutable default value>

The problem is that default values are evaluated only once: when the function is *defined*, not when it is called. This means that mutations of the value persist into subsequent calls to the function.

Even if a function doesn't mutate the object anywhere in the code *today*, sometime in the future someone might come along and change it so that it does, so just close the door to this potential bug as standard practice and do not give function parameters mutable values as defaults, in Python.

paolafer:
Ok, @jacg I'll change the code as you suggest, thanks for pointing this out. @gonzaponte I'll look into the code to find out where magics happens...

gonzaponte:
> I'll look into the code to find out where magics happens...

It probably happens in `base_cities.py`, but the point is that it must not depend on that magic.

paolafer:
While looking around to understand the magic behaviour, I actually realized that the algorithm wasn't finding the correct masked SiPMs at all. Therefore, I fixed this, introducing a new function in the xy_algorithms, `get_neighbours`, which returns all the SiPMs around a given one, irrespectively of the charge they have detected. Afterwards, a search is made on this list, to see if any of these SiPMs are masked (following @gonzaponte's suggestion of using `numpy.allclose`): I tried to do this as pythonically as possible, not sure If I succeded...

invisible_cities/reco/xy_algorithms.py
```diff
@@ -170,7 +170,7 @@ def corona(pos, qs,
             masked_sipm = []
         # find any masked channel in the first ring around the new local maximum of charge
         neighbours = get_neighbours(new_local_maximum, pitch)
-        n_masked_neighbours = sum(1 for p in neighbours if len(list(filter(lambda x: np.allclose(p, x), masked_sipm))) == 1)
+        n_masked_neighbours = sum(len(list(filter(lambda x: np.allclose(p, x), masked_sipm))) for p in neighbours)
```
jacg: Wrapping the filter in a list so that you can take its `len` in order to count the number of times a condition is satisfied, is probably not the clearest way of expressing the intent. So I would replace

    len(list(filter(lambda x: np.allclose(p,x), masked_sipm))

with

    sum(1 for x in masked_sipm if np.allclose(p,x))

and give `x` a meaningful name. But it would probably be even better to wrap this whole expression inside a function with a meaningful name, and call the function. Remember that such a function can be defined *inside* `corona` itself.

Try to make the code-self documenting: meaningful names attached to small functions or values of expressions often help to clarify what is going on, and why.

jacg:
> While looking around to understand the magic behaviour, I actually realized that the algorithm wasn't finding the correct masked SiPMs at all.

Have you recorded this discovery by adding a test that catches it?

jjgomezcadenas (commented):
This reads quite kosher! @jacg you are about to create a new monster (or shall we say monstrenca?)

paolafer:
> Have you recorded this discovery by adding a test that catches it?

The existing test wasn't catching the error, because the code was calculating erroneously **more** masked SiPMs. Therefore, since the condition to add a cluster is 

`if len(within_new_lm_radius) >= msipm - n_masked_neighbours`

having a larger number of `n_masked_neighbours` kept the number of reconstructed clusters (which is the assertion of the test) always the same and the correct one.

To test this behaviour, one should be able to retrieve the number of masked SiPMs found and to compare it with a known number, but the `corona` function does not return this value. 

Does it occur to you a way of doing it?

invisible_cities/reco/xy_algorithms.py
```diff
@@ -170,7 +170,11 @@ def corona(pos, qs,
             masked_sipm = []
         # find any masked channel in the first ring around the new local maximum of charge
         neighbours = get_neighbours(new_local_maximum, pitch)
-        n_masked_neighbours = sum(len(list(filter(lambda x: np.allclose(p, x), masked_sipm))) for p in neighbours)
+
+        def number_of_masked_neighbours(neighbours, masked_sipm):
+            return sum(1 for neigh in neighbours for m_sipm in masked_sipm if np.allclose(neigh, m_sipm))
+
+        n_masked_neighbours = number_of_masked_neighbours(neighbours, masked_sipm)
```
jacg: This still leaves a cryptic nested loop whose meaning is difficult to discern. I was thinking along the lines of

    def my_meaningful_counting_operation_name(sipm):
        return sum(1 for masked_sipm in masked_sipms if np.allclose(neigh, masked_sipm))

    n_masked_neighbours = sum(my_meaningful_counting_operation_name(neigh) for neigh in neighbours)

but after closer inspection, I think that what you actually mean is more like

    def is_masked(sipm):
        return any(np.allclose(sipm, masked_sipm) for masked_sipm in masked_sipms)

    n_masked_neighbours = len(tuple(filter(is_masked, neighbours)))

where an alternative spelling of the last line could be

    n_masked_neighbours = sum(1 for neigh in neighbours if is_masked(neigh))


IIUC your use of `np.allclose` here has the meaning

    def have_same_position_in_3D_space(a,b):
        return np.allclose(a,b)

which would lead to `is_masked` being


    def is_masked(sipm):
        return any(have_same_position_in_3D_space(sipm, masked_sipm) for masked_sipm in masked_sipms)

invisible_cities/reco/xy_algorithms.py
```diff
@@ -57,6 +57,9 @@ def get_neighbours(cs, pitch):
     comb = product(shift, shift)
     return [tuple([cs[0][0] + pitch*i, cs[0][1] + pitch*j]) for i, j in comb]
 
+def number_of_sipms_in_two_lists(list1, list2):
```
jacg: How about calling it `count_equivalent_positions_in_two_sets`?

jacg:
> This reads quite kosher!

Hmm ... I disagree. I don't think that

    good_name = <complex expression>

is improved by turning it into

    def good_very_similar_name(...):
        return <complex expression>

    good_name = good_very_similar_name(...)

I was looking for

    def good_name_for_substep(...):
        return <less complex sub-expression>

    good_name = <other less complex sub-expression using good_name_for_substep>

invisible_cities/reco/xy_algorithms.py
paolafer: I like this organization and I have implemented it in the code. By the way, I had to look up for 
> IIUC 

in google and the first entry I found was [the following](https://www.iiuc.ac.bd) :-)

invisible_cities/reco/xy_algorithms.py
```diff
@@ -174,7 +174,13 @@ def corona(pos, qs,
         # find any masked channel in the first ring around the new local maximum of charge
         neighbours = get_neighbours(new_local_maximum, pitch)
 
-        n_masked_neighbours = number_of_sipms_in_two_lists(neighbours, masked_sipm)
+        def is_masked(sipm):
+            print(sipm)
```
jacg: I trust that the `print` wil be removed.

gonzaponte:
> Does it occur to you a way of doing it?

At least you can test the function you have exported, which was the cause of the failure, wasn't it?

paolafer:
I've added a test for the function that returns neighbours SiPMs.

invisible_cities/reco/xy_algorithms.py
```diff
@@ -46,13 +50,24 @@ def get_nearby_sipm_inds(cs, d, pos, qs):
     """return indices of sipms less than d from (xc,yc)"""
     return np.where(np.linalg.norm(pos - cs, axis=1) <= d)[0]
 
+def get_neighbours(cs, pitch):
```
gonzaponte: may I suggest:

```
def get_neighbours(cs, pitch):
    x0, y0 = cs[0]
    shift  = np.array([-pitch, 0, +pitch])
    return product(x0 + shift,
                   y0 + shift)
    # surround product with tuple if the iterator isn't good enough
```

Also, leave *two* empty lines between functions, please

invisible_cities/reco/xy_algorithms.py
```diff
@@ -46,13 +50,24 @@ def get_nearby_sipm_inds(cs, d, pos, qs):
...
+    """return positions of all sipms around cs,
+       irrespectively of the charge detected"""
+    shift = [-1, 0, 1]
+    comb = product(shift, shift)
+    return [tuple([cs[0][0] + pitch*i, cs[0][1] + pitch*j]) for i, j in comb]
+
+def have_same_position_in_space(a,b):
```
gonzaponte: `a,b` -> `a, b`

invisible_cities/reco/xy_algorithms.py
```diff
@@ -150,8 +169,21 @@ def corona(pos, qs,
         # find the SiPMs within new_lm_radius of the new local maximum of charge
         within_new_lm_radius = get_nearby_sipm_inds(new_local_maximum, new_lm_radius, pos, qs)
 
-        # if there are at least msipms within_new_lm_radius, get the barycenter
-        if len(within_new_lm_radius) >= msipm:
+        if masked_sipm is None:
```
gonzaponte: you can use `()` as the default for `masked_sipm` and avoid these two lines.

invisible_cities/reco/xy_algorithms.py
paolafer: Ok.

invisible_cities/reco/xy_algorithms.py
paolafer: Done.

invisible_cities/reco/xy_algorithms.py
paolafer: Probably @jacg can give a more informed opinion here, but googling a bit it seems to me that the `if masked_sipm is None` option is the preferred one for clarity.

jacg: If the default value of `masked_sipm` is supposed to be an empty *sequence* and you don't rely on any specific list-like behaviour, then using an empty tuple as the default strikes me as the best solution.

The `None`-as-default-arg idiom is standard, but quite ugly. I'm all for avoiding it if it can be avoided elegantly. Empty tuple to represent immutable (constant) empty sequence is perfectly kosher.

paolafer: Ok, then!

jjgomezcadenas (dismissed):
This seems to be solved and we have procastinated long enough.

jjgomezcadenas:
Please @gonzaponte merge

jacg:
> This seems to be solved

This is at odds with my records, which show that incorrect results were getting through the tests, and that the tests needed improving. 

I haven't had the time to understand the issue and try to see what can be done about it, and it is possible that a solution has been provided in the meantime and I have missed it. But if that not the case, then approving this PR in its current state is a mistake. Let's be clear: my records show that the current tests have been demonstrated to fail to spot incorrect behaviour. This is not something that we want to find its way into master!

Please **do not merge** until it has been confirmed that a test which catches the known incorrect behaviour is in place.

paolafer:
I've added a test that ensures that the function `get_neighbours()` returns indeed all the expected neighbours. The test that is missing is one that ensures that the function `is_masked()` gives the correct result. In order to test this function, we should move it outside the `corona` function, right? Is it worth it? What do you think, @jacg ?

jacg:
Sorry, I'm about to go offline, not back until late this evening. Teaching all day tomorrow. If this is urgent, you'll have to do it without me.

jjgomezcadenas:
>Sorry, I'm about to go offline, not back until late this evening. Teaching all day tomorrow. If this is urgent, you'll have to do it without me.

Yes, it was urgent (pipeline overflowing) and this is why I approved the PR (there was a test ensuring correct behaviour of get_neighbourts() and I assumed that is_masked() if it needs further testing can be done in a different PR.

paolafer:
I decided to extract the `is_masked()` function from `corona` and added a test, which fails with the old code and does not with this PR's one. I think it's all done!

gonzaponte:
I have rebased and cleaned up the commit history.

gonzaponte:
@paolafer, please confirm you are happy with my cleanup. Has the issue raised by @jacg being addressed?

paolafer:
For me it's fine, and, to my knowledge, the issue has been solved.

gonzaponte (approved):
Extension to the corona algorithm to take into account the masked channels. Everything is tested and the code quite good-looking!

PR #431, opened 2018-01-25
Fix diomira masked channels

gonzaponte:
This PR:
- Demonstrates that `Diomira` doesn't work with masked pmts unless the input is already masked and fixes it.
- Demonstrates that there isn't any protection at all against triggering on masked channels. Now it raises an error.
- Speeds up Diomira a bit by not deconvoluting all waveforms (only those needed for trigger simulation).


paolafer (approved):
This PR addresses a bug in the dealing of masked PMTs in Monte Carlo simulation. The solution proposed works and a test has been added which demonstrates the need for this. Also, a protection against triggering on masked PMTs has been added. I approve it!

PR #417, opened 2018-01-09
Adapt IC to new PMAPs model

gonzaponte:
This PR adapts all existing code in IC to the new PMap interface. This includes a big refactor of `peak_functions`,  `pmap_functions` and `calib_sensor_functions`, which have been fully de-cythonized. The IO functions have also been rewritten to match the new representation.

Everything is thoroughly tested (at least I have tried to do so).

Closes #416

PR #432, opened 2018-02-01
Cities safety checks

gonzaponte:
This PR fixes a bug in the output of `Irene`, where the event number of empty (at the pmap level) events were being stored, creating an inconsistency between two different tables in the file. At the same time, `PCity` contained a check when reading the input that allowed the bug to pass silently. In #417 this check was removed, which revealed the bug.

This PR filters out empty events from `Irene`'s output and reintroduces the check in `PCity` for backward-compatibility (so we can run the city on old data). Three tests have been introduced to ensure this bug doesn't cause any further problems.


jjgomezcadenas (approved):
Good catch, and excellent fix.

PR #434, opened 2018-02-01
Implement monitoring-related functions

Aretno:
Hi everybody, this PR is to discuss and add the monitoring software to IC. As I have nearly zero idea in which would be the best way to add this to our scheme and before doing too much work that could not be useful I'd rather ask. 

First, I'd like to talk about the writer, starting from @andLaing work at io/hist_io.py I modified it so do not have to declare a writer instance for every histogram. In the meanway I added the information of under/over flow. It would be the following function (of course, it should be made clean, optimized with the help of those who know the ways of the python, etc., this is just a rough and starting base to start the discussion):

```
def hist_writer_var(file, *, compression = 'ZLIB4'):
    
    def write_hist(group_name : 'options: HIST, HIST2D',
                   table_name : 'histogram name',
                   entries    : 'np.array: RWF, CWF, SiPM',
                   bin_centres: 'np.array of bin centres',
                   out_of_range: 'np.array lenght=2 with events out of range'):
        
        try:                       hist_group = getattr          (file.root, group_name)
        except tb.NoSuchNodeError: hist_group = file.create_group(file.root, group_name)

        try:
            hist_table = getattr(hist_group, table_name)
        except tb.NoSuchNodeError:
            file.create_array(hist_group, table_name, entries)
            file.create_array(hist_group, table_name+'_bins', bin_centres)
            file.create_array(hist_group, table_name+'_outRange', out_of_range)
                    
    return write_hist
```

It basically creates an array with the variable name containing the entries, another with the name+'_bins' to containt the binning and another named with the variable name + '_outRange' to containt the under/over flow. Are you ok with something like this or is a terrible idea?

Moving on, my idea was to get the variables we want to plot on an file by file basis, plot them and refresh the variables so we do not end up with reaaaally long arrays in case we want to run the monitoring through a bunch of files. I know it will probably be run always on individual files but I though it would be better if we could accommodate the code for other cases.

For this I created a custom class MonPar (Monitoring Parameters, temporal name) which is initialized with an array of parameter names that it will contain. I also create an instance of that class with the binning of each parameter and another for the binned values, the output of the histogram. To bin the data I use a modification of the bin_waveform function that Andrew did (in one of his branches), so it returns the data binned in the bins given and the under/overflows, it would be the following:

```
def bin_data(data, histbins): #Copiado de Andrew en invisible_cities/calib/sipmpdf.py
    """
    Bins the current event data and adds it
    to the file level bin array
    """
    bin_data = lambda x: [np.histogram(x, histbins)[0], np.array([len(x[x<histbins[0]]), len(x[x>histbins[-1]])])]
    return np.apply_along_axis(bin_data, 0, data)
```
And here you have the MonPar class I was talking about:

```
class MonPar:
    def __init__(self, attrNames, values=None):
        for attribute in attrNames:
            if values:
                setattr(self, attribute, values[attribute])
            else:
                setattr(self, attribute, []) 
        self.attrNames = attrNames

    def BinValues(self, bins):
        for attribute in self.attrNames:
            attrVal = getattr(self, attribute)
            attrBin = getattr(bins, attribute)
            setattr(self, attribute, bin_data(attrVal, attrBin))        

    def Add(self, additive):
        for attribute in self.attrNames:
            attrVal = getattr(self, attribute)
            if len(attrVal) == 0:
                setattr(self, attribute, getattr(additive, attribute))
            else:
                setattr(self, attribute, attrVal + getattr(additive, attribute))
```

Here you have a working example that I have in mind with this structure:

```
S1Vars = {'Number', 'Width', 'Height', 'Energy', 'Charge', 'Time'}
rangeS1 = {'Number': [0, 10, 10], 'Width': [0, 1000, 10], 'Height': [0, 10, 10], 'Energy': [0, 50, 50],\
           'Charge': [0, 20, 20], 'Time': [0, 1300, 1300]} #Bin limits and number of bins

output = "/Users/ansimez/Documents/InvCit/hdf5/histos1_pmaps_001_4670_v0.9.3_20171128_double_kr1300.h5"
h5out = tb.open_file(output, "w")
writer = hist_writer_var(h5out)

files = glob.glob("/Users/ansimez/Documents/InvCit/hdf5/diffusion/4670/pmaps_00*_4670_v0.9.3_20171128_double_kr1300.h5")

S1Bin = MonPar(S1Vars, {attrName: np.arange(attrVal[0], attrVal[1] + (attrVal[1]-attrVal[0]) / attrVal[2], (attrVal[1]-attrVal[0]) / attrVal[2])\
                        for attrName, attrVal in rangeS1.items()})
S1Var = MonPar(S1Vars)
S1Binned = MonPar(S1Vars)

for file in files:
    pmaps = load_pmaps(file)
    S1Var = MonPar(S1Vars)
    for i in pmaps:
        fillS1var(pmaps[i].s1s, S1Var)
    S1Var.BinValues(S1Bin)
    S1Binned.Add(S1Var)

for attribute in S1Binned.attrNames:    
    attrValue = getattr(S1Binned, attribute)
    writer('HIST', 'S1_'+attribute, attrValue[0], shift_to_bin_centers(getattr(S1Bin, attribute)), attrValue[1])
``` 
This is what I currently have in a notebook,  the binning and range are set at the beginning. In S1Bin the bin arrays are stored, in S1Var we store the values, it is created from 0 at the beginning of each file so it does not grow too big and in S1Binned we accumulate the binned values. The fillS1Var just fills the S1 fields with the pmap information:

```
def fillS1var (s1s, S1Var):
    S1Var.Number.append(len(s1s))
    for s1 in s1s:
        S1Var.Width .append(s1.width)
        S1Var.Height.append(s1.height)
        S1Var.Energy.append(s1.total_energy)
        S1Var.Charge.append(s1.total_charge)
        S1Var.Time  .append(s1.time_at_max_energy)
``` 

So this would be the proposal, comments? Are you ok with a scheme like this or should it be done in another way? As I said, I really have no idea of what I'm doing so I'm prepared for the stones if something is a terrible bad idea.


gonzaponte:
> It basically creates an array with the variable name containing the entries, another with the name+'_bins' to containt the binning and another named with the variable name + '_outRange' to containt the under/over flow. Are you ok with something like this or is a terrible idea?

This depends a bit on the big picture: do we want to store ***any*** type of histogram, for instance with different bin sizes? If so, we need something like a table with 3 columns: `x_min`, `x_max` and `entries`. This is a bit more painful, but it is more flexible. 2D- or 3D-histograms would need different tables and the implementation of the out-of-range bins is not obvious to me. We may also need an extra field for storing uncertainties, which may not be poissonian.

> Here you have a working example that I have in mind with this structure:

The code in this snippet is a bit confusing to me, particularly the lines within the loop. Let me try to understand: first, you create an instance of `MonPar` in order to store the relevant pmap info; then you call a function that fills that instance; after that you bin the data you have filled based on a previous instance of the `MonPar` class; finally you add the binned data to another `MonPar` instance that you use as an accumulator. Right? If this is correct, it seems a bit too complicated, with too many instances of `MonPar`. This needs a bit of thought...

After a first iteration with myself, I think we should have the `MonPar` class with `attr_names` and `bins` as arguments to the constructor + a dictionary to use as an accumulator in the loop (instead of another instance of the class). The dict is then passed to `Add` which bins the data and adds it to a `values` attribute (which has been initialized as zeros). Opinions?

Aretno:
`This depends a bit on the big picture: do we want to store any type of histogram, for instance with different bin sizes? If so, we need something like a table with 3 columns: x_min, x_max and entries. This is a bit more painful, but it is more flexible. 2D- or 3D-histograms would need different tables and the implementation of the out-of-range bins is not obvious to me. We may also need an extra field for storing uncertainties, which may not be poissonian.`

Well, we are saving the bin centers, we could save the bin limits instead (just by not calling the shift_to_bin_centers when writing) so we have the possibility of any bin size. Completely agreed on the uncertainties.

`After a first iteration with myself, I think we should have the MonPar class with attr_names and bins as arguments to the constructor + a dictionary to use as an accumulator in the loop (instead of another instance of the class). The dict is then passed to Add which bins the data and adds it to a values attribute (which has been initialized as zeros). Opinions?`

That's good, I like it, I tried to have each set of interesting values in separate objects but I prefer this a lot more, I'll work on it and get back with a modification of the proposal.

Aretno:
After a bit more of conversation with @gonzaponte this is what I have written (I believe it would be easier to follow in a notebook), in the end you have a comparison between a histogram generated with the data and the loaded histogram after being saved, identical as we should expect.

https://github.com/Aretno/ICARO/blob/master/MonitorTest.ipynb

I'm pretty happy with how is it as of now, I would like to, at least, prepare executable scripts for each monitoring step, at least for the beginning of the run. With this working we would only need to make the fill functions, which are a no brainer.

At a later time we can make (or discuss) a set of cities as suggested by @jahernando but that would take more time (no idea how to make a city plus higher standards in the IC chain).

Comments on this?

jjgomezcadenas:
Seems pretty straight forward. I am OK with this approach. You may want to keep the name of functions consistent (add rather than Add, fill_svar rather than fillSvar, etc). Also, you may want to encapsulate your histo drawing in some pretty functions. 

But the main question. What sort of comparison are you doing? Drawing histos is nice, but we want to have an automatic procedure (bin-by-bin) to certify that they are equal or very close, no? With some complain if they are not. 

Agree?

Aretno:
Agreed on the names, not sure about the drawing. As it is presented the code is thought to bin data into n-dimension binning and saving that info but it does not plot anything. Plotting will be made later by loading the binned information. The plot in the nb is to compare this binned information once plotted with the actual histogram, without any saving. 

However, I agree that I could make some easy functions that load every histo and plots it so it's easier for the user; maybe this was what you were referring to. 

For the tests, yes, I agree, better with a bin by bin comparasion. I have checked now with np.allclose() and everything is okey. However, I wonder if we want this check done every time we save a histogram or we just have to check it works at the beginning. I mean, this would mean that we produce the actual plots for each file, and they could be a considerable number. 

Another thing, should we add an additional value with the axis labels? I'm thinking it may be useful, specially to specify units and probably should be a must if we do a function for loading and plotting the histos.

Aretno:
Here you have a new version of the code for monitoring. This is now a full example of the plots generated for PMap monitoring:

https://github.com/Aretno/ICARO/blob/master/MonitorTest.ipynb

I have added labels to the histo writer and two new functions, one to add histograms (as we will probably be generating them file by file) and another to plot histograms with its statistics. The latest one can plot all the histograms in one file or go through a list of histo names to print, it also can save each histogram as a pdf, to upload to our online monitoring website. 

The structure of the code we have is: a histogram writer (saves bins, bin content, overflow, errors and labels), a MonParam Class that stores binned data, a set of utility functions (plotting and adding) and a set of datatype-specific filling functions (in this case, for pmaps).

In addition there are some really rudimental tests in the end of the NB (I have to learn how to do proper tests). If you look at them you'll see that the labels are not saved exactly properly and the check returns false. For some reason I don't know the labels are saved as byte literals ( b' string ' ) instead of just the strings and the test fails when checking for equality. However, this is not worrisome as it is taken into account in the plot histogram function. The rest of the fields pass the tests.

As we start data taking shortly and I would like this to be available for shifters as soon as possible, I would like to do first scripts for each datatype to be monitored (waveforms, pmaps, kdst, hdst) so they can be added to the monitoring daemons (after iterating with Pau). I'm saying this because I would like this scripts to live in ICARO (lower standards) with no modifications unless you see something truly terrible. Meanwhile we can prepare the proper Monitoring Cities for each datatype but while we are at it, the shifters already have something to look at. On the other hand, I'd like to leave this PR open for implementation when we write the cities.

jjgomezcadenas:
@jahernando can you please review this?

jahernando:
First, this is a nice piece of code to deal with histograms, write them into hdf5, read them, add them or accumulate on them more data.

The idea is to have a writer/reader of a histogram into a h5 file. There is a MonParam class that internally stores all the components of the histograms. There are some extra functions to read and accumulate histogram from several files. 

There is a minimal but sensible collection of test to write/read and accumulate histograms.

In addition in this (dummy) PR there is specific code to monitor pmaps and wfs. For every data file, the data from pmaps is accumulated into dictionary of 'variables', which functions as a old-fashion 'ntuple', and the MonParam instance is smart enough to accumulate the data of that dictionary in the corresponding arrays. There is some 'specific' code for the pmaps and wfs monitoring.

My proposal is to approve this (dummy) PR. To convert the  code (histogram writer/reader, MonParam class, read and accumulate histograms from files) into a module in IC (most likely to be located into 'io' folder). And the specific pmaps, wfs code should go into a module in a different folder  in IC, named for example 'monitoring'.

But the structure behind the code proposed is, to my view, more general, and demands for a more 'generic' implementation or re-factoring. To not miss this point, we can open an issue, and discuss openly the implementation (@jjgomezcadenas, @gonzaponte , @jacg).

The re-factory proposal will go in the following, old-fashion tradition:

1. Define a class Histo, with attributes: name, labels, bins, contents, errors, where bins, contents, errors can be 1D or 2D arrays (completely  generic). Histo can know how to add.
2. Define a class HistoManager (that extend MonParam class), with a list or a dictionary of Histo(s). The class have the extra functionality: read/write from a file, accumulate or add histograms and plotting. 

This structure is generic enough that can be applied to other parts of the code, and monitoring, and specially to old-fashion write/read histograms into/from a file.

jjgomezcadenas:
>The re-factory proposal will go in the following, old-fashion tradition:

>Define a class Histo, with attributes: name, labels, bins, contents, errors, where bins, contents, errors can be 1D or 2D arrays (completely generic). Histo can know how to add.
Define a class HistoManager (that extend MonParam class), with a list or a dictionary of Histo(s). The class have the extra functionality: read/write from a file, accumulate or add histograms and plotting.

Not sure we want to do this!!! Ben there before! Histograms are nasty objects and have been invented many times, why do we need to invent them again? We have tried (and miserably failed) in the past.

For the very same reason, not sure that a HistoManager class is a good idea. I would tend to keep things simple in this end and focus our efforts somewhere else.

Aretno:
I have to say, I agree with JA as I think this refactor will keep things simpler and more readable. In fact as it was it kind of rested in the middle between a general utility and a more specific one.

In any case, I almost have the refactor done as of now, I'll finish it by tonight and show it to you tomorrow (once I have access again to internet). We can discuss which approach seems better once we have both at hand.

Aretno:
Here you have the result of the refactor:

https://github.com/Aretno/ICARO/blob/master/MonitorTest-Refactor.ipynb

There is a Histogram class that can be filled, plotted and added, an HistoManager class that stores the histograms, saves them, and allows for filling, addition and plotting of several of them. 

In addition there are three general functions, one to load the histograms from a file into an HistoManager object and two functions to directly add or plot histos from files so the user does not have to manage it.

I think the functionalities are almost equal but the latest seems a more flexible solution really similar to the pre-refactor version but I'm okay with whatever version the experts prefer.

Aretno:
I already have scripts ready to give them to Pau (under the assumption that the refactor is the way to go, otherwise i'll have to change some things here and there) but I was thinking that since Gonzalo is here maybe we should try and directly do a Monitoring city. Opinions?

jahernando:
@Aretno Your re-factoring seems to me quite good. Some minor things that maybe can simplify it:

1) The `HistoManager` have some methods `save_to_file`, I will add `read_from_file` and `read_from_files`. And keep functions such as `histograms_from_files` (should return the HistoManager and have the possibility to save the added histograms into a output file) and `plot_histograms_from_file`.
2) (This is a question of taste) I will move the plot functions from `Histogram` and `HistoManager` classes, and convert them into functions which take `HistoManager` as first argument.
3) I understand that you do not want to create `Histogram`s directly, and delegate to the `HistoManager`, because you reuse the labels and bins for the 2D histos, right?

I think this is the right time, now that you start the data taking, to explore how to do the monitoring, going for a 'liquid' city seems to me a good idea (@gonzaponte ).

Aretno:
> The HistoManager have some methods `save_to_file`, I will add `read_from_file` and `read_from_files`. And keep functions such as `histograms_from_files` (should return the HistoManager and have the possibility to save the added histograms into a output file) and `plot_histograms_from_file`.

I'm not sure we want the HistoManager being able to load itself, I find really more intuitive obtaining it from the general function so the user does not need knowledge of the interiors of the HistoManager. 

On the other hand, `add_histograms_from_files` already saves them into an output file by default, I understand you want this to be optional and that, by default, it returns the added histogram? If that's it I don't have any strong opinions but loading the result of the `add_histograms_from_file` is just an extra line (`get_histograms_from_file (sum_file)`)

> (This is a question of taste) I will move the plot functions from `Histogram` and `HistoManager` classes, and convert them into functions which take `HistoManager` as first argument.

Cool with this, already done. Now we have a `plot_histogram` function that takes an histogram as an argument and a `plot_histograms` function that takes an `HistoManager`, in addition to the `plot_histograms_from_file` one, of course.

> I understand that you do not want to create `Histograms` directly, and delegate to the `HistoManager`, because you reuse the labels and bins for the 2D histos, right?

Well, not exactly, the 2D histos label and binning is done in the "fill_pmaps" functions so I really have no important reason to declare the histograms within the HistoManager. The only reason is just to keep the loop over the variables hidden but I could make a function that declares Histograms and then make it so the HistoManager initializes with a list of Histograms instead of bins so it is not the one to initialize the Histograms. Dunno what is better.

gonzaponte:
> I will add read_from_file and read_from_files

I prefer those to be external functions.

> I will move the plot functions from `Histogram` and `HistoManager`

I agree with the first one. For some reason it seems reasonable in `Histomanager`, but I don't oppose externalizing it.

> I understand that you do not want to create Histograms

I think I am in favor of forcing the user to initialize `Histogram` instances, as it would be very readable, but I would like to see it in place before making the decision.

invisible_cities/evm/histos.py
```diff
@@ -0,0 +1,194 @@
...
+        additive     = Array or list with data to fill the histogram.
+        data_weights = Array or list with weights of the data.
+        """
+        additive = np.array(additive)
+
+        if len(data_weights) != len(additive.T):
+            data_weights = np.ones (len(additive.T))
+        else:
+            data_weights = np.array(data_weights)
+
```
gonzaponte: A few things:
1. The `!=` comparison is to check whether `data_weights` is empty? If so, then just check its size. An invalid length of that sequence should raise an error, I guess.
2. The default value for `data_weights` should be an empty tuple, as the list object is mutable.
3. You can use `a = some_value if condition else other_value` for assignments of this kind. If it is too long, leave it as it is, but put the code in the same line as the `if`/`else`.
4. I find more elegant `additive.shape[1]` than `len(additive.T)`

invisible_cities/evm/histos.py
```diff
@@ -0,0 +1,194 @@
...
+        then the square root of the counts is computed and assigned as error.
+
+        Arguments:
+        errors = List or array of errors.
+        """
+        if errors is not None:
+            self.errors = np.asarray(errors)
+        else:
+            self.errors = np.sqrt(self.data)
+
```
gonzaponte: Same as before. `a = some_value if condition else other_value` or put the code in the same line as the if/else.

invisible_cities/evm/histos.py
```diff
@@ -0,0 +1,194 @@
...
+    def __radd__(self, other):
+        return self + other
+
+    def __add__ (self, other):
+        if other is None:
+            return self
+        if len(self.bins) != len(other.bins) or not np.all(a == b for a, b in zip(self.bins, other.bins)):
```
gonzaponte: hide this `if` in a function `_check_valid_binning`

invisible_cities/evm/histos.py
```diff
@@ -0,0 +1,194 @@
...
+
+        Arguments:
+        additives: Dictionary with keys equal to the Histograms names.
+                   Values are the data to fill the Histogram.
+        """
+        for histoname, additive in additives.items():
+            if histoname in self.histos:
+                self[histoname].fill(np.asarray(additive))
+            else:
+                print("Histogram with name {} does not exist".format(histoname))
```
gonzaponte: You can use format strings, if you prefer (`f""`)

invisible_cities/evm/histos.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def save_to_file(self, file_out, mode='w', group='HIST'):
```
gonzaponte: Did we finally agreed on leaving this as a method rather than as a standalone function? I feel now more comfortable with a function that lives in `hist_io`

invisible_cities/evm/histos.py
```diff
@@ -0,0 +1,194 @@
...
+                   be written.
+        mode     = Writting mode. By default a new file will be created.
+        group    = Group name to save the histograms in the file.
+        """
+        if mode not in 'wa':
+            raise ValueError("Incompatible mode of writting, please use 'w' (write) or 'a' (append).")
+        with tb.open_file(file_out, mode, filters=tbl.filters('ZLIB4')) as h5out:
+            writer = hist_writer_var(h5out)
+            for histoname, histo in self.histos.items():
+                writer(group, histoname, histo.data,   histo.bins, histo.out_range,
```
gonzaponte: new line after `histoname,`

invisible_cities/io/hist_io.py
```diff
@@ -27,3 +27,36 @@ def write_hist(histo : 'np.array: RWF, CWF, SiPM'):
...
+def hist_writer_var(file, *, compression='ZLIB4'):
+
+    def write_hist(group_name  : 'string with folder name to save histograms',
+                   table_name  : 'histogram name'                            ,
+                   entries     : 'np.array with bin content'                 ,
+                   bins        : 'list of np.array of bins'                  ,
+                   out_of_range: 'np.array lenght=2 with events out of range',
+                   errors      : 'np.array with bins uncertainties'          ,
+                   labels      : 'list with labels of the histogram'):
+
```
gonzaponte: align `)` with the `,`s

invisible_cities/io/hist_io.py
```diff
@@ -27,3 +27,36 @@ def write_hist(histo : 'np.array: RWF, CWF, SiPM'):
...
+        try:                       hist_group = getattr          (file.root, group_name)
+        except tb.NoSuchNodeError: hist_group = file.create_group(file.root, group_name)
+
+        if table_name in hist_group:
+            raise ValueError("Histogram {} already exists".format(table_name))
```
gonzaponte: you can use format strings, if you prefer

invisible_cities/io/hist_io.py
```diff
@@ -27,3 +27,36 @@ def write_hist(histo : 'np.array: RWF, CWF, SiPM'):
...
+
+        vlarray = file.create_vlarray(hist_group, table_name + '_bins', atom=tb.Float64Atom(shape = ()), filters=tbl.filters(compression))
+        for ibin in bins:
+            vlarray.append(ibin)
+        add_carray       (hist_group, table_name                   , entries     )
+        add_carray       (hist_group, table_name + '_outRange'     , out_of_range)
+        add_carray       (hist_group, table_name + '_errors'       , errors      )
+        file.create_array(hist_group, table_name + '_labels'       , labels      )
+
```
gonzaponte: remove extra spaces before `,`s

invisible_cities/io/hist_io.py
```diff
@@ -27,3 +27,36 @@ def write_hist(histo : 'np.array: RWF, CWF, SiPM'):
...
+                   errors      : 'np.array with bins uncertainties'          ,
+                   labels      : 'list with labels of the histogram'):
+
+        try:                       hist_group = getattr          (file.root, group_name)
+        except tb.NoSuchNodeError: hist_group = file.create_group(file.root, group_name)
+
+        if table_name in hist_group:
+            raise ValueError("Histogram {} already exists".format(table_name))
+
+        vlarray = file.create_vlarray(hist_group, table_name + '_bins', atom=tb.Float64Atom(shape = ()), filters=tbl.filters(compression))
```
gonzaponte: split the `vlarray` line so it is shorter?

invisible_cities/io/hist_io.py
```diff
@@ -27,3 +27,36 @@ def write_hist(histo : 'np.array: RWF, CWF, SiPM'):
...
+        for ibin in bins:
+            vlarray.append(ibin)
+        add_carray       (hist_group, table_name                   , entries     )
+        add_carray       (hist_group, table_name + '_outRange'     , out_of_range)
+        add_carray       (hist_group, table_name + '_errors'       , errors      )
+        file.create_array(hist_group, table_name + '_labels'       , labels      )
```
gonzaponte: Why is different the labels array than the other ones?

invisible_cities/io/hist_io.py
```diff
@@ -27,3 +27,36 @@ def write_hist(histo : 'np.array: RWF, CWF, SiPM'):
...
+
+    def add_carray(hist_group, table_name, var):
+        array_atom  = tb  .Atom.from_dtype(var.dtype)
+        array_shape = var .shape
+        entry       = file.create_carray(hist_group, table_name, atom=array_atom,
+                                         shape=array_shape, filters=tbl.filters(compression))
```
gonzaponte: put each keyword argument in a different line

invisible_cities/reco/histogram_functions.py
```diff
@@ -0,0 +1,218 @@
...
+import tables            as tb
+import matplotlib.pyplot as plt
+
+from .. evm  .histos         import Histogram
+from .. evm  .histos         import HistoManager
+from .. core .core_functions import weighted_mean_and_std
+from .. icaro.hst_functions  import shift_to_bin_centers
+
+
+def create_histomanager_from_dicts(histobins_dict, histolabels_dict, init_fill_dict={}):
```
gonzaponte: `init_fill_dict` will have to be `None` by default, as the dictionary is mutable.

invisible_cities/reco/histogram_functions.py
```diff
@@ -0,0 +1,218 @@
...
+    Creates and returns an HistoManager from a dict of bins and a given of labels with identical keys.
+
+    Arguments:
+    histobins_dict   = Dictionary with keys equal to Histogram names and values equal to the binning.
+    histolabels_dict = Dictionary with keys equal to Histogram names and values equal to the axis labels.
+    init_fill_dict   = Dictionary with keys equal to Histogram names and values equal to an initial filling.
+    """
+    histo_manager = HistoManager()
+    for histotitle, histobins in histobins_dict.items():
+        histo_manager.new_histogram(Histogram(histotitle, histobins, histolabels_dict[histotitle],
```
gonzaponte: I would put each argument in a different line, I think

invisible_cities/reco/histogram_functions.py
```diff
@@ -0,0 +1,218 @@
...
+                                              init_fill_dict.get(histotitle, None)))
+    return histo_manager
+
+
+def get_histograms_from_file(file_input, group_name='HIST'):
+    histo_manager = HistoManager()
+    with tb.open_file(file_input, "r") as h5in:
+        name_sel = lambda x: (    ('bins'     not in x)
```
gonzaponte: use a `def` statement rather than a `lambda`

invisible_cities/reco/histogram_functions.py
```diff
@@ -0,0 +1,218 @@
...
+                              and ('labels'   not in x)
+                              and ('errors'   not in x)
+                              and ('outRange' not in x))
+        histogram_list = []
+        group = getattr(h5in.root, group_name)
+        for histoname in filter(name_sel, list(group._v_children)):
```
gonzaponte: do you need to wrap `group._v_children` into a list? If it is an iterable, it is not  needed

invisible_cities/reco/histogram_functions.py
```diff
@@ -0,0 +1,218 @@
...
+
+def join_histograms_from_files(histofiles, group_name='HIST', join_file=None):
+    """
+    Joins the histograms of a given list of histogram files. If possible,
+    Histograms with the same name will be added.
+
+    histofiles = List of strings with the filenames to be summed.
+    join_file  = String. If passed, saves the resulting HistoManager to this path.
+    """
+    if len(histofiles)<1:
```
gonzaponte: `if not histofiles`

invisible_cities/reco/histogram_functions.py
```diff
@@ -0,0 +1,218 @@
...
+        raise ValueError("List of files is empty")
+
+    final_histogram_manager = get_histograms_from_file(histofiles[0], group_name)
+
+    for file in histofiles[1:]:
+        added_histograms        = get_histograms_from_file(file, group_name)
+        final_histogram_manager = join_histo_managers     (final_histogram_manager, added_histograms)
+
+    if join_file is not None:
+        final_histogram_manager.save_to_file(join_file)
```
gonzaponte: The function `Histomanager.save_to_file` takes extra keyword arguments, you may want to accept them also as parameters of this function

invisible_cities/reco/histogram_functions.py
```diff
@@ -0,0 +1,218 @@
...
+
+    return final_histogram_manager
+
+
+def plot_histograms_from_file(histofile, histonames='all', group_name='HIST', plot_errors=False, out_path=None):
```
gonzaponte: Do we want these plotting functions in this module? Probably better somewhere else, but not sure.

invisible_cities/reco/histogram_functions.py
```diff
@@ -0,0 +1,218 @@
...
+    return new_histogram_manager
+
+
+def get_percentage(a, b):
+    """
+    Given two flots, return the percentage between them.
+    """
+    if b == 0:
+        return -100.
+    return 100. * a / b
```
gonzaponte: `return 100 * a / b if b else -100`

invisible_cities/reco/histogram_functions_test.py
```diff
@@ -0,0 +1,143 @@
...
+        assert histoname in list_of_names
+        if (histoname in histogram_manager1.histos) and (histoname in histogram_manager2.histos):
+            histo1 = histogram_manager1[histoname]
+            histo2 = histogram_manager2[histoname]
+
+            aux_histogram           = Histogram(histoname, histo1.bins, histo1.labels)
+            aux_histogram.data      =         histo1.data        + histo2.data
+            aux_histogram.errors    = np.sqrt(histo1.errors ** 2 + histo2.errors ** 2)
+            aux_histogram.out_range =         histo1.out_range   + histo2.out_range
+
```
gonzaponte: I prefer `true_histogram` rather than  `aux_histogram`

invisible_cities/reco/histogram_functions_test.py
```diff
@@ -0,0 +1,143 @@
...
+
+@given(lists(bins_arrays(), min_size=1, max_size=5))
+def test_create_histomanager_from_dicts(bins):
+    histobins_dict   = {}
+    histolabels_dict = {}
+    histograms_dict  = {}
+    print(bins)
+    for i, bins_element in enumerate(bins):
+        print(bins_element)
```
gonzaponte: Don't forget to remove debug prints! :)

invisible_cities/reco/histogram_functions_test.py
```diff
@@ -0,0 +1,143 @@
...
+        title =    "Histo_{}" .format(i)
```
gonzaponte: format strings would be nice here (more readable)

invisible_cities/reco/histogram_functions_test.py
```diff
@@ -0,0 +1,143 @@
...
+    file_out2 = os.path.join(output_tmpdir, 'test_save_histogram_manager_2.h5')
+    histogram_manager2.save_to_file(file_out2)
+
+    joined_histogram_manager1 = histf.join_histograms_from_files([file_out1, file_out2])
+    joined_histogram_manager2 = histf.join_histo_managers(histogram_manager1, histogram_manager2)
+
+    assert len(joined_histogram_manager1.histos) == len(joined_histogram_manager2.histos)
+    for histoname in joined_histogram_manager1.histos:
+        assert_histogram_equality(joined_histogram_manager1[histoname], joined_histogram_manager2[histoname])
+
```
gonzaponte, line 110: two blank lines

invisible_cities/reco/histogram_functions_test.py
```diff
@@ -0,0 +1,143 @@
...
+@settings(suppress_health_check=(HealthCheck.too_slow, HealthCheck.filter_too_much,))
+@given   (histograms_lists())
+def test_join_histograms_from_file_and_write(output_tmpdir, histogram_list):
+    _, list_of_histograms   = histogram_list
+    histogram_manager       = HistoManager(list_of_histograms)
+
+    file_out_aux = os.path.join(output_tmpdir, 'test_save_histogram_manager_1.h5')
+    histogram_manager.save_to_file(file_out_aux)
+
```
gonzaponte: `file_out_test`?

invisible_cities/reco/monitor_functions.py
```diff
@@ -0,0 +1,222 @@
...
+    """
+    var_bins   = {}
+    var_labels = {}
+
+    for k, v in config_dict.items():
+        if "_bins"     in k:
+            var_bins  [k.replace("_bins", "")]   = [ np.linspace(v[0], v[1], v[2] + 1) ]
+        elif "_labels" in k:
+            var_labels[k.replace("_labels", "")] = v
+
```
gonzaponte: - put the `var_*` in the same line as the `if`s.
- align the `, "")`
- remove the spaces surrounding the `np.linspace`

invisible_cities/reco/monitor_functions.py
```diff
@@ -0,0 +1,222 @@
...
+    exception = ['S2_Energy', 'S2_Number', 'S2_Time']
+    bin_sel   = lambda x: ('S1' not in x) and (x not in exception) and ('SiPM' not in x)
+    for param in filter(bin_sel, list(var_bins)):
+        var_bins  ['S2_Energy_' + param]  = var_bins  ['S2_Energy'] + var_bins  [param]
+        var_labels['S2_Energy_' + param]  = var_labels['S2_Energy'] + var_labels[param]
+    var_bins      ['S2_Time_S2_Energy']   = var_bins  ['S2_Time']   + var_bins  ['S2_Energy']
+    var_labels    ['S2_Time_S2_Energy']   = var_labels['S2_Time']   + var_labels['S2_Energy']
+    var_bins      ['S2_Energy_S1_Energy'] = var_bins  ['S2_Energy'] + var_bins  ['S1_Energy']
+    var_labels    ['S2_Energy_S1_Energy'] = var_labels['S2_Energy'] + var_labels['S1_Energy']
+    var_bins      ['S2_XYSiPM']           = var_bins  ['S2_XSiPM']  + var_bins['S2_XSiPM']
```
gonzaponte: alignment

invisible_cities/reco/monitor_functions.py
```diff
@@ -0,0 +1,222 @@
...
+    var_labels    ['S2_XYSiPM']           = var_labels['S2_XSiPM']  + ['Y (mm)']
+
+    del var_bins['S2_XSiPM']
```
gonzaponte: why is there just the `x` variable, but not the `y` one?

invisible_cities/reco/monitor_functions.py
```diff
@@ -0,0 +1,222 @@
...
+    fill_pmap_var_2d(var, 'S1')
+    fill_pmap_var_2d(var, 'S2')
+
+    del var['S2_XSiPM']
+    del var['S2_YSiPM']
+    del var['S2_SingleS1']
+    del var['S2_SingleS1_Energy']
+
+    return var
+
```
gonzaponte, line 150: two blank lines

invisible_cities/reco/monitor_functions.py
```diff
@@ -0,0 +1,222 @@
...
+    histo_manager = histf.create_histomanager_from_dicts(var_bins, var_labels)
+    SiPM_db = dbf.DataSiPM(run_number)
+
+    for in_file in glob.glob(in_path):
+        pmaps = load_pmaps(in_file)
+        for ti, pi in enumerate(pmaps):
+            var = fill_pmap_var(pmaps[pi], SiPM_db)
+            histo_manager.fill_histograms(var)
+    return histo_manager
+
```
gonzaponte, line 171: two blank lines

invisible_cities/reco/monitor_functions.py
```diff
@@ -0,0 +1,222 @@
...
+    Arguments:
+    in_path     = String with the path to the file(s) to be monitored.
+    run_number  = Run number of the dataset (used to obtain the SiPM database).
+    config_dict = Dictionary with the configuration parameters (bins, labels).
+    """
+    var_bins, var_labels = pmap_bins(config_dict)
+    histo_manager = histf.create_histomanager_from_dicts(var_bins, var_labels)
+    SiPM_db = dbf.DataSiPM(run_number)
+
```
gonzaponte: align equals

invisible_cities/reco/monitor_functions.py
```diff
@@ -0,0 +1,222 @@
...
+    for in_file in glob.glob(in_path):
+        pmaps = load_pmaps(in_file)
+        for ti, pi in enumerate(pmaps):
+            var = fill_pmap_var(pmaps[pi], SiPM_db)
+            histo_manager.fill_histograms(var)
+    return histo_manager
+
+def histogram_pmap_variable_creator(run_number):
```
gonzaponte: These functions will be introduced for the liquid city. They are not needed here.

invisible_cities/reco/monitor_functions.py
```diff
@@ -0,0 +1,222 @@
...
+    var_bins   = {}
+    var_labels = {}
+
+    for k, v in config_dict.items():
+        if "_bins"     in k:
+            var_bins  [k.replace("_bins", "")  ] = [ np.linspace(v[0], v[1], v[2] + 1) ]
+        elif "_labels" in k:
+            var_labels[k.replace("_labels", "")] = v
+
+    return var_bins, var_labels, config_dict['n_baseline']
```
gonzaponte: same comments as for `pmap_bins`

invisible_cities/reco/monitor_functions.py
```diff
@@ -0,0 +1,222 @@
...
+    rwf            = Raw waveforms
+    var_dict       = Dictionary that stores the variable values.
+    sensor_type    = Type of sensor('PMT' or 'SiPM')
+    """
+
+    bls = np.mean(rwf, axis=1)[:, np.newaxis]
+    rms = np.std (rwf, axis=1)[:, np.newaxis]
+    var_dict[sensor_type + '_Baseline']   .extend(bls.flatten())
+    var_dict[sensor_type + '_BaselineRMS'].extend(rms.flatten())
+
```
gonzaponte: `array[:, np.newaxis].flatten()` is the same as `array` :).

invisible_cities/reco/monitor_functions.py
```diff
@@ -0,0 +1,222 @@
...
+def fill_rwf_histos(in_path, config_dict):
```
gonzaponte: two blank lines

invisible_cities/reco/monitor_functions.py
```diff
@@ -0,0 +1,222 @@
...
+    histo_manager = histf.create_histomanager_from_dicts(var_bins, var_labels)
+
+    for in_file in glob.glob(in_path):
+        with tb.open_file(in_file, "r") as h5in:
+            var = defaultdict(list)
+            nevt, pmtrwf, sipmrwf, _ = get_rwf_vectors(h5in)
+            for evt in range(nevt):
+                fill_rwf_var(pmtrwf [evt,:,0:n_baseline], var,  "PMT")
+                fill_rwf_var(sipmrwf[evt]               , var, "SiPM")
+
```
gonzaponte: `pmtrwf[evt, :, :n_baseline]`

invisible_cities/reco/monitor_functions_test.py
```diff
@@ -0,0 +1,296 @@
...
+        nsipm = len(speak.sipms.ids)
+        assert     var_dict['S2_NSiPM']   [i] == nsipm
+        if len(s1s) == 1:
+            assert var_dict['S2_SingleS1_Energy'][i] == s1s[0].total_energy
+        assert np.allclose(var_dict['S2_QSiPM'] [counter:counter + nsipm], speak.sipms.sum_over_times                  )
+        assert np.allclose(var_dict['S2_IdSiPM'][counter:counter + nsipm], speak.sipms.ids                             )
+        assert np.allclose(var_dict['S2_XSiPM'] [counter:counter + nsipm], data_sipm.X.values[speak.sipms.ids].tolist())
+        assert np.allclose(var_dict['S2_YSiPM'] [counter:counter + nsipm], data_sipm.Y.values[speak.sipms.ids].tolist())
+        counter += nsipm
+
```
gonzaponte, line 57: two blank lines

invisible_cities/reco/monitor_functions_test.py
```diff
@@ -0,0 +1,296 @@
...
+    for i, speak in enumerate(s1s):
+        assert var_dict['S1_Energy_S1_Width'] [0,i] == speak.total_energy
+        assert var_dict['S1_Energy_S1_Width'] [1,i] == speak.width / units.mus
+        assert var_dict['S1_Energy_S1_Height'][0,i] == speak.total_energy
+        assert var_dict['S1_Energy_S1_Height'][1,i] == speak.height
+        assert var_dict['S1_Energy_S1_Charge'][0,i] == speak.total_energy
+        assert var_dict['S1_Energy_S1_Charge'][1,i] == speak.total_charge
+        assert var_dict['S1_Time_S1_Energy']  [0,i] == speak.time_at_max_energy /units.mus
+        assert var_dict['S1_Time_S1_Energy']  [1,i] == speak.total_energy
+
```
gonzaponte: `[a, b]` rather than `[a,b]`

invisible_cities/reco/monitor_functions_test.py
```diff
@@ -0,0 +1,296 @@
...
+        assert     var_dict['S2_Time_S2_Energy']  [0,i] == speak.time_at_max_energy /units.mus
+        assert     var_dict['S2_Time_S2_Energy']  [1,i] == speak.total_energy
+        if len(s1s) == 1:
+            assert var_dict['S2_Energy_S1_Energy'][0,i] == speak.total_energy
+            assert var_dict['S2_Energy_S1_Energy'][1,i] == s1s[0].total_energy
+        sipm_ids = speak.sipms.ids
+        assert np.allclose(var_dict['S2_XYSiPM'][0, counter:counter + len(sipm_ids)], data_sipm.X.values[speak.sipms.ids].tolist())
+        assert np.allclose(var_dict['S2_XYSiPM'][1, counter:counter + len(sipm_ids)], data_sipm.Y.values[speak.sipms.ids].tolist())
+        counter += len(sipm_ids)
+
```
gonzaponte, line 97: two blank lines

invisible_cities/reco/monitor_functions_test.py
```diff
@@ -0,0 +1,296 @@
...
+                 'S2_XSiPM_labels'  : [          'X (mm)' ],
+                 'S2_Time_labels'   : [   'S2 time (mus)' ]}
+
+    test_bins = {'S1_Energy': [    0.,    5.,             10. ],
+                 'S1_Width' : [    0.,    2.,  4.,         6. ],
+                 'S1_Time'  : [    0.,  1.25, 2.5, 3.75,   5. ],
+                 'S2_Energy': [    0., 7.5e3,          15.0e3 ],
+                 'S2_XSiPM' : [ -100.,  -50.,  0.,  50., 100. ],
+                 'S2_Time'  : [    5.,  6.25, 7.5, 8.75,  10. ]}
+
```
gonzaponte: - Don't use a dot in the numbers when not necessary. In python 3, `1` behaves as `1.0` when required.
- Align decimal points (those that are actually necessary).
- I think is better to put
```
[x, y, z      ]
[a, b, c, d, e]
```
rather than
```
[x, y,     , z]
[a, b, c, d, e]
```
unless there is a symmetry I am not seeing.

invisible_cities/reco/monitor_functions_test.py
```diff
@@ -0,0 +1,296 @@
...
+    out_bins, out_labels = monf.pmap_bins(test_dict)
+
+    list_var = [ 'S1_Energy', 'S1_Width', 'S1_Time', 'S2_Energy', 'S2_Time' ]
+
```
gonzaponte: remove spaces after/before `[` and `]`

invisible_cities/reco/monitor_functions_test.py
```diff
@@ -0,0 +1,296 @@
...
+    assert out_labels['S2_Energy_S1_Energy'][1] == test_dict['S1_Energy_labels'][0]
+
+    assert out_labels['S2_XYSiPM']          [0] == test_dict['S2_XSiPM_labels' ][0]
+    assert out_labels['S2_XYSiPM']          [1] == 'Y (mm)'
+
+    for k in out_bins:
+        assert k in [ 'S1_Energy', 'S1_Width', 'S1_Time', 'S2_Energy', 'S2_Time',
+                      'S1_Energy_S1_Width', 'S1_Time_S1_Energy', 'S2_Time_S2_Energy',
+                      'S2_Energy_S1_Energy', 'S2_XYSiPM' ]
+
```
gonzaponte: double blank line

invisible_cities/reco/monitor_functions_test.py
```diff
@@ -0,0 +1,296 @@
...
+    assert out_labels['S2_Energy_S1_Energy'][0] == test_dict['S2_Energy_labels'][0]
+    assert out_labels['S2_Energy_S1_Energy'][1] == test_dict['S1_Energy_labels'][0]
+
+    assert out_labels['S2_XYSiPM']          [0] == test_dict['S2_XSiPM_labels' ][0]
+    assert out_labels['S2_XYSiPM']          [1] == 'Y (mm)'
+
+    for k in out_bins:
+        assert k in [ 'S1_Energy', 'S1_Width', 'S1_Time', 'S2_Energy', 'S2_Time',
+                      'S1_Energy_S1_Width', 'S1_Time_S1_Energy', 'S2_Time_S2_Energy',
+                      'S2_Energy_S1_Energy', 'S2_XYSiPM' ]
```
gonzaponte: remove spaces after/before `[` and `]`

invisible_cities/reco/monitor_functions_test.py
```diff
@@ -0,0 +1,296 @@
...
+                        'S2_Height_labels' : [      "S2 height (pes)" ],
+                        'S2_Energy_labels' : [      "S2 energy (pes)" ],
+                        'S2_Charge_labels' : [      "S2 charge (pes)" ],
+                        'S2_Time_labels'   : [    r"S2 time ($\mu$s)" ],
+
+                        'S2_NSiPM_labels'  : [      'SiPM number (#)' ],
+                        'S2_IdSiPM_labels' : [              'SiPM id' ],
+                        'S2_QSiPM_labels'  : [    'SiPM charge (pes)' ],
+                        'S2_XSiPM_labels'  : [               'X (mm)' ]}
+
```
gonzaponte: - remove spaces after/before `[` and `]`
- Are the raw strings necessary?
- Check the spaces after the first comma in the bins section.
- You can use Unicode µ (alt + m) if it is handy and does not create problems.

invisible_cities/reco/monitor_functions_test.py
```diff
@@ -0,0 +1,296 @@
...
+    check_histo    = histf.get_histograms_from_file(test_checkfile)
+
+    assert set(check_histo.histos) ==  set(test_histo.histos)
+
+    for k, v in check_histo.histos.items():
+        assert np.allclose(v.data     , test_histo.histos[k].data     )
+        assert np.allclose(v.out_range, test_histo.histos[k].out_range)
+        assert np.allclose(v.errors   , test_histo.histos[k].errors   )
+        assert             v.title   == test_histo.histos[k].title
+        for i, label in enumerate(v.labels):
```
gonzaponte: If they are both lists of strings you can simply say `assert v.labels == test_histo.histos[k].labels` rather than using an explicit loop. Same for bins

invisible_cities/reco/monitor_functions_test.py
```diff
@@ -0,0 +1,296 @@
...
+            assert           label   == test_histo.histos[k].labels[i]
+        for i, bins in enumerate(v.bins):
+            assert np.allclose(bins,    test_histo.histos[k].bins  [i])
+
+
+def test_fill_rwf_var():
+    var_dict = defaultdict(list)
+    pmt_waveforms  = np.random.uniform(0., 10., size=(  12, 10000))
+    sipm_waveforms = np.random.uniform(0., 10., size=(1792, 10000))
```
gonzaponte: remove unnecessary `.`s

invisible_cities/reco/monitor_functions_test.py
```diff
@@ -0,0 +1,296 @@
...
+    monf.fill_rwf_var(pmt_waveforms , var_dict, "PMT" )
+    monf.fill_rwf_var(sipm_waveforms, var_dict, "SiPM")
+
+    assert np.allclose(var_dict['PMT_Baseline']    , np.mean(pmt_waveforms , axis=1)[:, np.newaxis].flatten())
+    assert np.allclose(var_dict['PMT_BaselineRMS'] , np.std (pmt_waveforms , axis=1)[:, np.newaxis].flatten())
+    assert np.allclose(var_dict['SiPM_Baseline']   , np.mean(sipm_waveforms, axis=1)[:, np.newaxis].flatten())
+    assert np.allclose(var_dict['SiPM_BaselineRMS'], np.std (sipm_waveforms, axis=1)[:, np.newaxis].flatten())
+
```
gonzaponte: `[:, np.newaxis].flatten()` equals doing nothing

invisible_cities/reco/monitor_functions_test.py
```diff
@@ -0,0 +1,296 @@
...
+
+def test_rwf_bins():
+    test_dict = {'PMT_Baseline_bins'       : [ 2300., 2700., 400 ],
+                 'PMT_BaselineRMS_bins'    : [    0.,   10., 100 ],
+                 'SiPM_Baseline_bins'      : [    0.,  100., 100 ],
+                 'SiPM_BaselineRMS_bins'   : [    0.,   10., 100 ],
+
```
gonzaponte: remove spaces before/after `[`/`]` and check the spaces before `:`

invisible_cities/reco/monitor_functions_test.py
```diff
@@ -0,0 +1,296 @@
...
+    aux = test_dict['PMT_Baseline_bins']
+    assert np.allclose(out_bins['PMT_Baseline']    , [ np.linspace(aux[0], aux[1], aux[2] + 1)])
+    aux = test_dict['PMT_BaselineRMS_bins']
+    assert np.allclose(out_bins['PMT_BaselineRMS'] , [ np.linspace(aux[0], aux[1], aux[2] + 1)])
+    aux = test_dict['SiPM_Baseline_bins']
+    assert np.allclose(out_bins['SiPM_Baseline']   , [ np.linspace(aux[0], aux[1], aux[2] + 1)])
+    aux = test_dict['SiPM_BaselineRMS_bins']
+    assert np.allclose(out_bins['SiPM_BaselineRMS'], [ np.linspace(aux[0], aux[1], aux[2] + 1)])
+
```
gonzaponte: - rename all `aux` to a better name (`bins`? actually maybe even better `min, max, nbin` instead of a single variable)
- spaces before/after `[`/`]`

invisible_cities/reco/monitor_functions_test.py
```diff
@@ -0,0 +1,296 @@
...
+    assert out_labels['SiPM_Baseline']   [0] == test_dict['SiPM_Baseline_labels']   [0]
+    assert out_labels['SiPM_BaselineRMS'][0] == test_dict['SiPM_BaselineRMS_labels'][0]
+    assert out_baseline                      == test_dict['n_baseline']
+
+def test_fill_rwf_histos(ICDATADIR):
+    test_config_dict = {'PMT_Baseline_bins'       : [ 2300., 2700., 400 ],
+                        'PMT_BaselineRMS_bins'    : [    0.,   10., 100 ],
+                        'SiPM_Baseline_bins'      : [    0.,  100., 100 ],
+                        'SiPM_BaselineRMS_bins'   : [    0.,   10., 100 ],
+
```
gonzaponte: spaces and decimal points

invisible_cities/reco/monitor_functions_test.py
```diff
@@ -0,0 +1,296 @@
...
+    for k, v in check_histo.histos.items():
+        assert np.allclose(v.data     , test_histo.histos[k].data     )
+        assert np.allclose(v.out_range, test_histo.histos[k].out_range)
+        assert np.allclose(v.errors   , test_histo.histos[k].errors   )
+        assert             v.title   == test_histo.histos[k].title
+        for i, label in enumerate(v.labels):
+            assert           label   == test_histo.histos[k].labels[i]
+        for i, bins in enumerate(v.bins):
+            assert np.allclose(bins,    test_histo.histos[k].bins[i])
```
gonzaponte: if they are lists use `assert list_a == list_b`, same as before.

gonzaponte (commented):
It took me a while but finally got there. Mostly ok, but some small cosmetics details still to be fixed. Also, please check that there are not unused imports in the files you've introduced.

invisible_cities/io/hist_io.py
Aretno: If I recall correctly, because it is not a numpy array. Doing a numpy array of that so the add_carray function works would return a numpy array of dtype object I believe.

invisible_cities/evm/histos.py
Aretno: I thought so but I'll see if I move it to hist_io then.

invisible_cities/reco/monitor_functions.py
Aretno: In the config_dict I do not pass any YSiPM bining, just get the same as XSiPM therefore in the end I only need to erase the X one. The XY plot I'm generating does not need a different binning in Y (nor the possibility, I would say).

invisible_cities/reco/monitor_functions_test.py
Aretno: The spaces before `:` are due to being aligned with `'SiPM_BaselineRMS_labels'`

invisible_cities/reco/monitor_functions.py
gonzaponte: Ok. Can we name it `whatever_XY`?

invisible_cities/reco/monitor_functions_test.py
gonzaponte: ok, I didn't realize

invisible_cities/reco/monitor_functions.py
Aretno: In the end I just added a Y field and erased both of them, thought it was clearer.

invisible_cities/evm/histos.py
```diff
@@ -0,0 +1,173 @@
+import numpy  as np
+import tables as tb
+
+from .. reco       import tbl_functions as tbl
```
gonzaponte: extra spaces

invisible_cities/evm/histos.py
```diff
@@ -0,0 +1,173 @@
...
+    def fill(self, additive, data_weights=None):
+        """
+        Given datapoints, bins and adds thems to the stored bin content.
+
+        Arguments:
+        additive     = Array or list with data to fill the histogram.
+        data_weights = Array or list with weights of the data.
+        """
+        additive = np.array(additive)
+        data_weights = np.ones(len(additive.T)) if data_weights is None else np.array(data_weights)
```
gonzaponte: align equals

invisible_cities/io/hist_io.py
```diff
@@ -1,6 +1,14 @@
...
 
-from .. reco import tbl_functions as tbl
+from hypothesis import settings
+from hypothesis import HealthCheck
+from hypothesis import given
+
+from .. reco    import tbl_functions as tbl
+
+from .. evm.histos import HistoManager
+from .. evm.histos import Histogram
```
gonzaponte: align imports

invisible_cities/io/hist_io.py
```diff
@@ -27,3 +35,91 @@ def write_hist(histo : 'np.array: RWF, CWF, SiPM'):
...
+
+        try:                       hist_group = getattr          (file.root, group_name)
+        except tb.NoSuchNodeError: hist_group = file.create_group(file.root, group_name)
+
+        if table_name in hist_group:
+            raise ValueError(f"Histogram {table_name} already exists")
+
+        vlarray = file.create_vlarray(hist_group, table_name + '_bins',
+                                      atom=tb.Float64Atom(shape=()),
+                                      filters=tbl.filters(compression))
```
gonzaponte: align equals with spaces surrounding them

invisible_cities/io/hist_io.py
```diff
@@ -27,3 +35,91 @@ def write_hist(histo : 'np.array: RWF, CWF, SiPM'):
...
+        add_carray       (hist_group, table_name + '_errors'  , errors      )
+        file.create_array(hist_group, table_name + '_labels'  , labels      )
+
+    def add_carray(hist_group, table_name, var):
+        array_atom  = tb  .Atom.from_dtype(var.dtype)
+        array_shape = var .shape
+        entry       = file.create_carray(hist_group, table_name,
+                                         atom=array_atom,
+                                         shape=array_shape,
+                                         filters=tbl.filters(compression))
```
gonzaponte: align equals with spaces surrounding them

invisible_cities/reco/histogram_functions.py
gonzaponte: Thoughts on this?

invisible_cities/reco/histogram_functions_test.py
```diff
@@ -0,0 +1,128 @@
...
+from .. reco                import histogram_functions as histf
+
+from .. io .hist_io         import save_histomanager_to_file
+from .. io .hist_io         import get_histograms_from_file
+from .. evm.histos          import HistoManager
+from .. evm.histos          import Histogram
+from .. evm.histos_test     import assert_histogram_equality
+from .. evm.histos_test     import histograms_lists
+from .. evm.histos_test     import bins_arrays
+
```
gonzaponte, line 20: two blank lines

invisible_cities/reco/histogram_functions_test.py
```diff
@@ -0,0 +1,128 @@
...
+@settings(suppress_health_check=(HealthCheck.too_slow, HealthCheck.filter_too_much,))
+@given(histograms_lists())
+def test_join_histo_managers(histogram_list):
+    args, list_of_histograms = histogram_list
+    histogram_manager        = HistoManager(list_of_histograms)
+    joined_histogram_manager = histf.join_histo_managers(histogram_manager, histogram_manager)
+
+    assert len(list_of_histograms) == len(joined_histogram_manager.histos)
+    for histoname, histogram in joined_histogram_manager.histos.items():
+        histo1 = histogram_manager[histoname]
```
gonzaponte, line 30: align equals

invisible_cities/reco/histogram_functions_test.py
```diff
@@ -0,0 +1,128 @@
...
+    histogram_manager1       = HistoManager(list_of_histograms1)
+    histogram_manager2       = HistoManager(list_of_histograms2)
+    joined_histogram_manager = histf.join_histo_managers(histogram_manager1, histogram_manager2)
+
+    list_of_names = set(histogram_manager1.histos) | set(histogram_manager2.histos)
+    assert len(list_of_names) == len(joined_histogram_manager.histos)
+
+    for histoname, histogram in joined_histogram_manager.histos.items():
+        assert histoname in list_of_names
+        if (histoname in histogram_manager1.histos) and (histoname in histogram_manager2.histos):
```
gonzaponte, line 52: maybe in two lines like:
```
if (histoname in histogram_manager1.histos and
    histoname in histogram_manager2.histos    ):
```

invisible_cities/reco/histogram_functions_test.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+@given(lists(bins_arrays(), min_size=1, max_size=5))
+def test_create_histomanager_from_dicts(bins):
+    histobins_dict   = {}
+    histolabels_dict = {}
+    histograms_dict  = {}
+    for i, bins_element in enumerate(bins):
+        title =    f"Histo_{i}"
+        labels = [ f"Xlabel_{i}", f"Ylabel_{i}" ]
```
gonzaponte, line 79: align equals and spaces before/after `[`/`]`

invisible_cities/reco/histogram_functions_test.py
```diff
@@ -0,0 +1,128 @@
...
+    save_histomanager_to_file(histogram_manager, file_out_test)
+
+    file_out      = os.path.join(output_tmpdir, 'test_join_histograms.h5')
+    _  = histf.join_histograms_from_files([file_out_test, file_out_test], join_file=file_out)
+    joined_histogram_manager1 = histf.get_histograms_from_file(file_out)
+    joined_histogram_manager2 = histf.join_histo_managers(histogram_manager, histogram_manager)
+
+    assert len(joined_histogram_manager1.histos) == len(joined_histogram_manager2.histos)
+    for histoname in joined_histogram_manager1.histos:
+        assert_histogram_equality(joined_histogram_manager1[histoname], joined_histogram_manager2[histoname])
```
gonzaponte, line 128: split into two lines and align

invisible_cities/reco/monitor_functions.py
```diff
@@ -0,0 +1,203 @@
...
+        if   "_bins"   in k: var_bins  [k.replace("_bins"  , "")] = [np.linspace(v[0], v[1], v[2] + 1)]
+        elif "_labels" in k: var_labels[k.replace("_labels", "")] = v
+
+    exception = ['S1_Energy', 'S1_Number', 'S1_Time']
+    bin_sel   = lambda x: ('S2' not in x) and (x not in exception)
+    for param in filter(bin_sel, list(var_bins)):
+        var_bins  ['S1_Energy_' + param] = var_bins  ['S1_Energy'] + var_bins  [param]
+        var_labels['S1_Energy_' + param] = var_labels['S1_Energy'] + var_labels[param]
+    var_bins      ['S1_Time_S1_Energy']  = var_bins  ['S1_Time'] + var_bins  ['S1_Energy']
+    var_labels    ['S1_Time_S1_Energy']  = var_labels['S1_Time'] + var_labels['S1_Energy']
```
gonzaponte, line 37: align plus signs

invisible_cities/reco/monitor_functions.py
```diff
@@ -0,0 +1,203 @@
...
+    var_bins      ['S2_XYSiPM']           = var_bins  ['S2_XSiPM']  + var_bins  ['S2_YSiPM']
+    var_labels    ['S2_XYSiPM']           = var_labels['S2_XSiPM']  + var_labels['S2_YSiPM']
+
+    del var_bins['S2_XSiPM']
+    del var_bins['S2_YSiPM']
+
+    return var_bins, var_labels
+
+
+def fill_pmap_var_1d(speaks, var_dict, ptype, DataSiPM=None):
```
gonzaponte, line 65: I didn't realize before, but the peak has its own type, so `ptype` is redundant. Instead of `if ptype == "S2"` you can say `if isinstance(speak, S2)` where `S2` comes from `evm.pmaps`.

gonzaponte: You can get the string representation with `type(speak).__name__`

invisible_cities/reco/monitor_functions.py
```diff
@@ -0,0 +1,203 @@
...
+    the pmaps variables.
+
+    Arguments:
+    var_dict = Dictionary that stores the variable values.
+    ptype    = Type of pmap ('S1' or 'S2')
+    """
+    param_list = ['Width', 'Height', 'Charge']
+    for param in param_list:
+        var_dict[ptype + '_Energy_' + ptype + '_' + param] = np.array([var_dict[ptype + '_Energy'], var_dict[ptype + '_' + param]])
+    var_dict    [ptype + '_Time_' + ptype + '_Energy']     = np.array([var_dict[ptype + '_Time']  , var_dict[ptype + '_Energy']])
```
gonzaponte, line 112: align plus signs

invisible_cities/reco/monitor_functions.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+    var_bins   = {}
+    var_labels = {}
+
+    for k, v in config_dict.items():
+        if   "_bins"   in k: var_bins  [k.replace("_bins"  , "")] = [np.linspace(v[0], v[1], v[2] + 1)]
+        elif "_labels" in k: var_labels[k.replace("_labels", "")] = v
+
+    return var_bins, var_labels, config_dict['n_baseline']
+
```
gonzaponte, line 188: two blank lines

invisible_cities/reco/olivia.py
```diff
@@ -0,0 +1,62 @@
...
+import invisible_cities.reco.histogram_functions as histf
+import invisible_cities.reco.monitor_functions   as monf
+import invisible_cities.database.load_db         as dbf
+
+
+ICDIR = os.environ['ICDIR']
+
+data_types = ['rwf', 'pmaps'  ]
+sources    = ['kr' , 'generic']
+config_files = {'pmaps_kr'      : os.path.join(ICDIR, 'config/PmapsKrConfig.json')     ,
```
gonzaponte: The city shouldn't know anything about the configuration files available. It should be itself an argument. Actually, the config file should be the only parameter and follow somehow the structure we already have.

invisible_cities/reco/olivia.py
```diff
@@ -0,0 +1,62 @@
...
+    if len(sys.argv) < 6:
+        print("Error: There are missing parameters")
+        print("Usage: python RunMonitoring.py IN_PATH OUT_PATH RUN_NUMBER DATA_TYPE SOURCE\n")
+        sys.exit(2)
+
+    in_path    = sys.argv[1]
+    out_path   = sys.argv[2]
+    run_number = int(sys.argv[3])
+    data_type  = sys.argv[4]
+    source     = sys.argv[5]
```
gonzaponte: align `sys.`

invisible_cities/reco/olivia.py
```diff
@@ -0,0 +1,62 @@
...
+
+    if data_type not in data_types:
+        print('Error: Data type {} is not recognized.\n'.format(data_type))
```
gonzaponte: format strings
remove `\n`

invisible_cities/reco/olivia.py
```diff
@@ -0,0 +1,62 @@
...
+        print('Supported data types: {}'.format(data_types))
+        sys.exit(2)
+
+    if source not in sources:
+        print('Error: Source {} is not recognized.\n'.format(source))
+        print('Supported sources: {}'.format(sources))
```
gonzaponte: format strings
remove `\n`

invisible_cities/reco/olivia.py
```diff
@@ -0,0 +1,62 @@
...
+        sys.exit(2)
+
+    if data_type == 'rwf':
+        with open(config_files[data_type]) as config_file:
+            config_dict = json.load(config_file)
+        histo_manager = monf.fill_rwf_histos(in_path, config_dict)
+        save_histomanager_to_file(histo_manager, out_path)
+
+    elif data_type == 'pmaps':
+        with open(config_files[data_type+'_'+source]) as config_file:
```
gonzaponte: spaces surrounding `+`

gonzaponte (commented):
Minor cosmetics + reorganize olivia arguments

invisible_cities/io/hist_io_test.py
```diff
@@ -0,0 +1,105 @@
...
+    save_histomanager_to_file(histogram_manager, file_out, group=group)
+
+    with tb.open_file(file_out, "r") as h5in:
+        file_group = getattr(h5in.root, group)
+        for histogram in list_of_histograms:
+            histoname    = histogram.title
+            saved_labels = [str(label)[2:-1].replace('\\\\', '\\') for label in getattr(file_group, histoname + "_labels")[:]]
+
+            assert histoname in file_group
+            assert np.all     (a == b for a, b in zip(histogram.bins, getattr(file_group, histoname + "_bins")[:]))
```
gonzaponte: You need to ensure the length of  `histogram.bins` and `getattr(file_group, histoname + "_bins")[:]`matches. `zip` truncates the output to the shorter of the inputs.

gonzaponte: I guess we have discussed it a few times, but why is it not enough `np.allclose(histogram.bins, getattr(file_group, histoname + "_bins")[:]`?

invisible_cities/io/hist_io_test.py
```diff
@@ -0,0 +1,105 @@
...
+            assert np.allclose(histogram.data     , getattr(file_group, histoname              )[:])
+            assert np.allclose(histogram.errors   , getattr(file_group, histoname + "_errors"  )[:])
+            assert np.allclose(histogram.out_range, getattr(file_group, histoname + "_outRange")[:])
+            assert             histogram.labels  == saved_labels
```
gonzaponte: Align `histogram`, `getattr` and `)[:])`

invisible_cities/io/hist_io_test.py
```diff
@@ -0,0 +1,105 @@
...
+
```
gonzaponte, line 79: same as above

invisible_cities/io/hist_io_test.py
```diff
@@ -0,0 +1,105 @@
...
+@settings(suppress_health_check=(HealthCheck.too_slow,))
+@given   (histograms_lists(), text(characters, min_size=1), text(characters, min_size=1, max_size=1))
+def test_save_histomanager_to_file_raises_ValueError(output_tmpdir, histogram_list, group, write_mode):
+    args, list_of_histograms = histogram_list
+    histogram_manager        = HistoManager(list_of_histograms)
+
+    file_out   = os.path.join      (output_tmpdir, 'test_save_histogram_manager.h5')
+    write_mode = write_mode.replace('w','')
+    write_mode = write_mode.replace('a','')
```
gonzaponte: It is better to exclude `"w"` and `"a"` from the input. Try `text(characters, min_size=1, max_size=1).filter(lambda x: x not in ["w", "a"])`

invisible_cities/reco/monitor_functions_test.py
```diff
@@ -38,7 +38,6 @@ def test_fill_pmap_var_1d(pmaps):
         assert     var_dict['S2_Energy']  [i] == speak.total_energy
         assert     var_dict['S2_Width']   [i] == speak.width / units.mus
         assert     var_dict['S2_Height']  [i] == speak.height
-        assert     var_dict['S2_Energy']  [i] == speak.total_energy
```
gonzaponte: why do you remove that?

invisible_cities/reco/monitor_functions_test.py
```diff
@@ -93,6 +92,28 @@ def test_fill_pmap_var_2d(pmaps):
...
+        times    = speak.times
+        energies = pmts .sum_over_times
+        heights  = np.max             (pmts.all_waveforms, axis=1)
+        times    = np.apply_along_axis(lambda wf: times[np.argmax(wf)], axis=1, arr=pmts.all_waveforms) / units.mus
+        npmts    = len(pmts.all_waveforms)
+
+        for j in range(npmts):
+            assert var_dict['PMT' + str(j) + '_S2_Energy'][i] == energies[j]
+            assert var_dict['PMT' + str(j) + '_S2_Height'][i] == heights [j]
+            assert var_dict['PMT' + str(j) + '_S2_Time'  ][i] == times   [j]
```
gonzaponte: It is shorter to use format strings

gonzaponte (commented):
Much shorter review than in previous occasions! 😄

invisible_cities/reco/monitor_functions_test.py
Aretno: It was duplicated (line 38).

invisible_cities/io/hist_io_test.py
Aretno: For 2D plots the bins don't have de same dimension so the bins are a array of dtype=object which I think fails when using the allclose.

gonzaponte:
ok!

invisible_cities/io/hist_io_test.py
gonzaponte: I meant, add the `filter` to the input and remove the `replace` calls here :)

invisible_cities/reco/monitor_functions_test.py
```diff
@@ -152,38 +154,11 @@ def test_pmap_bins():
...
-    assert out_labels['S2_Energy_S2_Height'][0] == test_dict['S2_Energy_labels'][0]
-    assert out_labels['S2_Energy_S2_Height'][1] == test_dict['S2_Height_labels'][0]
-
-    assert out_labels['S2_XYSiPM']          [0] == test_dict['S2_XSiPM_labels' ][0]
-    assert out_labels['S2_XYSiPM']          [1] == test_dict['S2_YSiPM_labels' ][0]
+    assert_bins_and_labels_ndim('S1_Energy_S1_Width' , ['S1_Energy', 'S1_Width' ], out_bins, out_labels, test_bins, test_dict)
+    assert_bins_and_labels_ndim('S1_Time_S1_Energy'  , ['S1_Time'  , 'S1_Energy'], out_bins, out_labels, test_bins, test_dict)
+    assert_bins_and_labels_ndim('S2_Time_S2_Energy'  , ['S2_Time'  , 'S2_Energy'], out_bins, out_labels, test_bins, test_dict)
+    assert_bins_and_labels_ndim('S2_Energy_S1_Energy', ['S2_Energy', 'S1_Energy'], out_bins, out_labels, test_bins, test_dict)
+    assert_bins_and_labels_ndim('S2_XYSiPM'          , ['S2_XSiPM' , 'S2_YSiPM' ], out_bins, out_labels, test_bins, test_dict)
```
gonzaponte, line 163: Like!

gonzaponte (approved):
The time has come. This PR is approved.

Lots of ugly (although necessary) functions to implement data monitoring have been added.
A review and refactor of them will probably be necessary at some point, but it is good to go.
Hard work from @Aretno!!!

gonzaponte:
The only change to the code is the definition of the RMS. The rest of the work concerns commit organization, so I am going to proceed and merge this PR as soon as the Travis build is complete.

PR #435, opened 2018-02-10
Liquid cities

jacg:
This is a major refactoring of the cities. The class and inheritance based structure has been replaced completely with one based on mostly side-effect free functions which are composed into pipelines in a dataflow network. This has a number of benefits, including

+ making the actions of the cities much easier to understand
+ increasing the signal to noise ratio of the city definitions by a factor approaching infinity
+ making the components that make up the cities much easier to test
+ making the components that make up the cities much easier to identify

Some of the components of the dataflow framework and many of the components of the city were developed as specific needs arose during the translation of the cities into the new paradigm. These new components are tested to the extent that they are used in client code which is tested, but many of them do not have their own unit tests. These should be added before this PR is approved. Such components are mostly found in 

1. `dataflow`
2. `cities.components`

Another prerequisite for approving this PR is verification of the new cities by running data tests and comparing the results with the old cities.

The head of this PR does not contain the old cities any more, but the tag `end-of-solid-cities` marks the last commit before the old cities disappear. Please use this tag to run the verifications.


jjgomezcadenas:
Notice that approving this PR is urgent, so that we can use the new cities for the incoming Run III. At the same time, approving this PR requires extensive testing of the new cities which should be a priority of the usual suspects next week. @jahernando to organise the show. Thanks to @jacg and @gonzaponte for the titanic work. As for @jjgomezcadenas, word is that he will be deleted together with the solid cities unless his level or karma raises to redemption level due to his contributions to the liquid ones.

gonzaponte:
Questions:
- Do we want to maintain the double import of `dataflow`? I mean `from . import dataflow as fl` and `from dataflow.dataflow import whatever`? I think we have already talked about this, but I forgot the reasoning behind it. I kinda like having `fl.` everywhere.
- Have you changed the rule about trailing underscores? In Diomira you use the trailing-underscored functions as the pipes and the other one as the component. It seems that when imported the other convention is maintained.
- Have you searched for unused imports?
- What is the plan for the TODOs?

Some remarks before we approve and merge:
- Diomira does not contain trigger simulation yet.
- `wf_from_files` -> `wfs_from_files` and `pmap_from_files` -> `pmaps_from_files`. Or do we prefer the former?
- We are missing a lot of `close_all = True` in the `fl.slice` calls.
- I do like having components that are unique to a city in the city file.
- Some whitespaces have been introduced.
- The spacing between functions is often too short.
- More cosmetics

to be continued...

jjgomezcadenas:
I will let the master reply to your points, but one important point. We would like the code to be tested against the old versions (e.g, running mini-productions of diomira, Irene, Dorothea and penthesilea and extract relevant histograms for comparison. I will rise this point on Wednesday coordination meeting. It is essential to certify that this version performs as the previous version before final approval.

Have you been able to rebase on top of master?

gonzaponte:
> We would like the code to be tested against the old versions

I am aware. I will start to work on that asap.

> Have you been able to rebase on top of master?

I have.

jjgomezcadenas:
Question for @jacg (@gonzaponte): 

In 

def map(op=None, *, args=None, out=None, item=None):
    if item is not None:

        if args is not None or out is not None:
            raise ValueError("dataflow.map: use of `item` parameter excludes both `args` and `out`")
        assert args is None and out is None
        args = out = item

Will it be more (less) elegant something like

try:
    assert args is None and out is None
except AssertionError:
    print(""dataflow.map: use of `item` parameter excludes both `args` and `out`"")
args = out = item

jacg:
@gonzaponte 

> Do we want to maintain the double import of `dataflow`?

`dataflow` defines some components with names which are identical to some builtins or well-known functions, such as `map`, `fliter` and `reduce`. These must *always* be qualified. In the case of other `dataflow` components the qualification can be dropped. When we were organizing the flow description of the cities horizontally and when we didn't have implicit `pick` etc. it was vital to save as many characters as possible. Now that we have multiple streams inside a single pipeline, the horizontal style is not so important, and we seem to have adopted a vertical style. In this case universal qualification of all `dataflow` components is probably perfectly fine.

> Have you changed the rule about trailing underscores?

Sort of. I wasn't aware that we had a rule. There's certainly something wrong about the way things were being done with the underscores, and we probably need to think a bit about the convention we want to adopt. There simply was no time available to dedicate to this when the work was being done.

> Have you searched for unused imports?

No, but while compiling the new branch, I started with empty files and only ever added an import if a test told me to, so there should be very few unused imports, if any.

> What is the plan for the TODOs?

Do them :-)

IIUC, this work needs to be available for production ASAP. In that case, I would leave any TODOs which aren't showstoppers, until later. We'll just have to take on a bit of technical debt here. We should probably immediately open an issue or PR entitled *Do Liquid Cities TODOs*.

> `wf_from_files` -> `wfs_from_files` and `pmap_from_files` -> `pmaps_from_files`. Or do we prefer the former?

Actually, IIUC, it looked like we could replace the vast majority of these functions with a single one, parametrized on the kind of data that it is supposed to extract from the files. (This is an idea that emerged after you had gone.) Another thing that we didn't have time for.

> I do like having components that are unique to a city in the city file.

The question is what we should do when a second city comes along that needs to use one of these. I think that @jjgomezcadenas already has an example of this up his sleeve.

jacg:
@jjgomezcadenas:

    if <boolean expression>:
        raise ValueError(<message>)
    <assertion on logical negation of previous boolean expression>
    <action>

vs

    try: 
        assert <boolean expression>
    except AssertionError:
        print(<message>)
        <action>

In the former, the assertion is completely superfluous, as the `if-raise` already eliminates this possibility.

The latter is an abomination in so many ways, the two most obvious being

1. It converts rejection of meaningless input into acceptance of meaningless input while making noise on `stdout`.
2. The `if` statement exists in multiple languages for the express purpose of making decisions on the basis of boolean values: replacing

       if <bool>:
          <action>

   with

       try:
          assert <bool>
       except AssertionError:
          <action>

  is just perverse!

jjgomezcadenas:
OK, understood.  But then: 

if args is not None or out is not None:
        raise ValueError("dataflow.map: use of `item` parameter excludes both `args` and `out`")
    assert args is None and out is None
    args = out = item

Includes a superfluous assert, right? Something to be changed in the code

jacg:
> [...] Includes a superfluous assert, right?

Which is why I wrote "In the former, the assertion is completely superfluous".

jacg:
@gonzaponte Have I understood correctly that you have rebased this on top of `master` which includes an implementation of #432 in the liquid cities? If so, could you push it please. In your fork would be fine.

@jjgomezcadenas Shall we try to unify the `*_from_files` functions, as we discussed during the last day of liquification?

jjgomezcadenas:
>Shall we try to unify the *_from_files functions, as we discussed during the last day of liquification?

I would prefer to approve the PR before doing anything else. We need to go through several tests and validations in the next few days. I would leave the code alone until all is done and PR is approved. Then we can start another cycle.

gonzaponte:
> Have I understood correctly that you have rebased this on top of master which includes an implementation of #432 in the liquid cities? If so, could you push it please.

It is already pushed in this fork. I have also set in my fork (I couldn't do it in yours) the tag `end-of-solid-cities` to the corresponding place in the rebased branch.

> Shall we try to unify the *_from_files functions, as we discussed during the last day of liquification?
> > I would prefer to approve the PR before doing anything else

This is pretty straightforward, but I don't have any problem with leaving it as it is.

gonzaponte:
> I wasn't aware that we had a rule. There's certainly something wrong about the way things were being done with the underscores, and we probably need to think a bit about the convention we want to adopt.

I agree. What do you think of the `getattr` (coroutine) / `attrgetter` (component) style?

jacg:
> It is already pushed in this fork.

As of 30 seconds ago, I was still unable to fetch the `liquid-cities` branch from your fork.

gonzaponte:
Sorry, I forgot to push.

jacg:
@gonzaponte The diff between my local branch and the one in the PR shows only the addition of two new `.h5` files. I was expecting to find the implementation of #432 in the liquid cities. What am I missing?

gonzaponte:
#432 included a fix to the old cities, which have been removed in this PR.

jacg:
So you are saying that this is an issue which does not even arise in the new cities? In that case, great.

jacg:
The `end-of-solid-cities` tag has been reset to the currently relevant commit in the PR branch.

jacg:
> What do you think of the getattr (coroutine) / attrgetter (component) style?

I'm sorry, you've lost me here.

gonzaponte:
> So you are saying that this is an issue which does not even arise in the new cities? In that case, great.

#432 solves two problems:
1. Irene was saving the event info data even when there is not a pmap asociated to it. This causes a mismatch when reading them and has been solved in that PR, but we have plenty of data processed with that bug.
2. In the `PCity`-derived cities we had some kind of protection against empty pmaps, which was removed in #417. In #432 we included a `try/except` clause to make it backward-compatible.

We have already done 1. in the liquified cities, but not 2. It is not obvious to me how to introduce the same protection as in 2. to the liquified cities without making a mess in the `pmap_from_files` function.

gonzaponte:
> I'm sorry, you've lost me here.

You were concerned about how to be consistent with the naming of the components and the coroutines and which convention we should adopt. We were using a trailing underscore to differenciate them, but that convention was used inconsistently because sometimes the component lived in the same file and sometimes in another file. I proposed to use the `getattr`/`attrgetter` style meaning that we have a function (`getattr`) that does something and its partial application (`attrgetter`) which returns the function applied to a particular case. For instance:
```
# This is the component. Its name is in the attrgetter style
def pmt_simulator(parameters):
    # This is the full function, and is in the getattr style
    def simulate_pmt(wf):
        # do something with wf and parameters
        return something
    return simulate_pmt

# This is the pipe. Its name can also be in the getattr style if we want to
simulate_pmt = fl.map(pmt_simulator(parameters))

jacg:
I really like this.

It would be slightly more correct to say that `attrgetter` is a *curried* version of `getattr` (forgeting the reversal of the parameters).

jjgomezcadenas:
>I really like this.

Me too!

gonzaponte:
Build failed (just in Linux) because of #441. I am triggering a rebuild.

gonzaponte:
Opinions about the last two commits?

jjgomezcadenas (commented):
I love this clock!

gonzaponte:
I have run the whole chain form MCRD to KDST/HDST with 100 events and I have validated the liquid cities.

- Diomira (MCRD to RWF): This part of the processing is not deterministic as it uses random numbers. I have compared the wfs produced with both liquid and solid cities and they are 100% compatible within statistical uncertainties.
- Irene (RWF to PMAP),  Dorothea (PMAP to KDST), Penthesilea (PMAP to HDST) and Zemrude (RWF to CAL): These cities are deterministic. I have compared the values in the output files and they are the same.

This is the notebook I've used for the validation:
https://github.com/gonzaponte/MyNotebooks/blob/master/validations/validate_liquid_cities.ipynb

jjgomezcadenas (commented):
After the checks, this PR is approved. 
We need to be careful before merging though. We have pending PR #440 which introduces several solid cities. PR #440 must be approved and merged before we merge this PR

jjgomezcadenas (approved):

jacg:
Timing in a city component strikes me as forcing the issue somewhat. I think it is more naturally done with a context manager. Preferably one which returns a future which can be used in the results dict of the pipeline. Maybe something along the lines of


    @contextlib.contextmanager
    def timing():
        future = Future()
        t0 = time()
        try:
            yield future
        finally:
            future.set_result(time() - t0)
    
    with timing() as duration:
        fl.push( ...
                results = dict(duration=duration, ...))

jjgomezcadenas:
Well, as usual I must bow to the Master and HIS elegant solutions.

gonzaponte:
I also prefer the context manager. It is much more elegant and explicit!

However, that doesn't allow you to put the future in the `result` argument of `push` as it will try to access its value too soon, right? If so, we may need to keep the value within the city and not visible externally, which may be perfectly fine for our purposes.

jjgomezcadenas:
>If so, we may need to keep the value within the city and not visible externally, which may be perfectly fine for our purposes.

I don't think so. Tipically, when we want to time the execution of the cities we will want to send this info to some external program... I think returning the time is important...

jacg:
> that doesn't allow you to put the future in the result argument of push as it will try to access its value too soon, right?

Hmm, yes, because it's `push` which gets the results of the futures, *before* it returns.

In which case we could do something like this:

    with timing() as duration:
        xxx = fl.push(...)
    xxx.duration = duration.get_result()
    return xxx

But the increased boilerplate is now starting to annoy me. Maybe hide it away in an asbstraction called `timed_push`. I recall that I already wanted to wrap `push` in something that hides a bunch of things that we do for *all* cities (I forget what they were, maybe event range and print mod and the like).

gonzaponte:
Even though this seems stale, I have progressed in my fork. My plan is now to:
- Include any remaining updates to the solid cities in the liquid ones (I am only one or two commits behind)
- Include some test that ensures the output of each city is exactly what we expect.

jjgomezcadenas:
oh yes! Dragons hydras and Gorgones, all of them!

PR #438, opened 2018-02-13
Adds functions to read HDF5 Nexus output and corresponding tests

jerenner:
These functions read MC hits and particles from Nexus HDF5 output.  They and their tests are similar to the existing functions for reading MC hits.  Note that at the moment the PDG code is assumed to be 0.  The test file was generated with only 1 event, though it seems to be quite large -
 about 13Mb in size.  The size of these files appears to change in "blocks" (generating 10 events instead of 1 does not seem to increase the output file size, but generating 1000 events does cause a noticeable increase in size).


jjgomezcadenas:
Tests are failing...

jerenner:
It is likely because h5py needs to be added to the installed libraries - I
will look into adding this

gonzaponte:
Why `h5py` and not `tables`?

Are you using compression when making the test file?

neuslopez:
@jerenner, I am using another hdf5 input file (created using current master of nexus and next100 macro) and I had to change: "track_indx" by "particle_indx" in your code to be able to read it (Nexus writer says so). Am I missing something here? is your code supposed to be reading another type of file?

jerenner:
I thought tables could only read/write Pytables.  Can it read hdf5 that has not been written by pytables?

At some point compression needed to be turned off because it was slowing down Nexus, though I suppose for 1 event I could turn it back on.

I may not have been using the most recent version of the Nexus writer - I will look into this

neuslopez:
I did read nexus output with tables.open_file, no problem.

jerenner:
Ok I'll try to do the reading with tables

jjgomezcadenas:
It should be pretty straight forward to read the tables with pYtables. You have to be careful and index them (@paolafer, @jmbenlloch, are the tables you are writing indexed? If not, this may be a reason why the writer is slow). If in doubt, ask the usual suspects. Indeed, *ask before implementing*. Didn't you suspect that hdf5.py was not needed since we had not installed it? And how do you think we read the tables that NEXUS and DETSIM (C++) are preparing for us? Again, please, sweet please, with sugar and candy **ask before you let your coding steak loose**. If feel terrible that you have spent a darned amount of time solving a problem the wrong way (using hdf5.py).

paolafer:
> I may not have been using the most recent version of the Nexus writer - I will look into this

Yes, I've been making some modification to the nexus code last week, please, use the last version.

I'm expanding Josh's code with the reader of the rest of the Monte Carlo information. Should I contribute to this PR or should you approve this PR and then I open a new one?

jjgomezcadenas:
Please add to the existing pR

paolafer:
This PR has to do also with the changes in the MCTrack class in the [nh5.py](https://github.com/nextic/IC/blob/master/invisible_cities/evm/nh5.py) file that we decided to implement. I volunteer myself to take care of this, too, after a briefing with @gonzaponte and/or @jjgomezcadenas to agree on what we decided...

jerenner:
Yes sorry I should have asked about this first rather than assuming something not written with pytables could not be read with pytables.  Luckily this appears to have been easy to fix (assuming it works now!) - I just needed to change the statements to open the file and retrieve the relevant HDF5 tables, and everything else was the same.

Some modifications:
- now uses pytables instead of h5py
- new (compressed) test file generated with updated version of nexus; conftest.py updated
- removed unnecessary declaration `current_event = {}` in `read_mctracks_nexus`
- `read_mctracks_nexus` updated to use "particle_indx" instead of "track_indx"

jjgomezcadenas (commented):
The code seems OK, although it is somewhat difficult to follow it. The test itself is clear, but perhaps a bit sparse. Anyway, it seems in good shape. I would recommend to write a notebook (we don't have NBS in IC but we definitively need something, meanwhile put it in some public place). The NB should simply read a few events and display the true hits. It also acts as a sort of documentation for this part of the code. 

Otherwise, the PR can be approved, unless some of the usual suspects have any major objection. @paolafer are u adding to this? Given the fact that this is pretty self-consistet (and brief) I think it would be a good idea to approve it sooner than later.

neuslopez:
@jerenner, I've started a notebook here: https://github.com/neuslopez/ICARO/blob/docu/nexusio-plots.ipynb, for now just plotting E of the event. I was not sure though if I should read the "dict" using any IC function, or simply like I did (probably better to keep this offline). Anyways, let me know if I could help you on this.

jacg:
@neuslopez I recommend replacing 

    e = []
    etot = 0
    for ev, hits in a.items():
        etot = 0
        for hit in hits:
            etot = etot + hit.E
        e.append(etot)

with

    Es = [sum(hit.E for hit in hits) for hits in a.values()]

invisible_cities/conftest.py
```diff
@@ -101,6 +107,23 @@ def mc_particle_and_hits_data(electron_MCRD_file):
...
     return efile, name, pdg, vi, vf, p, Ep, nhits, X, Y, Z, E, t
 
+@pytest.fixture(scope='session')
+def mc_particle_and_hits_nexus_data(electron_MCRD_nexus_file):
+    X = [0.07783413, 0.07652718, 0.04626106, 0.03018976, 0.02231551]
+    Y = [-0.00466795, -0.02718811, -0.03123637, -0.04843096, -0.04031885]
+    Z = [0.02206664, 0.08338148, 0.05198042, 0.0628169, 0.06507947]
+    E = [0.00328978, 0.00515397, 0.00501425, 0.00506388, 0.00735444]
+    t = [0.00088132, 0.00163837, 0.00221191, 0.00261107, 0.00284019]
+    name = b'e-'
```
jacg: Does something *demand* that `name` be `bytes` rather than `str`, or is this just an artifact?

invisible_cities/conftest.py
```diff
@@ -101,6 +107,23 @@ def mc_particle_and_hits_data(electron_MCRD_file):
     efile = electron_MCRD_file
     return efile, name, pdg, vi, vf, p, Ep, nhits, X, Y, Z, E, t
 
+@pytest.fixture(scope='session')
+def mc_particle_and_hits_nexus_data(electron_MCRD_nexus_file):
+    X = [0.07783413, 0.07652718, 0.04626106, 0.03018976, 0.02231551]
+    Y = [-0.00466795, -0.02718811, -0.03123637, -0.04843096, -0.04031885]
```
jacg: Tabulating these values according to our style would be good.

invisible_cities/io/mchits_io.py
```diff
@@ -63,6 +75,59 @@ def read_mctracks (mc_table: tables.table.Table,
...
+    all_events = {}
+    particles = {}
+
+    ihit = 0; ipart = 0
+    if(event_range[0] > 0):
+        ihit = h5extents[event_range[0]-1]['last_hit']+1
+        ipart = h5extents[event_range[0]-1]['last_particle']+1
+
+    for iext in range(*event_range):
+        if(iext >= len(h5extents)):
```
jacg: Do **NOT** put parentheses around conditions in `if` and `while` in Python!

invisible_cities/io/mchits_io_test.py
```diff
@@ -50,3 +52,45 @@ def test_load_mcparticles(mc_particle_and_hits_data):
     assert np.isclose(Z,hZ).all()
     assert np.isclose(E,hE).all()
     assert np.isclose(t,ht).all()
+
+def test_load_mchits_nexus(mc_particle_and_hits_nexus_data):
+    efile, _, _, _, _, _, _, _, X, Y, Z, E, t = mc_particle_and_hits_nexus_data
```
jacg, line 57: This is begging for a namedtuple.

jacg:
`read_mctrack_nexus` is completely meaningless to me without setting aside (at least) half an hour without distractions and carefully parsing every single line of it in order to understand minute details which hopefully will then add up to some coherent high-level whole.

It would be *really* helpful if the `read_mctrack_nexus` could be expressed in a way that promotes some high-level view of what is going on.

If this is approved without such a rewrite, then please open a *refactor `read_mctracks_nexus` to emphasize meaning* issue at the same time, and I'll get around to it sometime.

paolafer:
I'm expanding this writer to read the rest of MC true info, but I don't object to approving this PR first, so that people can use the code. In the meanwhile, we can improve style, names, and rewrite things at the same time I add other functionalities (e.g., I'm strongly against *mctracks* name...).

invisible_cities/conftest.py
paolafer: This is how the name of the particle is saved in the hdf5 file we create as the nexus output. This particular datum is a `char particle_name[20]`. I would prefer very much this to be a `str`, can you point me towards the correct way to do it?

paolafer:
About MC truth monitoring, ages ago I was asked to write a notebook, which is [here](https://github.com/paolafer/ICARO/blob/TruthMCMonitoring/icaro/analysis/monitoring/MCTruthMonitor.ipynb), in case you want to start from something. It will have to be updated to the new data format and expanded, anyway.

jjgomezcadenas:
> I'm expanding this writer to read the rest of MC true info, but I don't object to approving this PR first, so that people can use the code. In the meanwhile, we can improve style, names, and rewrite things at the same time I add other functionalities (e.g., I'm strongly against mctracks name...).

So I propose to approve PR, then open immediately a new one, as suggested by @jacg and let Paola try her hand at improving readability, style, and then adds missing information. Alternatively, we can wait and add Paola's work to this PR.
Opinions, in particular @jacg?

jacg:
If it works and is needed, approve it. But do open another PR immediately to register our intent to pay back the technical debt.

I do think that the minor issues which require no thought (alignment, superfluous parentheses) can be fixed before approval. I haven't had time to think about why the `bytes` are there.

jjgomezcadenas:
> If it works and is needed, approve it. But do open another PR immediately to register our intent to pay back the technical debt.

@jerenner do we need this approves straight away? If we can wait, I propose that @paolafer tries her hand at improving code (with the help of @jacg as needed) and also adds missing functionality (we really want to read back all the MC). 

@jmbenlloch can you take a look to the **bytes** issue? It could be that the field can be read directly as a string using pyTables. But if it introduces too much overhead, we could leave it at that. 


**everyone** We need to agree on a way to store documentation notebooks linked to IC. The issue is becoming pressing, notebooks are proliferating all over the place.

jerenner:
Ok - I will work on fixing the minor things Jacek has pointed out (parentheses in `if` and `while` statements and tabulation).  Note that I constructed the test and fixture similar to the ones that were already there for `test_load_mchits` and `test_load_mcparticles`, so these may also need to be looked at again since issues are being raised with these new tests.

jerenner:
@jjgomezcadenas Others have recently asked me for this code, so I think the sooner it is approved the better (that way they can use it directly from master and don't have to take it from my branch).  On the other hand, I don't think it needs to be approved right away if we're going to make important improvements soon.

jacg:
> Note that I constructed the test and fixture similar to the ones that were already there for test_load_mchits and test_load_mcparticles

I noticed that you copied the parametrization of the fixtures, but only ever provide one parameter in each. If you don't expect to add any more parameters then just get rid of the parametrization and hard-wire the value inside the fixture.

invisible_cities/conftest.py
paolafer: @jmbenlloch suggests to use `decode('utf-8')` to convert `bytes` to `str`. @jerenner, if you're still implementing the other changes, could you try this one, too?

jerenner:
I have made an attempt to fix some of the formatting issues brought up by @jacg and also removed the fixture parameters for the test file names.

In terms of reorganizing `read_mctracks_nexus`:
- the MC hit and particle information is stored by Nexus in 3 tables: `h5extents`, `h5hits`, and `h5particles`
- the hits (in `h5hits`) and particles (in `h5particles`) are stored with increasing index numbers (`ihit` and `ipart`) that are not reset after each event
- the indices of the final hit and particle of each event are stored in `h5extents` (1 entry per event)

Therefore the code attempts to go through the tables, picking out the hits and particles from each event as specified in `h5extents`:
1.  `ihit` and `ipart` are set to 0, or if the first event to be read is not the first one stored, they are set to the last indices retrieved from `h5extents` for the event just before the one to be read first
2.  A loop is made over each event in the event range:
2a)  For each event, an empty dictionary `current_event` is created 
2b) The indices of the last hit and particle in this event are extracted from `h5extents`
2c) For each particle in the event an entry in the `current_event` dictionary is created (indexed by its track ID `itrack`) and the corresponding information extracted from `h5particle`
2d) Each hit in the event is appended to the list `hits` of the particle to which it belongs
2e)  The event `current_event` is added to the `all_events` dictionary indexed by its event number, which is extracted from `h5extents`
3.  The dictionary `all_events` is returned

The only thing that comes to mind right now in terms of making this more readable would be adding individual functions to handle parts 2c and 2d.

invisible_cities/conftest.py
paolafer: We've found a problem using decode as it is, since our byte-strings have fixed length and the "empty" characters at the end are pointless hexadecimal ones. Apparently `decode` doesn't work well on them. So, I've found a (very ugly) way, which, nevertheless, works:

`import re`

`re.sub(r'[^\x20-\x7e].*', '', my_byte_string.decode('utf-8','ignore'))`

@jacg, any comment?

paolafer: In the end, @jmbenlloch found a way to write strings in nexus that are read as python-strings by pytable. Case closed, hurray!

jahernando:
@jerenner I found the code to read the MC events a little cumbersome. Maybe I am missing something, but I will propose something like this:

1.  The client can ask for a MC-event, providing either an event (an id), or a range of events, or either a list of events (a list of events ID), and a function returns an iterator instead of a dictionary (I guess we want to study the events iteratively).  This function is in fact the only one 'public' for the user.
2. The elements of the iterators are (event number, (list of MC-Particles)). In this case, we also do not need the second dictionary of MC-particles. Just simple a list of MC particles.

Unfortunately we need this PR soon. There is a master student at Santiago, another Gonzalo, who needs this PR to read the MC data.

jerenner:
@jahernando thanks for looking at the code.  The Nexus reader functions were written to match those that were already there before: `load_mchits`, `load_mcparticles`, and `read_mctracks` - the corresponding functions `load_mchits_nexus`, `load_mcparticles_nexus`, and `read_mctracks_nexus` return the same data structures but operate on Nexus HDF5 files.

I was discussing this today with @jjgomezcadenas and at the moment we were not clear on what each function returned.  To clarify, the two functions that would be called by the user are:
- `load_mchits_nexus`: returns a dictionary indexed by event number containing lists of MCHit objects for that event: {evt_number, [hits]}
- `load_mcparticles_nexus`: returns a dictionary indexed by event number containing dictionaries of MCParticle objects indexed by track number: {evt_number, {track_id, MCParticle}}

Comments on the specific points raised by @jahernando:
1.  One can already specify an event range for these two functions.  Why would we want to be able to specify a list of event IDs rather than a range?  If these are to be the actual "event numbers," we can't actually know these exactly before reading the events because these are the keys returned in the dictionary.  Also, @paolafer and I were talking about this and were wondering why we would need to return the iterator directly - for example as `load_mchits_nexus` is being used right now in "detsim", one can easily iterate over the dictionary, for example:
```
for evt_number, mchits in mchits_dict.items():
...
```
2.  If we do not return the dictionary of MCParticle objects indexed by track number but just return a list of MCParticle objects, we will lose that information (unless we choose to store it in the MCParticle itself).  However, since it is just an index, it may not be so important unless it corresponds to something specific that happened in the simulation that we may want to know about (does it correspond to the order in which the particles were produced?)

Based on this, what changes if any should be made?  Note that anything we change for the `_nexus` functions we should change also for the original MC readers (assuming we will still be using them).

jahernando:
Hi @jerenner, 

For me it is ok to return the dictionary {event:[hists]} or {event:{track_id, MCparticle}}. I was wondering if this requires to reserve a large amount of memory and could affect the performance of manipulating the data. I was thinking that an iterator that has in memory only the data of the event could be more time-computing efficient, but I am not an expert here.

Respect the id-track in the second dictionary, -my ignorance- why we need that info? there is a unique id-track for every MC-particle?

Maybe we should go ahead with the current interface {event: [hits]} and {event:{id-track, MC-particle}} and we can always re-factor later in the case it is needed. 

I will go for approving the PR and open an issue, as was discussed in the threat, for further improvements,  what still requires some work before approving @jerenner @paolafer?

paolafer:
I'm already working on an extension of this PR, which implies also a refactoring of some underlying structure: I will submit a PR when I have it functioning and we can discuss it. In the meanwhile, please, go ahead with the approval of Josh's PR.

jahernando (approved):
The PR is approved, subjected to re-factoring, @paolafer working on

jerenner:
> there is a unique id-track for every MC-particle?

I think there is a unique ID track for every MC particle within the same event but it is not unique across events.  I just realized that we store the particles in a dictionary based on particle ID so that later when we want to add the hits to each particle we can easily do so - the HDF5 hits contain the ID of the particle they belong to, and using a dictionary with the key = the ID, one can easily access the correct particle and add to its list of hits.  

After this has been performed, we could just add a list of particles to the event dictionary rather than the dictionary of particles itself.  In this case we would lose this ID information, and though I can't think of a good reason to keep it, if it is not costing us much to keep it then I can't think of a good reason to get rid of it either.  Is returning a dictionary much more costly than returning a list?

jjgomezcadenas:
Please @gonzaponte or @jmbenlloch, MERGE.

jerenner:
Paola has just noted that we should keep the particle ID information and store in class MCParticle the ID of the parent particle so that we know which particle was produced by which.

In this case, we would either have to keep the dictionary of particles or store the particle's own ID in the MCParticle class (or both).  At this point keeping the dictionary seems to make sense, as if we need to quickly access a particle by its ID we can do so (otherwise we would have to search through a list of particles to find it).

jjgomezcadenas:
I think there is nothing wrong with the dicts. Please usual suspects (@gonzaponte @abotas @jmbenlloch) merge this PR

gonzaponte:
Since there is another PR extending this one, ***please*** maintain the coding style we have been sticking to. In this case: separate functions with two blank lines (some of them have been removed), align values in arrays according to decimal points and commas, separate arguments in a function call with a comma ***and*** a blank space (`a, b` rather than `a,b`) and align brackets when appropriate (minor).

I will merge it so we don't get stuck with cosmetics but try to comply with these simple rules.

gonzaponte:
> Note that at the moment the PDG code is assumed to be 0.

@jerenner @paolafer, has this issue with the PDG code been solved or is it going to be merged as it is now?

paolafer:
Since we have already the particle name, the PDG code is not stored in the nexus files, since it's redundant. In the next PR, the MCParticle class will be changed accordingly.

PR #440, opened 2018-02-15
Add sensor calibration cities

andLaing:
New scripts and functions for the calibration of the pmts and sipms.

    1. adds pmtgain and corresponding test
    2. adds sipmgain and corresponding test
    3. functions common to all or some calibration 'cities' centralised
    in calib_functions.py
    4. tests for calib_functions added in calib_functions_test
    5. sensorSpeResponse class added to fit_functions with various fit
    functions relevant for extracting calibration parameters.


jjgomezcadenas (commented):
This PR is **huge**. It includes a large number of functions and several drivers (cities). A systematic review of this PR may take as long as the time that has been invested to develop the code.

At the same time, this PR needs to be urgently reviewed and approved since it implements essential code for calibration of sensors that we need very soon.

This leaves few options: either to delay approval until full review (and that can take a long time), or to approve with a light review (and that incurs in technical debt).

This situation could have been avoided by splitting the PR in smallish PRs and submitting them sequentially.

jjgomezcadenas (commented):
class sensorSpeResponse has been added to fit_functions.py

1. The name does not follow our usual convention for classes: it should be:

SensorSpeResponse

2. There are no tests exercising the class. Tests are essential here for two reasons:

a) To demonstrate that the functions are correct.
b) To illustrate what the functions do.

Without the tests, the reviewer is lest to her own devices to either parse the code, write tests, or both. 

Here is my proposal:

1) Rename fitting class to SensorSpeResponse
(suggestion: code readability improves with white space, for example:

>s2 = self.s0*self.s0 + self.nGau*s1*s1

reads worse than:

> s2 = self.s0 * self.s0 + self.nGau * s1 * s1

2) Prepare an example Notebook, where it is possible to follow an example, top-to-bottom of the works of the fitting class over a credible toy example (or a real example from data, or both). The NB will be very useful as "active" documentation of the code.

3) Translate the NB into tests that exercise and document the code. The tests are essential to illustrate the workings of your functions and to ensure correctness of the code.

andLaing:
I'm not sure I agree that the PR is huge, nor even particularly big. It's two, virtually identical, scripts/cities/whatever and a handful of useful functions.

As to the new fitting class, I'll get to the name changes and tidying. The tests are a bit awkward with fake data since the response function isn't so simple but in principle I could write them using the spectra (one or two probably enough) that I made when I was testing the functions. I have a script that uses one of the functions of the class on the sipms, I suppose that can be a starting point

jjgomezcadenas:
>I'm not sure I agree that the PR is huge, nor even particularly big. It's two, virtually identical, scripts/cities/whatever and a handful of useful functions.

Well, the fitting class and functions are not trivial! It takes an effort to understand what is going on (tests will help to document what the functions actually do). Then the calib_functions.py module contains some redundancy. And then two cities which need to be understood... it is taking this reviewer quite an effort to digest everything. Perhaps the PR is not long and I am too slow. 

>As to the new fitting class, I'll get to the name changes and tidying.
OK, thanks, this is useful. 

>I could write them using the spectra (one or two probably enough) that I made when I was testing the functions.

I think this is a good idea. Go ahead!

jjgomezcadenas:
@andLaing @jahernando @gonzaponte @jacg  I would like to discuss the calibration of PMTs and SiPMs.

For SiPMs we have:

    def calibrate_sipms(sipm_wfs, adc_to_pes, thr, n_MAU=100):
        """
        Subtracts the baseline as an average of the waveform and
        uses a MAU to set the signal threshold (thr, in PES).
        """
        thr  = np.full(sipm_wfs.shape[0], thr)[:, np.newaxis]
        bls  = subtract_baseline(sipm_wfs)
        cwfs = calibrate_wfs(bls, adc_to_pes)
        MAU  = np.full(n_MAU, 1 / n_MAU)
        mau  = signal.lfilter(MAU, 1, cwfs, axis=1)
        return np.where(cwfs > mau + thr, cwfs, 0)

The default for `subtract_baseline` is to compute the mean. On the other hand, since the SiPMs are positive signals, **the mean is always a biased estimator of the pedestal** (unlike the case of the Raw waveforms, where the mean is a good estimator of the pedestal because the total area is zero).

As @andLaing has already pointed out, it is dubious that the MAU threshold is helping us very much. To conclude, I think the approach here is not correct. 

@andLaing has proposed the use of the *mode*, which is a good estimator of the pedestal for positive signals, but the mode calculation is awfully long. I believe that we can use instead the *median*, which is fast and a good estimator in this case. 

In the case of the PMTs we first do the deconvolution (which includes the pedestal subtraction, in this case using the mean which is a good estimator), and then we compute the calibrated waveforms (and sums) with and without MAUS. I believe the calibrated waveforms with MAUS are not needed (any one has a use for them?) and thus we can avoid that calculation and save some time. I also believe that the calibrated sum with the MAU is still useful for S1 searches (not for S2). 

A discussion with toy examples is here:

https://github.com/jjgomezcadenas/ICDEV/blob/master/reco/calib_sensors_functions.ipynb

Comments?

jjgomezcadenas:
This discussion is of interest for (at least): @andLaing, @gonzaponte @jahernando @jacg @paolafer 

I propose the following changes in  *calib_sensors.py*:

Currently in code:

    class BlsMode(Enum):
        mean = 0
        mau  = 1

Propopsed:

    class BlsMode(Enum):
        mean   = 0
        median = 1
        mode   = 2

Currently in code:

    def mode(wfs):
          gen_m = lambda x: np.bincount(x[x>=0]).argmax() if x[x>=0].size != 0 else -1
          return np.apply_along_axis(gen_m, 1, wfs)

Proposed: 

    import scipy.stats as scs
    def mode(x, axis=0):
        m, c = scs.mode(x, axis=axis)
        return m

Currently in code:

    def subtract_mode(wfs):
          return wfs - mode(wfs)[:, np.newaxis]

    def subtract_mean(wfs):
          return wfs - np.mean(wfs, axis=1)[:, np.newaxis]

    def sipm_subtract_mode_and_calibrate(sipm_wfs, adc_to_pes):
          """Computes mode pedestal"""
          bls = subtract_mode(sipm_wfs)
          return calibrate_wfs(bls.astype(np.float), adc_to_pes)

    def sipm_subtract_mean_and_calibrate(sipm_wfs, adc_to_pes):
          bls = subtract_mean(sipm_wfs)
          return calibrate_wfs(bls, adc_to_pes)

Propose: replace by compact version (see below)

Currently in code 

    ## Not 100% sure if we want to keep this?!?!?
    def sipm_subtract_mean_mau_and_calibrate(sipm_wfs, adc_to_pes, n_MAU=100):
          bls = subtract_baseline(sipm_wfs, bls_mode=BlsMode.mau, n_MAU=n_MAU)
         return calibrate_wfs(bls, adc_to_pes)

Proposed: eliminate

Currently in code

    def sipm_subtract_mode_select_signal(sipm_wfs, adc_to_pes, thr):
          zwfs = sipm_subtract_mode_and_calibrate(sipm_wfs, adc_to_pes)
          return np.where(zwfs > thr, zwf, 0)

There is bug in the code above (it should have been zwfs). Code does not have a test. Replace by compact version

Currently in code

    def subtract_baseline_and_calibrate(sipm_wfs, adc_to_pes):
          """Computes pedetal as average of the waveform. Very fast"""
          bls = subtract_baseline(sipm_wfs)
          return calibrate_wfs(bls, adc_to_pes)

Yes it is fast because it uses the mean by default. Replace by version allowing to choose method of pedestal computation. 


Currently in code

    def subtract_baseline_mau_and_calibrate(sipm_wfs, adc_to_pes, n_MAU=100):
          """Computes pedetal using a MAU. Runs a factor 100 slower than previous"""
          bls = subtract_baseline(sipm_wfs, bls_mode=BlsMode.mau, n_MAU=n_MAU)
          return calibrate_wfs(bls, adc_to_pes)

Proposal: eliminate. Using MAU for subtraction leads to biases. 

**Proposed code:**

Provide uniform interface to compute means, medians and modes for waveforms (objects of type
wfms[sensors][waveform]

     def means_from_waveforms(wfms):
            return np.mean(wfms, axis=1)

    def medians_from_waveforms(wfms):
             return np.median(wfms, axis=1)

    def modes_from_waveforms(wfms):
          return mode(wfms, axis=1)

Subtract baseline

    def subtract_baseline(wfs, *, bls_mode=BlsMode.mean):
         if   bls_mode is BlsMode.mean:
             return wfs - array_to_col_vector(means_from_waveforms(wfs))
        elif bls_mode is BlsMode.median:
             return wfs - array_to_col_vector(medians_from_waveforms(wfs)) 
        elif bls_mode is BlsMode.mode:
             return  wfs - modes_from_waveforms(wfs) 
        else:
             raise TypeError(f"Unrecognized baseline subtraction option: {bls_mode}")
        return bls

where (defined in core, wraps somewhat odd numpy signature):

    def array_to_col_vector(x):
          return x[:, np.newaxis]


Finally, the crucial functions, used by the cities are:

    def calibrate_wfs(wfs, adc_to_pes):
          adc_to_pes = adc_to_pes.reshape(adc_to_pes.size, 1)
          ok         = adc_to_pes > 0
          return np.divide(wfs, adc_to_pes, out=np.zeros_like(wfs), where=ok)

This is OK

    def calibrate_pmts(cwfs, adc_to_pes,  n_MAU = 100, thr_MAU = 3):
   
(function body omitted). This is OK


    def pmt_subtract_mau(cwfs, n_MAU = 100):
   
        MAU = np.full(n_MAU, 1 / n_MAU)
        mau = signal.lfilter(MAU, 1, cwfs, axis=1)

        return cwfs - mau

(eliminate)

Then, for SiPMs:

    def calibrate_sipms(sipm_wfs, adc_to_pes, thr, n_MAU=100):
     """
        Subtracts the baseline as an average of the waveform and
        uses a MAU to set the signal threshold (thr, in PES).
     """
  
        thr  = np.full(sipm_wfs.shape[0], thr)[:, np.newaxis]
        bls  = subtract_baseline(sipm_wfs)
        cwfs = calibrate_wfs(bls, adc_to_pes)
        MAU  = np.full(n_MAU, 1 / n_MAU)
        mau  = signal.lfilter(MAU, 1, cwfs, axis=1)
        return np.where(cwfs > mau + thr, cwfs, 0)

As discussed, the mean should not be used here and the mode is expensive. Thus I propose to use the median. Also, as noted by @andLaing I think the MAU is of no help here, and time-expensive. Thus, a much simpler version:

    def calibrate_sipms(sipm_wfs, adc_to_pes, thr):
        """
        Subtracts the baseline as an average of the waveform and
        uses a MAU to set the signal threshold (thr, in PES).
        """
        thr  =  array_to_col_vector(np.full(sipm_wfs.shape[0], thr))
        bls  = subtract_baseline(sipm_wfs, bls_mode=BlsMode.median)
        cwfs = calibrate_wfs(bls, adc_to_pes)
        return np.where(cwfs > thr, cwfs, 0)

My proposal is:

- to test the mean vs median with actual data. 
- to refactor the module along the lines described above
- to add tests as needed.

As for the cities: PMTs remain unchanged and the interface of calibrate_sipms almost unchanged (n_MAU dissappears), thus it should be easy to adapt. I have not had the time yet to find out the calibration functions that @andLaing uses in his new calibration cities but it should be straight forward to adapt.

andLaing:
So I'll look at this in detail in a bit but as to the use of
scipy.stats.mode: I tested this and it is around an order of magnitude
slower than what I coded (I have a notebook somewhere where I tested it). I
checked around and it is well documented that it's slow. The bincount
method is a very similar speed to the mean calculation at least in its old
form. I'll rerun the notebook to check but I really don't think the mode
loses us much.

I think the method with the dict of functions is more pythony than your
suggested if else chain too...

jjgomezcadenas:
>So I'll look at this in detail in a bit but as to the use of
scipy.stats.mode: I tested this and it is around an order of magnitude
slower than what I coded (I have a notebook somewhere where I tested it).  I
checked around and it is well documented that it's slow. The bincount
method is a very similar speed to the mean calculation at least in its old
form. I'll rerun the notebook to check but I really don't think the mode
loses us much.

Yes, I just checked it and you can find my test in the same NB above, I have christened your imp (pretty smart, BTW, although it takes a wee bit to parse and decode), and is an order of magnitude faster than SciPy version, but still one order of magnitude slower than the median and the mean. Thus, we still need to figure out if the use of the mode (which appears to be the best estimator, as you pointed out) implies a performance penalty (in this case we could resort to the median). 

>I think the method with the dict of functions is more pythony than your
suggested if else chain too...

You mean that passing around:

    sipm_processing = {
        'gain'  : subtract_mode,
        'gainM' : subtract_mean,
        'pdf'   : sipm_subtract_mode_and_calibrate,
        'pdfM'  : sipm_subtract_mean_and_calibrate,
        'pdfMM' : sipm_subtract_mean_mau_and_calibrate,
        'data'  : sipm_subtract_mode_select_signal,
        'dataM' : calibrate_sipms
        }

Is more pythonic that using:

    def subtract_baseline(wfs, *, bls_mode=BlsMode.mean):
        if   bls_mode is BlsMode.mean:
            return wfs - array_to_col_vector(means_from_waveforms(wfs))
        elif bls_mode is BlsMode.median:
            return wfs - array_to_col_vector(medians_from_waveforms(wfs)) 
        elif bls_mode is BlsMode.mode:
            return  wfs - modes_from_waveforms(wfs) 
        else:
            raise TypeError(f"Unrecognized baseline subtraction option: {bls_mode}")
        return bls

Well, perhaps @gonzaponte and @jacg want to comment here. But my impression is that passing a dict around when you could get by with a simple enumeration is an overkill. Anyway, this is a minor point. 

BTW: it is better to comment directly in the GitHUb page (you can use markdown and such) than to reply to e-mail (I used to do the same and lost many Karma points before I finally got around to write in Github nice cells).

andLaing:
I did some tests of my own with various different data sets and while the mean is significantly faster the median doesn't gain so much and the gains don't scale so well, with the longest possible buffer it basically gives me the same as the mode. It doesn't hurt to add the option though and then I can do some tests with the pdf generation to see what sort of spectra it gives me.

I tried to attach the notebook but it doesn't like the file type. I don't want to push at the moment because I'm in the middle of sorting tests that I wrote yesterday and that are giving me results that I don't understand.

jjgomezcadenas:
> I did some tests of my own with various different data sets and while the mean is significantly faster the median doesn't gain so much and the gains don't scale so well, with the longest possible buffer it basically gives me the same as the mode. It doesn't hurt to add the option though and then I can do some tests with the pdf generation to see what sort of spectra it gives me.

Great. 

>I tried to attach the notebook but it doesn't like the file type. I don't want to push at the moment because I'm in the middle of sorting tests that I wrote yesterday and that are giving me results that I don't understand.

Notebooks are not supposed to go to IC (not yet at least). We must find a solution, in the sense that I would like very much to have documentation NBs in IC. They are extremely useful, for instance in a discussion like this one, but at the same time we have to be careful, they tend to proliferate.

May I suggest that you (and everybody) create a gitub repo call ICDOC or similar (I have something called ICDEV) where you put your NBs? Then send the link around.

jacg:
I am very busy with non-IC things, so you should expect very little input from me on this. Understanding the details of the calculations you are performing, why you do them, and what the best choice is, would require a significant investment on my part before I could make any sensible contribution, so I am not even going to try: I just don't have the time.

Fishing out two points on which am qualified to have an immediate opinion:

1. Using dispatching dictionaries to fake switch statements is a standard Python idiom. I prefer it to a chain of `if`-`else`s, as long as the dictionary's keys aren't contrived. This is a general opinion, which might be overridden by the details of any specific case. Given that the example above contrasts a dictionary with seven entries to an `if-else` with only 3 valid branches, I don't have enough to go on in this case.

2. Please don't create any new repositories for notebooks. Just create a branch in IC or ICARO and put them there. As long as such branches don't get merged into `nextic/master` it will be OK.

3. In the spirit of not falling victim to NIH syndrome and strengthening the community tools, can we understand *why* the scipy solution is an order of magnitude slower that the home-grown one, and whether ours can be contributed to scipy? Yes, it does require some of our precious resources to look into this, but please appreciate just how much of our precious resources have been saved by the existence of scipy, numpy, cython, etc. etc. and the vast resources that others pumped into them. By fostering a culture of feeding back to this community and ecosystem, we are saving our resources in the long run.

jjgomezcadenas:
>Using dispatching dictionaries to fake switch statements is a standard Python idiom. I prefer it to a chain of if-elses, as long as the dictionary's keys aren't contrived. This is a general opinion, which might be overridden by the details of any specific case. Given that the example above contrasts a dictionary with seven entries to an if-else with only 3 valid branches, I don't have enough to go on in this case.

OK, we will come up with one or another solution depending on how complex the code is, but it looks we can stay simple. 

>Please don't create any new repositories for notebooks. Just create a branch in IC or ICARO and put them there. As long as such branches don't get merged into nextic/master it will be OK.

Unless I hear a complains I will create a branch in IC called docs where I vote to commit documentation in general and notebooks in particular. 

>In the spirit of not falling victim to NIH syndrome and strengthening the community tools, can we understand why the scipy solution is an order of magnitude slower that the home-grown one, and whether ours can be contributed to scipy?

This is a very good point, I was discussing about this with @andLaing earlier. It turns out that the SciPy mode is a very generic algorithm (it computes the mode in all cases and provides also the frequency bins) while the home-solution just computes the mode for positive values (which are close-but-no-quite-always) the case for us. So our mode has some limitations in generality that most likely make it up for its better performance.

jacg:
An order of magnitude might be interesting to others. 

I'll just leave this linke here, for reference: [Contributing to SciPy](https://docs.scipy.org/doc/scipy/reference/hacking.html)

jjgomezcadenas:
Tests are failing...

andLaing:
I can't get it to fail in local but I've given a bit more room in the allclose tolerance just in case.

jjgomezcadenas:
I have run tests in local and get this failure:

```
 fn = <bound method SensorSpeResponse.min_integ_gaussians of     <invisible_cities.core.fit_functions.SensorSpeResponse object at 0x111004390>>
pars = array([  1.00000000e+04,   0.00000000e+00,   2.00000000e+01,
         1.50000000e+00,   5.00000000e+00,   1.00000000e+01]), b0 = [(0.0, -100.0, 0.0, 0.0, 0.001, 0.001), (1e+99, 100.0, 10000.0, 100.0, 10000.0, 10000.0)]

E       AssertionError:
E       Not equal to tolerance rtol=0.05, atol=0.05
E
E       (mismatch 16.66666666666667%)
E        x: array([  1.002977e+04,  -9.505391e-02,   1.985073e+01,   1.513685e+00,
E                4.965253e+00,   1.045889e+01])
E        y: array([  1.000000e+04,   0.000000e+00,   2.000000e+01,   1.500000e+00,
E                5.000000e+00,   1.000000e+01])

invisible_cities/core/fit_functions_test.py:283: AssertionError

andLaing:
Looks like the fit to the pedestal zero depends on the machine, might have to rethink the test instead of just allowing a worse fit.

jjgomezcadenas:
> Looks like the fit to the pedestal zero depends on the machine, might have to rethink the test instead of just allowing a worse fit.

Agreed, we need to come with a robust test here.

jjgomezcadenas:
>Looks like the fit to the pedestal zero depends on the machine, might have to rethink the test instead of just allowing a worse fit.

Any progress there? I have started to work in the refactoring of the calibration functions, but is more work than I thought. Therefore I would like to approve this PR as soon as we check that all tests pass (ncluding the one that fail randomly) and I will submit a separate PR for the refactoring of the calibration functions.

jjgomezcadenas:
I am done with my revision of this PR. The last set of commits target mostly calib_sensor_functions and related tests. The calib_sensors functions has been refactored. Three important points are to be taken into account:

1. The PMTs and SiPMs raw waveforms are int16 and tests must send int16 waveforms. This is important because the fast version of the mode expects an int array. 

2. When calibrating one is dividing by adc_constants which are floats, and thus the waveform must be first converted in float (this is done in calibrate_wfs

3. MAUS as no longer used for SiPMs (they always bias the energy measurement and are of no use in the case of SiPMs). They are still used in PMTs, where they can help for S1 searches.

The refactoring includes an expanded test suite. 

@gonzaponte you are in charge of the final tidying up, review and approval.

gonzaponte:
1. Can we rename  `mode` to `scipy_mode` (or something similar) and `amode` to `mode`? Since it is our implementation what we are going to use, I prefer to give it the "most obvious" name.
2. I don't think those functions `to_row/col_vector` are really useful. I am more in favor of people understanding `numpy`'s interface for array manipulation. But I can live with it.
3. I don't like the function `relative_difference` either. Too much code for such a simple thing and it is only used in one way.
4. Do we prefer to have the spe response functions in `fit_functions` rather than in their own module? They seem very specific, but it is not a big deal to leave them there.
5. In `sierpe/waveform_generator.py` the function `simulate_pmt_response` must be renamed, as this name is already in use somewhere else. Please, suggest another name.
6. Why did you went back to individual functions rather than a unique one when doing baseline subtraction?
7. I would like to give better (more readable) names to the keys of the dictionary `sipm_processing`. Ideas?
8. @andLaing Is the covariance matrix going to be used for something ***in this PR***? Otherwise I would put it in another PR to avoid confusion.

jjgomezcadenas:
> Can we rename mode to scipy_mode (or something similar) and amode to mode? Since it is our implementation what we are going to use, I prefer to give it the "most obvious" name.

OK

>I don't think those functions to_row/col_vector are really useful. I am more in favor of people understanding numpy's interface for array manipulation. But I can live with it.

I, on the other hand, had problems with parsing the numpy interface that appears not obvous to me. So thanks, leave with it. 

>I don't like the function relative_difference either. Too much code for such a simple thing and it is only used in one way.

Don't use it then. But there is nothing wrong with the function, it serves a purpose and I happen to like it. 

>Do we prefer to have the spe response functions in fit_functions rather than in their own module? They seem very specific, but it is not a big deal to leave them there.

I would prefer to have them in their own module. 

>In sierpe/waveform_generator.py the function simulate_pmt_response must be renamed, as this name is already in use somewhere else. Please, suggest another name.

Why? Names should be specific to a given module. They can always be imported with "as"

>Why did you went back to individual functions rather than a unique one when doing baseline subtraction?

To keep backward compatibility with code elsewhere. Passing a dict is not such a bad idea after all.

>I would like to give better (more readable) names to the keys of the dictionary sipm_processing. Ideas?

I agree with this. Here my five cents:


    sipm_processing = {
        'gain'  : subtract_mode,
        'gainM' : subtract_mean,
       'pdf'   : sipm_subtract_mode_and_calibrate,
       'pdfM'  : sipm_subtract_mean_and_calibrate,
       'pdfMe' : sipm_subtract_median_and_calibrate,
       'dataM' : calibrate_sipms
      }

gain -> Mode
gainM -> Median (we should pass around the median, rather than the mean)
pdf -> pdfMode
pdfM -> pdfMean
pdfMe -> pdfMedian
dataM -> Calibrate

But let's see what @andLaing says

andLaing:
Hi
I'll implement some of the simpler changes re: names etc, I have no problem with them.

As far as the SensorSpeResponse class goes, I put it there so that all the fitting functions were grouped together. Whatever you prefer, I can put it in a separate file in the same folder (I would put it in calib but I understand that with the liquid cities that folder is being removed).

waveform_generator.simulate_pmt_response seems completely unrelated to this PR and I'm not entirely sure what its use is to be honest. I'd suggest removing it for now JJ. It doesn't seem to do anything to do with calibration which was the idea of the PR. I can do the same with the addition of the covariance matrix to FitFunction. It's a small change with a generally related motivation which is why I put it but it can be added separately and approved in no time I assume.

I'm not particularly happy with JJ's suggestions for the key names. I think they should represent usage rather than the function name. If we want to extract gain we have to use the low level functions etc but a user of the configs for such things would probably be more motivated by a related term. Also, JJ, "we should pass around the median, rather than the mean", what? I have no idea what you want to say. How about:

sipm_processing = {
'mode_ped_gain' : subtract_mode,
'median_ped_gain' : subtract_median,
'mode_ped_pdf' : sipm_subtract_mode_and_calibrate,
'mean_ped_pdf' : sipm_subtract_mean_and_calibrate,
'median_ped_pdf' : sipm_subtract_median_and_calibrate,
'mode_ped_full_selection' : calibrate_sipms
}

They're maybe a bit long, that's the only thing

jjgomezcadenas:
>waveform_generator.simulate_pmt_response seems completely unrelated to this PR and I'm not entirely sure what its use is to be honest. I'd suggest removing it for now JJ. It doesn't seem to do anything to do with calibration which was the idea of the PR. 

Waveform generator is used for numerous tests of sensor_calibration.py. It is therefore useful and should be left in place.

>How about:

    sipm_processing = {
    'mode_ped_gain' : subtract_mode, 
    'median_ped_gain' : subtract_median,
    'mode_ped_pdf' : sipm_subtract_mode_and_calibrate,
    'mean_ped_pdf' : sipm_subtract_mean_and_calibrate,
    'median_ped_pdf' : sipm_subtract_median_and_calibrate,
    'mode_ped_full_selection' : calibrate_sipms
    }

I like this. Much better than mine.

andLaing:
Gonzalo and I had a bit of a discussion offline and we agreed on:

sipm_processing = {
    'subtract_mode'              : subtract_mode,# For gain extraction
    'subtract_median'            : subtract_median,# For gain extraction
    'subtract_mode_calibrate'    : sipm_subtract_mode_and_calibrate,# For PDF calculation
    'subtract_mean_calibrate'    : sipm_subtract_mean_and_calibrate,# For PDF calculation
    'subtract_median_calibrate'  : sipm_subtract_median_and_calibrate,# For PDF calculation
    'subtract_mode_zs'           : calibrate_sipms# For data processing
    }

Adding the comments for additional info.

I've made the name changes. What is the final decision on SensorSpeResponse and the covariance matrix addition?

A

gonzaponte:
> What is the final decision on SensorSpeResponse and the covariance matrix addition?

If it has no impact in this PR I would put it in a micro-PR with maybe some small test to ensure it is present in the output of `fit`.

andLaing:
OK, covariance matrix removed.

Anything else?

gonzaponte:
I am going to uniform style and refactor the spe response class into functions. Probably going into their own module in reco.

After I finish, I will ask for your ok and then I will rewrite commit history to make it more clear (without changing the code).

jjgomezcadenas (commented):
Lots of polishing work that really improve the readibility and overall quality of the code. 

Concerning input/output: often (but not consistently) we write what the function takes and what it returns. This is good and fine BUT:

1) As mentioned, not done consistently, we often skip for small functions (it would add to noise, actually).
2) When we do it, in fact, the doc string becomes verbose.

PROPOSAL: USE TYPE ANNOTATIONS consistently. At the very least they are equivalent to the doc string and forces us to think what come in and out of the function. And then, eventually we will be able to use tools one day to check the correctness of the types involved.

Otherwise I am quite happy with the way this PR is evolving, and the demonstration that team work is possible even when the team involves three alien races from Scotland, Galicia and Cartago.

jjgomezcadenas:
P.D. I am happy to approve this. Are you guys OK to proceed, @andLaing @gonzaponte

andLaing:
I think that there's missing functionality from the new versions of the fitting functions.

There doesn't seem to be any way to set the dark spectrum, only to use the default Gaussian which defeats the point of having the functions in the first place. It's also not obvious how the maximum number of Gaussians nor the minimum Gaussian integral can be set in advance of a fit nor read back after a fit which is required for control and to an extent display.

Unless there is some magic that I'm not seeing, the two functions dark_convolution and scaled_dark_pedestal are pointless in the current version.

gonzaponte:
> Are you guys OK to proceed

I am to discuss with @andLaing the last changes and check that docstrings and variable names are fine. After that I am also happy to approve this PR.

> not done consistently, we often skip for small functions (it would add to noise, actually).

I often skip documenting small functions with big docstrings precisely because of the signal-to-noise ratio.

> PROPOSAL: USE TYPE ANNOTATIONS consistently. At the very least they are equivalent to the doc string and forces us to think what come in and out of the function. 

I agree, however, I am concerned how this affects automatic documentation. Anyway, this is an issue for the whole repo, not just this PR. Are you ok with dealing with it in another PR?

P.S: @andLaing I will get in touch with you to discuss your comments shortly.

jjgomezcadenas:
> agree, however, I am concerned how this affects automatic documentation. Anyway, this is an issue for the whole repo, not just this PR. Are you ok with dealing with it in another PR?

Automatic doc should be able to handle type annotations which are now a Python Standard. But of course, I am not suggesting to type-annotate all the functions in this PR! I am just suggesting that we start to use type annotations consistently. 

@andLaing raises the points:

>here doesn't seem to be any way to set the dark spectrum, only to use the default Gaussian which defeats the point of having the functions in the first place. It's also not obvious how the maximum number of Gaussians nor the minimum Gaussian integral can be set in advance of a fit nor read back after a fit which is required for control and to an extent display.

>Unless there is some magic that I'm not seeing, the two functions dark_convolution and scaled_dark_pedestal are pointless in the current version.

Can you guys discuss and agree on the best way to proceed?

jjgomezcadenas:
@andLaing are you OK with the last refactoring? Can we go ahead and approve the PR?

andLaing:
It's currently failing tests on my machine, so no, not yet

jjgomezcadenas:
I just checked that tests pass in my machine (and they also do in Travis). Which test is failing?

andLaing:
It works again!

gonzaponte:
Ok, I've rewritten commit history. The code is the same. Opinions before approval?

andLaing:
I'm happy

gonzaponte:
I have just rewritten a commit message.

gonzaponte (approved):
Implementation of the necessary functions for calibration and the corresponding cities. It also includes a refactor of `calib_sensor_functions`, adding the mode for baseline calculation and removing the mau. Some utilities have also been added.

Great work.

gonzaponte:
Forgot to say that everything is tested!

PR #443, opened 2018-02-19
Parallelize dataflow

jacg:
(This is a dummy commit to enable starting a PR)

One of the advertised benefits of the liquefication of the cities is
easy parallelization.

This PR proposes making progress on this. Without having thought about
the details carefully yet, two distinct possibilities come to mind:

1. Write an alternative `dataflow` module, which tries to parallelize
   things automagically. Switching to it would simply involve
   replacing

       from dataflow import xxx
       import dataflow as fl

   with

       from dataflow_par import xxx
       import dataflow_par as fl

2. Add utilities to the existing `dataflow` for dispatching streams or
   tasks onto separate processors, and for joining them back together.

One obvious question is: do you care whether, after processing by a
city, the events are written in the same relative order as the one in
which they were read?

Does anyone have any data on what proportion of their time various
cities spend on I/O vs CPU-bound activites?

I think that it is worthwhile to write our own parallelizer first, in
order to be able to explore, without restriction, what it is that we
want, but it is also important to keep in mind that we do not want to
reinvent the wheel and become guilty of Not Invented Here, so we
should be looking to take advantage of existing frameworks, in the
long run.


jjgomezcadenas:
The benefits of parellelizin RECONSTRUCTION code is not completely clear to me (and I think @jmbenlloch can add to this discussion). The key issue is that we have a limited number of cores to run the pre-processing of the data (from Raw to PMAPS). This part MUST run in the machines at the LSC, because the LSC bandwith is so ridiculously low that transfer raw data would take for ever.

Since we are running in a limited number of CPUs... does it make a difference whether we run sequentially or in parallel?

Issue #420, opened 2018-01-10
Loosen coupling between code and database, and fake the database in tests


jacg:
Some of our code is tightly coupled to *The One And Only Database*. This makes all manner of things difficult, including testing. This coupling needs to be reduced.

gonzaponte:
Isn't this addressed in #447?

gonzaponte:
> Isn't this addressed in #447?

It is.

PR #445, opened 2018-02-20
Fix to potential error in NoiseSampler

andLaing:
Code added baseline to samples before conversion
to ADC which is an error in the case of a
baseline != 0.

Changed class to have access to sipm gains so
that conversion could be done in function.

Affected functions and tests altered to take
change into account.

### DIFF
--- a/invisible_cities/reco/sensor_functions.py
+++ b/invisible_cities/reco/sensor_functions.py
@@ -51,7 +51,5 @@ def simulate_pmt_response(event, pmtrd, adc_to_pes, run_number = 0):
 def simulate_sipm_response(event, sipmrd, sipms_noise_sampler, sipm_adc_to_pes):
     """Add noise to the sipms with the NoiseSampler class and return
     the noisy waveform (in adc)."""
-    # add noise (in PES) to true waveform
-    dataSiPM = sipmrd[event] + sipms_noise_sampler.sample()
-    # return total signal in adc counts
-    return wfm.to_adc(dataSiPM, sipm_adc_to_pes)
+    # add noise (in ADC) to true waveform and return total signal in ADC
+    return wfm.to_adc(sipmrd[event], sipm_adc_to_pes) + sipms_noise_sampler.sample()
--- a/invisible_cities/core/random_sampling.py
+++ b/invisible_cities/core/random_sampling.py
@@ -66,6 +66,7 @@ class NoiseSampler:
         self.nsamples    = sample_size
         self.smear       = smear
         self.active      = DB.DataSiPM(run_number).Active.values[:, np.newaxis]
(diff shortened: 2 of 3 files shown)


jjgomezcadenas:
@gonzaponte can you review this?

gonzaponte:
The addition of `adc_to_pes` to this class forces caching the database somewhat as loading stuff from it is very costly.

gonzaponte:
@andLaing. Do you prefer returning the noisy waveform in adc rather than pes in the sampler? This is a variation wrt what we had before. I don't have a problem with that, but I want to make sure this is clear.

invisible_cities/reco/sensor_functions.py
```diff
@@ -51,7 +51,5 @@ def simulate_pmt_response(event, pmtrd, adc_to_pes, run_number = 0):
 def simulate_sipm_response(event, sipmrd, sipms_noise_sampler, sipm_adc_to_pes):
     """Add noise to the sipms with the NoiseSampler class and return
     the noisy waveform (in adc)."""
-    # add noise (in PES) to true waveform
-    dataSiPM = sipmrd[event] + sipms_noise_sampler.sample()
-    # return total signal in adc counts
-    return wfm.to_adc(dataSiPM, sipm_adc_to_pes)
+    # add noise (in PES) to true waveform and return total signal in ADC
```
gonzaponte: In this comment `(in PES)` is incorrect.

gonzaponte:
Can you rename the new variables in `random_sampling_test` to meet our coding style? `a_variable` rather than `aVariable` and `variable_full_name` rather than `var_shrt_name`.

andLaing:
I spoke to @paolafer this morning about the change that means the sampler gives values in adc and she is happy as long as the implementation is consistent. I think that's the case with the change in sensor_functions.py (when we rebase the changes due to the sensor simulation it becomes a bit more complicated but I don't see any major problems to integrate this change). If anyone sees the possibility to need to generate pe spectra without baselines we could add an option/alternate function without too much overhead but at the moment I don't think it's necessary.

I've changed the names in the test and updated the comment in sensor_funcitons.py.

As to the database question, in principle I could change the NoiseSampler init so that the gain values are sent (from base_cities in the extant uses) at definition which would save a call to the database at run time if you think this is necessary. The only thing is that it forces random_sampling_test.py to make a call to the database to be able define an instance of the sampler, thoughts?

gonzaponte (approved):
I would like to have a test that demonstrates the bug, but since it depends on the database I reckon it will be too difficult to do now. However, as soon as #420 is implemented, we should dedicate some time to pay this debt.

I am happy with this PR otherwise. It introduces a bit of an overhead as it includes another call to the database, but it will be addressed in an upcoming PR.

PR #446, opened 2018-02-23
Cache database

gonzaponte:
The functions of the database are decorated with `functools.lru_cache`, which keeps the result of such functions calls associated with their parameters so they are not called every time. This assumes the output of the function is unchanged when called with the same parameters, as is the case for the database. This comes in response to #445 and to #435, which needed/considered caching the database to improve performance.

I am not very proud of the test I've added, but I am sure @jacg can come up with something more appropriate and fancy.

### DIFF
--- a/invisible_cities/database/load_db.py
+++ b/invisible_cities/database/load_db.py
@@ -2,10 +2,11 @@ import sqlite3
 import numpy as np
 import pandas as pd
 import os
-from operator import itemgetter
-
+from operator  import itemgetter
+from functools import lru_cache
 
 DATABASE_LOCATION =  os.environ['ICTDIR'] + '/invisible_cities/database/localdb.sqlite3'
+
 
 def tmap(*args):
     return tuple(map(*args))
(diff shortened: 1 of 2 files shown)


gonzaponte:
@jacg Are you free to take a look at this? I've used a standard function I've found, but you might know better options. I would like also some feedback regarding the test.

jacg:
I don't expect to have any free time until May. I'll be in Valencia during the last week of April, to give a course, rather than do IC stuff. I'm afraid I'm completely out of the picture at the moment.

jjgomezcadenas:
@gonzaponte @jmbenlloch please proceed. @jmbenlloch should review, and then rebase.

jjgomezcadenas:
@jmbenlloch, @gonzaponte, this is stale, please do not let it linger.

jmbenlloch (approved):
I have rebased this old branch and checked the code. A simple change that can improve DB times by a great factor (test shows a factor 1e6).

PR #447, opened 2018-02-23
Allow database file selection via parameter

gonzaponte:
This makes testing easier as it allows the use a different, maybe custom, database in the test suite.

This goes in line with #420. Tests still to come.


jjgomezcadenas:
@gonzaponte @jmbenlloch this needs to be reviewed (@jmbenlloch I know that you are busy but we need you here). Also rebased.

jmbenlloch (approved):
This is a simple change to ease the future tests involving the database. I have already rebased this PR and therefore I'll approve and merge.

PR #449, opened 2018-02-26
Read and write MC true info

paolafer:
In this PR the writing and reading code of MC true information has been updated to the new hdf5 format generated by nexus.

Three tables are now stored: one with the particle information, one with the hit information and a third one with the information needed to relate the first two tables to each event. These three tables are read and written again and the writing takes into account the possible filtering of some events.

Some of the MC test files have been generated again, to have the new hdf5 format, and some of them have been deleted because not in use anymore.

The code does what it's meant to (at the best of my and tests' knowledge): I'd be happy to improve it in efficiency and to discuss specific parts of the data model classes that I have modified to add the complete true MC information, if needed. I've also left out the storage of true waveforms for now, since we don't need them usually, but maybe we want to add them in a future PR.

Caveat: all the already generated MC files cannot be processed anymore with these changes. I couldn't find an easy way to keep the ability of reading and writing both formats.

A couple of observations regarding tests:
- **test_diomira_fee_table** has been marked as @mark_skip, because it needs to be updated to the latest code regarding FEE simulation. Can @gonzaponte or @jjgomezcadenas have a look at it at some point?
- **test_compare_cwf_blr** was failing when I changed the input file with the new one. I compared the two files and I realized that in the old one the 40 keV electron events were generated very close to the EL region (2.5 cm) instead of at a distance of 25 cm, as I expected from the name of the file. A new file generated with primary vertices at 2.5 cm of distance in z solves the problem (just a slight tuning of the test parameter has been needed). This, however, makes me think that the whole deconvolution process has bigger errors for larger drifts and we should be aware of that.


jjgomezcadenas:
> This, however, makes me think that the whole deconvolution process has bigger errors for larger drifts and we should be aware of that.

Is the simulation including the effect of the lifetime? If so, what we are seeing here is most likely related with it.

paolafer:
Yes, the simulation includes a lifetime of 1200 microseconds.

invisible_cities/cities/irene_test.py
```diff
@@ -206,7 +203,7 @@ def test_empty_events_issue_81(config_tmpdir, ICDIR, s12params):
     nrequired = 10
 
     conf = configure('dummy invisible_cities/config/irene.conf'.split())
-    conf.update(dict(run_number   = 0,
+    conf.update(dict(run_number   = 4714,
```
gonzaponte, line 206: I think we talked about this, but why did you change this run number?

invisible_cities/cities/irene_test.py
```diff
@@ -242,7 +239,7 @@ def test_irene_empty_pmap_output(ICDATADIR, output_tmpdir, s12params):
 
     nrequired = 3
     conf = configure('dummy invisible_cities/config/irene.conf'.split())
-    conf.update(dict(run_number   = -4714,
+    conf.update(dict(run_number   = 4714,
```
gonzaponte, line 242: why did you remove the minus sign?

invisible_cities/config/penthesilea.conf
```diff
@@ -10,7 +10,7 @@ include('$ICDIR/config/s12_selector.conf')
 
 # override the default input/output files:
 
-files_in = '$ICDIR/database/test_data/dst_NEXT_v1_00_05_Kr_ACTIVE_0_0_7bar_PMP_10evt.h5'
+files_in = '$ICDIR/database/test_data/Kr_7bar_pmaps.h5'
```
gonzaponte: I would like to maintain the naming of the previous file so we know the production it came from. It is just so we can go all the way back to the nexus files.

If you have produced a file specifically to replace the other one, please make sure we can go back to the nexus file. If it is not too big I think we can upload it here (10 evt is a good number). It is also useful to indicate metadata about the data (such as nexus version, #evt, data type, etc.). Don't feel obligated to follow the pattern of the previous file, but it is helpful to have something similar.

invisible_cities/conftest.py
```diff
@@ -62,65 +62,59 @@ def example_blr_wfs_filename(ICDATADIR):
...
 @pytest.fixture(scope  = 'session',
                 params = ['electrons_40keV_z250_MCRD.h5'])
 def electron_MCRD_file(request, ICDATADIR):
     return os.path.join(ICDATADIR, request.param)
 
+
 @pytest.fixture(scope  = 'session',
-                params = ['dst_NEXT_v1_00_05_Tl_ACTIVE_140_0_7bar_PMP_2.h5'])
-def thallium_DST_file(request, ICDATADIR):
+                params = ['kr.3evts.MCRD.h5'])
```
gonzaponte: naming again

invisible_cities/evm/event_model.py
```diff
@@ -43,6 +43,23 @@ def __str__(self):
 
     __repr__ = __str__
 
+class Waveform:
```
gonzaponte: This is probably better implemented as a `namedtuple`. @jjgomezcadenas has already implemented this kind of thing elsewhere (probably in another PR). Please discuss and agree on an implementation.

invisible_cities/evm/event_model.py
```diff
@@ -62,25 +79,31 @@ def __str__(self):
 class MCParticle:
     """A MC Particle """
 
-    def __init__(self, particle_name, pdg_code,
+    def __init__(self, particle_name, primary,
```
gonzaponte, line 83: I don't bother the class, but a `namedtuple` may be good here too.

invisible_cities/evm/nh5.py
```diff
@@ -29,24 +29,40 @@ class DataSensor(tb.IsDescription):
     adc_to_pes = tb.Float32Col(pos=3)
     noise_rms  = tb.Float32Col(pos=4)
 
+class MCExtentInfo(tb.IsDescription):
+    """Stores. the parameters used by the simulation as metadata using
```
gonzaponte: typo in docstring

invisible_cities/evm/nh5.py
```diff
@@ -29,24 +29,40 @@ class DataSensor(tb.IsDescription):
...
+    Pytables.
+    """
+    evt_number    = tb.Int16Col(pos=0)
+    last_hit      = tb.Int16Col(pos=1)
+    last_particle = tb.Int16Col(pos=2)
 
-class MCTrack(tb.IsDescription):
+class MCHitInfo(tb.IsDescription):
+    """Stores. the parameters used by the simulation as metadata using
```
gonzaponte: typo in docstring

invisible_cities/evm/nh5.py
```diff
@@ -29,24 +29,40 @@ class DataSensor(tb.IsDescription):
...
+    """
+    hit_position  = tb.Float32Col(    pos=0, shape=3)
+    hit_time      = tb.Float64Col(    pos=1)
+    hit_energy    = tb.Float32Col(    pos=2)
+    label         = tb. StringCol(20, pos=3)
+    particle_indx = tb.  Int16Col(    pos=4)
+    hit_indx      = tb.  Int16Col(    pos=5)
+
+class MCParticleInfo(tb.IsDescription):
     """Stores. the parameters used by the simulation as metadata using
```
gonzaponte: typo in docstring. All docstrings in the `MC*Info` classes are the same. They could be improved a bit.

invisible_cities/evm/nh5.py
```diff
@@ -29,24 +29,40 @@ class DataSensor(tb.IsDescription):
...
+    evt_number    = tb.Int16Col(pos=0)
+    last_hit      = tb.Int16Col(pos=1)
+    last_particle = tb.Int16Col(pos=2)
 
-class MCTrack(tb.IsDescription):
+class MCHitInfo(tb.IsDescription):
+    """Stores. the parameters used by the simulation as metadata using
+    Pytables.
+    """
+    hit_position  = tb.Float32Col(    pos=0, shape=3)
```
gonzaponte: The `hit_` prefix could be removed here (as the table is already `MCHitInfo` and the `hit_indx` column would be better placed next to the other `hit_` columns. But I guess this has to be synchronized with the other side

invisible_cities/io/indexation_test.py
```diff
@@ -36,7 +36,7 @@ def init_city(ICDIR, config_tmpdir):
 @mark.parametrize("writer group node column".split(),
                   [(    hits_writer, "RECO" , "Events"  , "event"     ),
                    (      kr_writer, "DST"  , "Events"  , "event"     ),
-                   (mc_track_writer, "MC"   , "MCTracks", "event_indx"),
+                   ( mc_info_writer, "MC"   , "extents" , "evt_number"),
```
gonzaponte, line 39: If this is the only table indexed, reading back the data will be extremely slow. The ones for hits and particles should also be indexed.

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,305 @@
...
+                        filters     = tbl.filters(self.compression))
+
+        # Mark column to index after populating table
+        self.extent_table.set_attr('columns_to_index', ['evt_number'])
+
+        self.hit_table = self.h5file.create_table(MC, "hits",
+                        description = MCHitInfo,
+                        title       = "hits",
+                        filters     = tbl.filters(self.compression))
+
```
gonzaponte: As mentioned before, you should index these tables.

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,305 @@
...
+    def _create_tables(self):
+        """Create tables in MC group in file h5file."""
+        if '/MC' in self.h5file:
+            MC = self.h5file.root.MC
+        else:
+            MC = self.h5file.create_group(self.h5file.root, "MC")
+
+        self.extent_table = self.h5file.create_table(MC, "extents",
+                        description = MCExtentInfo,
```
gonzaponte: Align the keyword arguments with the first character after the parenthesis, please.

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,305 @@
...
+                        filters     = tbl.filters(self.compression))
+
+    def __call__(self, mctables: (tables.Table, tables.Table, tables.Table),
+                     evt_number: int):
+
+        extents = mctables[0]
+        for iext in range(self.last_row, len(extents)):
+            if extents[iext]['evt_number'] < evt_number:
+                continue
+            if  extents[iext]['evt_number'] > evt_number:
```
gonzaponte: there is a double space after if

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,305 @@
...
+                break
+            if iext == 0:
```
gonzaponte: assign `this_row = extents[iext]` and use `this_row` to improve readability. Also put spaces surrounding the arithmetic operators. Aligning the `=` and making any other symmetry more obvious would also be nice.

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,305 @@
...
+                        title       = "hits",
+                        filters     = tbl.filters(self.compression))
+
+        self.particle_table = self.h5file.create_table(MC, "particles",
+                        description = MCParticleInfo,
+                        title       = "particles",
+                        filters     = tbl.filters(self.compression))
+
+    def __call__(self, mctables: (tables.Table, tables.Table, tables.Table),
+                     evt_number: int):
```
gonzaponte: check alignment

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,305 @@
...
+            new_row['final_volume']   = p[7]
+            new_row['momentum']       = p[8]
+            new_row['kin_energy']     = p[9]
+            new_row['creator_proc']   = p[10]
+            new_row.append()
+        self.particle_table.flush()
+
+
+def load_mchits(file_name: str,
+                    event_range=(0,int(1e9))) -> Mapping[int, MCHit]:
```
gonzaponte: check alignment and space after comma.

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,305 @@
...
+
+    with tables.open_file(file_name,mode='r') as h5in:
```
gonzaponte: space after comma. The parameter `mode='r'` is not needed as it is the default, but you can leave it, as you prefer.

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,305 @@
...
+        mcevents = read_mcinfo(h5in, event_range)
+        mchits_dict = compute_mchits_dict(mcevents)
+
+    return mchits_dict
+
+
+def load_mcparticles(file_name: str,
+                         event_range=(0,int(1e9))) -> Mapping[int, MCParticle]:
```
gonzaponte: check alignment, space after comma. Check this for all functions.

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,305 @@
...
+
+    with tables.open_file(file_name,mode='r') as h5in:
```
gonzaponte: space after comma. Check it for every `open_file` call, at least.

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,305 @@
...
+
+
+def load_mcsensor_response(file_name: str,
+                               event_range=(0,int(1e9))) -> Mapping[int, MCParticle]:
+
+    with tables.open_file(file_name,mode='r') as h5in:
+        return read_mcsns_response(h5in, event_range)
+
+
+def read_mcinfo_evt_by_evt (mctables: (tables.Table, tables.Table, tables.Table),
```
gonzaponte: I would name it `read_mcinfo_evt` rather than `read_mcinfo_evt_by_evt` as the latter suggests (at least it does to me) that it produces an iterator over the mcinfo, which is untrue.

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,305 @@
...
+    h5extents   = mctables[0]
+    h5hits      = mctables[1]
+    h5particles = mctables[2]
+
+    particle_rows = []
+    hit_rows = []
+
+    event_range = (last_row,int(1e9))
+    for iext in range(*event_range):
+        if h5extents[iext]['evt_number'] == event_number:
```
gonzaponte: again, assign `h5extents[iext]` to a variable for readability and efficiency.

jjgomezcadenas:
@paolafer tests are failing...

paolafer:
They are not on my laptop... two of them are failing because of hypothesis, and the third one is test_2d_corrections... They started failing after the rebase on the database fix of a few hours ago.

gonzaponte:
Launch rebuild if you can. If not I can do it in a while. Btw, I am preparing a review, but it will take me a while since I don't have much time to do this.

paolafer:
I've tried to launch the rebuild and the message "Hooray, the build was successfully restarted" has been prompted, but it doesn't seem to actually start.

gonzaponte:
I have tried to launch a rebuild, but I couldn't either. I will try later.

jjgomezcadenas:
Suggestion:  I would like to propose whenever possible simple "data classes" derived from NamedTuple. Thus, instead of:

```
class Waveform:
    """Transient class storing times and charges for a sensor"""
    def __init__(self, times, charges, bin_width):
        self.times     = times
        self.charges   = charges
        self.bin_width = bin_width

    def __str__(self):
        return """ Waveform: times = {}
                    charges = {}
                    bin width = {}
                    \n""".format(self.times,
                                     self.charges,
                                     self.bin_width)

    __repr__ =     __str__


I propose something like this:


```
    from typing      import NamedTuple

    class WaveformPmt(NamedTuple):
       times    : np.ndarray
       charges    : np.ndarray
       bin_width : int

Notice that the boiler plate is much reduced, the class is annotated and the repr and str method come from free (derived from namedtuple). Such objects represent faithfully "data stuff" and are clear and very lean.

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,305 @@
...
+    h5extents   = h5f.root.MC.extents
+    h5hits      = h5f.root.MC.hits
+    h5particles = h5f.root.MC.particles
+
+    mc_info = (h5extents, h5hits, h5particles)
+
+    all_events = {}
+
+    for iext in range(*event_range):
+        if iext >= len(h5extents):
```
gonzaponte: assign `len(h5extents)` to a variable before the loop.

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,305 @@
...
+            break
+
+        current_event = {}
+        evt_number = h5extents[iext]['evt_number']
+        hit_rows, particle_rows = read_mcinfo_evt_by_evt(mc_info, evt_number, iext)
+
+        for h5particle in particle_rows:
+
+            current_event[h5particle['particle_indx']] = MCParticle(h5particle['particle_name'].decode('utf-8','ignore'),
```
gonzaponte: `current_event[h5particle['particle_indx']]` is better read as
```
this_particle = h5particle[`particle_indx`]
current_event[this_particle]
```

also, check the alignment

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,305 @@
...
+
+    return all_events
+
+
+def compute_mchits_dict(mcevents:Mapping[int, Mapping[int, MCParticle]])->Mapping[int, MCHit]:
+    """Returns all hits in the event"""
+    mchits_dict = {}
+    for event_no, particle_dict in mcevents.items():
+        hits = []
+        for particle_no in particle_dict.keys():
```
gonzaponte, line 237: If you find this readable:
`hits = chain.from_iterable(map(attrgetter("hits"), particle_dict.values())`
`chain` comes from `itertools` and `attrgetter` from `operator`.

Otherwise leave it as is.

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,305 @@
...
+            hits.extend(particle.hits)
+        mchits_dict[event_no] = hits
+    return mchits_dict
+
+
+def read_mcsns_response(h5f, event_range=(0,1e9)) ->Mapping[int, Mapping[int, Waveform]]:
+
+    h5config = h5f.root.MC.configuration
+
+    bin_width_PMT  = 100. * units.nanosecond
```
gonzaponte: you can use `ns` and `mus` if you prefer to.

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,305 @@
...
+    h5extents = h5f.root.MC.extents
+    h5waveforms = h5f.root.MC.waveforms
+
+    last_line_of_event = 'last_sns_data'
+
+    all_events = {}
+
+    iwvf = 0
+    if(event_range[0] > 0):
+        iwvf = h5extents[event_range[0]-1][last_line_of_event]+1
```
gonzaponte: spaces around `+` and `-`. In python you can omit the parenthesis in an `if` statement.

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,305 @@
...
+        while (iwvf <= iwvf_end):
+            wvf_row = h5waveforms[iwvf]
+            sensor_id = wvf_row['sensor_id']
+
+            if (sensor_id == current_sensor_id):
+                time_bins.append(wvf_row['time_bin'])
+                charges.append(wvf_row['charge'])
+            else:
+               # current_event[current_sensor_id] = zip(time_bins, charges)
+                if 0 <= sensor_id < 1000:
```
gonzaponte: do you have a special case `sensor_id < 0` and do you want to assume it is a PMT? If so, it is ok, but I would raise an exception (`ValueError`, I guess) otherwise.

gonzaponte: I don't think we actually want to check every possible scenario (`sensor_id`'s between X064 and (X+1)000 would also be fake, for instance). So you can simply rephrase it as `bin_width = bin_width_PMT if sensor_id < 1000 else bin_width_SiPM`.

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,305 @@
...
+                    bin_width = bin_width_PMT
+                else :
+                    bin_width = bin_width_SiPM
+
+                times = [t * bin_width for t in time_bins]
```
gonzaponte: `times = np.array(t) * bin_width`?

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -0,0 +1,163 @@
...
+import tables as tb
+
+from . mcinfo_io import load_mchits
+from . mcinfo_io import load_mcparticles
+from . mcinfo_io import load_mcsensor_response
+from . mcinfo_io import mc_info_writer
+from . mcinfo_io import read_mcinfo_evt_by_evt
+
+from .. reco.tbl_functions import get_mc_info
+
```
gonzaponte: two spaces between imports and functions and between functions

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -0,0 +1,163 @@
...
+def test_non_consecutive_events(config_tmpdir, ICDIR):
+    filein  = os.path.join(ICDIR, 'database/test_data/', 'kr.3evts.MCRD.h5')
```
gonzaponte: there is an `ICDATADIR` fixture that gives you the path to the test data.

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -0,0 +1,163 @@
...
+    fileout = os.path.join(config_tmpdir, 'test_mcinfo.h5')
+    h5in    = tb.open_file(filein)
```
gonzaponte: use context managers:

```
with tb.open_file(filein) as h5in:
    with tb.open_file(fileout) as h5out:
        #do stuff with both files open
    # do stuff with only input file open

File: invisible_cities/io/mcinfo_io_test.py [hunk 33]
```diff
@@ -0,0 +1,163 @@
...
+    for evt in events_to_copy:
+        mc_writer(mc_info, evt)
+
+    events_out = np.unique(h5out.root.MC.extents[:]['evt_number'])
+
+    np.testing.assert_array_equal(events_to_copy, events_out)
+
+    h5out.close()
+
+    hit_rows, particle_rows = read_mcinfo_evt_by_evt(mc_info, event_number=2)
```
gonzaponte: It seems like this is a different test. Try to keep them as simple and self-contained as possible. This way, when a test fails the problem becomes even more obvious.

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -0,0 +1,163 @@
...
+    nhits_evt2  = len(hit_rows)
+
+    h5filtered = tb.open_file(fileout)
+    filtered_extents = h5filtered.root.MC.extents
+    hit_extent_evt2 = filtered_extents[1]['last_hit'] - filtered_extents[0]['last_hit']
+
+    assert hit_extent_evt2 == nhits_evt2
+
+    filt_mc_info = get_mc_info(h5filtered)
```
gonzaponte: And here begins another test. I think this is to reuse the file produced at the beginning of this test. If that is the case, split the tests and decorate them with `mark.serial` so they are run in order.

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+        for partr, filt_partr in zip(particle_rows, filt_particle_rows):
+            assert np.allclose(partr['initial_vertex'], filt_partr['initial_vertex'])
+            assert np.allclose(partr['final_vertex']  , filt_partr['final_vertex'])
+            assert np.allclose(partr['momentum']      , filt_partr['momentum'])
+            assert np.allclose(partr['kin_energy']    , filt_partr['kin_energy'])
+            assert partr['particle_name']            == filt_partr['particle_name']
+
+    #Now skip the first event
```
gonzaponte: If you have the same test for two different scenarios you can parametrize them using 
```
@mark.parametrize("name_of_par_1 name_of_par_2 etc".split(),
                  (values_1,
                   values_2,
                   ...))
def test_of_mine(name_of_par_1, name_of_par_2, ...):
    ...
```

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -0,0 +1,163 @@
...
+
+    for partr, filt_partr in zip(particle_rows    , filt_particle_rows):
+        assert np.allclose(partr['initial_vertex'], filt_partr['initial_vertex'])
+        assert np.allclose(partr['final_vertex']  , filt_partr['final_vertex'])
+        assert np.allclose(partr['momentum']      , filt_partr['momentum'])
+        assert np.allclose(partr['kin_energy']    , filt_partr['kin_energy'])
+        assert partr['particle_name']            == filt_partr['particle_name']
+
+
+def test_load_all_mchits(mc_all_hits_data):
```
gonzaponte: Maybe `test_load_mchits_correct_number_of_hits`?

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -0,0 +1,163 @@
...
+def test_load_mchits(mc_particle_and_hits_nexus_data):
+    efile, _, _, _, _, _, _, X, Y, Z, E, t = mc_particle_and_hits_nexus_data
+
+    mchits_dict = load_mchits(efile)
+    hX = [hit.X for hit in mchits_dict[0]]
+    hY = [hit.Y for hit in mchits_dict[0]]
+    hZ = [hit.Z for hit in mchits_dict[0]]
+    hE = [hit.E for hit in mchits_dict[0]]
+    ht = [hit.time for hit in mchits_dict[0]]
+
```
gonzaponte: check alignment

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -0,0 +1,163 @@
...
+    assert np.isclose(X,hX).all()
+    assert np.isclose(Y,hY).all()
+    assert np.isclose(Z,hZ).all()
+    assert np.isclose(E,hE).all()
+    assert np.isclose(t,ht).all()
+
```
gonzaponte: check spaces after commas. Also `np.allclose` is the same as `np.isclose(...).all()`

invisible_cities/reco/tbl_functions.py
```diff
@@ -152,9 +152,9 @@ def get_rd_vectors(h5in):
 
 
 
-def get_mc_tracks(h5in):
-    """Return MC Tracks bank"""
-    return h5in.root.MC.MCTracks
+def get_mc_info(h5in):
+    """Return MC info bank"""
+    return h5in.root.MC.extents, h5in.root.MC.hits, h5in.root.MC.particles
```
gonzaponte: returning a `namedtuple` would be nice and pretty

invisible_cities/reco/wfm_functions_test.py
```diff
@@ -57,6 +57,6 @@ def test_compare_cwf_blr():
 
     CWF = wfm.cwf_from_rwf(pmtrwf, range(NEVT), calib, deconv)
     diff = wfm.compare_cwf_blr(CWF, pmtblr,
-                               event_list=range(NEVT), window_size=200)
+                               event_list=range(NEVT), window_size=300)
 
-    assert max(diff) < 0.1
+    assert max(diff) < 0.15
```
gonzaponte, line 62: Do you understand why is there a need to increase the tolerance?

gonzaponte (commented):
Most of my comments are purely cosmetic and the code looks quite nice. I haven't commented every time a cosmetic detail is repeated, so while fixing them try to search for other places. Let me know if you need help with anything. Great work!

invisible_cities/cities/irene_test.py
paolafer: I changed the run number in order not to generate again the file with the new MC format. Since this test doesn't involve MC, it doesn't matter.

invisible_cities/cities/irene_test.py
paolafer: Same as above: in this way, the code doesn't look for MC true information and the old format does not prevent the test to work.

invisible_cities/config/penthesilea.conf
paolafer: Yes, I produced some new files specifically, since we haven't done an official production yet. The files are usually small, I can upload them here, too. Should I upload also the corresponding MCRD and the RWF files?

invisible_cities/evm/event_model.py
paolafer: Yes, I saw his comment, I'll implement it as suggested.

invisible_cities/io/mcinfo_io_test.py
paolafer: Ok,  I thought I could put them together because they use the same input file, but I can separate them in more tests.

invisible_cities/reco/wfm_functions_test.py
paolafer: I talked about this with @jjgomezcadenas and he told me that `0.1` had been chosen with fine-tuning for that particular file. My understanding is that `0.15` being of the same order is fine, but tell me if you think otherwise.

invisible_cities/evm/event_model.py
paolafer: Since there are other classes in the event model, I would leave an eventual consistent restyling for another PR.

gonzaponte: Ok. I agree

invisible_cities/config/penthesilea.conf
paolafer: In the end, since we have by now a first tag of nexus with hdf5 output, I decided to generate again those files and call them with a meaningful naming. I have also uploaded the original nexus files and the MCRD and RWF ones.

invisible_cities/evm/nh5.py
paolafer: I understand, but I would keep it this way, since we have already a tag in nexus that is being used. Also, the `hit_indx` attribute is better placed after `label`, because it is the ordered number of a hit *with a particular label*.

invisible_cities/cities/base_cities.py
```diff
@@ -398,20 +398,20 @@ def file_loop(self):
             with tb.open_file(filename, "r") as h5in:
 
                 events_info = self.get_run_and_event_info(h5in)
-                mc_tracks   = self.get_mc_tracks(h5in)
+                mc_info   = self.get_mc_info(h5in)
```
gonzaponte: align

invisible_cities/cities/base_cities.py
```diff
@@ -398,20 +398,20 @@ def file_loop(self):
...
                 dataVectors = 0
                 NEVT        = 0
 
                 if self.raw_data_type == 'RWF':
                     NEVT, pmtrwf, sipmrwf, _ = self.get_rwf_vectors(h5in)
                     dataVectors = DataVectors(pmt=pmtrwf, sipm=sipmrwf,
-                                             mc=mc_tracks, events=events_info)
+                                             mc=mc_info, events=events_info)
```
gonzaponte: align

invisible_cities/cities/base_cities.py
```diff
@@ -398,20 +398,20 @@ def file_loop(self):
...
 
                     self.event_loop(NEVT, dataVectors)
                 elif self.raw_data_type == 'MCRD':
                     NEVT, pmtrd, sipmrd     = self.get_rd_vectors(h5in)
                     dataVectors = DataVectors(pmt=pmtrd, sipm=sipmrd,
-                                             mc=mc_tracks, events=events_info)
+                                             mc=mc_info, events=events_info)
```
gonzaponte: align

invisible_cities/cities/diomira.py
```diff
@@ -65,7 +65,7 @@ def event_loop(self, NEVT, dataVectors):
         write       = self.writers
         pmtrd       = dataVectors.pmt
         sipmrd      = dataVectors.sipm
-        mc_tracks   = dataVectors.mc
+        mc_info   = dataVectors.mc
```
gonzaponte: align

invisible_cities/cities/diomira_test.py
```diff
@@ -20,16 +20,16 @@
 
 from .  diomira  import Diomira
 
-
+@mark.skip
```
gonzaponte: two blank lines between imports and function definitions (including decorators)

invisible_cities/cities/diomira_test.py
```diff
@@ -104,20 +104,20 @@ def test_diomira_copy_mc_and_offset(config_tmpdir):
...
             assert h5out.root.Run.events [0]['evt_number'] == start_evt
 
-            # check mctracks
+            # check mcextents
             # we have to convert manually into a tuple because MCTracks[0]
             # returns an object of type numpy.void where we cannot index
             # using ranges like mctracks_in[1:]
-            mctracks_in  = tuple(h5in .root.MC.MCTracks[0])
-            mctracks_out = tuple(h5out.root.MC.MCTracks[0])
+            mcextents_in  = tuple(h5in.root.MC.extents[0])
```
gonzaponte, line 113: align

invisible_cities/cities/irene.py
```diff
@@ -109,7 +109,7 @@ def write_parameters(self, h5out):
     def get_writers(self, h5out):
         writers = Namespace(
         run_and_event = run_and_event_writer(h5out),
-        mc            =      mc_track_writer(h5out) if self.monte_carlo else None,
+        mc            =      mc_info_writer(h5out) if self.monte_carlo else None,
```
gonzaponte: align

invisible_cities/cities/isidora.py
```diff
@@ -75,15 +75,15 @@ def event_loop(self, NEVT, dataVectors):
...
+                write.mc(mc_info, self.cnt.n_events_tot)
 
     def get_writers(self, h5out):
         RWF = partial(rwf_writer,  h5out,   group_name='BLR')
         writers = Namespace(
             run_and_event = run_and_event_writer(h5out),
             pmt  = RWF(table_name='pmtcwf' , n_sensors=self.sp.NPMT , waveform_length=self.sp.PMTWL),
             sipm  = RWF(table_name='sipmrwf' , n_sensors=self.sp.NSIPM , waveform_length=self.sp.SIPMWL),
-            mc            =      mc_track_writer(h5out) if self.monte_carlo else None,
+            mc            =      mc_info_writer(h5out) if self.monte_carlo else None,
```
gonzaponte: align

invisible_cities/cities/penthesilea.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, **kwds):
 
     def get_writers(self, h5out):
         return Namespace(dst = hits_writer    (h5out),
-                         mc  = self.get_mc_track_writer(h5out))
+                         mc  = self.get_mc_info_writer(h5out))
```
gonzaponte: align

invisible_cities/cities/penthesilea_test.py
```diff
@@ -105,15 +105,15 @@ def test_penthesilea_produces_tracks_when_required(KrMC_pmaps_filename, KrMC_hds
     conf.update(dict(files_in        = PATH_IN,
                      file_out        = PATH_OUT,
                      event_range     = (nevt_req,),
-                     write_mc_tracks = write_mc_tracks,
+                     write_mc_info = write_mc_info,
```
gonzaponte: align

invisible_cities/cities/penthesilea_test.py
```diff
@@ -105,15 +105,15 @@ def test_penthesilea_produces_tracks_when_required(KrMC_pmaps_filename, KrMC_hds
...
                      **KrMC_hdst.config))
 
     penthesilea = Penthesilea(**conf)
     penthesilea.run()
 
     with tb.open_file(PATH_OUT) as h5out:
-        assert write_mc_tracks == ("MC"          in h5out.root)
-        assert write_mc_tracks == ("MC/MCTracks" in h5out.root)
+        assert write_mc_info == ("MC"          in h5out.root)
```
gonzaponte: align

invisible_cities/cities/penthesilea_test.py
```diff
@@ -105,15 +105,15 @@ def test_penthesilea_produces_tracks_when_required(KrMC_pmaps_filename, KrMC_hds
...
+        assert write_mc_info == ("MC/particles" in h5out.root)
```
gonzaponte: you should also check the existance of the two other tables, right?

invisible_cities/evm/event_model.py
```diff
@@ -1,6 +1,8 @@
 # Clsses defining the event model
 
+import tables
```
gonzaponte: for uniformity, can you import it as `tb`?

invisible_cities/evm/event_model.py
```diff
@@ -44,6 +48,20 @@ def __str__(self):
     __repr__ = __str__
 
 
+class MCInfo(NamedTuple):
+    """Transient class storing the tables of MC true info"""
+    extents : tables.Table
+    hits : tables.Table
+    particles : tables.Table
```
gonzaponte: align colons

invisible_cities/evm/event_model.py
```diff
@@ -62,25 +80,32 @@ def __str__(self):
...
+        self.p              = momentum
+        self.E              = energy
+        self.process        = creator_proc
+        self.hits           = []
 
     def __str__(self):
-        return """ MCParticle: name = {} pdg = {}
-                    vi = {} vf = {}
-                    p =  {}  E = {}
+        return """ MCParticle: name = {}
```
gonzaponte: you can use format strings, if you prefer. We don't have any convention yet. For example:

```
x = 3
s = f"the value of x is {x}"
# s is now "The value of x is 3"
```

gonzaponte: It works also with expressions, so it recognizes `"{self.name}"`, `"{a_dict[a_key]}"`, `"{len(a_sequence)}"`, etc.

invisible_cities/evm/nh5.py
```diff
@@ -29,24 +29,38 @@ class DataSensor(tb.IsDescription):
...
-    nof_hits       = tb.  Int16Col(    pos= 9)
-    hit_indx       = tb.  Int16Col(    pos=10)
-    hit_position   = tb.Float32Col(    pos=11, shape=3)
-    hit_time       = tb.Float32Col(    pos=12)
-    hit_energy     = tb.Float32Col(    pos=13)
+    evt_number    = tb.Int16Col(pos=0)
+    last_hit      = tb.Int16Col(pos=1)
+    last_particle = tb.Int16Col(pos=2)
+
+class MCHitInfo(tb.IsDescription):
```
gonzaponte: doubel blank line separation

invisible_cities/evm/nh5.py
```diff
@@ -29,24 +29,38 @@ class DataSensor(tb.IsDescription):
...
+    """Stores the simulated hits as metadata using Pytables.
+    """
+    hit_position  = tb.Float32Col(    pos=0, shape=3)
+    hit_time      = tb.Float64Col(    pos=1)
+    hit_energy    = tb.Float32Col(    pos=2)
+    label         = tb. StringCol(20, pos=3)
+    particle_indx = tb.  Int16Col(    pos=4)
+    hit_indx      = tb.  Int16Col(    pos=5)
+
+class MCParticleInfo(tb.IsDescription):
```
gonzaponte: doubel blank line separation

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,311 @@
...
+    def __init__(self, h5file, compression = 'ZLIB4'):
+
+        self.h5file      = h5file
+        self.compression = compression
+        self._create_tables()
+        # last visited row
+        self.last_row = 0
+        self.last_written_hit = 0
+        self.last_written_particle = 0
+        self.first_extent_row = True
```
gonzaponte: aligning equal signs would be nice

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,311 @@
...
+    def __call__(self, mctables: (tables.Table, tables.Table, tables.Table),
+                 evt_number: int):
+
+        extents = mctables[0]
+        for iext in range(self.last_row, len(extents)):
+            this_row = extents[iext]
+            if this_row['evt_number'] < evt_number:
+                continue
+            if this_row['evt_number'] > evt_number:
+                break
```
gonzaponte: I think in this case, the `continue` and `break` statements are better read when put right after the colon, in the same line. Check it and if you like it commit the change.

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,311 @@
...
+            if iext == 0:
```
gonzaponte: align equals and review the spaces around arithmetic operators. This affects everything in the for loop.

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,311 @@
...
+            new_row['creator_proc']   = p[10]
+            new_row.append()
+        self.particle_table.flush()
+
+
+def load_mchits(file_name: str,
+                event_range=(0, int(1e9))) -> Mapping[int, MCHit]:
+
+    with tables.open_file(file_name, mode='r') as h5in:
+        mcevents = read_mcinfo(h5in, event_range)
```
gonzaponte: align

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,311 @@
...
+
+def load_mcsensor_response(file_name: str,
+                           event_range=(0, int(1e9))) -> Mapping[int, MCParticle]:
+
+    with tables.open_file(file_name, mode='r') as h5in:
+        return read_mcsns_response(h5in, event_range)
+
+
+def read_mcinfo_evt (mctables: (tables.Table, tables.Table, tables.Table),
+                            event_number: int, last_row=0) -> ([tables.Table], [tables.Table]):
```
gonzaponte: align

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,311 @@
...
+    h5extents   = mctables[0]
+    h5hits      = mctables[1]
+    h5particles = mctables[2]
+
+    particle_rows = []
+    hit_rows = []
```
gonzaponte: align

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,311 @@
...
+
+    event_range = (last_row, int(1e9))
+    for iext in range(*event_range):
+        this_row = h5extents[iext]
+        if this_row['evt_number'] == event_number:
+            # the indices of the first hit and particle are 0 unless the first event
+            #  written is to be skipped: in this case they must be read from the extents
+            ihit = 0; ipart = 0
```
gonzaponte: either `ihit = ipart = 0` or 
```
ihit  = 0
ipart = 0
```

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,311 @@
...
+            if iext > 0:
```
gonzaponte: alignments

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,311 @@
...
+            while ipart <= ipart_end:
+                particle_rows.append(h5particles[ipart])
+                ipart += 1
+
+            break
+
+    return hit_rows, particle_rows
+
+
+def read_mcinfo(h5f, event_range=(0, int(1e9))) ->Mapping[int, Mapping[int, MCParticle]]:
```
gonzaponte, line 184: review alignment in this function

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,311 @@
...
+    for event_no, particle_dict in mcevents.items():
+        hits = []
+        for particle_no in particle_dict.keys():
+            particle = particle_dict[particle_no]
+            hits.extend(particle.hits)
+        mchits_dict[event_no] = hits
+    return mchits_dict
+
+
+def read_mcsns_response(h5f, event_range=(0, 1e9)) ->Mapping[int, Mapping[int, Waveform]]:
```
gonzaponte, line 244: review alignment in this function

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,311 @@
...
+
+    h5config = h5f.root.MC.configuration
+
+    bin_width_PMT  = 100. * units.nanosecond
```
gonzaponte: why do you define these variables if you read them from file?

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,311 @@
...
+    bin_width_SiPM = 1. * units.microsecond
+    for row in h5config:
+        param_name = row['param_key'].decode('utf-8','ignore')
+        if param_name.find('time_binning') >= 0:
```
gonzaponte: If you only want to check containment use `if substring in string:`.

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,311 @@
...
+            wvf_row = h5waveforms[iwvf]
+            sensor_id = wvf_row['sensor_id']
+
+            if sensor_id == current_sensor_id:
+                time_bins.append(wvf_row['time_bin'])
+                charges.append(wvf_row['charge'])
+            else:
+                bin_width = bin_width_PMT if sensor_id < 1000 else bin_width_SiPM
+
+                #times = [t * bin_width for t in time_bins]
```
gonzaponte: remove commented out code

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -0,0 +1,146 @@
...
+from .. reco.tbl_functions import get_mc_info
+
+from pytest import mark
+parametrize = mark.parametrize
+
+
+@mark.serial
+@parametrize('skipped_evt, out_filename',
+                  ((0, 'test_mcinfo_skip_evt0.h5'), (1, 'test_mcinfo_skip_evt1.h5')))
```
gonzaponte: alignment. Put each set of parameters in a different line.

gonzaponte: Also, I would use `mark.parametrize` instead of
```
parametrize = mark.parametrize
@parametrize
```

because we use `mark.` anyway with `serial`.

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -0,0 +1,146 @@
...
+def test_non_consecutive_events(config_tmpdir, ICDIR, krypton_MCRD_file, skipped_evt, out_filename):
```
gonzaponte: use `ICDATADIR` to replace `os.path.join(ICDIR, "database/test_data/")`

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -0,0 +1,146 @@
...
+    filein  = os.path.join(ICDIR, 'database/test_data/', krypton_MCRD_file)
+    fileout = os.path.join(config_tmpdir, out_filename)
+
+    with tb.open_file(filein) as h5in:
+        with tb.open_file(fileout, 'w') as h5out:
+
+            mc_writer = mc_info_writer(h5out)
```
gonzaponte: alignment

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -0,0 +1,146 @@
...
+            mc_extents = h5in.root.MC.extents
+            events_in = np.unique(h5in.root.MC.extents[:]['evt_number'])
+
+            mc_info = get_mc_info(h5in)
+
+            #Skip the desired event (there are only 3 in the file)
+            events_to_copy = [evt for evt in events_in if evt != skipped_evt]
```
gonzaponte, line 35: `events_to_copy = list(filter(lambda x: x != skipped_evt, events_in))`

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -0,0 +1,146 @@
...
+            for evt in events_to_copy:
+                mc_writer(mc_info, evt)
+
+            events_out = np.unique(h5out.root.MC.extents[:]['evt_number'])
+
+            np.testing.assert_array_equal(events_to_copy, events_out)
+
+
+@mark.serial
+@parametrize('file_to_check, evt_to_be_read',
```
gonzaponte, line 46: same comment about `parametrize` + alignment + each parametrized case in a different line.

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -0,0 +1,146 @@
...
+                  (('test_mcinfo_skip_evt0.h5', 1), ('test_mcinfo_skip_evt1.h5', 0),
+                   ('test_mcinfo_skip_evt1.h5', 2)))
+def test_output_non_consecutive_events(config_tmpdir, ICDIR, krypton_MCRD_file, file_to_check, evt_to_be_read):
```
gonzaponte: `ICDATADIR` rather than `ICDIR`

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -0,0 +1,146 @@
...
+                # correctly
+                hit_rows, particle_rows = read_mcinfo_evt(mc_info, evt_to_be_read)
+                filtered_hit_rows, filtered_particle_rows = read_mcinfo_evt(filtered_mc_info,
+                                                                            evt_to_be_read)
+
+                for hitr, filtered_hitr in zip(hit_rows, filtered_hit_rows):
+                    assert np.allclose(hitr['hit_position'], filtered_hitr['hit_position'])
+                    assert np.allclose(hitr['hit_time']    , filtered_hitr['hit_time'])
+                    assert np.allclose(hitr['hit_energy']  , filtered_hitr['hit_energy'])
+                    assert hitr['label']                  == filtered_hitr['label']
```
gonzaponte: can you align `hitr` with the one in the other lines while maintaining? Same for `partr`

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -0,0 +1,146 @@
...
+def test_load_sensors_data(mc_sensors_nexus_data):
+    efile, pmt0_first, pmt0_last, pmt0_tot_samples, sipm_id, sipm = mc_sensors_nexus_data
+
+    mcsensors_dict = load_mcsensor_response(efile)
+
+    waveforms = mcsensors_dict[0]
+
+    sns_number = 0
+    wvf = waveforms[sns_number]
+    bins = [t/wvf.bin_width for t in wvf.times]
```
gonzaponte: alignment + spaces around `/`

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -0,0 +1,146 @@
...
+    samples = list(zip(bins, wvf.charges))
+
+    assert samples[0]   == pmt0_first
+    assert samples[-1]  == pmt0_last
+    assert len(samples) == pmt0_tot_samples
+
+    wvf = waveforms[sipm_id]
+    bins = [t/wvf.bin_width for t in wvf.times]
```
gonzaponte: alignment + spaces around `/`

gonzaponte (commented):
Again, mostly cosmetics. Just a couple of non-trivial comments.

invisible_cities/io/mcinfo_io_test.py
jacg: I disagree. Not strongly, but I disagree. The only reason we use =mark.parametrize= at all, is that it is impossible to import =parametrize= from =pytest.mark=, but morally speaking that's what we would want to do. =parametrize = mark.parametrize= is the next best thing.

gonzaponte: Ok. I don't have a strong opinion here, I just like the symmetry. I do also get the feeling that marking something with `serial` or with `parametrize` have different meanings.

invisible_cities/io/mcinfo_io.py
paolafer: I thought it was useful to have a default value, but maybe it's actually dangerous.

invisible_cities/io/mcinfo_io.py
paolafer: I tried it at the beginning, but it didn't work properly.

invisible_cities/io/mcinfo_io_test.py
paolafer: That was my other option :-), but I thought that this one was clearer.

invisible_cities/io/mcinfo_io.py
gonzaponte: Yes, I think it can be.

invisible_cities/io/mcinfo_io_test.py
gonzaponte: I think both of them are clear. Use what you like the most. `filter` tells you exactly what are you trying to do without actually reading the code, but in such a simple thing it doesn't make much of a difference.

invisible_cities/io/mcinfo_io_test.py
paolafer: I think the `ICDATADIR` variable is not defined in the setup done by `manage.sh`, could it be?

invisible_cities/io/mcinfo_io_test.py
jacg: Symmetry would be an argument for `serial = mark.serial` :-)

But, now that you mention it, I agree that `mark.serial` somehow feels better than just `serial` while the opposite is true for `parametrize`.

invisible_cities/io/mcinfo_io.py
paolafer: What would you suggest, then, in case the desired string is not found? One should realize that something is wrong.

gonzaponte: In that case you can set them to `None` and check after the `for` loop, raising an exception if case the variable's value is still `None`.

invisible_cities/io/mcinfo_io_test.py
gonzaponte: It is defined in `invisible_cities/conftest.py`.

paolafer:
I've implemented your comments and rebased the whole PR in master.

invisible_cities/cities/diomira.py
```diff
@@ -13,7 +13,7 @@
 import numpy as np
 
 
-from .. io.mc_io                import mc_track_writer
+from .. io.mcinfo_io                import mc_info_writer
```
gonzaponte: align

invisible_cities/cities/diomira.py
```diff
@@ -112,7 +112,7 @@ def get_writers(self, h5out):
         RWF = partial(rwf_writer,  h5out,   group_name='RD')
         writers = Namespace(
             run_and_event = run_and_event_writer(h5out),
-            mc            =      mc_track_writer(h5out) if self.monte_carlo else None,
+            mc            =      mc_info_writer(h5out) if self.monte_carlo else None,
```
gonzaponte: align

invisible_cities/cities/penthesilea_test.py
```diff
@@ -105,15 +105,17 @@ def test_penthesilea_produces_tracks_when_required(KrMC_pmaps_filename, KrMC_hds
...
     penthesilea = Penthesilea(**conf)
     penthesilea.run()
 
     with tb.open_file(PATH_OUT) as h5out:
-        assert write_mc_tracks == ("MC"          in h5out.root)
-        assert write_mc_tracks == ("MC/MCTracks" in h5out.root)
+        assert write_mc_info == ("MC"           in h5out.root)
+        assert write_mc_info == ("MC/particles" in h5out.root)
+        assert write_mc_info == ("MC/extents"   in h5out.root)
+        assert write_mc_info == ("MC/hits" in h5out.root)
```
gonzaponte: align

invisible_cities/io/mcinfo_io.py
gonzaponte: still there

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,316 @@
...
+        if param_name.find('time_binning') >= 0:
+            param_value = row['param_value'].decode('utf-8','ignore')
+            numb, unit = param_value.split()
+            if param_name.find('Pmt') > 0:
+                bin_width_PMT = float(numb)
+            elif param_name.find('SiPM') >= 0:
+                bin_width_SiPM = float(numb)
+
+
+    if not  bin_width_PMT:
```
gonzaponte: `if bin_width_PMT is None`

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,316 @@
...
+        raise SensorBinningNotFound
+    if not  bin_width_PMT:
```
gonzaponte: I guess you meant `bin_width_SiPM`

invisible_cities/io/mcinfo_io.py
```diff
@@ -0,0 +1,316 @@
...
+        raise SensorBinningNotFound
+
+
+    h5extents = h5f.root.MC.extents
```
gonzaponte: check the alignment in the whole function

invisible_cities/io/mcinfo_io_test.py
gonzaponte: still there. Review the whole function

invisible_cities/io/mcinfo_io_test.py
gonzaponte: still there. ignore the `parametrize` thing.

gonzaponte (commented):
still some cosmetic details

paolafer:
I'm asymptotically getting there :-D! How much is the tolerance ;-)?

paolafer:
Also, what do you mean with 

> still there. ignore the parametrize thing.

?

I understood that it was OK.

Ah, I've spotted it!

paolafer:
Squash done!

gonzaponte:
Please rebase

gonzaponte (approved):
Rearrangement of the mc info reader to accommodate new data structure produced by the simulation software. The test suite has also been adapted and extended, so everything is quite well tested.

Great work.

Issue #364, opened 2017-10-07
Open intervals for minrun-maxrun in database


gonzaponte:
When we search for data in the database based on some run number, we look for those rows between `MinRun` and `MaxRun` with an open interval. For example:

```
select * from ChannelMask where MinRun < {3} and {3} < MaxRun
```

Since the data runs have never matched those numbers so far, there has never been a problem with that. Do we want to change it to be a right-open interval like `[MinRun, MaxRun)`? I've been struggling to figure out why a test failed because of this subtlety.

jacg:
Sticking to the strong Python convention of semi-open intervals would certainly conform to the principle of least surprise.

What's really shocking in this case is that it's an *open* interval: usually deviations from the semi-open convention tend towards closed intervals, so open seems very surprising indeed.

The fact that you lost time on undertanding a test failure because of it, is an indication that it should be changed.

jmbenlloch:
I agree there is an error there, but I think the interval has to be closed `[MinRun, MaxRun]` due to two things:

1- That is the way data is stored in the DB now. For instance:

```
mysql> select distinct(MinRun), MaxRun from ChannelMapping;
+--------+--------+
| MinRun | MaxRun |
+--------+--------+
|      0 |   1273 |
|   1274 |   1754 |
|   1755 |   1765 |
|   1766 |   1978 |
|   1979 | 100000 |
|   1979 |   3292 |
|      0 | 100000 |
|   3293 |   3913 |
|   3914 | 100000 |
+--------+--------+
9 rows in set (0,91 sec)
```

2- We arre using the same DB as the ART software, if we change the DB to use semiopen intevals, we may break compatibility with ART.

So, in my opinion, we sould put: `select * from ChannelMask where MinRun <= {3} and {3} <= MaxRun`.

gonzaponte:
Ok. You mean that `MaxRun` is always the next `MinRun` minus one. Closed intervals make sense here.

gonzaponte:
This should't have been closed. There is no PR fixing it.

jjgomezcadenas:
Sorry, I don't get it. I think it is agreed that we cannot use semi-open interval in the case of the DB

gonzaponte:
We have agreed on the solution, but it is still not implemented.

jacg:
This was supposed to have been closed automatically by the merge of #450. This did not happen because the magical incantation was written in a PR comment. GitHub only looks for them in the PR description (the top comment) and in commit messages.

PR #450, opened 2018-02-27
Take the proper interval from database

jmbenlloch:
@andLaing pointed out to me that we are not taking the proper interval when querying the database as stated in #364. This PR fixes that issue.

### DIFF
--- a/invisible_cities/database/load_db.py
+++ b/invisible_cities/database/load_db.py
@@ -14,58 +14,31 @@ def tmap(*args):
 # 3012 was the first SiPM calibration after remapping.
 runNumberForMC = 3012
 
-def get_min_run_values(db, run_number, sensors):
-    ''' Due to ART database design MaxRun is not being used currently,
-    so we need to get the exact MinRun value for a given run number to
-    get only one row per sensor...
-    db is sqlite3 connection and sensors pmt or sipm.'''
-
-    if sensors == 'pmt':  bound = '<' #pmt id's are < 100
-    if sensors == 'sipm': bound = '>' #sipms id's are > 100
-
-    sql = '''select Max(MinRun) from ChannelGain
-where SensorID {} 100 and MinRun < {}'''.format(bound, abs(run_number))
-    cursor = db.execute(sql)
-    minrun_gain = cursor.fetchone()[0]
-
-    sql = '''select Max(MinRun) from ChannelPosition
-where SensorID {} 100 and MinRun < {}'''.format(bound, abs(run_number))
-    cursor = db.execute(sql)
-    minrun_position = cursor.fetchone()[0]
-
-    sql = '''select Max(MinRun) from ChannelMapping
-where SensorID {} 100 and MinRun < {}'''.format(bound, abs(run_number))
-    cursor = db.execute(sql)
-    minrun_map = cursor.fetchone()[0]
-
-    return minrun_gain, minrun_position, minrun_map
-
-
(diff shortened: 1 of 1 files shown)


gonzaponte:
Can you add a short test or should we leave it for another PR?

gonzaponte:
Closes #364.

jmbenlloch:
I don't have time for that test now, the queries are somehow complex and I don't think a toy example would be enough.

gonzaponte (approved):
Important fix that has been in our to-do list for a long time. We are accumulating a bit of technical debt by not adding a test, but there is a hurry to introduce this PR.

gonzaponte:
We should open an empty PR so we don't forget to add a test.

jmbenlloch:
TLDR;

**IMPORTANT**: Without this update, the code is **NOT** working with the current DB. So everyone should update to this commit after merge and **must run** `bash manage.sh download_test_db` to get the corrected DB.

Extended version:

There was another error in the DB code. It was assuming the mapping for all PMTs had to start at the same run. We have recently change only one PMT getting this status in the DB:

```
+--------+--------+--------+----------+
| MinRun | MaxRun | ElecID | SensorID |
+--------+--------+--------+----------+
|   5025 | 100000 |      0 |        0 |
|   5025 | 100000 |     12 |        1 |
|   5025 | 100000 |      1 |        2 |
|   5025 | 100000 |     13 |        3 |
|   5025 | 100000 |     19 |        4 |
|   5025 | 100000 |     18 |        5 |
|   5025 | 100000 |      7 |        6 |
|   5025 | 100000 |      6 |        7 |
|   5025 |   5153 |     17 |        8 |
|   5025 | 100000 |     16 |        9 |
|   5025 | 100000 |      5 |       10 |
|   5025 | 100000 |      4 |       11 |
|   5154 | 100000 |      2 |        8 |
+--------+--------+--------+----------+
13 rows in set (0,00 sec)
```

And the code was only returning the latest one (minrun 5154, sensor id 8). I have fixed the query to get the proper ones and tests are passing again.

For it to work I have also fixed some old values from the DB in the table `ChannelGain`, in particular this two queries:
```
UPDATE ChannelGain SET MaxRun = 2062 where MinRun = 1755;
UPDATE ChannelGain SET MaxRun = 2842 where MinRun = 2063;
```

The problem with them was that `MaxRun` was set to 100000 for both of them, so if we tried to get the values for a current run `MinRun <= 5168 and 5168 <= MaxRun`, we would get the current value but also two old and incorrect ones. I have check the status of all relevant tables to make sure `MaxRun` was correctly set in all of them and so is the case now. `ChannelGain` was the only one with this problem.

To summarize, the code can use now `MinRun` and `MaxRun` in a correct way after fixing some values in the DB.

gonzaponte:
Ok. PLEASE MERGE!

jmbenlloch:
Done, already merged!

jacg:
> everyone should update to this commit after merge and must run bash manage.sh download_test_db

Of course, this shouldn't need to be done manually at all! It should should happen automatically because the DB should be version controlled. How is the Git LFS work progressing?

jjgomezcadenas:
No progress. Jose Vicente proposed an scheme, as you remember, but no one has had any time to look into this. Right now we are terribly undermanned, with the usual suspects horribly over-committed.

jacg:
Is #210 the only issue or PR that we have specifically addressing this?

PR #452, opened 2018-03-09
Add covariance matrix to FitFunction's output

andLaing:
FitFunction tuple now contains the covariance
matrix of the fit to ease error propagation.

covariance matrix added in fit_functions.fit

### DIFF
--- a/invisible_cities/core/fit_functions.py
+++ b/invisible_cities/core/fit_functions.py
@@ -147,7 +147,7 @@ def fit(func, x, y, seed=(), fit_range=None, **kwargs):
     chi2, pval = get_chi2_and_pvalue(y, fitx, ndof, sigma_r)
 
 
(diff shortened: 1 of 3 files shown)


gonzaponte:
Great! Can you introduce a simple test to ensure the covariance matrix is what we expect it to be?

andLaing:
Not sure if this is enough but since we can't easily know the values only the shape...

jjgomezcadenas:
@andLaing @gonzaponte can you progress here?

gonzaponte:
@andLaing can you add a test that checks the values for a simple case like four points fitted to a straight line?

gonzaponte:
Nice!

gonzaponte (approved):
Nice addition to the output of the fit function. It also adds a couple of tests to ensure the output is correct.

PR #456, opened 2018-03-19
Replace implementation of profile functions with faster ones

gonzaponte:
In this PR, the functions `profile*` are replaced with a faster implementation. The new implementations use standard functions, which follows the principle of not reinventing the wheel. 

However, for simpler cases there is a better set of functions in `scipy.stats`: `binned_statistics`, `binned_statistics_2d` and `binned_statistics_dd`. Those functions are designed to do exactly what we are want to do: bin data on some axes and apply operations on the remaining one. But there is one caveat: they can only return one statistic per bin. We need to get 3 in our case: mean, std and count, which would imply three calls to those functions, worsening the overall performance of the function. Thus, I have explored a implementation using `pandas`'  dataframes, which is a factor 2-3 faster than our original implementation (in 1-d) and a factor 3-4 faster than the implementation using `scipy` functions. Those speed factors increase very rapidly with the number of dimensions of the original dataset.

### DIFF
--- a/invisible_cities/calib/pmtgain.py
+++ b/invisible_cities/calib/pmtgain.py
@@ -11,10 +11,10 @@ from functools import partial
 
 import numpy  as np
 
-from .                         import      calib_functions as cf
-from .. io   .         hist_io import          hist_writer
-from .. io   .run_and_event_io import run_and_event_writer
-from .. icaro.hst_functions    import shift_to_bin_centers
+from .                        import      calib_functions as cf
+from .. io  .         hist_io import          hist_writer
+from .. io  .run_and_event_io import run_and_event_writer
+from .. core.core_functions   import shift_to_bin_centers
 
 from .. core  .system_of_units_c import units
 from .. cities.base_cities       import CalibratedCity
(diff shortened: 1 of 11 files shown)


jjgomezcadenas (commented):
If I understand correctly the new imp passes the same tests as the old imp, correct? Is there a test to gurantee that the profile is correct and the errors as expected?

gonzaponte:
> If I understand correctly the new imp passes the same tests as the old imp, correct?

Yes, they pass.

> Is there a test to gurantee that the profile is correct and the errors as expected?

Yes, the original tests ensure the output is correct.

jjgomezcadenas (approved):
This is a very smart adoption of an existing resource to our case, avoiding the re-invetion of the wheel and saving time. Congrats.

jacg:
If reality lives up to the advertizing blurb (I haven't had time to verify) then this is a gem of a PR!

We want to make this sort of thing part of our core development culture.

jjgomezcadenas:
Indeed. But as the karma of @gonzaponte raises he demands more s&d&r&R. That's a bit of an issue, no?

PR #457, opened 2018-03-21
Patch 456: fix standard deviation and bug with `drop_nan=False`

gonzaponte:
This is a patch to #456. The test suite didn't catch two errors in the previous PR:
- The definition of the standard deviation changed, but it was subtle, particularly for large datasets (See the description of [97c2423](https://github.com/nextic/IC/commit/97c2423ed9b1f1c16d4a97f20882288f3c627933)).
- When the value of the keyword-argument `drop_nan` is `False`, the output arrays had their values shifted one position to the right.

The test suite has been extended to ensure these two bugs don't show up again. Also, the current definition of the standard deviation is the right one.

### DIFF
--- a/invisible_cities/core/fit_functions.py
+++ b/invisible_cities/core/fit_functions.py
@@ -197,10 +197,10 @@ def profileX(xdata, ydata, nbins=100,
 
     mean      = gb.mean().values
     deviation = gb.std () if std else gb.std() / gb.size()**0.5
-    indices   = deviation.index.values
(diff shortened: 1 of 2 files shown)


jacg:
May I propose "Patch 456: fix standard deviation" or something like that, as title for this PR?

jjgomezcadenas (approved):
A nice catch, nicely demonstrated and fixed. Good work.

jjgomezcadenas:
@gonzaponte please, merge.

PR #458, opened 2018-03-21
Truevoxels

jerenner:
Adds code to generate true voxels from Nexus MC hits:
- city detsim takes Nexus input, applies energy smearing and diffusion to the hits, and outputs true voxels in HDF5 format (see `VoxelsTable` in evm/nh5.py);
- voxel I/O is implemented in io/voxels_io.py, including a writer for true voxels and a function to load true voxels


jjgomezcadenas (commented):
Where are the tests demonstrating that your functions work?

jerenner:
The voxel io tests are at: https://github.com/jerenner/IC/blob/truevoxels/invisible_cities/io/voxels_io_test.py

The detsim function tests (only 1 function right now and an admittedly trivial test) are at:
https://github.com/jerenner/IC/blob/truevoxels/invisible_cities/detsim/detsim_functions_test.py

jjgomezcadenas (commented):
Yes, I can see the voxels - io tests, but should we try some tests to certify the voxelization itself, is possible?

jerenner:
I am using the voxelization function `voxelize_hits` from `reco/paolina_functions.py`. It looks like there are already relevant tests for this in: https://github.com/jerenner/IC/blob/master/invisible_cities/reco/paolina_functions_test.py

jjgomezcadenas (approved):
Yes, this is true. On this grounds I believe we can approve this PR.

jacg:
This was approved months ago, and the tests pass. Is there any good reason why it has not been merged?

jjgomezcadenas:
I think it went simply under the radar, please, merge.

PR #459, opened 2018-03-25
Reset MC writer attributes when reading a new file in Irene

paolafer:
This PR fixes an issue that happens when processing multiple input files with irene city.
First, a test has been added to catch the problem, then a fix has been implemented.


jjgomezcadenas (commented):
The subject of this PR is unclear. It seems to imply that irene has a problem when reading multiple files. The problem, in fact is related only with the MC and fixed by properly initcialization of variables, right?

jjgomezcadenas (commented):
The subject matter of this PR is confusing. It seems to imply that irene has a problem reading multiple files which is not the case. The problem arises with MC data and is related with variable initialization, correct?

jjgomezcadenas:
@paolafer please comment on my review

paolafer:
Yes, the problem is with Monte Carlo files only, and it's related to the true information being correctly read and written.

jjgomezcadenas (approved):
This is a necessary and well documented fix (although the subject of the PR was somewhat misleading). It includes a correct test. On the other hand we keep adding stuff to solid cities... @gonzaponte. But at this stage, this needs to be approved to proceed.

jjgomezcadenas:
@paolafer @gonzaponte please merge

jjgomezcadenas (commented):
This is a necessary and well documented fix (although the subject of the PR was somewhat misleading). It includes a correct test. On the other hand we keep adding stuff to solid cities... @gonzaponte. But at this stage, this needs to be approved to proceed.

jjgomezcadenas:
@gonzaponte @paolafer please merge

jacg:
> we keep adding stuff to solid cities

What is stopping replacement of the solid cities with the liquid ones?

jjgomezcadenas:
That it requires a major rebasing, and there is no one with time to work on this.

jacg:
And the rebase size just keeps on growing, as more stuff is added before it is merged.

If memory serves me, I had left it rebased. It's a great shame (and, above all) waste of resources that it has got out of hand.

jjgomezcadenas:
Yes it is. But detector switched on, man power was desperately needed somewhere else, and the total amount of expert programmers capable to implement the rebases was very small. So, as every poor family we had to go to the bank and borrow. Now we own, as usual. 

As far as I can see, only @jacg, @gonzaponte and perhaps yours truly could attack the rebase. None of us had the time for it. So now the rebase is way larger and the law of compound interest on technical debt applies.

I expect @gonzaponte to have some free time perhaps next week. This is the sooner we can consider attacking the rebase and the liquification of the cities.

gonzaponte:
I have renamed the PR.

My only comment is that we should hide all that code within a method of the MC writer class and call that method in the cities. But that is for another PR.

gonzaponte:
@paolafer rebase, please

paolafer:
Rebase done.

jacg:
>  we should hide all that code within a method of the MC writer class and call that method in the cities. But that is for another PR.

Please open an issue stating this, lest it be forgotten.

gonzaponte:
> Please open an issue stating this, lest it be forgotten.

Yes, I was going to open an empty PR directly as soon as this PR is merged.

PR #460, opened 2018-03-26
Fix missing `gain_mau` option in pmtgain

andLaing:
Re-added function for the second option which
had been removed accidentally in previous
PR review.

PR #462, opened 2018-03-27
Change types in MC true info

paolafer:
This PR is the adaption of the IC code to a problem of overflow in nexus MC true tables. 
The types of the columns of the `extents` table have been changed from `unsigned int` or `int` to `uint64` or `int64` in order to be able to represent a larger number of events. Strangely (to me), it turns out that, when adding a `uint64` to an `int`, python gives as a result a `float64` (see, for instance, [this thread](https://github.com/numpy/numpy/issues/7126)). Therefore, in this PR I simply cast my `uint64` variable to `int`, which, if I'm not wrong, is the same as a `long` in python, so it should be ok. Opinions?

### DIFF
--- a/invisible_cities/cities/irene_test.py
+++ b/invisible_cities/cities/irene_test.py
@@ -118,8 +118,9 @@ def test_irene_electrons_40keV(config_tmpdir, ICDIR, s12params, thr_sipm_type, t
 
             # check events numbers & timestamps
             evts_in     = h5in .root.Run.events[:nactual]
-            evts_out_u8 = h5out.root.Run.events[:nactual]
+            evts_out_i4 = h5out.root.Run.events[:nactual]
 
+            evts_out_u8 = evts_out_i4.astype([('evt_number', '<u8'), ('timestamp', '<u8')])
             np.testing.assert_array_equal(evts_in, evts_out_u8)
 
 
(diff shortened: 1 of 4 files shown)


jjgomezcadenas:
@paolafer, tests fail

paolafer:
It fails because of the same problem fixed in PR #457, which Andrew mentioned in #460. When it will be merged, I'll rebase and it will be fixed.

gonzaponte:
#457 has been merged. Please rebase.

paolafer:
Rebase done.

paolafer:
It fails again, with the following error: 

> hypothesis.errors.Flaky: Hypothesis test_all_elements_close_par(mu=-0.0, sigma=1.000000000000006) produces unreliable results: Falsified on the first call but did not on a subsequent one

gonzaponte:
Yes, I have launched a rebuild. Seems a rare failure.

paolafer:
Checks passed! Hurray!

gonzaponte:
> Opinions?

I guess it is ok, but I am not completely sure. It is probably harmless enough to be approved and corrected (if necessary) when @jacg finds the time to look at it.

paolafer:
Can we approve and merge this one? Maybe @jmbenlloch can have a second look at it from a technical point of view.

jmbenlloch:
I think this update is ok. It makes uniform the type of event numbers across the different tables and updates the table MC/extents to avoid overflows. We need this update to be able to continue with the development of related projects like Nexus. I have rebased this branch to the latest master. If all tests pass I will approve and merge.

jmbenlloch (approved):
This a needed update and everything is working ok.

PR #463, opened 2018-03-28
Add method to reset mc writer

gonzaponte:
The MC writer keeps track of what it has read and written from the mc table in the input file. However, when multiple files are read in the same city execution, the class attributes must be reset in order to read them correctly. We are currently reseting them by hand in the cities, but this is far from optimal.

In this PR we include a new method to the MC writer class so it manages the attributes when a new file is given.


jmbenlloch (approved):
This solves a bug when reading multiple files including several tests to ensure everything is working ok. We also need this to start a new montecarlo production so I will approve and merge.

PR #464, opened 2018-04-02
First version of channel parameter output

andLaing:
channel_param_io and associated test added.
Thought for generic use with any sensor type
or other numbered channel.

Updated names to be a bit more consistent with DB

improved names

added explicit sensor info in table title

### DIFF
--- a/invisible_cities/io/channel_param_io_test.py
+++ b/invisible_cities/io/channel_param_io_test.py
@@ -17,25 +17,30 @@ def test_generic_parameters(config_tmpdir):
     filename = os.path.join(config_tmpdir, 'test_param.h5')
 
     outDict = {}
+    val_list = []
+    n_rows = 5
     with tb.open_file(filename, 'w') as dataOut:
         pWrite = channel_param_writer(dataOut, sensor_type="test",
                                       func_name="generic",
                                       param_names=generic_params)
 
-        for i, par in enumerate(generic_params):
-            outDict[par] = [i, i / 10]
+        for sens in range(n_rows):
+            for i, par in enumerate(generic_params):
+                outDict[par] = [i, (i + sens) / 10]
 
-        pWrite(0, outDict)
+            pWrite(sens, outDict)
+            val_list.append(list(outDict.values()))
 
     col_list = ['SensorID'] + list(outDict.keys())
-    val_list = np.concatenate(list(outDict.values()))
+    val_list = np.concatenate(val_list)
     with tb.open_file(filename) as dataIn:
 
         tblRead = dataIn.root.FITPARAMS.FIT_test_generic
-        assert tblRead.nrows == 1
+        assert tblRead.nrows == n_rows
         assert len(tblRead.colnames) == len(generic_params) + 1
         assert tblRead.colnames == col_list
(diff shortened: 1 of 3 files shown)


gonzaponte:
have you considered writing the parameters as an array (rather than a table, I mean)?

andLaing:
I considered it for ease but it's useful to have the parameters names/descriptions in the output

gonzaponte:
The dictionary approach is not suitable for this case. Because it is mutable, it has a clear side effect, particularly if you try to use two instances of the writer as the second one will have every name from the first one. I suggest making a function that returns a dictionary or even a `tb.IsDescription`-derived class as in `evm/nh5.py`.

andLaing:
I've changed the dictionary definition to be part of the function which defines the writer.

It works as expected for two instances of the io in the same job. The class approach seemed to complicated when I first started looking at this but if it's better I can look again. I wanted the io to be as generic as possible to avoid having to define new classes etc every time someone wanted to try a new fit function etc

jjgomezcadenas:
This seems perfectly OK to me. But you could return a namedtuple instead (which is immutable and as simple and generic as a dictionary).

In fact, you could consider to derive the same type of little "data classes" that I am trying to standarize as our default:

As an example:

   class KrEvent(NamedTuple):
      E : np.array
      Q : np.array
      X : np.array
      Y : np.array
      Z : np.array

Notice that the definition of the class allow you to declare the type of the variable, which is useful.

gonzaponte:
I don't think we are free to choose what type to return as it has to be
something `tb.create_table` accepts. We can try and see if it works, though.

gonzaponte:
In `test_store_fit_values` you can use the `make_table_dictionary` function, as it has been already tested. It would also be nice to store several values with multiple sensor ids and ensure they are all correct. It would make the test more reobust.

Are you planning to add a reader for this type of data?

invisible_cities/io/channel_param_io_test.py
```diff
@@ -35,14 +37,46 @@ def test_generic_parameters(config_tmpdir):
...
 
 
+def test_generator_param_reader(config_tmpdir):
+    filename = os.path.join(config_tmpdir, 'test_param.h5')
+
+    outDict = {}
+    val_list = []
+    n_rows = 5
+    with tb.open_file(filename, 'w') as dataOut:
+        pWrite = channel_param_writer(dataOut, sensor_type="test",
```
jacg: Minor point: our style guide agrees with PEP8 that it should be `data_out` rather than `dataOut` and similarly for `dataIn`, `outDict` etc.

andLaing: easily fixed...

andLaing:
Anything else on this PR?

gonzaponte:
Please someone else review this PR, I don't have time to look at it with detail.

jjgomezcadenas:
Please @jmbenlloch review this PR

jjgomezcadenas:
Please @jmbenlloch review this PR (second attempt)

jmbenlloch (approved):
This PR includes a new functionality with the corresponding tests. I have added some minor cosmetic corrections. I think it is ready for approval.

PR #465, opened 2018-04-05
Test all cities with an exact result

gonzaponte:
This PR adds a test to each city comparing the output for a particular configuration with the true one, which is also stored. The true output is defined as the current one. This PR is devoted to make the transition to the liquid cities (#435) as safe as possible.


jjgomezcadenas:
This is a major and most important test that requires an urgent and dedicated review. Please @jmbenlloch, take it! (thanks for volunteering!)

jacg:
Sorry for being repetitive, but I have to point out that the addition of binary files in this PR once again highlights the need and urgency of **Git LFS.**

invisible_cities/core/testing_utils.py
```diff
@@ -134,3 +134,19 @@ def assert_PMap_equality(pmp0, pmp1):
...
+def assert_tables_equality(got_table, expected_table):
+    table_got      =      got_table[:]
+    table_expected = expected_table[:]
+    assert len(table_got) == len(table_expected)
+
+    for got, expected in zip(table_got, table_expected):
+        assert type(got) == type(expected)
+        try:
+            assert got == expected
+        except ValueError:
```
jacg: I struggle to see how `assert a == b` could ever raise `ValueError`

jjgomezcadenas:
@jmbenlloch, @gonzaponte can you coordinate with Jose Vicente and propose a scheme to implement LFS?

gonzaponte:
> I struggle to see how `assert a == b` could ever raise `ValueError`

`numpy` arrays do. Comparing two arrays with `==` returns another array
(filled with booleans). The `assert` statement tries to convert it to a
boolean and that operation raises a `ValueError` indicating that the truth
value of an array is ambiguous.

jacg:
I'm getting bored of my own voice here: 3 months ago this was declared by @jjgomezcadenas to be "a major and most important test that requires an urgent and dedicated review." Nothing has happened about it since then ... probably because it is **waiting for Git LFS** (#210) to be available.

jjgomezcadenas:
The main reason why there is no progress here is that this is in the hands of @gonzaponte, currently not available. I am also getting bored of repeating that we cannot do miracles. Currently our bus factor is lower than one in many corners of IC. This will be fixed by @gonzaponte, after he comes back to work full time for IC, in September.

jacg:
My point isn't to bore you with my repetitive voice, but to hammer home how overdue Git LFS (#210) is. I fully appreciated, at the time that I wrote my comment, that this was something that @gonzaponte will do upon his return to our universe. Do you think that Git LFS could be in place by then, so that, when he is ready to work on it, he isn't stalled by the continued absence of Git LFS?

jjgomezcadenas:
Yes I am asking @jmbenlloch and Jose Vicente Carrion (@jmbenlloch  please include him in IC) to implement LFS now.

gonzaponte:
Liquid cities (#435) are waiting for this to be accepted and merged so the transition is safe. How are we doing with LFS (#535)?

PR #467, opened 2018-04-06
Reset MC writer attributes correctly when reading more than one file with all cities 

paolafer:
This PR fixes and expand PR #459. When a second file was read in one job, the MC writer would write the table `extents` without updating the field `last_hit` and `last_particle`. I've added a test to show this behaviour, and fix the code. I've added the same kind of test also to diomira and penthesilea tests, in order to catch all the possibilities of making mistakes (I believe). In fact, the only two defined file loops in `cities` are in RawCity and in PCity: diomira and irene test the first, and penthesilea tests the latter. 
I suggest @jmbenlloch and/or @gonzaponte could be a reviewer, since they are familiar with the MC true information structure in IC.


gonzaponte:
We have two PRs with the same purpose, take a look at #463. From what I've seen in this PR #463 is not correct, but I think we can pick a few ideas from it.

paolafer:
Ok, sorry, I hadn't realized you had worked on it! We can merge the two PR into one, do you want to talk on the phone to see how it can be done? At that point, I'm afraid @jmbenlloch is the only one left to review the merge :-D

paolafer:
I, hereby, declare this PR closed, since the changes it enclosed have been merged into PR #463.

PR #468, opened 2018-04-13
Fix pmap io

gonzaponte:
@jmbenlloch has found a bug in the pmap reader. The SiPMs waveforms were not properly matched to each sensor. This was caused by `np.unique` which, besides removing duplicates, sorts the elements. This can also be a problem in the PMTs waveforms, although the PMTs are, in principle, always sorted.

A couple of tests have been added to demonstrate the situation and to ensure this doesn't happen again in the future.

The solution consists of replacing `np.unique` by `pd.unique`, which preserves the original order. There is one caveat, though: `pd.unique` removes only ***consecutive*** duplicate elements. This is ok as our data is stored on a sensor-by-sensor basis and there should not be non-consecutive duplicate elements.

Two other alternative implementations have been also considered, but the one using `pd.unique` is the fastest one as demonstrated in the plot attached.

image

### DIFF
--- a/invisible_cities/io/pmaps_io.py
+++ b/invisible_cities/io/pmaps_io.py
@@ -117,19 +117,25 @@ def load_pmaps(filename):
     return pmap_dict
 
 
+import itertools
+import operator
+def unique(values):
+    return np.array(tuple(map(operator.itemgetter(0), itertools.groupby(values))))
+
+
 def build_pmt_responses(pmtdf, ipmtdf):
-    times   =            pmtdf.time.values
-    pmt_ids = np.unique(ipmtdf.npmt.values)
-    enes    =           ipmtdf.ene .values.reshape(pmt_ids.size,
-                                                     times.size)
+    times   =         pmtdf.time.values
+    pmt_ids = unique(ipmtdf.npmt.values)
+    enes    =        ipmtdf.ene .values.reshape(pmt_ids.size,
+                                                times.size)
     return times, PMTResponses(pmt_ids, enes)
 
 
 def build_sipm_responses(sidf):
     if len(sidf) == 0: return SiPMResponses.build_empty_instance()
 
-    sipm_ids = np.unique(sidf.nsipm.values)
-    enes     =           sidf.ene  .values
+    sipm_ids = unique(sidf.nsipm.values)
+    enes     =        sidf.ene  .values
     n_times  = enes.size // sipm_ids.size
     enes     = enes.reshape(sipm_ids.size, n_times)
     return SiPMResponses(sipm_ids, enes)
--- a/invisible_cities/io/pmaps_io.py
+++ b/invisible_cities/io/pmaps_io.py
@@ -119,7 +119,7 @@ def load_pmaps(filename):
 
 def build_pmt_responses(pmtdf, ipmtdf):
     times   =            pmtdf.time.values
-    pmt_ids = np.unique(ipmtdf.npmt.values)
+    pmt_ids = pd.unique(ipmtdf.npmt.values)
     enes    =           ipmtdf.ene .values.reshape(pmt_ids.size,
                                                      times.size)
     return times, PMTResponses(pmt_ids, enes)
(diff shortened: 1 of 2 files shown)


jmbenlloch (approved):
This is a VERY important  bug fix. The solution chosen by @gonzaponte is the fastest one and also includes two simple tests to avoid this situation in the future. 

I think this can be merged and we need to find out whether this bug could have affected some of the data we have processed in the past months. Maybe @gonzaponte can think about the implications of this change for the rest of the cities (dorothea, penthiselea, etc).

jacg:
Leaving the other solutions that were considered, archied somewhere in history of this PR might be a good idea.

gonzaponte:
Is a comment  in the discussion good enough or do you recommend leaving it in the commit history?

jacg:
I would leave it in history. That way anyone can checkout the attempts and experiment. Here's an outline of how you could tell the story (if you copy-paste this into your shell, it should give you a repo which shows what I would like to see in history):

    # Start off with a repository with some history
    git init testrepo
    cd testrepo
    echo Ancient history > it
    git add it
    git commit -m "The master branch"
    
    # Add the test
    git checkout -b feature
    echo Demonstrate the problem in a test > test
    git add test
    git commit -m "Add test"
    
    # Try coming up with a fix
    echo First rejected solution > it
    git commit -am "First rejected solution"
    
    # Decide to reject it but archive it for posterity.
    git branch reject
    git reset --hard feature~
    git merge --no-ff --strategy=ours -m "Archive first rejected solution" reject
    git branch -d reject
    
    # Try a second solution
    echo Second rejected soultion > it
    git commit -am "Second rejected solution"
    
    # Decide to reject and archive it
    git branch reject
    git reset --hard feature~
    git merge --no-ff --strategy=ours -m "Archive second rejected solution" reject
    git branch -d reject
    
    # Come up with the correct solution on the third attempt
    echo The accepted solution > it
    git commit -am "The best solution"
    
    # Merge the feature
    git checkout master
    git merge --no-ff --no-edit feature master
    git branch -d feature

which should give you a history that looks like this:


    *   ad70165 (HEAD -> master) Merge branch 'feature'
    |\  
    | * e36216e The best solution
    | *   79ed8f8 Archive second rejected solution
    | |\  
    | | * 201a592 Second rejected solution
    | |/  
    | *   7c8a5b6 Archive first rejected solution
    | |\  
    | | * fd23f92 First rejected solution
    | |/  
    | * 4b2a076 Add test
    |/  
    * e005643 The master branch

gonzaponte:
Ok, I think I've done it correctly, but now it seems to complain in the web interface about not being able to rebase without conflicts. Is this a problem?

jacg:
I've fetched your branch and it looks just fine, sitting on top of master. A `merge --no-ff`seems to work fine too. It's not clear to me what it is complaining about.

jacg:
I think I get it. When you rebased it on top of master, you had to reselve conflicts, didn't you? (I certainly did when I tried it just now.) I think that the GitHub message is saying just that: it wouldn't be able to rebase it for you automatically, because it doesn't know what to do with those conflicts. As you have resolved those conflicts manually, you have done what GitHub was unable to do automatically.

So, I think you should just proceed with the merge.

jacg:
Incidentally, it's also clear why there are conflicts: when you created the archive merge commits, you had to resolve a conflict, but you did this automatically with the `--strategy=ours` option to `merge`. The rebase doesn't know that it can apply this strategy to the merge commits, so it presents you with the conflicts that arise.

gonzaponte:
I didn't rebase on top of master at any point because the branch is already there, but if I do it now explicitly it does complain. Your explanation makes sense, so we'll go ahead and merge. Thank you for everything.

jacg:
BTW, if you had had `rerere` switched on, then the conflict resolution would have been automatic during the rebase. (I have disabled `rerere` becasue, for teaching purposes, I need to be able to manually resolve the same conflicts over and over again, slowly and explicitly in demonstrations. For real life, it's probaly worth enabling. It's annoying if you messed up a conflict resolution and want to redo it ... if `rerere` automatically reuses the recorded resolution for you in such a situation (and you've forgotten that `rerere` is enabled) then you can get very confused and frustrated, so beware.)

gonzaponte:
I occasionally mess things up in complicated rebases so I prefer to have it disabled.

PR #470, opened 2018-05-01
Write MC true info of multiple files when the first event of the first file is filtered out

paolafer:
This PR fixes a bug in the MC true information writing, which only appears in a very specific case, i.e., when a job reads multiple files and the first event of the first file is filtered out by some selection that is done in the job.
A test has been added to show the incorrect behaviour and the code has been subsequently fixed.


jmbenlloch (approved):
Simple and concise PR, first includes a test that shows the failure and then a commit fixing it. Good job!

PR #472, opened 2018-05-28
Fix emacs init compilation error

jacg:
The current `init.el` fails to compile out of the box, because it uses a feature which seems to be supported no more. The first commit in this PR fixes that problem.

The second commit in this PR is not directly related, but as it's such a minor thing, I'm proposing to sneak it in, in order to save opening a separate PR for trivialities. It enables helm for buffer switching.


jjgomezcadenas (approved):
Good magic indeed. Agreed to sneak the small commit in this PR.

PR #473, opened 2018-05-30
Add number of entries to the stats printed in the histograms

Aretno:
Small change to add the number of entries to the printed statistic values when plotting histograms.


gonzaponte (approved):
Clear and simple.

PR #474, opened 2018-06-04
Changes in calibration cities for other setups

andLaing:
Reintroduced the copy of sensor information
from the raw files to the histogram files.
In this way a setup without reliable database
sensor mapping.

Protections introduced so that framework works
with data where there are no data from one
type of sensor.

centralised copy_sensor_table

Update to make the code work if there are no PMTs or Sipms

Fixed accidentally deleted 1 sensor type protections

Temporary patch to read files without sipmrwf

Improved sipmpdf adc_only mode

### DIFF
--- a/invisible_cities/calib/calib_functions_test.py
+++ b/invisible_cities/calib/calib_functions_test.py
@@ -10,7 +10,7 @@ from pytest        import raises
 from .                         import calib_functions as cf
 from .. reco                   import tbl_functions as tbl
 from .. core.system_of_units_c import units
-from .. evm.nh5                import DataSensor
+from .. evm.nh5                import SensorTable
 
 
 def test_bin_waveforms():
@@ -102,7 +102,7 @@ def test_filter_limits_outside():
 def test_copy_sensor_table(config_tmpdir):
 
     ## Create an empty input file to begin
-    in_name = os.path.join(config_tmpdir, 'test_copy_in.h5')
+    in_name  = os.path.join(config_tmpdir, 'test_copy_in.h5')
     out_name = os.path.join(config_tmpdir, 'test_copy_out.h5')
     with tb.open_file(in_name, 'w') as input_file:
         input_file.create_group(input_file.root, 'dummy')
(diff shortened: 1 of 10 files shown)


andLaing:
The changes to protect against data with different configurations, particularly with only one type of sensor were necessary for the work that is being done at IFIC at the moment but are probably insufficient for the future when we will have to run over several different detector configurations. That being said, @jmbenlloch and I thought they might open debate so that this can be implemented in a more general way.

gonzaponte:
We need to seriously think how to manage different detector configurations with our code automatically, but this patch is perfectly fine for the time being.

gonzaponte:
Can you add the following tests?

- The equivalent of `test_sipmpdf_sipmdarkcurrent` for the `adc_only = True` case. Parametrize the test if it is easy to do so and it keeps readable.
- A simple test that checks all scenarios for `copy_sensor_table`
- Test for both `get_vectors` and `get_rwf_vectors`. These should already be there way before you'd modified them.

andLaing:
Added an option (+test) for pmt calibration without deconvolution as requested by @Aretno since it's related to this PR as needed for the Ben Gurion setup.

invisible_cities/calib/calib_functions_test.py
```diff
@@ -92,3 +92,26 @@ def test_filter_limits_outside():
     assert len(filtered_dlimits) < len(unfiltered_dlimits)
     assert len(filtered_llimits) % 2 == 0
     assert len(filtered_dlimits) % 2 == 0
+
```
gonzaponte, line 100: I would like this commit ("Protect changes") and "Added tests for changes" to be just one commit. You can squash the latter into the former. Ask me if in doubt.

invisible_cities/calib/calib_functions_test.py
```diff
@@ -94,24 +99,42 @@ def test_filter_limits_outside():
...
+    with tb.open_file(out_name, 'w') as out_file:
+
+        ## Nothing to copy
+        cf.copy_sensor_table(in_name, out_file)
+
+        ## Sensor group with no table
+        with tb.open_file(in_name, 'a') as input_file:
+            sens_group = input_file.create_group(input_file.root, 'Sensors')
+        cf.copy_sensor_table(in_name, out_file)
+        assert 'Sensors' in out_file.root
```
gonzaponte, line 120: `test_copy_sensor_table` should be split into 3 different tests. The first one should be up to here, where you simply check that the node is properly created.

invisible_cities/calib/calib_functions_test.py
```diff
@@ -94,24 +99,42 @@ def test_filter_limits_outside():
...
+        out_file.remove_node(out_file.root.Sensors)
+
+        ## Only PMTs
+        with tb.open_file(in_name, 'a') as input_file:
```
gonzaponte: Here starts another test for the same function `test_copy_sensor_table`, where you check that the table you expect the function to create is there.

invisible_cities/calib/calib_functions_test.py
```diff
@@ -94,24 +99,42 @@ def test_filter_limits_outside():
...
+            sens_group = input_file.root.Sensors
+            input_file.create_table(sens_group, "DataPMT", DataSensor,
+                                    "", tbl.filters("NOCOMPR"))
+        cf.copy_sensor_table(in_name, out_file)
+        assert 'DataPMT' in out_file.root.Sensors
+        out_file.remove_node(out_file.root.Sensors, recursive=True)
+
+        ## All sensors
```
gonzaponte: And here starts the other test for `test_copy_sensor_table`, doing the same for SiPMs.

invisible_cities/calib/calib_functions_test.py
```diff
@@ -94,24 +99,42 @@ def test_filter_limits_outside():
...
+        with tb.open_file(in_name, 'a') as input_file:
+            sens_group = input_file.root.Sensors
+            input_file.create_table(sens_group, "DataSiPM", DataSensor,
+                                    "", tbl.filters("NOCOMPR"))
+        cf.copy_sensor_table(in_name, out_file)
+        assert 'DataPMT'  in out_file.root.Sensors
+        assert 'DataSiPM' in out_file.root.Sensors
+
```
gonzaponte: These tests are fine, but you also need to ensure the data you tell the function to write is there in the output file, that would be another test :)

invisible_cities/calib/pmtgain.py
```diff
@@ -113,7 +119,7 @@ def event_loop(self, NEVT, dataVectors):
             # LED (correlated)
             led_ints  = cf.spaced_integrals(pmt_adc, self.l_limits)[:, ::2]
             pmt_spe  += cf.bin_waveforms(led_ints, self.histbins)
-
+
```
gonzaponte: remove these whitespaces

gonzaponte (commented):
Nice. Almost there!

andLaing:
Happy?

jjgomezcadenas:
@gonzaponte please speak now or keep your peace, if I don't hear from you I will approve this PR late in the day

gonzaponte:
> Happy?

I guess. I don't have time to look at this in detail, but it seems to be perfectly fine.
Please remove the whitespaces in `pmtgain.py` before merge.

jjgomezcadenas (approved):

PR #476, opened 2018-06-08
Nullify waveforms of masked sensors

gonzaponte:
Mask sensors in the code. We removed it somehow from the code.

### DIFF
--- a/invisible_cities/calib/pmtgain.py
+++ b/invisible_cities/calib/pmtgain.py
@@ -92,7 +92,7 @@ class Pmtgain(CalibratedCity):
         events_info = dataVectors.events
 
         ## Where we'll be saving the binned info for each channel
-        shape    = len(self.pmt_active), len(self.histbins) - 1
+        shape    = len(self.pmt_active_list), len(self.histbins) - 1
         pmt_spe  = np.zeros(shape, dtype=np.int)
         pmt_dark = np.zeros(shape, dtype=np.int)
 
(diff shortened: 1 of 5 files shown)


paolafer:
Hi Gonzalo! I'm a bit confused about this. When was the masking code removed? We had to modify the corona algorithm to take into account masked channels, and I think we tested it with the whole code, right?

gonzaponte:
> When was the masking code removed?
I don't know...

> We had to modify the corona algorithm to take into account masked channels, and I think we tested it with the whole code, right?
Yes, but I've just realized that the changes affect only to the corona algorithm when it is not configured for barycenter.

jjgomezcadenas (commented):
This PR kills a monster that has been killed, I think a number of times and then has resurrected. Hopefully it will stay dead from now on.

jjgomezcadenas (approved):

jacg:
Why was the monster's resurrection not detected by a test?

gonzaponte:
> Why was the monster's resurrection not detected by a test?

I don't think we've paid too much attention to this monster. And I think we had only amputated a few members, but never killed it. It should be dead now.

Now in technical terms:
I think we've never fully masked the dead sensors. My feeling is that we just masked them in some particular scenarios (corona algorithm), but not generally (when corona is configured to perform a barycenter, for instance).
In this PR the waveforms of the masked sensors are nullified (set to 0). This way, they can never be used for anything.

gonzaponte:
By the way, I am aware this whole thing needs a careful review, but we need this now, so we will pay the technical debt later.

jacg:
> I am aware this whole thing needs a careful review, but we need this now, so we will pay the technical debt later.

Fair enough. But I note that we have been taking on a lot of technical debt over the last months, without repaying any: we **really** need to start repaying some of it soon.

I guess that our next sprint should be largely dedicated to repayment of technical debt. Starting with the total liquification of the cities.

jjgomezcadenas:
Agree, we are accumulating TD. Worst, our (expert) man power is dwindling. But yes, agreed, we need to devote the next Sprint to liquify and also to make IC detector-configurable. Among a zillion other things, I guess.

PR #477, opened 2018-06-08
Fix bug reading the MC response of the last sensor written in one event

paolafer:
The function that reads the Monte Carlo waveforms of the sensors was not reading the last sensor in each event. I have added a test which reveals this bug and fixed the code accordingly.

### DIFF
--- a/invisible_cities/io/mcinfo_io_test.py
+++ b/invisible_cities/io/mcinfo_io_test.py
@@ -214,3 +214,10 @@ def test_load_sensors_data(mc_sensors_nexus_data):
     samples = list(zip(bins, wvf.charges))
 
     assert np.allclose(samples, sipm)
+
+    with tb.open_file(efile, mode='r') as h5in:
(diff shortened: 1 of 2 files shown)


invisible_cities/io/mcinfo_io_test.py
```diff
@@ -214,3 +214,10 @@ def test_load_sensors_data(mc_sensors_nexus_data):
     samples = list(zip(bins, wvf.charges))
 
     assert np.allclose(samples, sipm)
+
+    with tb.open_file(efile, mode='r') as h5in:
```
jacg: This seems to have been added to an existing test, which suggests that this test now (and perhaps even before) tests for more than one thing. Would it be too much trouble to split this big test into separate smaller tests?

paolafer: Done!

jmbenlloch (approved):
This PR fixes a bug when reading MC info and includes a test that shows everything is working ok.

PR #479, opened 2018-06-10
Allow the saving of events with 0 S2 or 0 S2 in dorothea

gonzaponte:
Allow the saving of events with 0 S2 or 0 S2 in dorothea.
Useful to keep track of detection efficiencies.


jjgomezcadenas (approved):
All straight forward and badly needed for immediate analysis of data popping out for the machine.

PR #480, opened 2018-06-10
Flatten mask in display_matrix

jjgomezcadenas:
The display function `display_matrix` takes an optional mask array that allows to plot only those values of the matrix accepted by the mask. This feature had not been previously tested, and thus we had not noted that the mask had to be flattened. This is now fixed.


gonzaponte (approved):
Straightforward fix. Approved.

PR #482, opened 2018-06-13
Reduction of memory usage for monitor plotting

Aretno:
Quick fix to reduce memory usage for histogram plotting for the monitoring. Kind of urgent as online monitoring currently crushes every now and then due to a memory leak.


gonzaponte (approved):
Nice. Desperately needed to avoid crashes on a daily basis.

PR #483, opened 2018-06-17
Fix bug introduced in #482

Aretno:
If-related fuck-up, fixing it. (The if was mean to be independent of the reference histogram).

### DIFF
--- a/invisible_cities/icaro/histogram_plot_functions.py
+++ b/invisible_cities/icaro/histogram_plot_functions.py
@@ -157,11 +157,11 @@ def plot_histograms(histo_manager, histonames='all', n_columns=3, plot_errors=Fa
         fig, axes = plt.subplots(n_rows, n_columns, figsize=(8 * n_columns, 6 * n_rows))
 
     for i, histoname in enumerate(histonames):
+        if out_path:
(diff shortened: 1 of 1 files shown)


jacg:
Obligatory questions:

1. Why was this not caught by a test?
2. Where is the test which points out the bug now?

Aretno:
Well, the answer to both of them is that it is a plotting function and I understood that we were still not clear on how to properly test them but I guess I am wrong. No problem in adding a test, I just can't figure out which could be.

jacg:
Sorry, I missed the fact that it was about plotting. Two points about plotting in general, and testing plotting functions specifically:

1. Please make sure that generation of the data and actually presenting it graphically are decoupled steps in separate functions. One of the *many* advantages of doing it this way is that the first step becomes easily testable.

2. We *do* have a proof of concept for testing graphics: #119. Once we get Git LFS up and running, we should start making this standard practice.

Aretno:
Oh, I didn't now about the testing graphics possibility, I agree that we should get that running, of course. On the other hand, yes, we generate the data with separate functions (properly tested) and then we give the data to the function to plot it.

gonzaponte (approved):
Ok.

PR #486, opened 2018-06-21
Fix bug last written sensor

paolafer:
There was a bug in the code of `read_mcsns_response`: when storing a new `Waveform` object, the bin width of the following one was taken. A test has been added to show this bug, and the code has been fixed. In this process, I've realized that the bin width was actually being saved without units, so I have also added this feature. The test added can serve also as a test for the units being saved correctly.


carmenromo (approved):
We have reviewed the code, there was indeed a bug and it has been successfully fixed including a test that shows it.

jmbenlloch (approved):

Issue #484, opened 2018-06-18
Silence warnings


jacg:
5c7125f2a669439d64d3d2bfbb331514830212bf (the second commit of #434) introduces screenfuls of warnings like this:

    envs/IC3.6/lib/python3.6/site-packages/tables/path.py:112: NaturalNameWarning: object name is not a valid Python identifier: '0BsIflpz5M2'; it does not match the pattern ``^[a-zA-Z_][a-zA-Z0-9_]*$``; you will not be able to use natural naming to access this object; using ``getattr()`` will still work, though
    NaturalNameWarning)

This pretty much ensures that if a meaningful warning ever appears somewhere nearby, it will not be seen, as it will be completely drowned out by this noise. So these warnings need to be silenced.

This might be an interesting task for someone who has not yet had a PR accepted into nextic: it will probably not require involved coding, but will serve as practice in the nextic PR cycle.

jjgomezcadenas:
Nothing is being done here, and the warnings are extremely annoying. This needs to be fixed urgently, if no other volunteers by the author of of the commit.

Aretno:
At @jjgomezcadenas request I patched the thing at #487. (Sorry, didn't see this issue before aside from some email from him).

PR #487, opened 2018-06-26
Improve histograms tests

Aretno:
Optimization of the hypotheses generation to make tests faster. 
Removed deadline for 4 tests that were borderline (it was giving a warning that the current value was to be changed).
Silenced the NaturalNameWarning to avoid noise running the tests. This will be revised in the future.

### DIFF
--- a/invisible_cities/evm/histos_test.py
+++ b/invisible_cities/evm/histos_test.py
@@ -20,7 +20,8 @@ from hypothesis.strategies  import one_of
 from .. evm.histos  import HistoManager, Histogram
 
 
-characters = tuple(string.ascii_letters + string.digits + "-")
+characters = tuple(string.ascii_letters + string.digits)
+letters    = tuple(string.ascii_letters)
 
 def assert_histogram_equality(histogram1, histogram2):
     assert np.all     (a == b for a, b in zip(histogram1.bins, histogram2.bins))
@@ -57,7 +58,7 @@ def filled_histograms(draw, dimension=0, fixed_bins=None):
     else:
         bins = draw(bins_arrays(dimension=dimension))
 
-    title  = draw(text(characters, min_size=1))
+    title  = draw(text(letters, min_size=1)) + draw(text(characters, min_size=1))
     labels = draw(lists(text(characters, min_size=1), min_size=dimension, max_size=dimension))
     shape  = draw(integers(50, 100)),
     data   = []
@@ -77,7 +78,7 @@ def empty_histograms(draw, dimension=0):
     if dimension <= 0:
         dimension = draw(sampled_from((1,2)))
     bins   = draw(bins_arrays(dimension=dimension))
-    title  = draw(text(characters, min_size=1))
+    title  = draw(text(letters, min_size=1)) + draw(text(characters, min_size=1))
     labels = draw(lists(text(characters, min_size=1), min_size=dimension, max_size=dimension))
     args   = title, bins, labels
     return args, Histogram(title, bins, labels)
--- a/invisible_cities/evm/histos_test.py
+++ b/invisible_cities/evm/histos_test.py
@@ -20,8 +20,9 @@ from hypothesis.strategies  import one_of
 from .. evm.histos  import HistoManager, Histogram
 
 
-characters = tuple(string.ascii_letters + string.digits)
-letters    = tuple(string.ascii_letters)
+characters = string.ascii_letters + string.digits
+letters    = string.ascii_letters
+
 
 def assert_histogram_equality(histogram1, histogram2):
     assert np.all     (a == b for a, b in zip(histogram1.bins, histogram2.bins))
@@ -31,17 +32,24 @@ def assert_histogram_equality(histogram1, histogram2):
     assert             histogram1.title   == histogram2.title
     assert             histogram1.labels  == histogram2.labels
 
+
+@composite
+def titles(draw):
+    return draw(text(letters, min_size=1)) + draw(text(characters, min_size=5))
+
+
 @composite
 def bins_arrays(draw, dimension=0):
     if dimension <= 0:
         dimension = draw(sampled_from((1,2)))
 
-    bin_margins   = draw(arrays(float, (dimension, 2),
-                         floats(-1e3, 1e3, allow_nan=False, allow_infinity=False)))
-    for bin_margin in bin_margins:
-        assume(round(bin_margin.min(), 2) < round(bin_margin.max(), 2))
+    bin_lower_margins = draw(arrays(float, dimension,
+                             floats(-1e3, 1e3, allow_nan=False, allow_infinity=False)))
+    bin_additive      = draw(arrays(float, dimension,
+                             floats(1.1 , 1e3, allow_nan=False, allow_infinity=False)))
+    bin_upper_margins = bin_lower_margins + bin_additive
 
-    bins = [ np.linspace(bin_margin.min(), bin_margin.max(), draw(integers(2, 20))) for bin_margin in bin_margins ]
+    bins = [ np.linspace(bin_lower_margins[i], bin_upper_margins[i], draw(integers(2, 20))) for i, _ in enumerate(bin_lower_margins) ]
 
     return bins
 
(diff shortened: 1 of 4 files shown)


jacg:
Two things bother me at first glance:

1. This single commit mixes two orthogonal concerns:
    + Silencing warnings
    + Optimizing hypothesis generation

2. Concerning the warnings, the solution cures the symptoms rather than the cause. The warnings are generated and then suppressed. In general it is better no write code that doesn't generate warnings, rather than silencing the warnings that are generated by the code: make sure that the warnings aren't generated in the first place! Do we really need names which are not valid identifiers? Where do these names come from? Will suppressing the warnings suppress other important warnings?

jacg:
Removing the deadline altogether seems a bit ham-fisted? Wouldn't it be better to set the deadline to the future default of 200 ms, or to some other value that is reasonable. I don't think that we want to disable it completely.

Aretno:
Yeah, regarding the first concern you're right, I'll look into it and try to separate the two in two commits.

The names come from random generated names for the histograms, basically... but I think that should be solvable by making the names a tiny bit less generic (basically removing the possibility of the first character to be a number), I'll look into that.

The deadline time... mmm I have to look into more detail but at a quick sight the only way was to speed up the tests which didn't seemed as easy to me (the ones giving the deadline warning were those with both histogram and pmaps generation or with histogram generation + reading and/or writing from file)

Anyway, I see a test failed (weird, didn't failed at all in my compute) so I'll check the log and see what happens; if this is not that urgent I'll check all issues altogether.

gonzaponte:
Concerning the silenced warnings, the idea is to silence them temporarily
while we find someone suitable for your proposal in #484. This is certainly
not a fix.

O 23:52, Mar., 26 de xuño de 2018, Aretno <notifications@github.com>
escribiu:

Aretno:
Btw, I haven't been able to reproduce Travis errors in my computer even with the seed given (`--hypothesis-seed=302934307671667531413257853548643485645`). No idea on what to do in this case.

jjgomezcadenas:
Travis is failing...

Aretno:
So I think I addressed @jacg comments, commits are separated, deadline is set high but not None (this can be investigated further but I suspect that it just is slow), the warning silence was removed and warnings were fixed.

However, Travis is still failing and I don't know why, I still haven't been able to get this error on my computer even using the same seed. Anybody knows why? Without being able to locally evaluate the failure it may take me forever to understand what may be happening.

jacg:
> Concerning the silenced warnings, the idea is to silence them temporarily while we find someone suitable for your proposal in #484. This is certainly not a fix.

@gonzaponte The trouble with that is: out of sight, out of mind. If you silence them, they won't annoy anyone so nothing will be done about it, and the cure of the symptoms rather than the cause will stay forever. But if I have understood @Aretno reply, then it looks like it has been fixed properly already.

invisible_cities/evm/histos_test.py
```diff
@@ -20,7 +20,8 @@
 from .. evm.histos  import HistoManager, Histogram
 
 
-characters = tuple(string.ascii_letters + string.digits + "-")
+characters = tuple(string.ascii_letters + string.digits)
+letters    = tuple(string.ascii_letters)
```
jacg: No need for `tuple` here. Strings are perfectly good iterables, and `text` can consume them directly.

invisible_cities/evm/histos_test.py
```diff
@@ -57,7 +58,7 @@ def filled_histograms(draw, dimension=0, fixed_bins=None):
     else:
         bins = draw(bins_arrays(dimension=dimension))
 
-    title  = draw(text(characters, min_size=1))
+    title  = draw(text(letters, min_size=1)) + draw(text(characters, min_size=1))
```
jacg: Instead of defining `title` in multiple places, why not define it once as a strategy, and reuse it multiple times?

invisible_cities/io/hist_io_test.py
```diff
@@ -79,7 +79,7 @@ def test_save_histomanager_to_file_append_mode(output_tmpdir, histogram_list, gr
 
 
 @settings(suppress_health_check=(HealthCheck.too_slow,))
-@given   (histograms_lists(), text(characters, min_size=1), text(characters, min_size=1, max_size=1).filter(lambda x: x not in ["w", "a"]))
+@given   (histograms_lists(), text(letters, min_size=1), text(letters, min_size=1, max_size=1).filter(lambda x: x not in ["w", "a"]))
```
jacg: I see that you haven't quite grasped Duck Typing and its consequence for sequences and iterables. `str`, `list`, `tuple` etc. are all sequences. All sequences are iterable, but there are many more iterable things besides. When you need a sequence or an iterable its *concrete* type is not important.

Consequently `x not in ["w", "a"]` is just a very noisy and inconvenient way of spelling `x not in 'wa'`!

Both `["w", "a"]` and `'wa'` are sequences of length 2, containing the two elements `'w'` and `'a'`. From the perspective of the semantics of the `in` operator, they are identical.

jacg: Oh, just to be sure, there is **absolutely no semantic difference** between `'` and `"`: both create strings of exactly the same type.

jacg:
Remember that, by default, we run the tests with the `dev` hypothesis profile, which only tries 10 examples, in order to keep the edit-test cycle nice and quick. On Travis we run the tests with the `hard` profile, which tries 1000 examples, in order to do more thorough testing.

Running the tests on my machine with the default `dev` profile, repeatedly passed all tests. Running with the `hard` profile

    HYPOTHESIS_PROFILE=hard <your usual test invocation command>

caused the test in question to fail.

Once hypothesis has found a failure (on your machine), it will store it in its database, and subsequent runs of the same test should fail immediately, even in the `dev` profile.

I haven't managed to get the seed to reproduce the error though.

Anyway, the hard profile should give you something to work with.

Aretno:
Oh ok, thanks! That works and I seem to be able to reciprocate the error with what you gave me, I'll work on it and try to provide a solution ASAP. I fixed the other concerns you had about the code, thanks for the input.

jacg:
The last commit seems to have fixed the test failure.

Now, it's unfortunate that this PR contains two orthogonal bits of work, but as they are both pretty small, I guess we can live with it.

The title should probably be changed to something like 'improve histogram tests', and it should probably contain just two commits:

1. silence warnings

2. optimize hypothesis generation

Don't forget to put `close #484` or similar in the commit message or the description of the PR. I guess it belongs in the commit message of the 'silence warnings' commit.

jacg:
@Aretno I see that you have written `Close Issue 484` in the commit message, presumably in response to an earlier request of mine. I was alluding to https://help.github.com/articles/closing-issues-using-keywords/

I don't think that what you wrote will work, I think that the `#<number>` syntax is required, so you would have to write `Close #484`.

jacg:
@Aretno I've just tried running `histos_test.py`, using the `hard` profile, before and after the optimization commit. Timings for all the tests to complete:

+ Before optimization: 206 seconds.
+ After optimization: 221 seconds!

Please check whether the 'optimization' also slows down the tests on your machine and report back.

I don't want this to delay acceptance of the silencing of the warnings, so I'll extract that work into a separate PR, and merge it ASAP. (Which is why these orthogonal bits of work should have been in separate PRs in the first place!)

jacg:
@Aretno I need a bit of help in confirming whether the warning suppression has worked.

On the one hand I can't trigger the `NaturalNameWarning` on my machine. I have tried running the `histos_test.py` tests with the `hard` profile and with `-s` and I don't get them, even for commits prior to the one that is supposed to solve the problem.

On the other, I get a bazillion warnings that look like

    Warning: Histogram titles are different.
                          ['aGba', 'a'], ['appa', 'a6aALO']

The latter don't appear in pytests's warnings summary (only when running whith `-s`). Also, they don't look like real warnings, so I suspect that they aren't real warnings, but just prints. Do you have any idea where they come from?

jacg:
@Aretno OK, I've found the test that was generating the `NaturalNameWarnings`. Still puzzled by the printouts.

Aretno:
Hi Jacek, these are prints, you're right:

```
Warning: Histogram titles are different.
                      ['aGba', 'a'], ['appa', 'a6aALO']
``` 

are normal and appear when trying to join two histograms together that have different title, the print is intended to appear and it is part of the `__add__` of the Histogram class (`evm/histos.py`).

Regarding the time of the test... I'm gonna check and get back to you. The issue was that lots of not valid examples were generated due to a couple of assumes that were removed, the idea was that reduce the number of rejected draws so I am surprised that it actually slows down the whole thing.

Aretno:
In my computer is also slower, 230s vs 260s. But the problem is that before the change there was some chance that when running the test a HealthCheck failed (too many invalid examples generated) and the test failed. This should ensure that the HealthCheck passes. Other way to go would be to ignore the HealthCheck. Or optimizing the code in some way I haven't think of (probably the best way to go).

jacg:
OK, this PR definitely needs splitting.

I have already submitted the warnings silencing as #491. Please open a separate PR for the hypothesis optimization.

PR #489, opened 2018-06-27
Fixes so that both ZS data and MC can be processed

andLaing:
Designed to avoid bad baseline calculations
when in ZS mode and to bring MC diomira ouput
in line with ZS data format.

there is currently a fail in the test test_mean_for_pmts_fee_is_unbiased
sometimes but other times not. I can't find any reason why it should be
affected.

### DIFF
--- a/invisible_cities/reco/calib_sensors_functions.py
+++ b/invisible_cities/reco/calib_sensors_functions.py
@@ -34,14 +34,34 @@ def mode(wfs, axis=0):
     positive waveforms.
     """
     def wf_mode(wf):
-        positive = wf >= 0
-        return np.bincount(wf[positive]).argmax() if np.count_nonzero(positive) else -1
+        positive = wf > 0
+        return np.bincount(wf[positive]).argmax() if np.count_nonzero(positive) else 0
     return np.apply_along_axis(wf_mode, axis, wfs).astype(float)
 
 
-def means  (wfs): return to_col_vector(np.mean  (wfs, axis=1))
-def medians(wfs): return to_col_vector(np.median(wfs, axis=1))
-def modes  (wfs): return to_col_vector(   mode  (wfs, axis=1))
+def median(wfs, axis=0):
+    """
+    The numpy median but with protection
+    so that we get the correct answer in case of 
+    zero suppressed data
+    """
+    mask_wfs = np.ma.masked_where(wfs == 0, wfs)
+    return np.ma.median(mask_wfs, axis=axis).filled(0)
+
+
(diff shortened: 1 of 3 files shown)


invisible_cities/reco/calib_sensors_functions.py
```diff
@@ -34,14 +34,34 @@ def mode(wfs, axis=0):
...
+
+
+def mean(wfs, axis=0):
+    """
+    The numpy mean but with protection
+    so that we get the correct answer in case of 
+    zero suppressed data
+    """
+    mask_wfs = np.ma.masked_where(wfs == 0, wfs)
+    return np.ma.mean(mask_wfs, axis=axis).filled(0)
```
jacg: `median` and `mean` are just *identical* thin wrappers around `np.ma.median` and `np.ma.mean`. As such the signal to noise ratio in them isn't all that great.

I would prefer something along the lines of 

    from functools import wraps

    def masked_version(fn):
        @wraps(fn)
        def proxy(wfs, *args, **kwds):
            maks_wfs = np.ma.masked_where(wfs == 0, wfs)
            return fn(maks_wfs, *args, **kwds).filled(0)
        proxy.__doc__ = "Masked version blah blah \n\n" + proxy.__doc__
        return proxy

    median = masked_version(np.ma.median)
    mean   = masked_version(np.ma.mean)

The exact mangling of the `__doc__` would require a few more seconds of thought.

invisible_cities/reco/calib_sensors_functions_test.py
```diff
@@ -247,6 +247,7 @@ def test_wf_baseline_subtracted_is_close_to_zero(gaussian_sipm_signal):
 def test_mean_for_pmts_fee_is_unbiased(square_pmt_and_sipm_waveforms):
     _, _, pmts_fee, _, _, _, _ = square_pmt_and_sipm_waveforms
     pmts_bls = csf.subtract_mean(pmts_fee) # ped subtracted near zero
+    print(pmts_bls)
```
jacg: I assume that this `print` is a hangover that will be eliminated before merging.

andLaing: Yep, sorry, was trying to understand the fail I was getting and forgot to remove it

jmbenlloch (approved):
This is a required update to make MC and data alike when using zero suppression. Needed update to test the new configuration at Canfranc.

jmbenlloch:
Test `test_mean_for_pmts_fee_is_unbiased` is now marked as flaky to avoid future issues with Travis.

PR #490, opened 2018-06-28
Dst io

jjgomezcadenas:
This PR introduces a protection in the function that reads dorothea DSTs, throwing an exception if the operation is not successful. This protection is necessary to prevent the case in which a spurious file appears in a directory containing DSTs (and being read out using glob). Without the exception, the function will read files until it finds the wrong file, then die ungracefully, and the concatenation of files will not occur.


invisible_cities/io/dst_io.py
```diff
@@ -1,13 +1,29 @@
...
 import pandas as pd
-
+from .. core.exceptions import UnknownDST
 
 def load_dst(filename, group, node):
     with tb.open_file(filename) as h5in:
-        table = getattr(getattr(h5in.root, group), node).read()
+        try:
+            table = getattr(getattr(h5in.root, group), node).read()
+        except:
```
jacg: `except` without an exception specification is rarely the correct thing to do. You are saying, "I don't care what message is being sent here, whatever it is I know that the way to deal with it is to print a message to standard out and to translate the exception to `UnknownDST`.

What if the message is "Incoming nuclear missile, run to the nuclear shelters right now". Who cares, let's print a message to standard out and translate it to `UnknownDST`.

Please handle the exceptions that you are expecting, and let all others propagate down the stack to whoever *really* knows how to deal with them.

jjgomezcadenas:
@jacg or @jmbenlloch could you please take a look to this? I am afraid I have made a mess, since I see a number of commits that could not possibly belong to this PR added right before my last commit.  Thanks and apologies, working from Korea, at funny times, not always very sharp.

jacg:
@jjgomezcadenas Have a look now. Have the intruding commits disappeared?

jacg:
> The proposed behavior of the function when an exception of type
NoSuchNodeError is found is simply to print an error message, to
inform the user of the existence of an offending file and then ignore
the file. This is a case where it is justified not to abort the concat
of good files (which can take many minutes) because a spurious or
corrupted file is found. No need to throw a new exception either,
although it could be done if one wanted to let the user decide what to do.

This should be a warning, rather than a print.

jacg:
I hope that it goes withot saying, but just in case: Some cleanup will have to be done before merging:

+ The contents of the commit titled `Add correct exception` have nothing whatsoever to do with the title.
+ Squash the three consecutive commits with the same title.
+ `Fix an accidental tab` needs to be squashed into wherever the accident occurred.
+ Spurious whitespace appears at some point at the end of `dst_io_test.py`.

jjgomezcadenas:
@jacg THANKS, intruding commits have disappeared!

As for the rest of the corrections. I feel insecure working with github at this moment (I have spent too much time in the real world and find myself seriously handicapped now that I am trying to go back to Platonic software World). Would you mind to lend me a hand here?

jacg:
@jjgomezcadenas I'd be happy to help, but I'll be on the road for the next few days, and generally mostly unavailable for the next few weeks. Just get the code to do what it is supposed to, and then I'll tidy up the commits for you when I have a spare moment. If getting it merged is urgent, then you'll have to ask @jmbenlloch or @gonzaponte to help.

jjgomezcadenas:
@jmbenlloch can you do the git manipulation suggested by @jagc, then review the PR?

jjgomezcadenas:
@jmbenlloch please check this PR!

jmbenlloch (approved):
Simple update including a test.

PR #491, opened 2018-06-29
Silence NaturalNameWarnings

### LINKED ISSUE #484: Silence warnings

jacg:
Close #484.

### DIFF
--- a/invisible_cities/evm/histos_test.py
+++ b/invisible_cities/evm/histos_test.py
@@ -20,7 +20,9 @@ from hypothesis.strategies  import one_of
 from .. evm.histos  import HistoManager, Histogram
 
 
-characters = tuple(string.ascii_letters + string.digits + "-")
+characters = string.ascii_letters + string.digits
+letters    = string.ascii_letters
+
 
 def assert_histogram_equality(histogram1, histogram2):
     assert np.all     (a == b for a, b in zip(histogram1.bins, histogram2.bins))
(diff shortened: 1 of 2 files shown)


jacg:
@Aretno The majority of the warnings (the `NaturalNameWarning`s) have been removed by this commit.

We still get the occasional `NaturalNameWarning` when a keyword is picked by hypothesis. The keyword always seems to be `as`: I haven't observed `is`, `in` or `or`, or any others. This is a bit puzzling, especially for the short ones. Maybe the hypothesis distribution favours text which starts nearer the beginning of the alphabet. (The strategy can be extended to filter out keywords, to get rid of this noise.)

However, we still have quite a few `HypothesisDeprecationWarning`s. These fall in two categories:

+ deadline
+ timeout

I have insufficent time and insufficient familiarity with the tests that trigger these, to be able to decide how to proceed.

jacg:
This is really @Aretno 's PR, but it is officially recorded by GitHub as mine, so I cannot officially review it. I will just override the restriction, so that we can reduce the noise in the test without further delay.

The `NaturalNameWarning`s have been silenced. There is still an excessive number of hypothesis-related warnings, but those are caused by an unrelated issue, and need a completely unrelated fix.

I hereby approve this PR.

Issue #455, opened 2018-03-17
Change conda activation command


jacg:
Conda now provides a new way of activating environments. Instead of 

    source activate IC3.6

you can now do

    conda activate IC3.6

Enabling this requires a change in `.bashrc`: replacing

    export PATH=~/miniconda/bin:$PATH

with

    . ~/miniconda/etc/profile.d/conda.sh

The new method has all sorts of advantages, including making it possible to have `conda` in your `PATH` without having to activate a conda environment and thereby shadowing (parts of) the standard system python, which may be important for all sorts of non-IC things what you might want to do on your machine.

We should change `manage.sh` to use the new activation style, at which point the `manage.sh` commands will stop working for anyone who already has a local conda installation, until the aforementioned change is introduced to `.bashrc`.

jacg:
Sorry, the previous description was accidentally submitted when it was half-finished. I have edited it, so it should make sense now: to read it you will have to look at this issue in GitHub.

jacg:
Python 3.7 was released 2 days ago. I will soon create a branch that starts migrating IC to 3.7. The conda activation command in `manage.sh` will be changed on that branch.

jacg:
This will be done as part of #496.

PR #493, opened 2018-06-29
Keep an eye on Python 3.7

jacg:
Added Python 3.7 as an allowed failure to the Travis build matrix. Like this we will know when all our dependencies are available on conda in Python 3.7.

I have also taken this opportunity to:

+ Update the pinned versions of our dependencies to the most recent ones
+ Update conda to the most recent version, and change the acivation command to the one which has been available and recommended since 4.4 (#455).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ matrix:
   fast_finish: true
   allow_failures:
   - os: osx
+  - env: IC_PYTHON_VERSION=3.7
 
 env:
   - IC_PYTHON_VERSION=3.7
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ matrix:
   - os: osx
 
 env:
+  - IC_PYTHON_VERSION=3.7
   - IC_PYTHON_VERSION=3.6
 
 os:
--- a/bash_manage.sh
+++ b/bash_manage.sh
@@ -27,7 +27,7 @@ manage.sh()
     if [[ ${prev2} == "manage.sh" ]] ; then
             case "${prev}" in
                         install_and_check|install|work_in_python_version|make_environment)
-                        local versions="3.6"
+                        local versions="3.6 3.7"
                     COMPREPLY=( $(compgen -W "${versions}" -- ${cur}) )
                     return 0
                     ;;
@@ -73,7 +73,7 @@ source_manage.sh()
     if [[ ${prev2} == "manage.sh" ]] ; then
             case "${prev}" in
                         install_and_check|install|work_in_python_version|make_environment)
-                        local versions="3.6"
+                        local versions="3.6 3.7"
                     COMPREPLY=( $(compgen -W "${versions}" -- ${cur}) )
                     return 0
                     ;;
--- a/manage.sh
+++ b/manage.sh
@@ -72,7 +72,7 @@ function install_conda {
     fi
 }
 
-CONDA_ENV_TAG=2018-06-30
+CONDA_ENV_TAG=2018-07-28
 CONDA_ENV_NAME=IC-${PYTHON_VERSION}-${CONDA_ENV_TAG}
 
 function make_environment {
@@ -90,15 +90,15 @@ dependencies:
 - networkx=2.1
 - notebook=5.5.0
 - numpy=1.14.5
-- pandas=0.23.1
-- pymysql=0.8.1
+- pandas=0.23.3
+- pymysql=0.9.2
 - pytables=3.4.4
 - pytest=3.6.2
 - scipy=1.1.0
 - sphinx=1.7.5
 - tornado=5.0.2
 - flaky=3.4.0
-- hypothesis=3.57
+- hypothesis=3.59.1
 - pytest-xdist=1.22.2
 EOF
 
--- a/manage.sh
+++ b/manage.sh
@@ -72,7 +72,7 @@ function install_conda {
     fi
 }
 
-CONDA_ENV_TAG=2018-07-28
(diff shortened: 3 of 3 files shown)


jacg:
I have a feeling that @jjgomezcadenas will like [dataclasses](https://www.python.org/dev/peps/pep-0557/) which have been added in Python 3.7.

jacg:
@jjgomezcadenas `test_spe_to_adc` is the only remaining test to fail after the package upgrades. Can you have a look to see if you can spot what has changed?

jacg:
@jjgomezcadenas The culprit seems to be `scipy` which has changed from version 0.19 to 1.1, so an API change is to be expected.

jjgomezcadenas:
@jacg to be sure: an API change from our side? Sorry about being thick, travelling right now (Korea) and with jet lag.

jjgomezcadenas:
Yes, I like dataclasses very much, they extend the funcionality I am using all the type with named tuples.

jacg:
@jjgomezcadenas No, you can expect an API change in scipy when its version number changes from 0. something to 1.whatever. `test_spe_to_adc` is now failing, but if you roll back scipy from 1.1.0 to 0.19.1 then `test_spe_to_adc` passes.

No point in hanging about in the ancient past (we'll leave that sort of stuff to CERN), so we need to adapt `test_spe_to_adc` to deal with whatever change occurred in scipy between 0.19.1 and 1.1.0.

jjgomezcadenas:
OK, I will look at it as soon as I can.

jacg:
Oh, and before you start looking into this, a word of warning: testing this is a huge PITA, because you need a *different conda environment* for this PR (and anything that is rebased on top of it) while anything that comes before this PR needs the old version of the conda env. Our `manage.sh` utilities cannot (at the moment) distinguish between these two environments, they use the same name for both.

It's not immediately obvious how to deal with it nicely, and to what extent it is worthwhile to build the capability into `manage.sh`. This is a problem we will have to face repeatedly but rarely: every time we want to bump the versions of dependencies.

As an ad-hoc solution, you could try something along the lines of

1. Rename the old env:
    + `conda create -n IC3.6-old --clone IC3.6`
    + `conda remove -n IC3.6 --all`
2. Create a new env for post version-bump work:
    + open new shell
    + `cd $ICDIR`
    + `git checkout python-3.7` (ensure that you have the new versions in `manage.sh`)
    + `source manage.sh make_environment 3.6`

This should give you two conda envs `IC3.6` and `IC3.6-old`, and you'll need to make sure that you are using the appropriate one, depending on whether you are working on a commit which comes before or after the version bump.

jacg:
@jjgomezcadenas #496 should make it easier to investigate the problem with `test_spe_to_adc`, so I suggest you work on it there.

jacg:
The commits in this PR which address dependency evolution, have a more natural place in #496, so **do not merge this** until #496 has been merged or rejected.

jacg:
It seems that Python 3.7 is IC-ready: all our tests pass on Python 3.7.

Do we want to keep Python 3.6 for a while, or shall we switch to 3.7 straight away?_

jjgomezcadenas:
I need to test that 3.7 works in my system, will do it later today or tomorrow (today is Sunday, we are out visiting Chicago)

jjgomezcadenas:
we are using conda 4.5.4, the latest version is 4.5.8. Since we are a it, should we bump conda too?

jjgomezcadenas (approved):
python 3.7 passes all tests in my machine. As an extra check I have run a complex analysis notebook that performs all kind of fits to the lifetime of the detector and the results are identical in 3.6 and 3.7. Thus, it seems that the upgrade is a success. 

I recommend keeping both versions of python for a short while. The current software in production is very stable and I am reluctant to introduce any changes which are not needed in the prod version, but the dev version of the software can already go to the bumped-package, 3.7 python. Over the next couple of weeks we can launch a test production of kdsts with dev and verify that everything is fine. 

However I would recommend all developers to move to the bumped 3.7 version, so that once we move the dev (3.7) to prod, we don't need to support 3.6 anymore (don't see the point, once we have checked everything is fine and transition has been completed. 

Please, @jacg if you can merge right away (let @carmenromo learn with the PR we have assigned to her)

jacg:
> we are using conda 4.5.4, the latest version is 4.5.8. Since we are a it, should we bump conda too?

Yes, I was thinking along those lines too. Of course, the conda version can't really get switched each time we switch historical environments, there's some scope here for inonsistency. I don't see a way around it, though.

jacg:
> we are using conda 4.5.4, the latest version is 4.5.8. Since we are a it, should we bump conda too?

There is no installer for 4.5.8, the most recent one is 4.5.4, so I suggest we leave this until later.

jjgomezcadenas:
Agreed, let's leave 4.5.4 but something went wrong here...

jacg:
> something went wrong here...

What goes wrong (most often) is that hypothesis sometimes finds counterexamples in `test_inverse_cdf_index_hypothesis_generated`. I get the impression that it finds counterexamples much more frequently now, than it did in the past, but because 'the past' is before #496 (Dependency versioning) it's *extremely* painful to track down and understand exactly what is responsible for this effect (and if it really exists ... the inherent randomness and slow turnaround times really don't help either).

Anyway, I'm just going to keep pressing the *Restart job* button on Travis, until I get a pass, so that it's formally approved by Travis. Then I'll open another PR and try to get to the bottom of this. But I need to decouple that issue from this PR, otherwise I'll never make any progress.

Sometimes if fails because of hypothesis health checks, but that seems to be much rarer, and it happens in completely different tests.

PR #494, opened 2018-06-30
Reduce number of non-valid histo examples for tests

Aretno:
Every now and then a HealthCheck error in this direction appeared (only a
few valid examples out of 50), this should address this issue at the
cost of slower test generation.

### DIFF
--- a/invisible_cities/evm/histos_test.py
+++ b/invisible_cities/evm/histos_test.py
@@ -43,12 +43,13 @@ def bins_arrays(draw, dimension=0):
     if dimension <= 0:
         dimension = draw(sampled_from((1,2)))
 
-    bin_margins   = draw(arrays(float, (dimension, 2),
-                         floats(-1e3, 1e3, allow_nan=False, allow_infinity=False)))
-    for bin_margin in bin_margins:
-        assume(round(bin_margin.min(), 2) < round(bin_margin.max(), 2))
+    bin_lower_margins = draw(arrays(float, dimension,
+                             floats(-1e3, 1e3, allow_nan=False, allow_infinity=False)))
+    bin_additive      = draw(arrays(float, dimension,
+                             floats(1.1 , 1e3, allow_nan=False, allow_infinity=False)))
+    bin_upper_margins = bin_lower_margins + bin_additive
 
-    bins = [ np.linspace(bin_margin.min(), bin_margin.max(), draw(integers(2, 20))) for bin_margin in bin_margins ]
+    bins = [ np.linspace(bin_lower_margins[i], bin_upper_margins[i], draw(integers(2, 20))) for i, _ in enumerate(bin_lower_margins) ]
 
     return bins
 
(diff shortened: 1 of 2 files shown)


jacg:
We want to speed up our tests, rather than slow them down. So in order to be able to judge whether we are prepared to take this hit, we need to see some numbers. Can you give an indication of how much this slows down

+ the specific tests concerned (times before / after)
+ the whole test suite (times before / after)

And ideally we would have these numbers for 

+ the dev profile
+ the hard profile

jacg:
I would like to discuss with you how you approached the problem, but my time available for IC is about to run out, so I fear that this will have to wait, unless someoene else picks it up with you.

Aretno:
Ok, I understand, I'll do an extensive profiling.

Regarding the discussion, do you mean all the histogram thing or just this test thing? If this results being too slower when compared to what it was we could go with an ignore of the HealthCheck and open an issue as something to address when you're available or maybe @gonzaponte could take a look at it as he reviewed the code the first time.

jacg:
Heh :-) I'd like to find the time to discuss *all* of it with you, but in this instance I was just referring to the content of this PR, which, unfortunately, is already more that I have time for.

Aretno:
Yeah, I can imagine, some other time then. 

Here are the timing tests, the results of the whole test suit for dev and hard, before (first number) and after (second) the commit:

Dev  :   291s - 231s 
Hard: 2170s - 2136s

For each of the three test files that depend on the changes made:

histos_test:

- Dev  : 11.03s - 6.12s
- Hard:    279s - 270s

histograms_functions_test:

- Dev  : 57.71s - 6.68s
- Hard:    196s -  229s

hist_io_test:

- Dev  : 11.56s - 5.01s
- Hard:  221s   - 215s

So overall, on the dev profile the improvement is quite good although for hard mode in the histograms_functions_test the velocity takes a hit.

jacg:
Thanks @Aretno.

Is that really an improvement by a factor of almost 10 in the `histograms_functions_test` case?

Anyway, overall the timings seem to be improving, so it looks good.

Sorry, I really don't have time to look at the code before my trip. But there seems to be no problem from the timings perspective.

Aretno:
Yup, it is indeed a factor 10, I was surprised by that. I was assuming there would be an improvement because before we tipically kept around 20% of the generated hypothesis and now all the generated are valid but still, a bit shocking of how bad it was before.

jacg:
Any insight into the discrepancy between `dev` and `hard`, where one gains a factor of 10 while the other loses 15%, on the same tests?

Aretno:
No, I have no clue, it is puzzling to me. `hard` seems to be pretty similar in both `histos_test` and `hist_io_test` but in `histograms_functions_test` is considerably slower. And that's just the case where `dev` sees the greatest improvement... All the tests (except the full suite ones) were run twice to make sure there wasn't anything weird, that's all I can say at this point.

jacg:
My initial guess is that `hard` might end up trying a greater proportion of complex cases than `dev` as it gets to search more obscure corners of the example space. If I were to try to get to the bottom of this, that's the first hypothesis I would try to refute.

gonzaponte:
Can the hypothesis saved examples (or whatever it does behind the scenes)
be a factor here?

O 12:34, Dom., 1 de xul. de 2018, jacg <notifications@github.com> escribiu:

jacg:
Yes, I was thinking along those lines too ...

jjgomezcadenas (approved):
In my opinion this PR is mature enough, after the long discussion to be approved.

jacg:
Contrary to the automatic status report, there are conflicts here. This appears to have been approved 24 days ago, and left unmerged ... and now we have to waste some time on resolving the conflicts.

I'm sorry, I'm packing and about to travel, so I can't do this. Could we please merge this before the conflicts grow!

Issue #495, opened 2018-06-30
Strategy for bumping pinned versions of dependencies


jacg:
Since #349 we have pinned versions of the packages we install into the IC conda environment we create with `manage.sh`. This is good, as it makes it much easier to reproduce eariler conditions. Without the pinning, the evolution of our code was decoupled from the evolution of the packages on which it relies.

However, we do not want to stick to the old versions of these packages forever. #493, whose primary purpose is to start the transition to Python 3.7, is taking the opportunity to update the dependency versions. This leaves us with the following problem: Any commits that predate (in our Git DAG, not in time) the version bump, need to use a different conda environment from commits that come after the version bump. According to `manage.sh` both of these environments have the same name. This makes it difficult to run and test code on either side of such version bumps.

Two obvious alternatives come to mind

1. Endow `manage.sh` with the ability to deal with a pre-version-bump and a post-version-bump variant of the environments it manages.
2. Use git hooks to automatically modify the environment whenever you checkout a commit which needs different versions.

Whatever solution we adopt, it would require some careful thought.

jacg:
Of course, we shouldn't re-invent the wheel.

Here's a wheel: https://github.com/SciTools/conda-gitenv

I haven't got the time at the moment to see whether it is square enough for us.

jacg:
This has been incarnated as #496, so I'm closing this.

In the past I would have preferred the merging of #496 to close this automatically, if and when it happens. But I am in the process of moving all open issues and PRs to the project workflow board, and that board will be useless if there is an overwhelming number of cards on it, so I'm closing this sooner rather than later.

PR #496, opened 2018-06-30
Dependency versioning

jacg:
An attempt at addressing #495.

The idea is:

+ Include a dependency configuration tag in a committed file (`manage.sh` for now).
+ Povide `manage.sh` commands which
   + create conda environments which have the tag in their name and contain the correct set of dependencies
   + switch conda environment
+ Make the `manage.sh` testing commands automatically switch to the conda env which is appropriate for the currently checked out code.
+ The environments are made lazily: no environment is created until it is needed, which will only happen if
   + A commit which needs that environment is checked out, and
   + The user asks for the environment to be created, ether
      + explicitly, or
      + implicitly, by using `manage.sh` to run tests.

The tags are, for now, of the form `YYYY-MM-DD`, but we could use relase tag names instead. It might be a good idea to update the dependencies on every release: that way we'll never get too far behind the march of progress.

An example: If you clone and install the repo containing this branch today, you will create a conda env called `IC-3.6-2018-06-30`. If you then check out a commit which corresponds to an earlier dependency configuration, as soon as you run the tests with `source manage.sh run_tests(_par)`, a second conda environment called `IC-3.6-2018-05-03` will be created and activate implicitly. Checking out the tip of the branch once more, and running the tests (with `manage.sh`) will automatically switch to the first environment.

The whole lot needs to be cleaned up, made more robust, and the database should be updated automatically too. We also need to think about how to deal with commits which predate the addition of this feature.

I can imagine that better approaches could be invented. Let me know what you think.

It seems that we don't have the manpower to think about this carefully and the only way to prevent this from being ingnored and festering forever, is to approve and commit the first thing that looks like it works.

### DIFF
--- a/invisible_cities/sierpe/fee_test.py
+++ b/invisible_cities/sierpe/fee_test.py
@@ -53,55 +53,10 @@ def test_fee_params():
     assert FE.f_LPF2 == 10 * units.MHZ
     assert FE.OFFSET == 2500   # offset adc
 
-def test_show_scipy_nasty_feature():
+
+def test_show_signal_decimate_signature():
     """
-    Scipy 1.1.0 changes shamelessly the default behaviour of signal.decimate
-    From the manual:
-
-    v 0.19
-
-    scipy.signal.decimate(x, q, n=None, ftype='iir', axis=-1, zero_phase=None)[source]
-    Downsample the signal after applying an anti-aliasing filter.
-
-
-    Parameters:
-    x : ndarray
-    The signal to be downsampled, as an N-dimensional array.
-    q : int
-    The downsampling factor.
-
-    n : int, optional
-    The order of the filter (1 less than the length for ‘fir’).
-    Defaults to 8 for ‘iir’ and 30 for ‘fir’.
-
-    v 1.1
-
-    n : int, optional
-    The order of the filter (1 less than the length for ‘fir’).
-    Defaults to 8 for ‘iir’ and 20 times the downsampling factor for ‘fir’.
-
-    Thus the behaviour of n changes completely and the default is no longer valid.
-    """
-    ipmt = 0
-    spe = FE.SPE()
-    fee = FE.FEE(noise_FEEPMB_rms = 0 * units.mA, noise_DAQ_rms = 0)
-
-    spe_i = FE.spe_pulse(spe, t0 = 100 * units.ns, tmax = 200 * units.ns)
-    spe_v     = FE.signal_v_fee(fee, spe_i,ipmt)
-
-    scale = 25
-    spe_adc = signal.decimate(spe_v     * FE.v_to_adc(), scale, ftype='fir', zero_phase=False)
-    adc_to_pes     = np.sum(spe_adc    [3:7])
-
-    try:
-        assert 23 < adc_to_pes     < 23.1
-    except AssertionError:
-        pass
-
-def test_fix_scipy_nasty_feature():
-    """
-    This is fixed by setting n to the same value it has in version 0.2
-    (which was 30 by default)
+    This test shows explicitly the signature os signal.decimate, including the need to set n=30
 
     """
     ipmt = 0
--- a/invisible_cities/sierpe/fee_test.py
+++ b/invisible_cities/sierpe/fee_test.py
@@ -1,7 +1,7 @@
 import numpy as np
 
 import pytest
-
+from scipy import signal
 from flaky import flaky
 
 from .. core import system_of_units as units
@@ -53,6 +53,73 @@ def test_fee_params():
     assert FE.f_LPF2 == 10 * units.MHZ
     assert FE.OFFSET == 2500   # offset adc
 
+def test_show_scipy_nasty_feature():
+    """
+    Scipy 1.1.0 changes shamelessly the default behaviour of signal.decimate
+    From the manual:
+
+    v 0.19
+
+    scipy.signal.decimate(x, q, n=None, ftype='iir', axis=-1, zero_phase=None)[source]
+    Downsample the signal after applying an anti-aliasing filter.
+
+
+    Parameters:
+    x : ndarray
+    The signal to be downsampled, as an N-dimensional array.
+    q : int
+    The downsampling factor.
+
+    n : int, optional
+    The order of the filter (1 less than the length for ‘fir’).
+    Defaults to 8 for ‘iir’ and 30 for ‘fir’.
+
+    v 1.1
+
+    n : int, optional
+    The order of the filter (1 less than the length for ‘fir’).
+    Defaults to 8 for ‘iir’ and 20 times the downsampling factor for ‘fir’.
+
+    Thus the behaviour of n changes completely and the default is no longer valid.
+    """
+    ipmt = 0
+    spe = FE.SPE()
+    fee = FE.FEE(noise_FEEPMB_rms = 0 * units.mA, noise_DAQ_rms = 0)
+
+    spe_i = FE.spe_pulse(spe, t0 = 100 * units.ns, tmax = 200 * units.ns)
+    spe_v     = FE.signal_v_fee(fee, spe_i,ipmt)
+
+    scale = 25
+    spe_adc = signal.decimate(spe_v     * FE.v_to_adc(), scale, ftype='fir', zero_phase=False)
+    adc_to_pes     = np.sum(spe_adc    [3:7])
+
+    try:
+        assert 23 < adc_to_pes     < 23.1
+    except AssertionError:
+        pass
+
+def test_fix_scipy_nasty_feature():
+    """
+    This is fixed by setting n to the same value it has in version 0.2
+    (which was 30 by default)
+
+    """
+    ipmt = 0
+    spe = FE.SPE()
+    fee = FE.FEE(noise_FEEPMB_rms = 0 * units.mA, noise_DAQ_rms = 0)
+
+    spe_i = FE.spe_pulse(spe, t0 = 100 * units.ns, tmax = 200 * units.ns)
+    spe_v     = FE.signal_v_fee(fee, spe_i,ipmt)
+
+    scale = 25
+    spe_adc = signal.decimate(spe_v     * FE.v_to_adc(), scale, n = 30, ftype='fir', zero_phase=False)
+    adc_to_pes     = np.sum(spe_adc    [3:7])
+    print(adc_to_pes)
+
+    assert 23 < adc_to_pes     < 23.1
+
+
+
 def test_spe_to_adc():
     """Convert SPE to adc values with the current FEE Parameters must be."""
     ipmt = 0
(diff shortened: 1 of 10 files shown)


jjgomezcadenas:
@jacg shall I clone your branch in an independent area to try it and see if everything works as expected? Willing to help but afraid to mess things up

jacg:
I don't think there's any need for another clone. Try the following

+ Fetch and check out the branch
+ Open a new shell
+ `cd </location/of/your/IC/repo>`
+ `source manage.sh work_in_python_version 3.6` (This should create and activate a new conda env called `IC-3.6-2018-06-30`, and run the tests. You should see `test_spe_to_adc` fail. You should also see that a newer version of pytest is being used: it shows progress percentages in the form `[ 23%]`.

That is enough to set you up for trying to undstand the bug, but if you want to take the new dependency versioning machinery for a spin, then

+ `git checkout HEAD~4` (or whatever it takes to checkout the commit just *before* `Update pinned package versions to most recent`. This doesn't have to be done on the CLI, you could perfectly well do it in Magit.)
+ `source manage.sh compile_and_test_par`

At which point you should automatically find yourself in a new conda env called `IC-3.6-2018-05-30` with the tests being executed. `test_spe_to_adc` (and all the other tests, subject to flakiness) should now pass. You should also see that you are using an older version of pytest: the percentage progress markers are missing.

Another `git checkout historical-dependency-versions && source manage.sh compile_and_test_par` should get you back to your `test_spe_to_adc` failure.

If you want to switch between the two versions, without having to wait for the tests to complete, then use `source manage.sh work_in_python_version_no_tests 3.6` after the checkout.

jacg:
While this PR is perhaps not of the highest short-term priority, I think that it is a vital one for the long-term health of the project and the mental sanity of the developers. A decent implementation will save us lots and lots of developer effort as we go along. I think that it is a mistake to continue to ignore this PR.

jjgomezcadenas:
I agree, this PR needs to be addressed ASAP, but we face three problems:

1. The strongest programmers in the team are either over committed (the case of @jmbenlloch ),  missing in combat (@gonzaponte) or sparsely available (@jacg). 
2. The chamber is in full swing, and many of us (@jahernando, @jjgomezcadenas, @jerenner, @paolafer) have barely the time to cope with the data, develop the analysis code or produce the Monte Carlo. 
3. This PR may be technically tricky and could be handled best working with peras. 

I will be in Geneva for a conference starting on 17 and finishing on 21, I can manoeuvre to arrive to Geneva earlier or leave later. If we we can find one or two days to work together we could fix this PR.

jacg:
@jjgomezcadenas could you at least have a look at why `test_spe_to_adc` is failing with

    >       assert 23 < adc_to_pes     < 23.1
    E       assert 23 < 0.030926200939492603

This has nothing to do with the technical issues that this PR addresses directly, rather with some change in the underlying library versions. As you are intimately familiar with this test, you should be in the best position to understand why it is failing. This failure will need to be addressed before this PR can be approved, but it's orthogonal to the thrust of the PR.

jjgomezcadenas:
@jjgomezcadenas will review this, dead line August 3. 
@jacg tests are failing, I understand that the reason is the famous test_spe_to_adc

jjgomezcadenas:
@jacg the problem should be fixed. Could you add to our environment jupyter-lab

conda install -c conda-forge jupyterlab 

and @dataclasses (I think we need pip for that). 

For the system to work the user needs to remove conda from the path and add a line to source a conda script. In my case:

source /Users/jjgomezcadenas/miniconda/etc/profile.d/conda.sh

When the user tries for the first time, the system complains if the user has conda in the path but gives the wrong instructions. 

Also, when you open a new terminal and ask to work in python 3.6, the system tries to download conda then complains because the installation already exists, I think this should be fixed.

jjgomezcadenas:
@jagc the test fails for a reason that has nothing to do with the old test, already fixed. Seems to be related with hypothesis.

invisible_cities/sierpe/fee_test.py
```diff
@@ -53,6 +53,73 @@ def test_fee_params():
...
+
+
+    Parameters:
+    x : ndarray
+    The signal to be downsampled, as an N-dimensional array.
+    q : int
+    The downsampling factor.
+
+    n : int, optional
+    The order of the filter (1 less than the length for ‘fir’).
```
jacg: The signal to noise ratio here is awful. Rather than copypasting the whole manual entries for the two versions, I would just describe the change: `n` used to have these defaults, now it has these, and maybe leave a link to the manual.

invisible_cities/sierpe/fee_test.py
```diff
@@ -53,6 +53,73 @@ def test_fee_params():
...
+    spe_i = FE.spe_pulse(spe, t0 = 100 * units.ns, tmax = 200 * units.ns)
+    spe_v     = FE.signal_v_fee(fee, spe_i,ipmt)
+
+    scale = 25
+    spe_adc = signal.decimate(spe_v     * FE.v_to_adc(), scale, ftype='fir', zero_phase=False)
+    adc_to_pes     = np.sum(spe_adc    [3:7])
+
+    try:
+        assert 23 < adc_to_pes     < 23.1
+    except AssertionError:
```
jacg: `except Whatever: pass` is not helpful at all. It fails to convey the meaning and is full of holes.

While I totally agree with the intent and motivation of this test and its partner, I don't think that there is any sensible way of implementing it, because to demonstrate the change requires running tests in two different conda envs containing two different versions of scipy, in a single execution of the test suite, which just isn't going to happen.

I propose that we just bump the version of scipy and fix the concomitant bug in a single commit with a good commit message, and forget about the test that demonstrates. The in-step evolution of the test and version number will just have to do as documentation.

jacg:
> Could you add to our environment jupyter-lab

In general, feel free to add anything that you feel is necessary. The environment is specified in `manage.sh`. But why are we suddenly using jupyterlab?

> and @dataclasses (I think we need pip for that).

Dataclasses are a new feature in Python 3.7. There is a backport in pip. I'd be tempted to wait until 3.7 with the dataclasses. First get this PR sorted out. Then rebase the Python 3.7 PR on top, and see whether all our dependencies are already met. If they are, we can switch to 3.7 and nothing more needs to be done.

As for the problems with the conda installation, my bucolic location affords me a bandwidth of close to zero, so anything to do with installing conda itself or creating new conda environments currently takes me many hours per iteration. In short: I can't do this from where I am. Please ask someone who has a decent internet connection, to do this. I can offer moral support and technical assistance, but I just can't go through the bandwidth-heavy motions of testing potential solutions.

jacg:
Question: as the top comment states, we need some kind of label (substring in the name of the env) to distinguish environments which correspond to different versions of the code. Two obvious schemes come to mind:

+ The date the version specification was changed
+ A Git tag labelling the commit in which the spec was changed

Which should we use, or can you think of a better idea?

jacg:
@jjgomezcadenas 

> the test fails for a reason that has nothing to do with the old test, already fixed. Seems to be related with hypothesis.

That seems to have been some statistical fluctuation (two different ones, one on Linux and a different one on OS X!). I reran the build and it has passed (after another failure because of a network timeout on Travis).

invisible_cities/sierpe/fee_test.py
jjgomezcadenas: I can change pass by a print that states that the error is skipped on purpose. The test demonstrates the problem independently of the conda environment, because in the first conda environment it works and in the second fails but is caught. I do believe that documenting the problem is useful. Actually I included the test because I was sure you were going to request it.

jjgomezcadenas:
>In general, feel free to add anything that you feel is necessary. The environment is specified in manage.sh. But why are we suddenly using jupyterlab?

I am using jupyterlab because I like it better than the classical notebook. It offers some nice extra features. 

>Dataclasses are a new feature in Python 3.7. There is a backport in pip. I'd be tempted to wait until 3.7 with the dataclasses. First get this PR sorted out. Then rebase the Python 3.7 PR on top, and see whether all our dependencies are already met. If they are, we can switch to 3.7 and nothing more needs to be done.

I thought we would like to support 3.6 and 3.7 for a short while. If not, I am OK with this solution.

jacg:
> I thought we would like to support 3.6 and 3.7 for a short while. If not, I am OK with this solution.

If we are all as overworked as you say, then nobody will have time to add any new features that require `dataclass` until 3.6 has long been replaced completely by 3.7 :-)

We really need to clear the backlog, which means no *new* development unless it is absolutely strictly necessary in the short term. I doubt/hope that anything that falls into that category will not depend crucially on `dataclass`.

invisible_cities/sierpe/fee_test.py
jacg: I was going to reply, but now I see that you have removed it. OK, let's leave it like that. We have more urgent things to do.

jacg:
Re: jupyterlab: I would like to separate the dependencies we need to pass the tests, from those we like to have in our interactive environments. But no time to think about that right now.

jacg:
> For the system to work the user needs to remove conda from the path and add a line to source a conda script. In my case:
>
>    source /Users/jjgomezcadenas/miniconda/etc/profile.d/conda.sh

Yes, the system does use the modern conda installation style, and is likely not to work if you have an old style conda installation. Please just pretend that you don't have an old-style conda installation by


1. Renaming the directory containing your conda
1. Commenting out the old-style conda PATH-munging in your `.profile` or whatever is relevant on your system.

(I don't think we can afford to make it completely stable in the presence of old-style conda installations right now: if we verify that it can correctly install in the new style, we'll just have to ask people to wipe their old-style conda configurations.)

> When the user tries for the first time, the system complains if the user has conda in the path but gives the wrong instructions.



I'll need more details.

> Also, when you open a new terminal and ask to work in python 3.6, the system tries to download conda then complains because the installation already exists, I think this should be fixed.

I cannot reproduce this locally. In zsh it works perfectly. In bash it does spuriously download the conda install script, but then continues without any problems. Maybe it's something to do with the way things work on the Mac? If so, it's beyond my reach.

I'll have a look at the spurious download on bash.

I also noticed that `work_in_python_version` only runs the serial tests, so that will need fixing too.

manage.sh
```diff
@@ -89,6 +89,9 @@ dependencies:
 - matplotlib=2.2.2
 - networkx=2.1
 - notebook=5.5.0
+- jupyterlab=0.32.1=py36_0
+- jupyterlab_launcher=0.10.5=py36_0
+- python.app=2=py36_8
```
jacg: Tests fail because it doesn't like `python.app==2=py36_8`

jacg:
I have removed jupyterlab (at least temporarily), as it is not required to pass the tests, but the attempt to install it crashes the Travis build.

jacg:
Strictly speaking, this should take care of the database version too! But we can't do that properly until we have Git LFS (#210), so I propose we approve this without database management for now, and add it once LFS is available.

jjgomezcadenas (approved):
I have tried the system and works. It may create a bit of hassle when conflicting with the old version of conda, but it can be easily solved.

jjgomezcadenas:
@jacg I suggest that you merge this one, in order to speed things out.

jacg:
The environment definition needs to be pulled out into a file external to `manage.sh` with a big fat comment at the top stating that the tag has to be changed whenever anything in the configuration changes.

Ideally we would find a way of automating this, so that it is impossible for the human to forget to do it.

But I don't think that any of this is good enough reason to delay the merge, under the current circumstances. So I'll go ahead and do it.

PR #497, opened 2018-07-01
Add functions to filter based on Event ID

Aretno:
This PR adds two functions to filter data based on the event id, one for DSTs and another for pmaps. The latest one uses a more generic function (also added in the PR) which filters dictionaries based on their keys. There was a previous dictionary filter but only on the values, this function has been renamed to clearly state the difference between both cases.


jjgomezcadenas (approved):
Excellent! the PR is concise and the tests are satisfactory, the functionality is really necessary.

PR #500, opened 2018-07-05
Addition of simulation of low frequency noise

andLaing:
Simulation of low frequency noise effects
seen in the PMTs to the sensor simulation.

Involves changes to the database to allow
parameters from the detector to be updated
whern needed.

load_db.py:
New funciton added 'PMTLowFrequencyNoise'
which reads the mapping of PMTs to
front end boxes and the frequencies
and magnitudes to be used in the simulation.

low_frequency_noise.py:
New function that randomises contributions
and returns a function to access these
according to front end board mapping

### DIFF
--- a/invisible_cities/sierpe/low_frequency_noise_test.py
+++ b/invisible_cities/sierpe/low_frequency_noise_test.py
@@ -20,17 +20,18 @@ def test_buffer_and_limits():
 
     buffer_len        = 40 ## dummy 1 mus buffer
     buffer_sample_wid = 25e-9
-    frequencies       = np.arange(15000, 20000, 500) ## subset of frequences
+    freqency_bin      = 500
+    frequencies       = np.arange(15000, 20000, frequency_bin)
     
     freq_cont, lowf, highf = lfn.buffer_and_limits(buffer_len       ,
                                                    buffer_sample_wid,
                                                    frequencies      )
 
     assert np.all(freq_cont.keywords['time_bins'] == buffer_sample_wid * np.arange(buffer_len))
-    assert np.all(np.diff(lowf) == 500)
-    assert np.all(np.diff(highf) == 500)
-    assert lowf[0] == frequencies[0] - 250
-    assert highf[-1] == frequencies[-1] + 250
+    assert np.all(np.diff(lowf) == frequency_bin)
+    assert np.all(np.diff(highf) == frequency_bin)
+    assert lowf[0] == frequencies[0] - frequency_bin / 2
+    assert highf[-1] == frequencies[-1] + frequency_bin / 2
 
 
 def test_low_frequency_noise():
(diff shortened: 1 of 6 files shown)


andLaing:
This PR is a tidied and rebased version of the code I used for this: http://next.ific.uv.es/DocDB/0007/000706/001/lowFreqNoiseSim.pdf study.

It's quite physicsy so maybe @jjgomezcadenas and/or @paolafer should take a look.

jjgomezcadenas:
@andLaing can you rebase please?
@paolafer can you review?

paolafer:
I can try, but I'm leaving for a week of holiday on August 1. If I don't manage before, I'll complete it when I'm back.

paolafer (changes requested):
I've read the code, and I have a couple of questions, which I can ask easily to Andrew in person. Apart from that, the main thing here is that there is no test to ensure that the simulation of the low frequency noise is done correctly. I find difficult to understand if the code behaves properly, because I don't know exactly what to expect from it, and maybe a test could show that.

invisible_cities/sierpe/low_frequency_noise_test.py
```diff
@@ -0,0 +1,56 @@
...
+    freq_contribution = lfn.frequency_contribution(1, 1, 0, time_bins)
+
+    assert_allclose(freq_contribution, 2 * np.cos(time_bins))
+
+
+def test_buffer_and_limits():
+
+    buffer_len        = 40 ## dummy 1 mus buffer
+    buffer_sample_wid = 25e-9
+    frequencies       = np.arange(15000, 20000, 500) ## subset of frequences
```
paolafer: I think that the code would be clearer if you use a variable for the pace of `arange` (500), so that you can use the same variable below in the set of `assert`.

invisible_cities/sierpe/low_frequency_noise_test.py
```diff
@@ -0,0 +1,56 @@
...
+    buffer_len = 32000 ## 800 mus buffer
+
+    noise_func = lfn.low_frequency_noise(run_no, buffer_len)
+
+    ## Get an array with all pmts
+    pmt_noise = np.array(list(map(noise_func, np.arange(12))))
+
+    assert pmt_noise.shape[1] == buffer_len
+
+    ## Which should be the same and which not depends on
```
paolafer: This comment is totally unintelligible to me. Could you please add some punctuation, correct typos and use a name instead of "they"?

paolafer (changes requested):
The code looks fine to me, and the tests added verify that the results are consistent with what is given as an input. I've just made a couple of comments to improve the readability of the code.

paolafer (approved):
Great! I approve this PR. Please, rebase and see if Travis is merciful with you...

PR #503, opened 2018-07-17
Kdst io

jjgomezcadenas:
This PR improves over a previous PR in which basic protections were added to load_dst function, by adding two additional cases in which exceptions are intercepted and warning send without interrupting execution.


jmbenlloch (approved):
Improved version on #490 including new tests.

PR #504, opened 2018-07-24
Copy trigger information from RD to pmaps

jmbenlloch:
This PR copies the trigger information written by the decoder in the raw waveforms files. The structure is the following:

- `/Trigger/events` is an `EArray` with the information about the channel(s) that triggered the event. The first dimension is the event and the second one the PMTs, which are sorted in the same order as the waveforms, DataPMT from load_db, etc.

- `/Trigger/trigger` is a table with the trigger type given by the DAQ. Under normal operation there could be two codes: 1 (corresponding to trigger 1) and 9 (corresponding to trigger 2). There is one entry per event in this table.

With the changes in this PR, `Irene` will propagate the information to the `pmaps` files. Two new tests are added to check both of the previous data structures are properly propagated.

It is important to say that this functionality has been implemented in a backward-compatible way. `Irene` will still be able to process files that has no trigger information.


jjgomezcadenas (commented):
This PR extends the functionality of IC to incorporate trigger info to the reco chain. All the necessary methods are provided. most of the time extending existing structures. Two tests are provided to ensure consistency.

jjgomezcadenas (approved):

PR #505, opened 2018-07-24
Split output based on trigger type

jmbenlloch:
This PR builds on #504. **DO NOT merge** before #504, after merging #504 this will be rebased.

This PR adds to `Irene` the ability of writing two separate output files in case of double trigger operation. Since the changes included here are only relevant for Irene, most of the new code has been added to `irene.py` instead of `base_cities.py` to avoid making changes that can affect other unrelated cities.

Four new parameters are to `Irene`:

- `split_triggers`: Boolean, tells `Irene` whether to split the output or not in two files. **This parameter is mandatory**.
- `trg1_code`: Int, trigger code for trigger1 given by the DAQ. Now should be 1, no changes foreseen in the near future.
- `trg2_code`: Int, trigger code for trigger2 given by the DAQ. Now should be 9, no changes foreseen in the near future.
- `file_out2`: Path, output filename for events of trigger 2.

The three latter parameters are only needed if `split_triggers` is `True`.

With this PR `Irene` will be able to write all the events to one file or classify the events based on their trigger type and write each of them to the proper file. The usual `file_out` will be for trigger 1 and `file_out2` for trigger 2.

If for some reason there is an event with a trigger type different from `trg_code1` or `trg_code2` a warning will be printed.

A new test has been added to check all that is done properly.

Again, **DO NOT MERGE** before #504.


jjgomezcadenas (commented):
Everything appears rather straight forward, I will review again after the previous PR is merged

jmbenlloch:
I have already merged the previous one and rebased this on top the current master.

jjgomezcadenas (approved):
Continuation of previous PR, allows splitting the stream in two independent streams, according to trigger.

PR #506, opened 2018-07-24
Copy generator info in HDF5 MC files

msorel:
This is to pass through MC generator info through the various IC processing steps. This is the 'generators' dataset in the 'MC' group, and contains 'evt_number', 'atomic_number', 'mass_number' and 'region' information, that changes event-by-event. 

This additional information is necessary for the background MC production, mixing many background generators in the same file.

The test 'test_copy_mc_generator_info' has been created. It  is meant to cover three cases:

    1. mcfile_withgeneratorinfo: MCRD file where 'MC' group has 'generators' dataset with non-zero dimension, equal to number of events in file. Produced with GATE version v1_03_00 or later, and from nexus file where GATE event string store contains "/Generator/IonGun/atomic_number", "/Generator/IonGun/mass_number" and "/Generator/IonGun/region" information

    2. mcfile_withemptygeneratorinfo: MCRD file where 'MC' group has 'generators' dataset with zero dimension. Produced with GATE version v1_03_00 or later, and from nexus file where GATE event string store does NOT contain "/Generator/IonGun/atomic_number", "/Generator/IonGun/mass_number" and "/Generator/IonGun/region" information

    3. mcfile_withoutgeneratorinfo: MCRD file where 'MC' group has NO 'generators' dataset. Produced with GATE version prior to v1_03_00


jjgomezcadenas:
@paolafer can you review this PR?

paolafer:
Ok, I will.

invisible_cities/io/mcinfo_io.py
```diff
@@ -195,42 +195,47 @@ def read_mcinfo_evt (mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
...
-    h5extents    = h5f.root.MC.extents
-    h5hits       = h5f.root.MC.hits
-    h5particles  = h5f.root.MC.particles
-    h5generators = h5f.root.MC.generators
+    mc_info = tbl.get_mc_info(h5f)
+
+    h5extents = mc_info.extents
+    h5ehits = mc_info.hits
+    h5particles = mc_info.particles
+    h5generators = mc_info.generators
```
paolafer: It seems that the `h5generators` variable is not used in this method, right? If so, remove it.

paolafer (changes requested):
Nice workaround, Michel! It looks like it works fine.
I've just a minor comment regarding a variable in the code.

msorel:
thanks @paolafer , you are right, i have removed the unused variable

paolafer (approved):

PR #508, opened 2018-07-25
Thekla

jerenner:
Does not yet include hit-by-hit corrections, which were going to be included in a modified version of Penthesilea but will now be included as the first step of this city.


jacg:
@jerenner Could you please edit the top comment so that it contains the text that you would like to appear in the merge commit once this PR is approved. If you want to add some WIP caveats (which is what the whole top comment looks like to me at the moment) then please do so in a section at the top or the bottom, and leave a WIP label somewhere.

Like this, the newly-appointed mergers (#509) will quickly be able to use the meat of the top comment as the author's description in the merge commit. This will make their lives much easier.

@carmenromo & Ryan (still no GitHub handle?) When you merge a PR and you find WIP-like caveats and warnings in the top comment, please edit the top comment to remove them, as they should no longer be valid once the PR has been accepted.

jjgomezcadenas (commented):
A problem with this PR is that you introduce a huge quantity of code in commit one, making very difficult to follow what is going on. In general, small commits make life easier for the reviewer. 

As it is I don't think the PR can be reviewed, let alone approved. See my proposal below

jjgomezcadenas:
Please, @jerenner, @jahernando, @gonzaponte, @jacg, @jmbenlloch, @paolafe, @msorel, @Aretno   I need your attention here.

1. The main goal of this PR is to introduce the city Thekla, which takes over after Penthesilea.
2. Remember that Penthesilea reads PMAPS and produces raw (uncorrected hits) stored in a HDST (hit dst). To visualize the info in the HDST look at this snippet:

            row["event"] = self.event
            row["time" ] = self.time
            row["npeak"] = hit.npeak
            row["Xpeak"] = hit.Xpeak
            row["Ypeak"] = hit.Ypeak
            row["nsipm"] = hit.nsipm
            row["X"    ] = hit.X
            row["Y"    ] = hit.Y
            row["Xrms" ] = hit.Xrms
            row["Yrms" ] = hit.Yrms
            row["Z"    ] = hit.Z
            row["Q"    ] = hit.Q
            row["E"    ] = hit.E
  
This is filled for each hit (e.g, for each time slice). The energy (E) and SiPM charge (Q) of each hit is not corrected by lifetime or geometrical effects.

3. We have agreed that Penthesilea should not do anything else. The main reason is that, at the time of production of HDSTs, the corrections maps needed for geometry or lifetime for this particular run are not guaranteed to be available. 

4. Therefore we agree that a city running after Penthesilea (Thekla) will read HDSTs and carry forward the analysis. 

5. However, in the current attempt, I believe Thekla does too many things. I want to propose a very simplified version of Thekla that simply reads the HDST, produces corrected hits and writes back the HDST. 

6. Why not do anything else, like for instance calling Paolina? Because after the corrected HDST, we have at least two analysis. A classical analysis that requires Paolina tracks, and a DNN analysis that does not require Paolina tracks. In addition, RESET analysis may produce similar HDSTs. 

7. Therefore, we can write other cities after Thekla to produce Paolina tracks (we could read HDST corrected and write TDST dsts, where T stands for Track), and even summary ntuples, a la kdst. But this is for later.

8. For now I want to propose that we focus in a simple Thekla, that reads and writes HDSTs. I propose to modify the hit structure above to include:

           row["event"] = self.event
            row["time" ] = self.time
            row["npeak"] = hit.npeak
            row["Xpeak"] = hit.Xpeak
            row["Ypeak"] = hit.Ypeak
            row["nsipm"] = hit.nsipm
            row["X"    ] = hit.X
            row["Y"    ] = hit.Y
            row["Xrms" ] = hit.Xrms
            row["Yrms" ] = hit.Yrms
            row["Z"    ] = hit.Z
            row["Q"    ] = hit.Q
            row["E"    ] = hit.E
            row["Ql"    ] = Q corrected by lifetime
            row["El"    ] = E corrected by lifetime
            row["Qc"    ] = Q fully corrected
            row["Ec"    ] = E fully corrected.

The new fields are not filled by Penthesilea (or are filled with NN or zeros), and is the role of Thekla to fill them. In principle Thekla can overwrite the HDST produced by Penthesilea, is up to us to decide if we want to do this or not. 

9. Thekla needs a correction function, such that, for each hit h, it calls:
          def hit_correction(h : BHit) -->HitEnergyMap

where:

     @dataclass
    class HitEnergyMap
         E   :  EVector             
         Q   : EVector

and:

     @dataclass
    class EVector
        raw   :  float            # raw value    
        lt      :  float            # corrected by lifetime
       cor    : float             # fully corrected

Notice that I am proposing along the way the use of @dataclass, these are extremely nice objects which @jahernando and myself are using extensively for the new version of the Krypton analysis (also called the New Testament version, because it happened after @gonzaponte ascended to Valhalla or descended to Hades, we have not yet found out). @dataclass is available in python 3.7 and can be easily installled with pip in python 3.6 (@jmbenlloch  could you prepare a small PR to add @dataclass to our IC setup, with @jacg blessings? We really need them). 

10. The function should handle internally, the access to the geometry and lifetime maps, and therefore all those details can be hidden to Thekla and left to a separate PR (which will be done by @jahernando and yours truly) defining the correction function. 

Please, let me hear your thoughts. Unless I hear complains I will ask @jerenner to submit two simple PRs, one modifying Penthesilea to add the new fields to hit, and another one creating the minimum Thekla structure. 

Thanks.

msorel:
hi, my comments:

- most of you know my love for modularity at analysis stage. so, lets not mix hit corrections and tracking in same city, please. agreed with putting paoline elsewhere

- your scheme for hit corrections could work but is not very flexible/general and may need to be revisited in near future. say we introduce a 3rd correction that goes with event timestamp, for example. a more general scheme to apply corrections would be beneficial. if not too much overburden, maybe all penthesilea hits can be kept as they are, and a 2nd instance of the same table hit structure duplicated in the event, but with corrections applied (whatever they are) by thekla. then the analyzer goes back to the config file of thekla to know exactly what corrections have been applied and how

- if we do stick to this non-optimal (as in non-general) scheme, we should at east add Zc to this list and have thekla do the drift velocity correction on hit Z position, too

jjgomezcadenas:
@msorel I think that the scheme I propose also accommodates a time correction, we don't need a variable for correction, I think one variable for lifetime and the second accumulating all other corrections (e.g, geometry x time).  

The problem with the scheme that you propose is that it indeed creates over-burden. Duplicating the table for "generality" is not worth, I think, since at the end of the day, the corrections are numbered (we have three of them). On the other hand, if we keep things simple, we have a chance of implementing them in a finite time give the reduced man power resources.

jjgomezcadenas:
@msorel  the correction for drift velocity indeed makes sense.

msorel:
@jjgomezcadenas  i am fine with this solution. it is already a big step forward toward a centralised and trackable way of applying corrections within the collaboration (as opposed to the current "black market" of correction files being exchanged by email)

lets see what others have to say

msorel:
@jjgomezcadenas and yes, i thought about the drift velocity correction, too, as typically the same Kr calibration run used to compute (1) lifetime, (2) geometric, (3) time corrections, is also used to compute (4) drift velocity. so all these corrections can potentially be applied at the same time.

jacg:
I'll be brief, as *my* time is very limited.

Liquid cities were approved over 5 months ago, after being acknowledged as an infinitely superior and vastly more productive approach than solid cities, yet here we are being presented with a new solid city.

I'm sorry, I cannot justify the effort required to look at this.

The amounts of technical debt that are being taken on by developing new solid cities strike me as absolutely insane.

jjgomezcadenas:
- Liquid cities were approved, but never implemented, chiefly due to lack of man power. As I wrote in a previous comment, three major developers (@gonzaponte, @jacg  and @abotas) have not been available over the last few months. The rest of us can only work 48 hours a day.
-  Most of what needs to be done in Thekla is independent of the solid or liquid state of the city. We just write functions, and will be easy enough to liquify  the part that handles book keeping. Structurally, Thekla reuses everything developed for Penthesilea. 
- I have been reluctant to add a new solid city, but not doing so has resulted in an effective solid city to be developed a a parallel effort, *because progress in analysis cannot happen without correcting hits** . As @msorel pointed out, we are now in a black market economy. Technical debt is a lesser evil than that. 
- Moving the whole system once and for ever to liquid cities require no less than one week sustained effort of a team including @jahernando, @jacg, @jmbenlloch, @gonzaponte and @jjgomezcadenas. The presence of @jacg is essential, and thus understanding your availability is of uppermost importance.

- As much as I would like to pay back the huge TD that we are accumulating, we need your help for that.

jacg:
> Liquid cities were approved, but never implemented,

My recollection is that liquid cities were, implemented, tested, verified, and only approved after the verification. @gonzaponte reported the verification [here](https://github.com/nextic/IC/pull/435#issuecomment-366545303). And then they were never merged. I'm hazy on the reason for that.

> As I wrote in a previous comment, three major developers (@gonzaponte, @jacg and @abotas) have not been available over the last few months. The rest of us can only work 48 hours a day.

I am 98% confident that the 5 month delay in merging and switching to liquid cities has vastly increased the work that needs to be done by everyone. At the beginning of that period, the effort required of those you mention, in order to complete the switch, would have been minimal. Now ... it seems much bigger.

[Here](https://github.com/nextic/IC/pull/435#issuecomment-376840797) @gonzaponte reports on his effort to keep the unmerged liquid cities hot on the tail of the advancing development in the solid ones. Note that he on Mar 28 he was only 2 commits behind. I bet you that it's a lot more today. If we value @gonzaponte's time, perhaps we shouldn't waste it on making him keep an approved but unmerged branch up to date with runaway development on a dead end!

Stop moving his target !!!!!!!!!

> As much as I would like to pay back the huge TD that we are accumulating, we need your help for that.

@gonzaponte was the engine room that powered the liquification of the cities which existed back in February. His understanding of the physics which motivates the cities was, at that point, far more valuable than my understanding of the dataflow framework. His gaps in understanding dataflow are far smaller than my gaps in uderstanding the physics you are trying to do. @jahernando and @jjgomezcadenas were also present, and both have valuable understanding of dataflow issues.

I think that you are overstating my importance here: The three of you have understood dataflow, and from now on it's just more of the same, as far as dataflow is concerned. The expertise required to implement the new cities is more in the ideas that you are trying to express in the framework, rather than the framework itself.



On a slightly different track, the continued lack of Git LFS is another huge and inexplicable mystery to me. We all agree that it is important, every few months @jjgomezcadenas gives instructions that something must be done about it immediately, **but nothing ever gets done**. The excuse that the core developers aren't available just doesn't cut the mustard, as the person who is going to do this (Jose Vicente) is not on that list. Sure, precious core developers will have to help verify it, but there continues to be absolutely nothing to verify!


And to complete the boring structural trinity: why don't you all stop what you are doing for 2 hours and have a look at #496. I mean, really actually, have a look: check out the branch and try to use the facilities I have provided to switch environments/dependency versions. If nobody sees any obvious flaws in the idea, let's just approve and merge it. Even if it's not ideal, it's going to be much, much better than **not having any means to verify or compare old runs** because of our complete inability to match dependency versions with IC source code evolution.

In general, there is a problem in IC with PRs getting ignored for far too long. We've played the *we have to take on technical debt, because we don't have resources* card **far too often**. It's becoming the indiscriminant go-to excuse for every occasion. Remember: technical debt **eats up your resources**, so if we are short on resources we **have to REDUCE** technical debt, rather than taking on more. Really. I mean it. Read that sentence again and think about it.

My availability before (and after) the sprint is very limited. I have many things besides IC vying for my time, and the rising deluge of technical debt makes me very reluctant to spend my precious time on IC.

jjgomezcadenas:
>My recollection is that liquid cities were, implemented, tested, verified, and only approved after the verification. @gonzaponte reported the verification [here](https://github.com/nextic/IC/pull/435#issuecomment-366545303). And then they were never merged. I'm hazy on the reason for that.

Things were just accumulating, I guess, and @gonzaponte never got to actually do it. 

>I think that you are overstating my importance here: The three of you have understood dataflow, and from now on it's just more of the same, as far as dataflow is concerned. The expertise required to implement the new cities is more in the ideas that you are trying to express in the framework, rather than the framework itself.

I am totally, absolutely sure that I am not overstating your importance. But I agree, @jahernando, @gonzaponte and I can really try to rescue liquid cities... when @gonzaponte is available, which will be after he deposits his Ph.D. Thesis (early September). Do you have any availability in September? We could try to adapt to it (perhaps an expedition to Geneva).

>On a slightly different track, the continued lack of Git LFS is another huge and inexplicable mystery to me. We all agree that it is important, every few months @jjgomezcadenas gives instructions that something must be done about it immediately, but nothing ever gets done. The excuse that the core developers aren't available just doesn't cut the mustard, as the person who is going to do this (Jose Vicente) is not on that list. Sure, precious core developers will have to help verify it, but there continues to be absolutely nothing to verify!

@jmbenlloch can you tell us what is needed to implement LFS? Can we switch in on, say in September at the latest?

>And to complete the boring structural trinity: why don't you all stop what you are doing for 2 hours and have a look at #496. I mean, really actually, have a look: check out the branch and try to use the facilities I have provided to switch environments/dependency versions. If nobody sees any obvious flaws in the idea, let's just approve and merge it. Even if it's not ideal, it's going to be much, much better than not having any means to verify or compare old runs because of our complete inability to match dependency versions with IC source code evolution.

Yes, I will work on that starting today. I am dropping everything else until I fix this. You are darned right. 

>In general, there is a problem in IC with PRs getting ignored for far too long. We've played the we have to take on technical debt, because we don't have resources card far too often. It's becoming the indiscriminant go-to excuse for every occasion. Remember: technical debt eats up your resources, so if we are short on resources we have to REDUCE technical debt, rather than taking on more. Really. I mean it. Read that sentence again and think about it.

I am going to work on fixing this. From now on I will be sure to assign reviewers and dead lines to review. It will also help to have mergers.

>My availability before (and after) the sprint is very limited. I have many things besides IC vying for my time, and the rising deluge of technical debt makes me very reluctant to spend my precious time on IC.

A shame. Without your help things become much more difficult. We are few and have also many things in our platter.

jacg:
>> [Liquid cities were implemented but not merged]

> Things were just accumulating, I guess, and @gonzaponte never got to actually do it.

I think that this was a crucial point at which we took the turn down the road to debtors prison. Please try to learn from this mistake and avoid similar ones at all costs in the future.

> I am totally, absolutely sure that I am not overstating your importance.

Well, then we (you) are in deep shit. IC is an (expensive) hobby for me. Up to the end of the year I have one free week that I have reserved for your sprint. As you know, my professional situation is going to change in 2019, with the likely consequence that I will have zero time available for IC from then on. For the rest of this year I *must* dedicate whatever spare minutes I have outside of my fully-booked course schedule, to prepare for the seismic change in my life.

Please start getting used to the idea that my contribution, and hence my importance, is soon going to be zero.

> We are few and have also many things in our platter.

Which is why it is especially important not to take on technical debt.

paolafer:
Sorry to interrupt this discussion, but I feel the need of protesting formally and vehemently against the translation to English of the names of the cities :-). I'd rather keep the original one, I don't think that they are so difficult to pronounce in an English speaking environment. Let's be Tecla!

PR #511, opened 2018-07-27
Add protections for true MC reading

paolafer:
Inspired by Michel's last PR, I fixed a bug, which caused a crash if the MC true information was accessed, in files where no true hits were saved. A few more exceptions have been added, to give meaningful error messages in the following cases:
1. In a file with no sensor response information, the user tries to access sensor responses.
2. In a file with **only** sensor response information (no hits, neither particles stored), the user tries to access true info. This is the case, for instance, of the nexus output of look-up table generation.

### DIFF
--- a/invisible_cities/io/mcinfo_io_test.py
+++ b/invisible_cities/io/mcinfo_io_test.py
@@ -276,13 +276,14 @@ def test_copy_mc_generator_info(output_tmpdir, ICDATADIR, in_filename, out_filen
 
 
 def test_read_file_with_no_hits(nohits_sim_file):
+    """
+    This test ensures that, even if there are no true hits in a file,
+    loading the true information doesn't make the program crash.
+    """
 
     filein = nohits_sim_file
+    load_mcparticles(filein)
 
-    try:
-        load_mcparticles(filein)
-    except IndexError:
-        raise
 
 def test_access_to_particles_in_sns_response_only_file_raises_IndexError(sns_only_sim_file):
 
--- a/invisible_cities/io/mcinfo_io_test.py
+++ b/invisible_cities/io/mcinfo_io_test.py
@@ -11,7 +11,8 @@ from .  mcinfo_io import load_mcsensor_response
 from .  mcinfo_io import mc_info_writer
 from .  mcinfo_io import read_mcinfo_evt
 
-from .. core import system_of_units as units
+from .. core            import system_of_units as units
+from ..core.exceptions  import NoParticleInfoInFile
 
 from .. reco.tbl_functions import get_mc_info
 
@@ -287,5 +288,5 @@ def test_access_to_particles_in_sns_response_only_file_raises_IndexError(sns_onl
 
     filein = sns_only_sim_file
 
-    with raises(IndexError):
+    with raises(NoParticleInfoInFile):
         load_mcparticles(filein)
--- a/invisible_cities/io/mcinfo_io_test.py
+++ b/invisible_cities/io/mcinfo_io_test.py
@@ -15,6 +15,7 @@ from .. core import system_of_units as units
 
 from .. reco.tbl_functions import get_mc_info
 
+from pytest import raises
 from pytest import mark
 parametrize = mark.parametrize
 
@@ -271,3 +272,20 @@ def test_copy_mc_generator_info(output_tmpdir, ICDATADIR, in_filename, out_filen
             events_out = h5out.root.MC.generators[:]['evt_number']
 
             np.testing.assert_array_equal(events_in, events_out)
+
+
+def test_read_file_with_no_hits(nohits_sim_file):
+
+    filein = nohits_sim_file
+
+    try:
+        load_mcparticles(filein)
+    except IndexError:
+        raise
+
+def test_access_to_particles_in_sns_response_only_file_raises_IndexError(sns_only_sim_file):
+
+    filein = sns_only_sim_file
+
+    with raises(IndexError):
+        load_mcparticles(filein)
--- a/invisible_cities/conftest.py
+++ b/invisible_cities/conftest.py
@@ -74,6 +74,16 @@ def electron_MCRD_file(request, ICDATADIR):
 def krypton_MCRD_file(request, ICDATADIR):
     return os.path.join(ICDATADIR, request.param)
 
+@pytest.fixture(scope  = 'session',
+                params = ['mcfile_nohits.sim.h5'])
+def nohits_sim_file(request, ICDATADIR):
+    return os.path.join(ICDATADIR, request.param)
+
+@pytest.fixture(scope  = 'session',
+                params = ['mcfile_sns_only.sim.h5'])
+def sns_only_sim_file(request, ICDATADIR):
+    return os.path.join(ICDATADIR, request.param)
+
 
 @pytest.fixture(scope='session')
 def mc_all_hits_data(krypton_MCRD_file):
--- a/invisible_cities/core/exceptions.py
+++ b/invisible_cities/core/exceptions.py
@@ -67,3 +67,6 @@ class UnknownParameter(ICException):
 
 class SensorBinningNotFound(ICException):
     pass
+
+class NoParticleInfoInFile(ICException):
+    pass
--- a/invisible_cities/io/mcinfo_io.py
+++ b/invisible_cities/io/mcinfo_io.py
@@ -212,7 +212,8 @@ def read_mcinfo_evt (mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
             ipart_end = this_row['last_particle']
 
             while ihit <= ihit_end:
-                hit_rows.append(h5hits[ihit])
+                if len(h5hits) != 0:
+                    hit_rows.append(h5hits[ihit])
                 ihit += 1
 
             while ipart <= ipart_end:
@@ -313,6 +314,12 @@ def read_mcsns_response(h5f, event_range=(0, 1e9)) ->Mapping[int, Mapping[int, W
 
 
     h5extents   = h5f.root.MC.extents
+
+    try:
+        h5f.root.MC.waveforms[0]
+    except IndexError:
+        print('Error: this file has no sensor response information.')
+
     h5waveforms = h5f.root.MC.waveforms
 
     last_line_of_event = 'last_sns_data'
--- a/invisible_cities/io/mcinfo_io.py
+++ b/invisible_cities/io/mcinfo_io.py
@@ -233,8 +233,6 @@ def read_mcinfo(h5f, event_range=(0, int(1e9))) ->Mapping[int, Mapping[int, MCPa
     mc_info = tbl.get_mc_info(h5f)
 
     h5extents = mc_info.extents
-    h5ehits = mc_info.hits
-    h5particles = mc_info.particles
 
     events_in_file = len(h5extents)
 
--- a/invisible_cities/io/mcinfo_io.py
+++ b/invisible_cities/io/mcinfo_io.py
@@ -211,10 +211,10 @@ def read_mcinfo_evt (mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
             ihit_end  = this_row['last_hit']
             ipart_end = this_row['last_particle']
 
-            while ihit <= ihit_end:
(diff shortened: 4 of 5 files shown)


jjgomezcadenas:
Good job, a few comments follow:

In 

      try:
           h5in.root.MC.particles[0]
       except IndexError:
           print('Trying to access particle information: this file has sensor response only')
           raise
    
you could consider using a warning.


     import warnings
      try:
           h5in.root.MC.particles[0]
       except IndexError:
          warnings.warn('Trying to access particle information: this file has sensor response only', RunTimeWarning)

This will not cause the program to crash immediately, which may be useful, unless you really want to crash the program and exit right away

jjgomezcadenas (commented):
def test_read_file_with_no_hits(nohits_sim_file):
 +
 +    filein = nohits_sim_file
 +
 +    try:
 +        load_mcparticles(filein)
 +    except IndexError:
 +        raise
 +

Don't you need a with raises(IndexError) above?

jjgomezcadenas:
@jmbenlloch can you add @msorel to the list of approved reviewers?

@msorel please, can you review this PR? I have already reviewed it and is straight forward and very much in the line of your recent work, it will be useful for you to review as an exercize and we need more developers involved in the code (see previous threads)

@carmenromo once this PR is approved you can try merge, (with @jmbenlloch)

jmbenlloch:
> @jmbenlloch can you add @msorel to the list of approved reviewers?

I have already added him as a member of nextic/IC, he should receive an email with the invitation. As soon as he accepts it he should appear in the list of possible reviewers.

jacg:
try:
        load_mcparticles(filein)
    except IndexError:
        raise

is semantically equivalent to 

    load_mcparticles(filein)

jacg:
This PR adds a couple of binary files and ... Git LFS is still missing!

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -271,3 +272,20 @@ def test_copy_mc_generator_info(output_tmpdir, ICDATADIR, in_filename, out_filen
...
+    try:
+        load_mcparticles(filein)
+    except IndexError:
+        raise
+
+def test_access_to_particles_in_sns_response_only_file_raises_IndexError(sns_only_sim_file):
+
+    filein = sns_only_sim_file
+
+    with raises(IndexError):
```
jacg: `IndexError` seems an odd thing to raise here. Don't we want some kind of meaningful IC exception here?

paolafer: `IndexError` is the error raised in `tbl_functions.py` when I try to access the first element of a zero-entries array:

`h5in.root.MC.particles[0]`

which is the test I do to see if the `particle` table is empty. 
Of course, we could add a new exception such as `NoParticleInfoInFile`, I just thought that a readable message could be of use. Can we add readable messages in our IC exceptions?

paolafer:
Thanks for the suggestions!

@jjgomezcadenas 

> This will not cause the program to crash immediately, which may be useful, unless you really want to crash the program and exit right away

Actually, I wanted the program to crash, because it makes no sense to access true info about particles in files where no such info is contained.

@jacg 

> ` try:
    load_mcparticles(filein)
except IndexError:
    raise`
 is semantically equivalent to
`load_mcparticles(filein)`

Yes, you're right. Is it ok, for a test, to just write `load_mcparticles(filein)`? I thought that this syntax would be clearer.

msorel:
@jjgomezcadenas @paolafer 
sure. i will have a detailed look early next week and get back to you in this thread. 

i think i can review well logic and functionality of PR, less so IC coding standards as i am not very familiar with them, yet

jjgomezcadenas:
@msorel has agreed to review, dead line August 3

invisible_cities/io/mcinfo_io_test.py
jacg: You can add any Python objects as messages inside any Python exception. So use `NoParticleInfoFile` and put whatever string you like as a message inside it.

paolafer: Great, I've implemented it.

jacg:
>> This will not cause the program to crash immediately, which may be  useful, unless you really want to crash the program and exit right away

> Actually, I wanted the program to crash, because it makes no sense to  access true info about particles in files where no such info is contained.

Then raise an exception: Exactly the type of exception you want (make a new type, if a suitable one doesn't exist) containing exactly the embedded information you want. (As we have already said elsewhere.) But just commenting on your original code: you handled the original exception, printed a message to stdout and then re-raised the original exception. It would have been much better to embed that message in the 

In summary:

+ If it makes no sense to continue, then raise an exception of exactly the type you want, containing exactly the detailed message you want. But only add the detailed info if it somehow enhances the pure exception type. Avoid nonsense like this: `raise NoseTooBigError('The nose is too big')`. The type already says it all, no need to repeat yourself. Contrast this with `raise BodyPartMissingError('nose')`. In the latter case the type only tells you that *some* body part is missing, so the extra embedded information (that it is the nose rather than an ear which is missing) is useful.
+ If it makes sense to continue, then don't print to stdout, but use the warnings system. That way user has the option of upgrading it to an exception.


>> `try: load_mcparticles(filein) except IndexError: raise` is semantically equivalent to load_mcparticles(filein)

> Yes, you're right. Is it ok, for a test, to just write load_mcparticles(filein)? I thought that this syntax would be clearer.

What is it that you are trying to say with this code? Sorry, I've lost track of the context where it appears, and can't find it any more.

paolafer:
I've added an exception and a detailed message that adds a piece of information to the error (it gives the user a hint about what this file could contain, given that no particle information is there).
The other piece of code you are referring to

`    try:
        load_mcparticles(filein)
    except IndexError:
        raise
`

is a test that wants to ensure that, in the case where no hits are found in the file, the program doesn't crash. This is a sensible situation (we can have particles without hits, but we cannot have hits without particles, because a hit is always generated by a particle).
If I understand your suggestion, I could add another specific exception (NoHitsFile) and raise that. My only concern is that in this case the exception should only be used in this test, which wants to ensure that no exception is raised...

jacg:
As much as I believe that code should be self-documenting and that you should therefore not write comments (by writing code which makes comments unnecessary), I think this is a situation where writing a comment is the best solution. Just write the line that shouldn't crash, without any `try` and comment that it should not crash.

msorel (changes requested):
hi @paolafer in IC/invisible_cities/reco/tbl_functions.py, get_mc_info function, I would specify column names for the 'hits' dataset even if it has zero rows:

```
diff --git a/invisible_cities/reco/tbl_functions.py b/invisible_cities/reco/tbl_functions.py
index c85c29d..ef22c3b 100644
--- a/invisible_cities/reco/tbl_functions.py
+++ b/invisible_cities/reco/tbl_functions.py
@@ -179,6 +179,7 @@ def get_mc_info(h5in):
``` 

```python
     if len(h5in.root.MC.hits) == 0:
         hits = np.zeros((0,), dtype=('3<f4, <f4, <f4, S20, <i4, <i4'))
+        hits.dtype.names = ('hit_position', 'hit_time', 'hit_energy', 'label', 'particle_indx', 'hit_indx')
 
     if 'generators' in h5in.root.MC:
         generator_table = h5in.root.MC.generators
```

msorel:
In IC/invisible_cities/io/mcinfo_io.py, read_mcinfo_evt function, why not putting the same protection you put for 'hits', also for 'particles'? And perhaps swap if and while statements. I mean something like this:

```python
            if len(h5hits) != 0:
                while ihit <= ihit_end:
                    hit_rows.append(h5hits[ihit])
                    ihit += 1

            if len(h5particles) != 0:
                while ipart <= ipart_end:
                    particle_rows.append(h5particles[ipart])
                    ipart += 1
```

Also, in read_mcinfo, I guess you can get rid of two unused variables:
```
@@ -233,8 +234,6 @@ def read_mcinfo(h5f, event_range=(0, int(1e9))) ->Mapping[int, Mapping[int, MCPa
     mc_info = tbl.get_mc_info(h5f)
 
     h5extents = mc_info.extents
-    h5ehits = mc_info.hits
-    h5particles = mc_info.particles
 
     events_in_file = len(h5extents)
```

msorel:
@paolafer I have nothing else to add on this particular PR, it looks good to me except the small suggestions I make above. Get back to me and we approve the merge, provided @jacg and @jjgomezcadenas are also happy with it

However, unrelated to this PR but possibly related to this same part of the code: as you can see from your emails, Corey's crashes in our current background MC production point to 
len(MC/extents) < len(Run/events)
for certain files. This should never happen, I believe: both should have as many rows as number of events

This could be due:
-  either to MC/extents not being copied over correctly from time to time from input file to output file. Meaning the bug would be in IC, likely in this part of the code.
- or to MC/extents being wrong (too short) to begin with, missing rows from the IC processing start. Meaning the bug would be in the detsim stage and GATE's HDF5Writer

Let's try to find out...

paolafer:
Hi Michel, I've implemented your suggestions. As for

> In IC/invisible_cities/io/mcinfo_io.py, read_mcinfo_evt function, why not putting the same protection you put for 'hits', also for 'particles'?

the rationale behind this asymmetric behaviour is that we can have simulation with particles without hits, but not with hits without particles (since a hit, by construction, is generated by the energy deposition of a particle). 
When no particles, nor hits are present, one shouldn't use the method load_mcparticles, and this is the reason why I added the error message in `tbl_functions.py`.

Regarding 

> And perhaps swap if and while statements.

I understand that the behaviour is the same,  the difference being that you don't increment the `ihit` variable in the no-hit case, thus saving time.

jacg:
> provided @jacg and @jjgomezcadenas are also happy with it

I need to drive a round trip to a distant airport today, starting in about 5 minutes. Please assume that I'm out of the picture today.

msorel (approved):
I agree with your latest changes @paolafer This PR improves robustness of code and can be merged. 

We may still have other problems in this area of the code, see my comments at the end of this PR comments. If that's the case, we open a new PR

jjgomezcadenas:
Please rebase

paolafer:
I cannot do the rebase, I'm afraid, because my IC code is not working.

msorel:
@paolafer @jjgomezcadenas paola is leaving for vacations and her IC is not working, is rebasing something brais and i can do for paola?

jacg:
> is rebasing something brais and i can do for paola?

I have rebased this on my machine. There were no conflicts, so it's trivial to do. Would you like me to push it, or shall I leave it so that you can have a go?

msorel:
thanks @jacg i am with brais and we will give it a try now

msorel:
no permission:

```
Pushing to https://github.com/paolafer/IC.git
remote: Permission to paolafer/IC.git denied to msorel.
fatal: unable to access 'https://github.com/paolafer/IC.git/': The requested URL returned error: 403
```

same thing trying as bpalmeiro. can you do it, @jacg ? thanks

jmbenlloch:
@msorel, I have given you write permissions, I think it should work now. Brais should have also received an email with an invitation from github, when he accepts it he will have write permissions.

msorel:
@jmbenlloch thanks, that worked, we just pushed with @bpalmeiro from my account
@carmenromo you can now merge

carmenromo:
ok!

jacg:
> @carmenromo you can now merge

If you could also move the PR's card to the *Merge* column on the [project board](https://github.com/nextic/IC/projects/1) in these situations, it will help to prevent these things from getting missed or forgotten.

carmenromo:
I don't know why I have this error while merging:

128 git … commit --amend --only --author\=msorel
fatal: --author 'msorel' is not 'Name <email>' and matches no existing author

jacg:
Ah, yes, I was going to warn you about that before, but there were already too many things to say.

I am not sure of the exact reasons, but I have observed that for some people, specifying their GitHub userid as the `--author` does not work. So far, specifying their full name as it appears in commits made by them, has worked for me. I suggest you try that. If that doesn't work, we'll have a closer look.

jacg:
I can confirm that `=AMichel Sorel` in Magit, did the right thing for me locally.

carmenromo:
Yes, it worked! Thanks!

PR #512, opened 2018-07-27
New wrapper to fix parameters for fitting

andLaing:
Adds a wrapper so that functions can have
parameters fixed to known values and
still work as expected with `scipy.curve_fit`.


andLaing:
There is currently no functionality in scipy to fix parameters in functions for fits strangely.

Since this is something useful for calibration and fundamental for higher level analysis we will need this functionality at some point.

There are packages which have this functionality as well as others like LMFIT and symfit (probably more but I found those ones) but would require a more fundamental change to how we do the fits (maybe something for the future). The wrapper is general (partial only works in this case to fix the first parameters not any) but is a bit clunky maybe...

jjgomezcadenas:
@gonzaponte I know that you are busy but that follows rignt into your turf. Please review this.

jjgomezcadenas:
@gonzaponte to review, dead line August 3.

gonzaponte:
Seriously, I have **absolutely no time** for this. I would like to, but I can't. If it is urgent, approve and merge and I will deal with it later. Sorry.

invisible_cities/core/fit_functions_test.py
```diff
@@ -286,6 +286,18 @@ def test_fit_with_errors(reduced):
     assert_allclose(f.values, pars)
 
 
+def test_fixed_parameters():
+    pars = np.array([3.0,  2.0, 0.50])
```
jacg, line 290: Is it necessary to wrap this in `np.array`? Won't any sequence do?

jacg:
> partial only works in this case to fix the first parameters not any

I admit I'm glancing over this quickly, so I may have missed something. You seem to be fixing parameters by name (keyword arguments), rather than by arbitrary numerical position. `partial` is perectly able to do this.

jacg:
I have just replaced

    fixed_mu = fitf.fixed_parameters(fitf.gauss, mu = pars[1])

with

    from functools import partial
    fixed_mu = partial(fitf.gauss, mu = pars[1])

(which amounts to replacing your `fixed_parameters` with `partial`) in your test, and it passes. This suggests that `partial` is up to the task, and you don't need to write your wrapper.

jacg:
... or maybe not ... now I'm getting errors. I need to dig a bit more.

invisible_cities/core/fit_functions_test.py
jacg: I see: `seeds` which you create by slicing `pars` needs to be an array. But `pars` itself does not.

jacg:
OK, I think I see the added value over `partial`: providing a keyword arg removes that arg from the set of positional ones?

invisible_cities/core/fit_functions_test.py
andLaing: I just copied the setups from the other tests in fit_functions_tests.py.

No preference nor even thought went into the use of np.array vs. a list or whatever

invisible_cities/core/fit_functions.py
```diff
@@ -4,16 +4,56 @@
...
 
 
+def fixed_parameters(fn, **kwargs):
+    """
+    Wrapper to fix parameters of a function
+    which can then be used in curve_fit.
+    """
+    fn_sig      = insp.signature(fn)
+    fn_args     = np.array(list(fn_sig.parameters))[1:]
+    par_gen     = (kwargs[k] if k in kwargs.keys() else np.nan for k in fn_args)
```
jacg, line 27: Strictly speaking arguments are values (data, objects ... the stuff you send in to a function when you call it) while parameters are names (the names which appear in the parameter list in the definiton of the function), so your choice of the variable `fn_args` for a bunch of names, is rather confusing.

andLaing:
> OK, I think I see the added value over partial: providing a keyword arg removes that arg from the set of positional ones?

yes, `partial` obviously can set things with keyword arguments very well but when I tried to use the returned function with the fitter for a PMT fit I got errors that it was trying to reset values etc. Believe me, I did not want to write the wrapper I just couldn't find another general way to do it.

invisible_cities/core/fit_functions_test.py
jacg: Aaah, copy 'n' paste programming ... which is exactly how code accumulates lots of pointessly misleading noise :-)

Actually, there is a serious point: *careful* thought is required when programming.

invisible_cities/core/fit_functions.py
andLaing: As a coding pleb I am not sensitive to these things. I'll change and update.

invisible_cities/core/fit_functions_test.py
andLaing: I have to admit that this seems like a minor point, I had thought myself out doing the coding itself and as that which is written by @gonzaponte in tests tends to be the accepted correct way to do things I chose the copy 'n' paste from my betters over waste more time with boring things

invisible_cities/core/fit_functions.py
jacg: We live and learn. Don't bother changing anything. I'll figure it all out eventually, and then I'll either tell you why you don't need to write it, or clean it up myself.

invisible_cities/core/fit_functions_test.py
jacg: Most *single* instances of this sort of thing are indeed minor. The problem is when thoughtless copy 'n' paste programming is endemic and becomes standard practice for everything that everybody writes ... as is the case in most HEP codes. Down the years, hundreds of individually insignificant thoughtless copy 'n' pastes accumulate to make roughly 80% of the code pure, distracting, noise. Then you end up with, incomprehensible, crappy, impenetrable, buggy, pointlessly slow code.

You can't do physics today without programming. If you think that this crucial part of your professional activity is 'boring' and not worth doing well, then you're setting yourself up for pain followed by failure to achieve your goals.

I trust that you think that it is worthwhile to build detectors carefully and thoughtfully, as I trust that you think that it is worthwhile doing mathematical analyses carefully and thoughtfully. Papers you write deserve to be carefully written. The code which enables to get your work done is the same.

jacg: > As a coding pleb I am not sensitive to these things.

Being a coding pleb is a natural starting point for all humans, and it's not an insurmountable problem as long as there's a will to, slowly, gradually, decrase one's plebishness.

Being a coding pleb and proud of it, is a serious problem for any project in which one takes part which requires notrivial amounts of programming.

andLaing: Much as it's probably not worthwhile to get into this discussion...

At no point did I say nor imply that the coding was either boring nor pointless. One small line of code which I needed to test if the actual function which I had been working on worked seemed pointless to waste time on especially since it was taken from another test which, I assumed, had been through review is the offending thing here.

I'll try not to do it again, but, in reply to your other comment:

Being strict with reviews is great but being passive aggressive and expecting perfection from the offset is not

jacg:
>  Believe me, I did not want to write the wrapper I just couldn't find another general way to do it.

Well, there's always `lambda` which can do the job, doesn't require you to implement anything, and whose meaning should be clear to *everyone*. But for functions with many parameters, I can imagine that the wrapper can end up looking much better.

andLaing:
Yes, I looked at lambda and thought it was the best way forward at first but since I wanted something as general as possible (not just a few examples for calibration) since this is something that should be useful for all problems involving fitting and for the determination of systematic errors for example. It's not as easy to write multiple lambdas as it is to be able to do wrapper(function, set these ones) when ever the thought ocurres.

I'm sorry about the constant spams but I worked quite hard on this and want to learn from the mess that it ended up being because I got to the point where I couldn't find anything online to help.

I considered everything I found in examples but nothing could do everything I wanted. I'm sure someone who is more experienced in python can improve it though but if it seems better to add one of the packages which have this functionality then we should do that, it just seemed like more of a fundamental change since it would require changes in the fit function to wrap around LMFIT or other. Not that they seemed particularly straightforward to use either.

The use of inspect.signature rather than inspect.getargspec was forced by some oddity in the returned functions of spe_response.dark_convolution etc which made them look like they had no arguments. I forgot to mention that in my first comment

invisible_cities/core/fit_functions_test.py
jacg: > At no point did I say nor imply that the coding was either boring nor pointless.

My profound apologies. You did not imply it, I erroneously inferred it from your use of the words 'waste more time with boring things'.

> Being strict with reviews is great but being passive aggressive and expecting perfection from the offset is not

I don't have any problems with your first version of the code, and I certainly don't expect perfection. But I do want to *understand* how the code works, which involves being curious about the things that appear in it. I am asking questions because you, as author, might know the answer, and thus save me precious time.

Please understand that code review is not an attack on your person, but a collaborative effort to improve not only the code but also both your and my skill at implementing stuff.

When I make comments about how things can be done better, it is not because I expected you to do it that way in the first place, but because I think you (and people who use and read your code) might benefit from it in the future. So please try to take it as an act of collaboration rather than aggression. I like to think that in technical circles we can cut the crap and make technical comments without pussyfooting around in case someone interprets technical comments as personal attacks.

jacg:
> I looked at lambda and thought it was the best way forward at first but [...]

Just to be clear: I think that writing this was probably a good idea. `lambda` can do the job, but having this wrapper is likely to make life easier for the user, and therefore probably worthwhile.

jacg:
> The use of `inspect.signature` rather than `inspect.getargspec` was forced by some oddity in the returned functions of `spe_response.dark_convolution` etc which made them look like they had no arguments.

If there is a specific reason which forces you to use a less obvious implementation, then it's a good idea to include a test that explicitly highlights the failure of the more obvious solution. Without such a test, someone who is ignorant of the problem could come along and simplify the solution without realizing that a failure to cater for the issue is being introduced in the code.

One could argue that other tests which use the feature indirectly, will catch it, but there are at least two problems with this:

+ Becasue such tests are testing something else, the feature is used indirectly and understanding what the problem is will be more costly.
+ Anyone working specifically on the feature (such as me, right now) is likely to be running just the direct tests for the feature, in a tight loop. The danger is that everything seems to be working during this focused process, and after considerable effort the problem only becomes apparent at the end when the whole test suite is run.

This is not a criticism, just a thought for the future.

invisible_cities/core/fit_functions_test.py
andLaing: I don't interpret anything as a personal attack. I would, however, prefer that a comment about one small section in a test not be interpreted as my disdain for the project. I actually quite enjoy coding even if I'm more interested in the solving the problem than the specifics. Maybe I'm too sensitive, too many years of everything I say being passed through amor/odio filters.

Please feel free to give me homework, I learn better by doing, I can make changes to try and make the code clearer etc.

jacg: > Please feel free to give me homework,

I have to do my homework on this, before I can give you any sensible homework.

andLaing:
Should I rebase this or are you taking care of everything @jacg ?

PR #514, opened 2018-07-28
Fix Irene bug caused by Numpy upgrade

jacg:
We were using Numpy 1.14.5. 1.15.0 is available, but `test_irene_split_trigger` breaks if we use the new version of Numpy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ matrix:
   fast_finish: true
   allow_failures:
   - os: osx
+  - env: IC_PYTHON_VERSION=3.7
 
 env:
   - IC_PYTHON_VERSION=3.7
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ matrix:
   - os: osx
 
 env:
+  - IC_PYTHON_VERSION=3.7
   - IC_PYTHON_VERSION=3.6
 
 os:
--- a/bash_manage.sh
+++ b/bash_manage.sh
@@ -27,7 +27,7 @@ manage.sh()
     if [[ ${prev2} == "manage.sh" ]] ; then
             case "${prev}" in
                         install_and_check|install|work_in_python_version|make_environment)
-                        local versions="3.6"
+                        local versions="3.6 3.7"
                     COMPREPLY=( $(compgen -W "${versions}" -- ${cur}) )
                     return 0
                     ;;
@@ -73,7 +73,7 @@ source_manage.sh()
     if [[ ${prev2} == "manage.sh" ]] ; then
             case "${prev}" in
                         install_and_check|install|work_in_python_version|make_environment)
-                        local versions="3.6"
+                        local versions="3.6 3.7"
                     COMPREPLY=( $(compgen -W "${versions}" -- ${cur}) )
                     return 0
                     ;;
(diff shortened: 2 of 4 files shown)


jacg:
@jmbenlloch I've assigned this to you, as you seem to be the author of the test which is failing.

jmbenlloch:
The problem here is that there are new warnings:

```
(Pdb) record[0].message
FutureWarning('Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.',)
(Pdb) record[1].message
FutureWarning('Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.',)
(Pdb) record[2].message
UserWarning('Event 2 has an unknown trigger type (5)',)
(Pdb) record[3].message
FutureWarning('Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.',)
```

The test checks irene raises a warning for unkown trigger (`record[2]`) (and only one). The rest of then come from:

```
  /home/jmbenlloch/miniconda/envs/IC-3.6-2018-07-28/lib/python3.6/site-packages/scipy/signal/signaltools.py:1344: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
    out = out_full[ind]
```

I don't know if this is going to be fixed soon by scipy...

jacg:
I have modified the test, so that, rather than requiring exactly one warning with exactly that message, the test now requires at least one warning, exactly one of which has the required message. I think that is logically more correct: it's not this test's business whether any other warnings appear or not.

@jmbenlloch Please review.

jacg:
Unfortunately the tests fail elsewhere about 50% of the time, probably because of spurious hypothesis problems, but that's a job for a separate PR.

jmbenlloch:
> I have modified the test, so that, rather than requiring exactly one warning with exactly that message, the test now requires at least one warning, exactly one of which has the required message. I think that is logically more correct: it's not this test's business whether any other warnings appear or not.

I agree, that seems to be a good solution!

jacg:
In which case, please approve the PR, if there's nothing else to consider. Of course it's a bit unfortunate that this one is on top of #493 which is still awaiting approval ...

jmbenlloch (approved):
You're right. This is approved, the bug has been successfully fixed.

jacg:
@carmenromo I'm moving this to the *Merge* column, but **don't merge it until AFTER #493**

jacg:
@carmenromo #493 has been merged, so please go ahead and merge this as soon as Travis gives its approval.

carmenromo:
ok!

carmenromo:
Done

jacg:
@carmenromo Travis has given the go-ahead. Please merge ... ah I see you did it before I finished typing. Well done!

A minor comment about the merge commit text: The two parts of the PR description that appear in brackets aren't really parts of the description: they were warnings about not merging too soon. Now that it is ready to merge, these warnings don't apply, so they should not have been added to the PR description in the merge commit message, and they need to be edited out of the PR description on GitHub.

There's nothing that can be done about the merge commit now that it has been pushed (at least nothing that is worth the trouble it causes). But please do edit the PR description on GitHub, to remove the obsolete warnings.

carmenromo:
Oh sorry... I edited it on GitHub

PR #515, opened 2018-07-29
Fix dependency versioning

jacg:
This addresses a bug in #496: it (sometimes?) crashed when needing to switch to an environment that had not been created yet.

The fix consits of ensuring that, if the environment does not yet exist, it is created before the switch is attempted.

(I have no idea how to write automated tests for the stuff in `manage.sh`, especially the ones related to switching environment ... which makes working on this very painful and error-prone.)

### DIFF
--- a/manage.sh
+++ b/manage.sh
@@ -105,8 +105,16 @@ EOF
     conda env create -f ${YML_FILENAME}
 }
 
+function make_environment_if_missing {
+    if ! conda env list | grep ${CONDA_ENV_NAME}
+    then
+        make_environment
+    fi
+}
+
 function switch_to_conda_env {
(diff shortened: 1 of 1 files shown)


jacg:
@jjgomezcadenas Could we have a quick approval here, please? I think that it's obvious that it makes dependency versioning more robust. What is not clear is how many more bugs there are still lurking there. Without being able to write automated tests for this, getting it right is a nightmare.

jjgomezcadenas (approved):
The code is straight forward and can be approved right away. I share the worry about automated tests here, but let's move this onwards.

jjgomezcadenas:
@jacg you may want to merge this and the 3.7 PRs, while @carmenromo learns the trade.

PR #516, opened 2018-07-30
Add padding around above threshold SiPM samples

andLaing:
Makes the baseline calculation more reliable
and MC more similar to data.

Fixes bug noticed in MC runs.

### DIFF
--- a/invisible_cities/cities/diomira.py
+++ b/invisible_cities/cities/diomira.py
@@ -30,7 +30,7 @@ class Diomira(MonteCarloCity):
     2. the response of the trigger
     """
 
-    parameters = tuple("""trigger_type""".split())
+    parameters = tuple("""filter_padding trigger_type""".split())
 
     def __init__(self, **kwds):
         """Diomira Init:
@@ -46,7 +46,9 @@ class Diomira(MonteCarloCity):
         self.cnt.init(n_events_tot = 0,
                       nevt_out     = 0)
 
-        self.sipm_noise_cut   = conf.sipm_noise_cut
+        self.sipm_noise_cut = conf.sipm_noise_cut
+
+        self.filter_padding = conf.filter_padding
 
         # thresholds in adc counts
         self.sipms_thresholds = self.sipm_noise_cut *  self.sipm_adc_to_pes
@@ -84,7 +86,9 @@ class Diomira(MonteCarloCity):
 
             thr_with_base   = self.sipms_thresholds[:, np.newaxis]
             thr_with_base  += self.noise_sampler.baselines
-            dataSiPM        = wfm.noise_suppression(dataSiPM_noisy, thr_with_base)
+            dataSiPM        = wfm.noise_suppression(dataSiPM_noisy     ,
+                                                    thr_with_base      ,
+                                                    self.filter_padding)
 
             RWF = dataPMT.astype(np.int16)
             BLR = blrPMT.astype(np.int16)
--- a/invisible_cities/cities/diomira.py
+++ b/invisible_cities/cities/diomira.py
@@ -50,8 +50,9 @@ class Diomira(MonteCarloCity):
 
         self.filter_padding = conf.filter_padding
 
-        # thresholds in adc counts
+        # thresholds in adc counts with baselines
         self.sipms_thresholds = self.sipm_noise_cut *  self.sipm_adc_to_pes
+        self.sipms_thresholds[:, np.newaxis] += self.noise_sampler.baselines
 
         self.trigger_filter   = TriggerFilter(self.trigger_params)
         self.trigger_type = conf.trigger_type
@@ -84,11 +85,9 @@ class Diomira(MonteCarloCity):
             dataPMT, blrPMT = self.simulate_pmt_response(evt, pmtrd)
             dataSiPM_noisy  = self.simulate_sipm_response(evt, sipmrd)
 
-            thr_with_base   = self.sipms_thresholds[:, np.newaxis]
-            thr_with_base  += self.noise_sampler.baselines
-            dataSiPM        = wfm.noise_suppression(dataSiPM_noisy     ,
-                                                    thr_with_base      ,
-                                                    self.filter_padding)
+            dataSiPM        = wfm.noise_suppression(dataSiPM_noisy       ,
+                                                    self.sipms_thresholds,
+                                                    self.filter_padding  )
 
             RWF = dataPMT.astype(np.int16)
             BLR = blrPMT.astype(np.int16)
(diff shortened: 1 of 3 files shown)


msorel:
thanks @andLaing can you give more details in words on what was the bug and how it was fixed?

should we make a new MC production using andLaing:diomiraFix before merging? I could tell corey to do it for the background MC. or has @paolafer already run the necessary checks on the calibration MC?

andLaing:
The filter which runs on the simulated SiPM waveforms was literally just a if < threshold = 0.

This didn't leave enough information to calculate the baselines so it tended to be overestimated, sometimes by a large amount. The change is to take a bit before and after each identified above threshold sample which is also more similar to what the FPGA does for the SiPMs in zero suppression mode. As a default I've put ±50 samples in the example configs like the current settings of the Java do for data although the thresholds are not identical. This results in larger output files for the same data and slower running (not by much I don't think but I've only checked 5 events) but I don't see another way around it. Maybe someone more expert in python could improve the code but in the short term we can try this as it works.

Paola is checking if this fixes all the oddities she was seeing and will approve when happy.

msorel:
sounds good to me. @jjgomezcadenas : @paolafer reviewer for this?

jacg:
@andLaing 5 PRs have been merged since you created this branch. I suggest that you rebase sooner rather than later. I did it for you, but you seem to have undone it.

andLaing:
Ah, that’s why I got a push error… sorry.

@paolafer  asked me to hold off on the rebase since the merges from yesterday force the need for changes in the environment etc.

We'll rebase and check once she's happy

jacg:
> Ah, that’s why I got a push error…

Indeed. And the fact that `--force-with-lease` didn't work, should have alerted you to the fact that someoene else had done something on the branch. (I trust that you are in the habit of using `--force-with-lease` rather than `--force` In magit this is `f` as opposed to `F`. There are **very** few situations in which `push --force` is the right thing to do!)

> sorry

No, I should have warned you. I was going to, but there were too many things going on at the same time, and I didn't get around to it.

> @paolafer asked me to hold off on the rebase since the merges from yesterday force the need for changes in the environment etc.

I sympathize, but I suspect that biting the bullet and getting it over and done with as soon as possible, will minimize the total pain in the middle and long term. But then I don't know what you need in your local environment to do the testing of which you speak, so do whatever you think is best.

andLaing:
The push worked with -f in magit, the error came when I pushed without any extra options. I assumed it was related to the squash I had done the time before so added the --force-with-lease.

I think paola wanted to be able to download and run, as soon as she's happy (by 19:20 so she can get the tram) I'll sort things, I'm already reinstalling the environment for my master.

jacg:
> The push worked with -f in magit, 

That's odd.

paolafer:
Andrew and myself have passed a bunch of events through diomira and irene, and the bug seems solved: the SiPM waveforms looks fine (while before they were not). Therefore, from the point of view of functionality it seems ok. However, I haven't processed a large statistics, due to lack of time, I'll do it next week.

andLaing:
I've rebased and pushed.

It seems to work happily although the runs where we don't have baselines in the database aren't great (pre run 5906 at the moment). If we really want to rerun anything for comparison for old runs we can calculate baselines or use the current ones maybe, I doubt there will be much difference.

Maybe worthwhile that @jjgomezcadenas has a quick look at the code and makes some comments before we approve?

jjgomezcadenas:
We got another hypothesis error... needs to be re-run. I am starting to be a bit skeptical about hypothesis, it is becoming a real pain...

jjgomezcadenas:
Concerning the code, it is too complex for a "quick look". I would prefer that @paolafer runs large stats and certifies consistent behaviour before approving, unless is urgent.

jacg:
> I am starting to be a bit skeptical about hypothesis, it is becoming a real pain...

Hypothesis is wonderful. We need to be less ham-fisted when using it. That's why I have opened (not one, but at least 2, it seems) issues, reminding me that I have to investigate the situation and give you good advice. Not had time to do it yet, sadly.

msorel:
hi @andLaing and @paolafer is software development on this branch done, at least until review?

After having fixed a number of other bugs and features in background MC production, I am now confronted with the fact that half of the events have zero SiPMs touched, despite having visible energy > 500 keV. And the ones that have some SiPMs, have only  a few. In short, the output makes no sense. I hope this will be fixed by this pull request. Let me know and I can try it out.

andLaing:
The last push removes an error that I had introduced which was causing the threshold to grow each event (stupid use of addition with a self. variable due, in part, to not using the correct run number when testing where to do the addition, sorry to all involved). I ran over a full MC data file of 9200 events and the basic plots look good.

@paolafer wanted to check things when she got back (tomorrow, I think) and she is the official reviewer so, of course, there won't be a merge until she's happy but have a look at the branch and run some events, that way any bugs I've missed can be found earlier and we'll know that it's working for you too.

jjgomezcadenas:
@msorel if you run some sample events and are happy with the outcome, I'd be happy to approve this PR

msorel:
@jjgomezcadenas @andLaing i want to confirm that i am happy with MC outcome of this branch. By processing some background MCRD events through diomira, irene, dorothea, penthesilea with this branch, I see:

- Back to ~5% reconstruction inefficiency for irene+dorothea or irene+penthesilea processing, using standard configurations. This is OK, and should be compared with ~50% earlier inefficiency due to zero SiPMs touched

- Number of SiPMs touched per S2 is now in the 10-100 range, typically, and not 0-3 as in the buggy master. This also looks reasonable

- For events with multiple S2s in the same event and S2s close in Z (few mm apart), I have verified that their X and Y positions are also nearby (few mm apart). So the fixed SiPM waveforms seem to give reasonable X/Y outputs.

In short, all good. We will keep validating MC with more in-depth studies and report residual oddities. But this is good enough  to be merged, as far as I am concerned.

jjgomezcadenas:
@andLaing could you please rebase? Then I will approve

andLaing:
Done.

Probably it should be revisited for possible improvements to the code at some point.

jjgomezcadenas (approved):
Extensively tested first by @paolafer then by @msorel, with consistent results.

jjgomezcadenas:
@gonzaponte, please merge this one too!

paolafer:
I haven't tested it yet.

jjgomezcadenas:
@msorel run a number of tests, please go ahead and run extra tests, if we find any issue will amend the PR, but it was somewhat urgent to approve (AND MERGE, PLEASE @gonzaponte ) for MC productions at Fermilab

PR #517, opened 2018-07-30
Fix test_inverse_cdf_index_hypothesis_generated

jacg:
`test_inverse_cdf_index_hypothesis_generated` fails on, very roughly, half of our Travis builds.

This probably means one of two things:

1. There is a subtle bug in our code (which needs fixing).
2. The hypothesis generators used in the test find spurious examples (which need eliminating).

Either way, this wastes a lot of developer time when it comes to merging PRs and Travis builds in general (it has wasted about half of my waking hours over the last 2 days), so getting it sorted out sooner rather than later is quite important.

The original author of the test was @gonzaponte: in his absence, is there anyone else who is familiar with it?

The chances are that you don't see the failure when running tests locally, as it only shows up in about 50% of `hard` profile test runs, so the chances of our (very weak) `dev` profile finding it are quite slim. To work on this problem, run the test with the `hard` profile locally, repeatedly, until the test fails. After that the hypothesis database (which is stored locally on your machine) should ensure that the counterexample that was found gets used by the `dev` profile every time, allowing you to get a much quicker turnaround.

Let me reiterate: this is a huge PITA, and it will annoy you with spurious Travis failures of your bulids, so it is en evyrone's interest to get it sorted out ASAP. Does anyone feel qualified to have a look at this?


jjgomezcadenas:
This falls directly into the domain of @gonzaponte. It is part of a set of tests which do not affect the main stream of current development. It is complex and totally his responsibility. It is not justified that other developers loose essential time (we are really up to our neck here) with this stuff. 

It is my understanding that @gonzaponte is not available until some time in September. Therefore, and unless he can find a couple of hours to work in this problem, please **comment out** the test until he can fix it. We can then approve a PR implementing this (no)-solution, but curing the symptoms and open a new one that will be closed only when @gonzaponte can fix the problem.

jacg:
> please **comment out** the test until he can fix it.

Nononononnooooooo!

Mark it with `@skip`, or similar.

jjgomezcadenas:
This is what I meant, sorry: 

comment = @skip.

jacg:
> please [skip] the test until he can fix it.

I agree.

I have taken the liberty to push a commit that does exactly this, without creating a PR and waiting for review. Happy Travising.

jjgomezcadenas:
Perfect, so that we can then keep this PR open until @gonzaponte has the chance to fix it.

jacg:
BTW, it would be good if we all got into the habit of running `pytest -v -m skip -r s` once a month ... and seeing if we are in a position to unskip anything that shows up.

PR #520, opened 2018-07-31
Hits corrections

mmkekic:
Adds hits_corrections.py and hits_corrections_test.py.
**To discuss:**
1. HitEnergyMap is added in hits_corrections but should probably be elsewhere
2.  There is a function 'merge_NN_hits' that redistribute energy of NN hits to surrounding non-NN hits 
3. A function hitc_corrections is applied to HitCollection since functions that read correction maps are much faster when applied to lists (instead of reading it hit by hit)


jjgomezcadenas:
@mmkekic tests fail because you use @dataclass (which I use extensively). This will work on python 3.7 but needs to be backported to python 3.6. I will try to look into this.

mmkekic:
Another option is to skip totally dataclass and just change HitCollection inside hitc_corrections function.

jjgomezcadenas:
There is a better solution: We can use a class derived from NamedTuple for the time being. It is quite similar to dataclass in the sense that it allows you to create a typed little data object. With this solution you cannot rewrite the fields (is a tuple), but in general you don't want to rewrite the field of dataclasses objects rather you create a new object, so this is OK. Thus I propose to use:

```
     from typing import NamedTuple

     class EVector (NamedTuple):
         raw : float            # raw value
         lt  : float            # corrected by lifetime
         cor : float            # fully corrected

And similar for any other dataclass that we need until move to python 3.7

Otherwise the PR looks pretty nice, clean and tested. Please, add this change, rebase and I will be happy to approve.

**Comment by mmkekic** [comment]
Can someone check that the corrections formulas for E and Q are correct?

**Comment by jjgomezcadenas** [comment]
The formulas appear correct to me, but a even more important, your test seems to demonstrate that they **are** correct, no?

**Comment by jjgomezcadenas** [comment]
Oh, I see, the text is a toy, you get what you put there. @jahernando, can you check the correction formulas (one to last commit). They appear correct to me, anyway.

**Comment by jjgomezcadenas** [comment]
Looking at the definition of functions:

```
     qlt    = q_hits  * QLTcor(z_hits, x_hits, y_hits).value**(ftlife)
     qcor   = qlt     * QXYcor(x_hits,y_hits)         .value
     elt    = e_hits  * ELTcor(z_hits, x_hits, y_hits).value**(ftlife)
     ecor   = elt     * EXYcor(x_hits,y_hits)         .value
```
I don't understand the factor **(ftlife) in qlt and elt. It is intended as a scale factor, but I don't understand why it goes with a power (**fitlife) and also I think it is dangerous. I would assume for the time being that each map gives the correct scale and remove ftlife. 

Furthermore, why don't you try to write all the functions with type hints? It makes for a more readable code and in the future we may be able to type check our functions. In the case of the functions that you are writing it would be something like this:

```
     def hitc_corrections(hitc : HitCollection, 
                                        EXYcor :  FCorrection, 
                                       ELTcor   :  FCorrection,
                                       QXYcor  : FCorrection, 
                                      QLTcor    : FCorrection )->HitEnergyMap
```
You need to import the names of all the types you reference (e.g, FCorrection).  The point is that the reader can understand the signature of the functions that you are using in your functions.

jahernando:
I am sorry. I think all this can be more simple:

We need an interface that takes arrays of X, Y, Z, T (absolute time), S2e (energy), S2q (tracking), and returns E (energy), Q (energy at tracking). In addition, with a flag (full=True), it returns also the Elt, Qlt, the corrected energy only by lifetime. 

The client will create a class i.e 'Calibration' with a filename. The class internally will figure out which is the best available map for geometry and lifetime. Clients do not need to know anything about what is done internally by this class.

The interface must take numpy arrays to speed up. It will return numpy arrays. The client then can store the values in the Event model hit, if needed.

The client will do something:

At initialization, one per job, 
calibrate = Calibration(filename)

Then per hits in a event, or per run:

E, Q = calibrate(X, Y, Z, T, E, Q) 

or in full:

E, EL, Q, QL = calibrate(X, Y, Z, T, E, Q, full=True)

The best is that the people who are preparing the maps, take care of the implementation of the interface.

jahernando:
@mmkekic we can propose an interface and a implementation in the next days. Assume for the moment that you will get the corrected quantities, for Elt, E, Qlt, Q.

mmkekic:
@jjgomezcadenas the formulas are copied from @jerenner script. Maybe he can explain better the factor?

@jahernando, if I understand, you are proposing to correct penthesilea what was decided not to be done? I dont have any preferences and was just following the suggestions given by @jjgomezcadenas  in #508  but I believe it is open for discussion. I will wait for a consensus on implementation of post-penthesilea analysis.

jahernando:
@mmkekic No, I am not proposing to change Penthesilea, She should produce un-corrected hits. What I am proposing is an interface valid in any step of the analysis, to apply the corrections, without exposing its mechanics to the client.

jjgomezcadenas:
@mmkekic, @jahernando : the point is that the current IC implementation (more or less copied by @jerenner in his scripts), assumes a specific setup, e.g, files containing a single correction map. In the most general case, we may need to interpolate among several maps, for example, although we will probably start by using a single map per run. But the point is that we want to hide all this to the user. The interface will probably require the run number and/or a tuple of time-stamps and will return the correction functions. We will try to produce such an interface,plus the zero order implementation in the next few days.

jacg:
> What I am proposing is an interface valid in any step of the analysis, to apply the corrections, without exposing its mechanics to the client.

Good, good ...

> But the point is that we want to hide all this to the user. The interface will probably [...]

... excellent stuff. The review system is doing an good and vital job here. Keep up the good work.

mmkekic:
@jahernando  Ah I see, I didnt read it carefully last night, sorry. So basically instead of using directly functions inside reco.dst_functions you will provide an universal interface? Yes, that sound good. 
Another question I have is if adding energies of NN hits to the closest localized hit makes sense to everyone?

jjgomezcadenas:
>Another question I have is if adding energies of NN hits to the closest localized hit makes sense to everyone?

Yes, this is our standard approach (not necessarily very correct, but the only one we have,so far)

jerenner:
> I don't understand the factor **(ftlife) in qlt and elt. It is intended as a scale factor, but I don't understand why it goes with a power (**fitlife) and also I think it is dangerous. I would assume for the time being that each map gives the correct scale and remove ftlife.

Sorry about my late response to this - the factor `ftlife` scales a given map to a different average lifetime, so the geometric variations given by that map are applied but the average lifetime matches that of a different run.

It needs to be applied by exponentiating `**(ftlife)` because the factor itself is an exponential `exp(t/tau)` and to get `exp(t/tau')` one needs to perform:

`exp(t/tau') = exp(t/tau)**(tau/tau')`.

So the factor we call `ftlife` is equal to the ratio `tau/tau'`.

paolafer:
> Another question I have is if adding energies of NN hits to the closest localized hit makes sense to everyone?

I have taken a look at the distance between the z position of the NN hits and the z position of the closest hit its energy is assigned to, I attach a plot for run 6202.
nn_hits_r6202

As you can see, there is a tail where clearly something has gone wrong in the reconstruction. I would suggest to filter out events of this type in the analysis.

I attach also the same plot, for MC.

nn_hits_mc

In MC the distribution decreases constantly, while in data the tail is sort of flat. Also, the number of hits at larger distances is lower.

mmkekic:
Do you have any idea where the threshold should be @paolafer ?

paolafer:
It is not clear from the plot, maybe 100 mm? 
A typical situation where this approach is sensible should be when the first or last slices of a waveform are not well reconstructed in the SiPMs. In this case, it is ok that the closest slice is not necessarily contiguous, but if too many slices are not reconstructed, I think that we should have a look at those events. The same goes, if it is a slice that is not connected to a waveform.

paolafer:
I've tried to merge the NN hits in a slightly different way. 
Instead of looking for the closest hits among all the hits of an event, I take into account only the hits that belong to the same peak as the NN hit's one. If there are no good hits in a given peak, this peak must be rejected. In the current code, the NN hits are assigned, erroneously, to the hits of another peak.
With this code, the distribution of the distance between an NN hit and its closest one looks much more reasonable.

nn_hits_r6202_new

I attach also a plot of the energy that gets lost with this procedure:
lost_peak_energy

It's mainly X rays events that give probably a too low signal.

jerenner:
In this case the "lost" energy is due to entire peaks being discarded for containing no good hits?  And this energy is no longer included in the event energy (not saved in any way)?  Thanks for looking into this!

paolafer:
@jerenner Yes, exactly.

PR #524, opened 2018-08-05
Change load_pmaps function to improve performance

mmkekic:
load_pmaps is very slow for large pmaps files I believe due to the lookup times in dataframe. I propose minimal change that uses groupby object so that load_pmaps scales linearly with the number of events.
The comparison of current and proposed load_pmaps performance for 3 different files (~300, ~10k, ~40k events) is shown in figure. 

figure_1

### DIFF
--- a/invisible_cities/io/pmaps_io.py
+++ b/invisible_cities/io/pmaps_io.py
@@ -95,23 +95,46 @@ def _build_ipmtdf_from_sumdf(sumdf):
     ipmtdf['npmt'] = -1
     return ipmtdf
 
-
 def load_pmaps(filename):
     pmap_dict = {}
     s1df, s2df, sidf, s1pmtdf, s2pmtdf = load_pmaps_as_df(filename)
-
     # Hack fix to allow loading pmaps without individual pmts
     if s1pmtdf is None: s1pmtdf = _build_ipmtdf_from_sumdf(s1df)
     if s2pmtdf is None: s2pmtdf = _build_ipmtdf_from_sumdf(s2df)
-
     event_numbers = set.union(set(s1df.event), set(s2df.event))
+    s1df_grouped   =s1df   .groupby('event')
+    s1pmtdf_grouped=s1pmtdf.groupby('event')
+    s2df_grouped   =s2df   .groupby('event')
+    s2pmtdf_grouped=s2pmtdf.groupby('event')
+    sidf_grouped   =sidf   .groupby('event')
     for event_number in event_numbers:
-        s1s = s1s_from_df(s1df   [s1df   .event == event_number],
-                          s1pmtdf[s1pmtdf.event == event_number])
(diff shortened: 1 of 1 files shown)


mmkekic:
I got this error from Travis in my fork:

   @given(dictionaries(keys=integers(min_value=-1e5, max_value=1e5), values=pmaps(), max_size=5), lists(integers(min_value=-1e5, max_value=1e5)))
>   def test_pmap_event_id_selection(pmaps, events):
E   hypothesis.errors.FailedHealthCheck: Data generation is extremely slow: Only produced 6 valid examples in 1.04 seconds (5 invalid ones and 0 exceeded maximum size). Try decreasing size of the data you're generating (with e.g.max_size or max_leaves parameters).
E   See https://hypothesis.readthedocs.io/en/latest/healthchecks.html for more information about this. If you want to disable just this health check, add HealthCheck.too_slow to the suppress_health_check settings for this test.
invisible_cities/reco/pmaps_functions_test.py:83: FailedHealthCheck
---------------------------------- Hypothesis ----------------------------------
You can add @seed(319847072161555484591984935677630821855) to this test or run pytest with --hypothesis-seed=319847072161555484591984935677630821855 to reproduce this failure.

Does somebody know why is this happening?

jjgomezcadenas (commented):
@mmkekic I like this very much, but I wonder if you could hide the groupby implementation inside s1s_from_df / s2s_from_df. Also, you may want to add a test (if you hide your implementation behind s1s_from etc, existing tests will probably be enough...

Good work!

jjgomezcadenas:
>Does somebody know why is this happening?

Seems to be a known quirk of hypothesis. Nothing to worry for now.

mmkekic:
> I like this very much, but I wonder if you could hide the groupby implementation inside s1s_from_df / s2s_from_df
It can not be inside event loop cause it makes no sense to group it more than once per full dataframe.  I dont see any easy way  of implementing it in others existing functions except maybe in load_pmaps_as_df but maybe is not very nice.
I can probably try to make it prettier but it will require much more changes... Any other suggestions?

jjgomezcadenas:
>I can probably try to make it prettier but it will require much more changes.

Actually, I think you are right, I had just overlooked the event loop! The code is quite clear as it is, but we still need a test to check that what you get before your changes and what you get after your changes is the same.

mmkekic:
But isnt test_load_pmaps doing exactly that check?

jjgomezcadenas:
>But isnt test_load_pmaps doing exactly that check?

Yes, you are right. So we can approve this PR: Good job. @carmenromo please merge.

jjgomezcadenas (approved):
Boosts performance, smart and concise. Good job.

jjgomezcadenas:
This needs to be merged asap: @carmenromo is in vacation, please, @gonzaponte, take 5 minutes and merge it if you may.

PR #525, opened 2018-08-07
Improve mc copy

msorel:
This PR is for fixing the copying of MC truth information, in particular the MC generator information on decaying isotope and decay volume in the detector. In the master branch, the copy is wrong if the two following conditions are met at the same time:

1. Some event is being filtered out between input and output. This can be the case, for example, in irene, dorothea or penthesilea cities.

2. Event numbers are not in increasing order in the input file. This can be the case, for example, in background MC files.   

When back from vacations, @paolafer is probably the best person to review this.

### DIFF
--- a/invisible_cities/io/mcinfo_io.py
+++ b/invisible_cities/io/mcinfo_io.py
@@ -223,10 +223,10 @@ def read_mcinfo_evt (mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
             while ipart <= ipart_end:
                 particle_rows.append(h5particles[ipart])
                 ipart += 1
-            
-            # It is possible for the 'generators' dataset to have a different length compared to the 'extents' dataset. In particular, it may occur that the 'generators' dataset is empty. In this case, do not add any rows to 'generators'.
-            if len(h5generators) == len(h5extents):
-                generator_rows.append(h5generators[last_row])
+
+            # It is possible for the 'generators' dataset to be empty. In this case, do not add any rows to 'generators'.
+            if len(h5generators) != 0:
+                generator_rows.append(h5generators[iext])
 
             break
 
--- a/invisible_cities/io/mcinfo_io.py
+++ b/invisible_cities/io/mcinfo_io.py
@@ -85,40 +85,43 @@ class mc_info_writer:
             self.reset()
 
         extents = mctables[0]
+        # Note: filtered out events do not make it to evt_number, but are included in extents dataset
         for iext in range(self.last_row, len(extents)):
             this_row = extents[iext]
-            if this_row['evt_number'] < evt_number: continue
-            if this_row['evt_number'] > evt_number: break
-            if iext == 0:
-                if self.first_file:
-                    modified_hit          = this_row['last_hit']
-                    modified_particle     = this_row['last_particle']
+            if this_row['evt_number'] == evt_number:
+                if iext == 0:
+                    if self.first_file:
+                        modified_hit          = this_row['last_hit']
+                        modified_particle     = this_row['last_particle']
+                        self.first_extent_row = False
+                        self.first_file       = False
+                    else:
+                        modified_hit          = this_row['last_hit'] + self.last_written_hit + 1
+                        modified_particle     = this_row['last_particle'] + self.last_written_particle + 1
+                        self.first_extent_row = False
+
+                elif self.first_extent_row:
+                    previous_row          = extents[iext-1]
+                    modified_hit          = this_row['last_hit'] - previous_row['last_hit'] + self.last_written_hit - 1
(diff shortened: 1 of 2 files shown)


paolafer:
I will review this PR in the next days. However, I have a couple of questions. 
The first one concerns the way the difference between the previous file and the new file is shown in the commit tab: in the first commit of this PR is not that easy to understand which line(s) have been changed. What should we take into account to make commits cleaner?
The second questions is: can you implement a test that shows the bug with the current code and is passed with this change? I think it would help the review.

msorel:
thanks @paolafer !

concerning your two points:

- you are right, commit seems to indicate that a lot has changed, but really mostly the indentation has changed from

```python
            if this_row['evt_number'] < evt_number: continue
            if this_row['evt_number'] > evt_number: break
            if iext == 0: # unindented
```

to

```python
            if this_row['evt_number'] == evt_number:
                if iext == 0: # indented
```

I do not know an easy way to fix this, I really want stuff inside this if statement, it seems to me.

- Yes, I was afraid you were going to ask for a test. So I started working on one already. But it takes some  time and effort. I will add an additional commit with a test today or tomorrow.

msorel:
hi @paolafer i have expanded the test "test_mc_info_writer_non_consecutive_events" to include the case where event numbers are not ordered in ascending order in the input file. please have a look. i have verified that this test would fail (twice) in nextic:master, while it passes in msorel:improve-mc-copy

as a minor point, i have removed the numpy.unique() ordering for events_in and events_out in this test, as random ordering was precisely one thing i wanted to test. i do not know why you put that in the first place, probably you have a good reason? if so, i am happy to put it back. The test would still fail in nextic:master and would still pass in msorel:improve-mc-copy, so it would still catch this particular problem.

paolafer:
Hi Michel, I think that the `np.unique()` can be safely removed, since we have always different IDs for different events.

The code looks fine to me: in the call method of the `mc_info_writer` class,  the condition `if this_row['evt_number'] == evt_number:` + `break` at the end of the event is more concise and general and allows to take into account the case where the events are not in increasing ordering. Also, a test has been added, to show the failure. 

However, I don't understand why the condition `if len(h5generators) == len(h5extents)`, in the `read_mcinfo_evt()` function is not correct in the case of event filtering. Aren't both tables the ones read from the input file? To my understanding there should be one row per event in both tables, always, shouldn't it? Also, correct me if I'm wrong, the test checks the first commit change, but not the second one. If you prefer, we can speak it on the phone.

msorel:
hi @paolafer, thanks for reviewing! 

OK, i leave out np.unique(), then

On the reason why I have changed from:

```python
            if len(h5generators) == len(h5extents):
```

to

```python
            if len(h5generators) != 0:
```

for appending a new generator row. This is not related to event filtering, it just seems more correct in general to me. Two cases should be correct here:

- calibration MC: len(h5extents) equals number of input events before filtering, len(h5generators) = 0
- background MC: both len(h5extents) and len(h5generators) equal number of input events before filtering

Suppose you have a buggy background MC file, with len(h5extents) different from len(h5generators) and both non-zero. Previous code would not copy h5generators to the output but would copy h5extents, in  a way assuming the problem is in h5generators and not in h5extents. This assumption does not need to be correct. It seems better to me to allow copying of h5generators to allow debugging downstream. 

On the other hand, the more minimal condition  len(h5generators) != 0 is necessary in order  to access an existing array element in h5generators, in

```python
                generator_rows.append(h5generators[iext])
```

That's why I made this change. But yes, it is unrelated to event filtering.

msorel:
Hi again @paolafer 

> Also, correct me if I'm wrong, the test checks the first commit change, but not the second one. If you prefer, we can speak it on the phone.

You are right. I guess I could extend the test `test_mc_info_writer_output_non_consecutive_events` to check the input and output contents match, not only for hit rows and particle rows, but also for generator rows. Or make a separate test to this effect. Any preference?

jjgomezcadenas:
A separate test is normally more readable. Also, you need to rebase

msorel:
@paolafer @jjgomezcadenas i have created a new test `test_mc_info_writer_generatoroutput_non_consecutive_events`. I have verified that this test actually fails prior to commit [b8f9ad0](https://github.com/nextic/IC/pull/525/commits/b8f9ad0e95155f28d9f456bb4dec5d36d672b12f) while the test passes after the commit.

I have rebased on top of nextic/master. First time rebasing on my own without git expertise from @bpalmeiro , so please check!

paolafer (approved):
Ok, I think that this PR is ready to be approved! Go along with the merging, then!

carmenromo:
The branch is out-of-date, I cannot merge. @msorel could you please review?

msorel:
hi @carmenromo i have rebased again. merges into master occurred since my rebase three days ago. please try again once checks are completed, it should work now!

carmenromo:
Thanks! I will try again when Travis finishes

PR #530, opened 2018-08-28
Add seaborn to our environment.

jjgomezcadenas:
Seaborn is a high level, very useful, statistics visualization library
based on matplotlib and making extensive use of numpy and pandas. It
deserves to live in our environment.

### DIFF
--- a/manage.sh
+++ b/manage.sh
@@ -83,24 +83,24 @@ function make_environment {
     cat <<EOF > ${YML_FILENAME}
 name: ${CONDA_ENV_NAME}
 dependencies:
-- python=${PYTHON_VERSION}
-- cython=0.28.4
-- jupyter=1.0.0
-- matplotlib=2.2.2
-- networkx=2.1
-- notebook=5.6.0
-- numpy=1.15.0
-- pandas=0.23.3
-- seaborn=0.9.0
-- pymysql=0.9.2
-- pytables=3.4.4
-- pytest=3.6.3
-- scipy=1.1.0
-- sphinx=1.7.6
-- tornado=5.0.2
-- flaky=3.4.0
-- hypothesis=3.59.1
-- pytest-xdist=1.22.2
+- python       = ${PYTHON_VERSION}
+- cython       = 0.28.4
+- jupyter      = 1.0.0
+- matplotlib   = 2.2.2
+- networkx     = 2.1
+- notebook     = 5.6.0
+- numpy        = 1.15.0
+- pandas       = 0.23.3
+- seaborn      = 0.9.0
+- pymysql      = 0.9.2
+- pytables     = 3.4.4
+- pytest       = 3.6.3
+- scipy        = 1.1.0
+- sphinx       = 1.7.6
+- tornado      = 5.0.2
+- flaky        = 3.4.0
+- hypothesis   = 3.59.1
+- pytest-xdist = 1.22.2
 EOF
 
     conda env create -f ${YML_FILENAME}
--- a/manage.sh
+++ b/manage.sh
@@ -91,7 +91,7 @@ dependencies:
 - notebook=5.6.0
 - numpy=1.15.0
 - pandas=0.23.3
-- seaborn-0.9.0
+- seaborn=0.9.0
 - pymysql=0.9.2
 - pytables=3.4.4
 - pytest=3.6.3
--- a/manage.sh
+++ b/manage.sh
@@ -72,7 +72,7 @@ function install_conda {
     fi
 }
 
-CONDA_ENV_TAG=2018-07-29
+CONDA_ENV_TAG=2018-08-29
 CONDA_ENV_NAME=IC-${PYTHON_VERSION}-${CONDA_ENV_TAG}
 
 function make_environment {
--- a/manage.sh
+++ b/manage.sh
@@ -91,6 +91,7 @@ dependencies:
 - notebook=5.6.0
 - numpy=1.15.0
 - pandas=0.23.3
+- seaborn-0.9.0
 - pymysql=0.9.2
 - pytables=3.4.4
 - pytest=3.6.3


jjgomezcadenas:
So, the obvious thing that I thought it would work (I just added seaborn to the list of packages, after checking that I could install it with conda) did not work. I guess is a stupid error, trivial to fix by @jacg

jacg:
"did not work" doesn't really give me much to go on.

jacg:
I hope that you recall that the environments are now versioned (#496). As written in the description of #496 (which is probably the closest thing we have to documentation for this feature) there is a environment version tag in `manage.sh`

    CONDA_ENV_TAG=2018-07-29

Whenever you change the environment configuration, you need to change this tag too. Then make a commit. With that done, `manage.sh` commands such as those which run the tests, should automatically switch to the environment that corresponds to the currently checked out commit.

Please change the tag and try again.

I notice that in `manage.sh` the command `switch_to_conda_env` is listed in the help feature of the command dispatcher, but is not actually enabled in the dispatcher itself. This should be fixed. Perhaps you have the manpower available to deal with this (I certainly do not). The probable solution is trivial: just add that command to the dispatcher `case` block. Verifying that the solution works will require more effort. (Let me state once again that `manage.sh` is something I'm not at all proud of, and would love to replace the whole thing something much more sane ... but what?)

Please also recall that, in order to reduce the chance that issues and PRs fester in oblivion, I have created a [project board](https://github.com/nextic/IC/projects/1). I recommend placing PRs and issues that you create on there ASAP, in order to reduce the chance that they be forgotten. The board should probably automatically grab new issues and PRs, but doing this naively will merely create another infinitely long list of items which will be ignored by everyone, so it requires a bit of thought. Perhaps we need a *Triage* column.

jjgomezcadenas:
Still fails miserably. @jagc I would be very grateful it you could look at this.

jacg:
"fails miserably" doesn't really give me much to go on.

Hint: I'm trying to make a point here. It is my second attempt.

jacg:
... and the point I am trying to make is that, in the context of software problems in general, but *especially* in the context of open source projects, error reports that essentially say something like "it didn't work", which is just a convenient shorthand for **"It didn't work for me that one time I tried it on my machine, but as my time is more precious than yours, I'll let you figure out how I screwed it up without giving you any of the information you need to be able to figure out what local configuration problem of mine is behind it all"** minimize the chances of your problem being solved (for a long list of reasons, which I'll not detail here and now).

Giving helpful error reports is a very useful thing to do for many reasons. One of these is that going through the steps of explaining your problem to the person you expect to solve it for you, has a lot in common with rubber duck debugging (look it up, if you don't know what it is), which, often, miraculously solves your problem for you without the need to bother anyone else. 

Had you rubber-duck-debugged this problem (or, equivalently, tried to give me a better error report than "fails miserably") then I suspect that your rubber duck would have pointed out very quickly that you used a `-` rather than a `=` in the seaborn package specification. (And your rubber duck would have told you this many hours before I even saw your request!)

Incidentally, this bug would have been much less likely, if the package spec had been JCK-1 conformant (align lines to make symmetries obvious, and asymmetries stick out like a sore thumb). So I have committed an JKCization. Fortunately the YML syntax admits this.

Applying this fix locally, gets me past the problem of failure to find the conda package `seaborn-0.9.0` and thus failing miserably to create the new environment, but I cannot confirm that there are no further problems related to the addition of seaborn, as the test suite fails miserably.

Oh, you'd like me to give you more useful information than "fails miserably"? Sure. All the tests that fail, do so because of database issues. Did I try to download the relevant database? Yes, I did, and I repeatedly get 

    Can't connect to MySQL server on 'neutrinos1.ific.uv.es' (timed out)

At which point

+ I am stuck
+ The time I can afford to give to IC today, has run out.

jjgomezcadenas:
Thanks @jacg, and I am sorry for being so thick here. The data base is not accesible because neutrinos has been down for a few days, courtesy of the computing services at IFIC. I expect the problem to be solved anywhere between few days and one week. I will move the data base out of IFIC as soon as I can.

jacg:
> sorry for being so thick

Being thick is not the problem. It is important to get into the habit of helping people to help you. It makes for far more efficient use of everyone's time, and for faster progress overall. And if you do make the effort, the rubber duck often steps in and solves the problem even more quickly.

> I will move the data base out of IFIC as soon as I can.

Probably a very good idea, but it is likely to make it very difficult to run pre-database-migration versions of the code. If the database location is hard-wired somewhere in the code, then getting the correct corresponding version of the database for pre-migration code is going to fail (miserably :-) after the old server goes offline. I don't know how important it is going to be for you to be able to run old code.

This, of course, brings us back to Git LFS once more. Given how no progress is being made there, I'm wondering whether perhaps we should simply put all the binaries in plain Git, and use shallow cloning to avoid huge downloads. It's morally wrong, but perhaps the most pragmatic and perhaps even most robust solution. I'm nervous about how it could come back to haunt us, but I haven't had the time to think more carefully about the potential problems.

jjgomezcadenas:
>This, of course, brings us back to Git LFS once more.

I discussed with Jose Maria @jmbenlloch and we agreed, the issue will take priority as soon as we all return to life after August break.

jacg:
I don't see what the type hints correction commit is doing in this PR.

paolafer:
I've done nothing this time :-). Git is starting having free will...

carmenromo:
Could it be caused by my merge?

jacg:
> Git is starting having free will...

Many SciFi authors have scared us with this sort of scenario ... though I can't think of one that predicted a version control system as the culprit.

> Could it be caused by my merge?

Hmm ... your merge seems to be fine. (Though our agreed on policy was that single-commit PRs would just be rebased, without adding an extra merge commit; but that really isn't important, and adding the merge commit is absolutely fine, perhaps even better than the alternative).

Unless, in the process of doing the merge, you accidentally pushed that commit to `jj/add-seaborn` in addition to everything else you did, which seems correct. It's possible.

Anyway, we all agree that that commit should be removed from this branch, right?

carmenromo:
Then this is not the problem, I couldn't accidentally pushed that commit to jj/add-seaborn in addition to everything else I did for the simple reason that I don't have JJ's remote haha

jjgomezcadenas:
It may be my fault, I fetched from IC, then rebased my seaborn branch to IC, then feched from my remote, then rebased into my remote. Most likely I have managed to introduce the SciFi effect (it is wonderful how incompetence may result in black magic). 

Travis still failing, but this time nothing to do with seaborn, the failures are always associated with hypotthesis, and appear to follow the usual random pattern.

jjgomezcadenas:
It seems that Travis let's us go.  @jacg can we approve this?

jacg:
If we're changing the envirnoment, should we take the opportunity to bump to the latest versions of the packages?

jjgomezcadenas:
>If we're changing the envirnoment, should we take the opportunity to bump to the latest versions of the packages?

No, I wouldn't.  Bumping the packages may result in further errors which need to be investigated, as it happened that last time, and right now we are critically short of developers and overly overcomitted with analysis. We could do the bump easily during the sprint, when we will have a few people with time reserved for iC work.

jjgomezcadenas:
Incidentally, all my analysis work of this summer is heavily based in the use of @dataclasses, so I would like to bump also our Python version to 3.7

jacg:
> I would like to bump also our Python version to 3.7

You won't get any resistance from me, on this point.

jacg (approved):
This PR adds a new package, which in has no impact on the rest of the project other than slightly increasing our build time. If it is useful for analysis, let's go ahead.

jjgomezcadenas:
Shall we go ahead and bump python 3.7 straight away in this PR? If no complains, we can then outphase 3.6 by sprint...

jacg:
> Shall we go ahead and bump python 3.7 straight away in this PR? If no complains, we can then outphase 3.6 by sprint...

What do you mean? remove 3.7 from `allow_failures`? Put 3.6 in `allow_failures`? Both? Something else?

jjgomezcadenas:
@carmenromo can you please rebase?

jjgomezcadenas:
>What do you mean? remove 3.7 from allow_failures? Put 3.6 in allow_failures? Both? Something else?

I guess I mean both. We want to be sure that IC has no complains concerning 3.7 and allow the uses of new features (@dataclasses) that will result in failures in 3.6

jacg:
I don't see the point of keeping 3.6 around at all, if we put it in `allowed_failures`. The point of 3.7 in `allowed_failures` is to know when it is production-ready. Once that is the case and we remove 3.7 from `allowed_failures` I think we signal that we are now working in 3.7, and no longer care about 3.6, so I don't see the point of clogging up Travis queues with a couple of builds that don't interest us.

In any case, I think that the switch to python 3.7 should happen clearly and loudly in a separate PR, rather than being smuggled in here.

carmenromo:
Done!

jacg:
@carmenromo Small mistake. My username rather than full name appears in the human-readable part of the email on the merge commit. I'll fix it.

carmenromo:
Sorry I don't understand the mistake, what do you mean?

jacg:
The commiter's email (on the merge commit) had the form `github-user-id <email@domain.whatever>` while it should have been `Real Name <email@domain.whatever>`.

I've pushed a fix (and doing so is a pain in the backside, as it requires branch protection to be switched off temporarily, and undoing that requires the old protection settings to be enabled one by one).

Anyway, I've done it now.

PR #532, opened 2018-09-03
Set correct size to voxels

paolafer:
I've noticed that in the current code of `voxelize_hits` in `paolina_functions`, the voxels are given as size the dimensions requested by the client. However, the code actually changes slightly these dimensions, in order to fit better the hits, event by event. Therefore, the `make_track_graphs` algorithm assumes the wrong size when building the tracks.
This PR fixes this issue.  I invite @jacg to be a reviewer of this PR, since he's familiar with the code.

### DIFF
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -65,10 +65,15 @@ def voxelize_hits(hits             : Sequence[BHit],
 
     def centres(a : np.ndarray) -> np.ndarray:
         return (a[1:] + a[:-1]) / 2
+    def   sizes(a : np.ndarray) -> np.ndarray:
+        return  a[1:] - a[:-1]
 
-    cx, cy, cz = map(centres, edges)
+    (   cx,     cy,     cz) = map(centres, edges)
+    size_x, size_y, size_z  = map(sizes  , edges)
     nz = np.nonzero(E)
-    return [Voxel(cx[x], cy[y], cz[z], E[x,y,z], voxel_dimensions) for (x,y,z) in np.stack(nz).T]
+    true_dimensions = np.array([size_x[0], size_y[0], size_z[0]])
+
+    return [Voxel(cx[x], cy[y], cz[z], E[x,y,z], true_dimensions) for (x,y,z) in np.stack(nz).T]
 
 
 class Contiguity(Enum):
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -34,10 +34,9 @@ def bounding_box(seq : BHit) -> Sequence[np.ndarray]:
 def voxelize_hits(hits             : Sequence[BHit],
                   voxel_dimensions : np.ndarray,
                   strict_voxel_size: bool = False) -> List[Voxel]:
-    """1. Hits are enclosed by a bounding box.
-       2. Bounding box is discretized (via a histogramdd).
-       3. The energy of all the hits inside each discrete "voxel" is added.
-     """
+    # 1. Find bounding box of all hits.
+    # 2. Allocate hits to regular sub-boxes within bounding box, using histogramdd.
+    # 3. Calculate voxel energies by summing energies of hits within each sub-box.
     if not hits:
         raise NoHits
     hlo, hhi = bounding_box(hits)
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -35,8 +35,8 @@ def voxelize_hits(hits             : Sequence[BHit],
                   voxel_dimensions : np.ndarray,
                   strict_voxel_size: bool = False) -> List[Voxel]:
     """1. Hits are enclosed by a bounding box.
-       2. Boundix box is discretized (via a hitogramdd).
-       3. The energy of all the hits insidex each discreet "voxel" is added.
+       2. Bounding box is discretized (via a histogramdd).
+       3. The energy of all the hits inside each discrete "voxel" is added.
      """
     if not hits:
         raise NoHits
--- a/invisible_cities/reco/paolina_functions_test.py
+++ b/invisible_cities/reco/paolina_functions_test.py
@@ -160,8 +160,8 @@ def test_voxelize_hits_flexible_gives_correct_voxels_size(hits, requested_voxel_
         return np.isclose(n, round(n))
 
     for pos1, pos2 in combinations(positions, 2):
-        for i in range(3):
-            assert is_close_to_integer((pos1[i] - pos2[i]) / voxel_size[i])
+        for separation, size in zip(pos2 - pos1, voxel_size):
+            assert is_close_to_integer(separation / size)
 
 
 @given(bunch_of_hits, box_sizes)
--- a/invisible_cities/reco/paolina_functions_test.py
+++ b/invisible_cities/reco/paolina_functions_test.py
@@ -157,11 +157,11 @@ def test_voxelize_hits_flexible_gives_correct_voxels_size(hits, requested_voxel_
     voxel_size = voxels[0].size
 
     def is_close_to_integer(n):
-        return np.isclose(n, round(n))
+        return np.isclose(n, np.rint(n))
 
     for pos1, pos2 in combinations(positions, 2):
-        for separation, size in zip(pos2 - pos1, voxel_size):
-            assert is_close_to_integer(separation / size)
+        separation_over_size = (pos2 - pos1) / voxel_size
+        assert is_close_to_integer(separation_over_size).all()
 
 
 @given(bunch_of_hits, box_sizes)
--- a/invisible_cities/reco/paolina_functions_test.py
+++ b/invisible_cities/reco/paolina_functions_test.py
@@ -57,6 +57,8 @@ box_sizes = builds(np.array, lists(box_dimension,
                                    min_size = 3,
                                    max_size = 3))
 
+eps = 3e-12
+
 
 def test_voxelize_hits_should_detect_no_hits():
     with raises(NoHits):
@@ -137,7 +139,7 @@ def test_voxelize_hits_respects_voxel_dimensions(hits, requested_voxel_dimension
 def test_voxelize_hits_gives_maximum_voxels_size(hits, requested_voxel_dimensions):
     voxels = voxelize_hits(hits, requested_voxel_dimensions, strict_voxel_size=False)
     for v in voxels:
-        assert (v.size <= requested_voxel_dimensions).all()
+        assert (v.size <= requested_voxel_dimensions + 2 * eps).all()
 
 @given(bunch_of_hits, box_sizes)
 def test_voxelize_hits_strict_gives_required_voxels_size(hits, requested_voxel_dimensions):
--- a/invisible_cities/reco/paolina_functions_test.py
+++ b/invisible_cities/reco/paolina_functions_test.py
@@ -135,11 +135,13 @@ def test_voxelize_hits_respects_voxel_dimensions(hits, requested_voxel_dimension
         assert (np.isclose(off_by, 0   ) |
                 np.isclose(off_by, unit)).all()
 
+
 @given(bunch_of_hits, box_sizes)
 def test_voxelize_hits_gives_maximum_voxels_size(hits, requested_voxel_dimensions):
     voxels = voxelize_hits(hits, requested_voxel_dimensions, strict_voxel_size=False)
     for v in voxels:
         assert (v.size <= requested_voxel_dimensions + 2 * eps).all()
+
 
 @given(bunch_of_hits, box_sizes)
 def test_voxelize_hits_strict_gives_required_voxels_size(hits, requested_voxel_dimensions):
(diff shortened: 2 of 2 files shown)


invisible_cities/reco/paolina_functions.py
```diff
@@ -35,8 +35,8 @@ def voxelize_hits(hits             : Sequence[BHit],
                   voxel_dimensions : np.ndarray,
                   strict_voxel_size: bool = False) -> List[Voxel]:
     """1. Hits are enclosed by a bounding box.
-       2. Boundix box is discretized (via a hitogramdd).
-       3. The energy of all the hits insidex each discreet "voxel" is added.
+       2. Bounding box is discretized (via a histogramdd).
```
jacg: @paolafer Thanks for fixing all the typos in this docstring.

@jjgomezcadenas The purpose of docstrings is to document the behaviour of a component, in order to enable a user to understand how the component should be used: docstrings should document the interface rather than the implementation. This docstring does exactly the opposite, it describes *how* `voxelize_hits` achieves its goals, but it doesn't describe at all what those goals are or how a user is expected to use `voxelize_hits`. This shouldn't be a docstring at all, it should be a comment.

Please understand that docstrings and comments serve different purposes, and don't confuse the two. Docsrtings are for clients, comments are for implementors.

Docstrings describe to users of magic black boxes, what they can expect the black box to do for them. The whole point of magic black boxes is that they work by magic: the user isn't supposed to understand or care about what goes on inside. This is a cornerstone of good program design. Consequently docstrings shouldn't talk about what happens in the box.

jacg:
@paolafer Is there a test that points out the problem?

paolafer:
No, there's not. I'm thinking about a way to do it!

jacg:
OK, for now I've split your commit into three separate ones

1. Modify the existing test
2. Do the work in `voxelize_hits`
3. Fix the typos in the docstring.

Furthermore I've followed my own advice, and changed the docstring into a comment.

If you come up with a test, please put it in before the commit that fixes the problem.

I'll try to find some time to think carefully about what you have done sometime when there are fewer distractions.

jacg:
You are using the first dimension of `size_x` and friends to calculate `true_dimensions`, this looks weird but I guess it's OK, because, for now at least we expect all the elemens of these arrays to be equal. If you add

    assert np.allclose(size_x, size_x[0])
    assert np.allclose(size_y, size_y[0])
    assert np.allclose(size_z, size_z[0])

to `voxelize_hits` and run all tests, then the assertions are always satisfied. But it would be more correct to replace your `true_dimensions` with

    np.array([size_x[x], size_y[y], size_z[z]])

in each voxel. This comes at the cost of calculating and storing a separate array (rather than reusing a single one) for each voxel produced. I don't know whether this is worth it. Any thoughts?

paolafer:
I've added a test, before the commit to fix the bad behaviour.
As for setting a size for each voxel, I'm not sure if the rest of the code is ready to deal with voxels with different sizes, as it is now. What I mean is that if we want to generalize, we should go through the rest of the code to see if it ready to be generalized.

jacg:
@paolafer I can't find this PR on the project board. Please place it there. Put it in the *Needs review* column.

Test names reflect truths that the tests aim to verify. Choosing the name `test_voxelize_hits_flexible_does_not_give_required_voxels_size` implies that the correct behaviour is for the correct voxel size **NOT** to be given!

The magic numbers that you used in the test are too cryptic for me. Can you explain why you chose those numbers, or give any hint about how I should understand them? This sort of hint should probably appear in a comment in the test .

paolafer:
Exactly! The `test_voxelize_hits_flexible_does_not_give_required_voxels_size` test wants to test that, when `strict_voxel_size` is `False`, the voxels in the example have a different size than the requested one (and a smaller one). This was the incorrect behaviour of the code: even when we allowed a flexible voxel size, it was always the one given at the beginning, no matter what the centres of the voxels were.

paolafer:
The magic number is `eps`, right? This is the small offset we allow to the range of the histogram used to make the voxels, to include the hits in the borders. @gonzaponte suggested to use a more meaningful name, which I will change. What puzzles me is that I would expect that, using `+ 2 * eps` would be enough to make the test pass, since it is exactly the amount we add in `voxelize_hits`. However, today the test has decided not to pass. I've found that using `+ 4 * eps` makes the trick, but I don't know why. Anyway, we don't really need to assure that inequality, we just want to be sure that the voxels have not a size significantly bigger than the requested one.

jacg:
> The magic number is `eps`, right?

Nope! `eps` has a name, so it's not magic! I mean all these

    [BHit(72.32, 117.04, 317.16, energy),
     BHit(61.78, 143.76, 317.16, energy),
     BHit(66.53, 122.86, 319.02, energy),
     BHit(83.90, 114.32, 319.02, energy)]

(BTW, I'm not entirely happy about making `eps` global in `paolina_functions` and importing it. I understand and appreciate your motivation, but it smells wrong.)

Test names reflect something that you want to be true in your code. Think of it as meaning "the code is correct when *test name*": By calling the test `test_voxelize_hits_flexible_does_not_give_required_voxels_size` you are saying "the code behaves correctly when it does NOT give the required size" which, while strictly speaking correct from your perspective, is a very confusing statement for someone who is reading it for the first time. So I would look for a name that conveys the sentiment 'flexible mode gives voxel sizes which do not exceed the required one'.

Your test contains a strict inequality, which I think is wrong. Flexible mode gives you the freedom (but not the obligation) to diverge from the requested size. Hmm, but that's a general statement, while this test uses very specific values which (presumably) force the inequality ... OK. That makes sense. But now I'd like to know why you picked those magic numbers. I would still like to come up with a better test, but this will have to do if I don't find the time and/or inspiration soon enough.

BTW, I keep getting the feeling that we have already fixed this very problem and written this test months ago, but I can't find a PR that does it. There is #386, but we don't seem to have done it there.

jacg:
Hmm `RANGE_EPS_INCREASE` doesn't really speak to me.

How about something like `BOUNDING_BOX_MARGIN`?

paolafer:
I passed through the same steps as you did :-) (I feel encouraged by this!). 
At the beginning I didn't want to use a specific example, but I found difficult to test the fact that actual voxel sizes should not be **always** equal to the required sizes (but sometimes they could be). So, I decided to use an example that I know for sure that produces actual voxel sizes different than (15., 15., 15.), when (15., 15., 15.) is given as required size. How did I choose these particular hits? Well, they are the first bunch of hits of the specific Monte Carlo event that made me dig into this stuff during the weekend and brought me to find this misbehaviour... I know, it's not very elegant, but I haven't come out with a workaround so far...

paolafer:
And yes, I had also the strong feeling that we fixed this... but maybe we just fixed part of the issue and didn't realize that the actual construction of the `Voxel` objects was using the requested dimensions.

paolafer:
What is the wrong smell you see in using `eps` as a global variable?

paolafer:
Ok, I think I found a better test. What we want to do to is to ensure that the centres of the voxels are separated by exactly the size of the voxel. This doesn't happen in the current code, since the centres are built with a flexible size, but the voxel is given the required one. I'll do it asap.

jacg:
Nice. Far simpler than the ideas I has fermenting in the back of my mind.

Except that, dealing with floats, it's not going to be 'exactly' the same, but I'm sure you know that already.

jacg:
Coming back to the topic of your choice of random-looking numbers (which will be a non-issue once you've implemented your last idea, but it's a point for the future), I would argue that you should pick less random looking numbers, preferably with some obvious feature, so that the human who tries to understand the example has some chance of spotting the feature. For example, the `voxel_spec` fixture in that same file has a very obvious structure, and a determined human shouldn't have too much trouble noticing that adjacent items are separated by one unit in either 1, 2 or 3 dimensions, and that the energy levels are all powers of 2, except for the two massive ones at the ends. These obvious features give the human something to grab on to, rather than being lost in the apparent randomness. The random numbers give me no easy path towards understanding *why* they exhibit the required feature.

This is one of the amazing and useful things about hypothesis: it first finds a random counterexample, but before presenting it to the human, it tries to simplify the example and usually ends up presenting the human with something that highlights the essence of the problem.

jacg:
> What is the wrong smell you see in using `eps` as a global variable?

It's an implementation detail of `voxelize_hits`. As such it has no business being global, let alone being imported and used outside the whole module. 

I understand that you want it global so that you can import it, and you want to import it so that you can use the same value in the test, but that smells: you are testing a black box, and having the test depend on an *implementation detail* means that the test is not testing an objective truth.

Imagine doing this with TDD. First you think about the desired behaviour, and describe it in a test. Only later do you write the code that passes the test, and there are many different implementations that could work. It makes no sense to define that test in terms of some detail that is internal to one of the implementations that hasn't been written yet.

Maybe, just maybe, this is a problem which fundamentally depends on some tolerance parameter (in which case parametrizing both the test and the black box with that parameter is probably better than making it global), but until I am convinced that the desired behaviour cannot be expressed without such a parameter, it will continue to smell.

paolafer:
I understand what you say and, for the moment, I just kept the variable local, and define another variable in the test function. In fact, the two variable were not the same anyway, since, for some reason, twice the value of the variable was not enough for the test to pass. Therefore, the value of that variable must be fine tuned anyway...
If you are reasonably satisfied with this, and the code and the test are ok, could you approve this PR? Thanks!

jacg:
I think it's important that a test pointing out a problem should be committed in official history *before* the solution. Preferably in the commit that immediately precedes the commit that contains the solution. That way it is easy to check out the test-without-solution, and observe how it fails; then check out the solution, and observe that the very same test now passes.

So I have reordeded and squashed the commits to make it so, and also cleaned up some other bits of trial-and-error noise in the history.

I have also renamed `is_integer` to `is_close_to_integer` which seems to be a more correct description of what it does.

If I understand your test correctly, it states that any two voxels must be separated by a (close to) integer multiple of voxel sizes, in any dimension. That has a certain pleasing feel to it.

I have added two refactorings of the test (between your introduction of the test and the solution). See what you think of them.

Maybe the test should be called something like `test_voxelize_hits_flexible_voxel_separation_is_multiple_of_size`?

I was about to comment on what-used-to-be-called-`eps` being global, but I see that you have just pushed something possibly relating to that ...

jacg:
Very quickly, you seem to have reverted to copy'n'paste, which is even worse! But isn't this `eps` business orthogonal to the mani point of this PR? If so, I suggest we leave it out of here.

paolafer:
It is actually not a copy and paste, since the value in the paolina function and in the test are not the same... They are both fine tuned.
For some reason I cannot see your changes to the code... Which numbers are your commits?

jacg:
I have now pushed the changes I mentioned a few minutes ago. This has made your last (`eps`-related) push disappear. I've kept it in a local branch. You'll be able to find it in your reflog, if necessary.

Can we wrap up this PR without the `eps` business?

paolafer:
Yes and no. We need to use a margin in the `test_voxelize_hits_gives_maximum_voxels_size` test, otherwise it won't succeed. It wasn't failing before, because the voxel size was always equal to the required dimensions by construction.

jacg:
Yes, OK, I can reproduce the faliure if I use the `hard` hypothesis profile. Lemme think about it ...

In the meantime, please have a look at my rephrasing of your test.

paolafer:
I have made a fetch of this branch, and rebase my local branch on top of that. The problem is that my local branch has 5 commits more than this one and I am not sure if I have to discard them, because you already pushed the same changes or not.
I attach a screenshot of my magit to show this.
screen shot 2018-09-07 at 11 06 07

jacg:
Don't rebase anything! Remember, rebasing branch A on top of branch B is an alternative way of merging branch B into branch A. We don't want to merge what I did into what you did, because they are two alternative approaches to closing off this PR. So just look at my commits. Check them out and run the tests. But don't merge anything.

If you want to add to or modify what I did, you'll have to put a local branch of your own on it, and work on that branch. Then you can push that local branch to this PR's branch on your remote.

There are two qualitatively distinct ways you could do this:

1. Use your `set-correct-vxl-size` branch. In this case your version of the branch will disappear from view. You'll be able to get it back by looking in the reflog. Alternatively you could make recovery easier, by placing a new local branch on top of it, before moving it to my version. You might call your new branch something like `set-correct-vxl-size-paola`.
2. Leave your branch where it is, and create a new one for fiddling with my version. You might call that branch `set-correct-vxl-size-jacg`.

Remember that you can push either of these branches to `origin/set-correct-vxl-size` (or anywhere else, for that matter).

jacg:
This is the hypothesis counterexample I see, when I remove the tolerance:


    hits = [BHit(10.0, 10.0, 10.0, E=10.0), BHit(10.100000000000016, 10.100000000000016, 54.99999999999405, E=10.0)], requested_voxel_dimensions = array([1., 1., 1.])
    
        @given(bunch_of_hits, box_sizes)
        def test_voxelize_hits_gives_maximum_voxels_size(hits, requested_voxel_dimensions):
            voxels = voxelize_hits(hits, requested_voxel_dimensions, strict_voxel_size=False)
            for v in voxels:
    >           assert (v.size <= requested_voxel_dimensions).all()
    E           assert False
    E            +  where False = <built-in method all of numpy.ndarray object at 0x7f0753ac2a30>()
    E            +    where <built-in method all of numpy.ndarray object at 0x7f0753ac2a30> = array([0.1, 0.1, 1. ]) <= array([1., 1., 1.]).all
    E            +      where array([0.1, 0.1, 1. ]) = Voxel(10.050000000000008, 10.050000000000008, 10.499999999997005, E=10.0).size

Adding 6e-12 to `requested_voxel_dimensions` in the assertion, makes the test pass.

I'm struggling to understand what this means, and I have about 5 minutes left to think about it for now. I do not see why adding this tolerance here would be the correct thing to do (or how it makes the test pass in this case, for that matter).

paolafer:
Yes, I agree with you.
As for your rephrasing, I'm fine with your changes.

paolafer:
@jacg, could we leave the problem of the tolerance for another PR and approve this one? We could use a value that works and move on.

jacg:
> could we leave the problem of the tolerance for another PR and approve this one?

Yes, that was exactly what I was proposing.

But for that we need to get the tests to pass. Please pick a value that works, and remove all other  commits that fiddle with the tolerance. Then open an issue that records the fact that the tolerance needs to be understood, and place in on the project board in the Todo column. Let me know when you're done, so that I can give GitHub approval to merge.

I've got two weeks of courses, so expect looong delays in my replies.

paolafer:
Done.

jacg (approved):
A clear mistake in the flexible mode of `voxelize_hits` has been identified. A test demonstrating the problem has been added. The problem has been fixed and the test passes.

PR #535, opened 2018-09-13
(WIP) Progress on LFS

jmbenlloch:
I have made some tests with git lfs and it seems to be working well. Below you have a more detailed report, most important thing: git lfs does **NOT** support ssh authentication.

The branch included in this PR has already everything setup. The server url included in this test is not the final one, we are updating the gitlab server in our main server, which will be the one used for this. Should be ready by next week. 

When we have the final server working I will run the tests again with different machines (at least a linux and a  Mac) and if everything is fine we can make the transition.

Server configuration
--------------------

We can use a gitlab server as an external LFS server. To do that, we need to create a repository that will hold the files, for example: https://nextdesk1.ific.uv.es:4433/jmbenlloch/IC. There are 3 requisites for this repository:

- Gitlab must have LFS enabled (https://docs.gitlab.com/ee/workflow/lfs/lfs_administration.html)
- The repository must have public visibility if we want people to be able to checkout anonymously (this is also useful for travis).
- Every developer will need an account on this server with writing permissions for this repository to push commits (besides the their github account).

Client configuration
--------------------

To configure LFS for the repository every user must have git lfs installed on their machines. On debian-like linux can be done with:

    curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | sudo bash
    sudo apt-get install git-lfs

For Mac OS X it is possible to install it using brew:

	brew install git-lfs

For other systems or alternatives downloads check this: https://git-lfs.github.com/

For git lfs to work there are some hooks that must be setup by running the following command on your repository dir:

    git lfs install

If you clone a repositoy which is already using lfs I don't think you need to run that command, but we can do more tests on that.

Setting up the external server
------------------------------
We need to add a `.lfsconfig` file with the address of our external lfs server:

```
[lfs]
    url = https://nextdesk1.ific.uv.es:4433/jmbenlloch/IC.git/info/lfs
```

Issues:

- So far we are using a self-signed certificate for the gitlab server, so unless we fix that (maybe https://letsencrypt.org/) we need to add this to `.bashrc`:

	`export GIT_SSL_NO_VERIFY=1`

- Git LFS does **NOT** support ssh yet (https://github.com/git-lfs/git-lfs/issues/1044). So developers will need to input their gitlab's user&pass after every commit.

Tracking files
--------------

To track files the command is:

	git lfs track '*.h5'

that will add a new line to `.gitattributes` like `*.h5 filter=lfs diff=lfs merge=lfs -text`

Migration
---------

In principle it is possible to migrate all the binary files in an existing repository using a tool provided by the git-lfs client:

	git lfs migrate import --include="*.h5" --include-ref=refs/heads/master

That will migrate all the h5 files from the master branch history and will rewrite all the commits. There is a caveat, though: since we are using an external server for lfs, the file .lfsconfig is needed to checkout the files and that file won't be present in older commits, so in principle it wouldn't work unless we put that file somehow in the past commits.

In any case, we could use LFS for the new files and keep the old ones in the repository if there is no simple solution for this issue.

You can find an already-migrated branch here: https://github.com/jmbenlloch/IC-1/tree/lfs-migrate

Travis is working well with that branch as you can see here: https://travis-ci.org/jmbenlloch/IC-1/jobs/425180907

Checking out a branch with lfs for the first time
-------------------------------------------------
The first time a user checkouts a branch using lfs, errors will appear becuase lfs won't be configured yet until he gets that branch.

The procedure to avoid that error is the following:

```
export GIT_LFS_SKIP_SMUDGE=1
export GIT_SSL_NO_VERIFY=1
checkout the branch (either with git or magit)
git lfs pull
```


jacg:
Good work!

> developers will need to input their gitlab's user&pass after every commit.

Urgh!

> That will migrate all the h5 files from the master branch history and will rewrite all the commits. There is a caveat, though: since we are using an external server for lfs, the file .lfsconfig is needed to checkout the files and that file won't be present in older commits, so in principle it wouldn't work unless we put that file somehow in the past commits.

If we are going to rewrite history on the master branch anyway, then injecting that file deep in the history isn't going to cause any significant extra hassle.

jacg:
Also, before we go ahead and use our own wheel, we should make sure that there isn't a perfectly usable wheel already available. Have we given enough thought to whether we could solve this issue without taking on the burden of managing our own LFS server, simply by migrating to github.com?

jmbenlloch:
> If we are going to rewrite history on the master branch anyway, then injecting that file deep in the history isn't going to cause any significant extra hassle.

How would you do that? Is there a command to do so? Or we should create a new commit including the LFS config an rebase the whole master history on top of that commit?

> Have we given enough thought to whether we could solve this issue without taking on the burden of managing our own LFS server, simply by migrating to github.com?

I think we have already discussed that in the past. The price scheme has not changed since then. So unless someone has any good reason I don't think is a good idea to commit ourselves to an option where the cost can increase easily as we start to add more and more files.

jacg:
> The price scheme has not changed since then.

Sorry, my mistake, I meant to write git**LAB** not git**HUB**. (It's a mistake I keep making both in writing and in speech.)

I don't think we ever looked into the LFS pricing scheme on git**LAB**

jacg:
> How would you do that? Is there a command to do so? Or we should create a new commit including the LFS config an rebase the whole master history on top of that commit?

Hmm, the numerous merge commits we have might make it a pain to do with `rebase -i`. I'm sure a suitable spell can be cast with `filter-branch`, but I wouldn't know what it is off the top of my head.

jmbenlloch:
> I don't think we ever looked into the LFS pricing scheme on gitLAB

I have been looking around in gitlab.com's documentation and I think there is only a limit in the total repository size of 10GB (https://docs.gitlab.com/ee/user/gitlab_com/#repository-size-limit). There is no mention to any bandwith limit. In [this issue](https://gitlab.com/gitlab-com/www-gitlab-com/issues/1003#note_19874603) from a year ago they mention that as long as the traffic is "acceptable" there won't be any problem.

You can see their pricing scheme here: https://about.gitlab.com/pricing/ LFS is included in all options, but there are some important features missing in the free version, for example [merge request approvals](https://docs.gitlab.com/ee/user/project/merge_requests/merge_request_approvals.html). Without that we can not set any restrictions for PR approvals. The cheapest option including that is $4/month/user, so I still think using our own server is the best option for us.

jacg:
I have a proof of concept on my [insert-file-into-history](https://github.com/jacg/IC/tree/insert-file-into-history) branch. The file `INSERT-FILE-DEEP-INTO-HISTORY.txt` should appear throughout the history of that branch. The branch was made by running

    git filter-branch --index-filter "cp ~/INSERT-FILE-DEEP-INTO-HISTORY.txt . && git add INSERT-FILE-DEEP-INTO-HISTORY.txt" -- --all

in my local clone of IC, and then pushing `master` to `origin/insert-file-into-history`.

jmbenlloch:
I think there is another collateral damage if we modify the entire history of the current master to include the new file (`.lfsconfig`) and to remove the binary files. After that operation all commits' hashes will be different. As an example compare this two:

https://github.com/jacg/IC/commit/d26c5defcd210c3f0c0eee235dc7514ce377e634
https://github.com/nextic/IC/commit/0baf5371dec90e98aad08d722a280c918663e44d

They are the same commit but with different hashes due to the updates. This can potentially (and probably, I think) break our PRs history since commits from PRs won't have the same hashes as those in the master. When github finds a match between those it mark the PR as merged, if all those commits disappear, probably those PRs will appear as closed.

jacg:
> I think there is another collateral damage if we modify the entire history of the current master to include the new file (.lfsconfig) and to remove the binary files. After that operation all commits' hashes will be different.

I'm sorry, I thought that this was so obvious that it didn't need mentioning explicitly.

Yes, **any** rewriting of history necessarily changes all the hashes of all modified commits.

 > This can potentially (and probably, I think) break our PRs history

We have already rewritten history (with concomittant hash changes) when I regularized email addresses a while ago. So check any PRs which were opened/closed/merged before that was done to get an idea of how PRs will be affected by this change.

Whatever that cost is, I'm sure it will be outweighed by the benefits of having LFS.

PR #537, opened 2018-09-25
Fix redundancy in load_db file

carmenromo:
This PR removes dbfile parameter because it was an useless variable.


jmbenlloch (approved):
Simple change that will keep our code a little bit cleaner. Approved!

PR #540, opened 2018-10-08
Seeds and bounds for calibration fits

carmenromo:
This PR adds several functions to provide seeds and bounds for sensor calibration fits. For gain and sigma it provides either the option to take their values from previous runs in the database, or it looks for the pedestal and 1 pe peaks. Corresponding tests are also added.


invisible_cities/calib/calib_functions_test.py
```diff
@@ -168,4 +169,40 @@ def test_copy_sensor_table3(config_tmpdir):
...
+
+    min_stat = 10
+
+    for ich, (led, dar) in enumerate(zip(specsL, specsD)):
+        valid_bins = np.argwhere(led>=min_stat)
+        b1 = valid_bins[0][0]
+        b2 = valid_bins[-1][0]
+        pD = find_peaks_cwt(dar, np.arange(2, 20), min_snr=2)
+        if len(pD) == 0:
+            print('no peaks in dark spectrum, spec ', ich)
```
jacg: The `print`should be removed.

invisible_cities/calib/calib_functions.py
```diff
@@ -136,9 +141,78 @@ def scaled_spectrum(x, mu):
     return scaled_spectrum
 
 
+def seeds_db(sensor_type, run_no, n_chann):
+    """
+    Take gain and sigma values of previous runs in the database
+    to use them as seeds.
+    """
+    if sensor_type == 'sipm':
```
jacg: You should relally use some kind of symbol, rather than a string, to denote values belonging to a limited set of possibilitiies. Perhaps an `Enum`.

carmenromo: What do you mean by a kind of symbol?

jacg: Strings are types with an almost infinite number of values. By accepting strings you are implying that any of these might be OK. In this context, only a very limited set of values is acceptable, so accept a type that has only that limited set.

That way, there is automatic documentation of what the allowed values are, inadvertent misspelling is more difficult, static tools can more easily help you out, etc., etc..

Furthermore, strings usually need to be compared by looking through all their elements, one by one, until a mismatch is found, or the end is reached. Symbols, can be compared by identity, so comparison is faster.

I know that it is widespread practice in Python to accept strings where symbols would be *The Right Thing*, but that doesn't mean we should perpetuate bad practices. This bad practice was understandable (and maybe even excusable) when Python didn't have anything resembling a symbol type that was easy to use, but now that we have `enum` there really is no excuse.

gonzaponte:
Before the review; Travis complains about the following things:
- `test_poisson_mu_seed`: `NameError: name 'dark_scaler'
- `test_sensor_values`: `NameError: name 'dark_scaler'`
- `test_pedestal_values` :`NameError: name 'pedestal_values'`
- `test_seeds_db`: `NameError: name 'seeds_db' is not defined`
- `test_seeds_and_bounds`: Some error I cannot still discern.

Did the tests pass in your computer, @carmenromo?

PR #541, opened 2018-10-10
Remove annoying print that adds noise to dorothea's output

gonzaponte:
A leftover `print` call introduces a lot of noise to dorothea's output. Remove it.


carmenromo (approved):
Easy!

PR #543, opened 2018-10-18
Test all cities with an exact result

gonzaponte:
This is a LFS version of #465


jmbenlloch (approved):
This is a very important PR that will check the result of every city, it is specially useful if we want to change how we implement cities, like we are doing in #548.

PR #544, opened 2018-10-18
Add git lfs to Readme.rst

jmbenlloch:
Update `Readme.rst` to include a first step about `git lfs`installation, which is needed now.


jacg (approved):
A crucial step of the installation, that has become necessarry as a consequence of our use of Git LFS.

PR #545, opened 2018-10-18
Drop support for Python 3.6

jmbenlloch:
Drop support for Python 3.6


jacg (approved):
Travis has been testing Python 3.7 for a few months now, with no problems. High time to leave Python 3.6 behind and start taking advantage of dataclasses.

gonzaponte:
For some reason, this PR is not tagged as merged, but it is merged. Probably caused by an lfs-related workaround.

PR #547, opened 2018-10-18
Adds fix for Mac in init.el

jerenner:


gonzaponte:
What does it do exactly?

jerenner:
It allowed access of git lfs from within Magit which was not working previously on my (Mac) machine

jacg:
> What does it do exactly?

Window-based applications on the Mac do not inherit the environment from any shell. This includes Emacs, when it is launched by clicking on an icon. This PR extents the `docs/init.el` so that itinstalls the emacs package `exec-path-from-shell` which ensures that Emacs uses the `PATH` specification that you configure in your shell resource files.

This has the efffect of fixing all sorts of things that should work in Emacs but are broken on Macs, among them `git lfs`. Specifically, it enables you to use magit with the LFS-ized version of IC.

jacg (approved):
Makes Emacs work more sensibly for Mac users.

As it's not IC Python code, no tests are needed or sensibly possible.

PR #548, opened 2018-10-18
Implement Liquid cities

gonzaponte:
This is a LFS version of #435


jacg:
When can we expect a fix for the failures? The typos in the names are easily fixed, but that only clears the way to a failure to compare a `minmax` with an `int` using `<`.

gonzaponte:
The typos are already fixed. The other failure may need a bit of discussion. As soon as I arrive to the office.

jacg:
Apparently it's unclear whether the office is open, or whether public transport is working. We seem to be staying here.

In the meantime, I could really do with a liquid cities branch which does not contain any test failures, by yesterday, in order to make any progress on getting rid of the infinite noise that comes out of our test suite at the moment. Today is likely to be the last scrap of time I have to work on IC for a loong time (perhaps for ever).

paolafer:
We're all here!!! Don't leave us alone!

jacg:
This is miserably and persistently failing on OS X !

jacg:
I'm trying to replace solid cities with liquid ones, and eradicate all trace of the former
(in a PR that will follow once this one is merged). I get test failures because the (now defunct) solid
cities are still imported in

1. `invisible_cities/calib`
2. `invisible_cities/detsim`

   I'm guessing that both of these groups have been reincarnated under
   new city names in the `liquid_cities`.

   Can I just delete these?

3. `config_test.py` creates a `class DummyCity(City)`

   Can I assume that the equivalent test exists somewhere and that this
   can be deleted?

jacg:
I get the impression that LFS is not properly configured on our Travis OS X, but given that OS X is an allowed failure, we've never noticed it. Every post-LFS Travis build has failed on OS X, I think, in a way that seems consistent with having LFS pointers rather than the files of interest. As it does seem to be working on our personal Macs, I propose that we go ahead, and deal with the Travis problem later.

jacg:
@jjgomezcadenas I propose that you give us a suitably poetic formal approval on this momentous occasion, and that we go ahead and merge this.

jjgomezcadenas (approved):
After a titanic struggle, and at the very long last, a small step forward for the software community but a   huge leap for us. Liquid cities, encouraging functional programming and implementing a flow paradigm that favour modularity and avoids code duplication, have passed with flying colours all our tests and replace thus dumb, repetitive and inefficient solidi cities. The King is dead, long live the King!

gonzaponte:
image

PR #549, opened 2018-10-19
New wrapper to fix parameters for fitting

andLaing:
Adds a wrapper so that functions can have
parameters fixed to known values and
still work as expected with scipy.curve_fit.

Slightly updated from old PR to clarify some
arguments.

This is an LFS version of #512.

### DIFF
--- a/invisible_cities/core/fit_functions.py
+++ b/invisible_cities/core/fit_functions.py
@@ -24,12 +24,14 @@ def fixed_parameters(fn, **kwargs):
     """
     fn_sig    = insp.signature(fn)
     fn_pars   = np.array(list(fn_sig.parameters))[1:]
-    par_gen   = (kwargs[k] if k in kwargs.keys() else np.nan for k in fn_pars)
+    par_gen   = (kwargs.get(k, np.nan) for k in fn_pars)
     all_args  = np.fromiter(par_gen, np.float)
     free_pars = np.isnan(all_args)
     
     if np.all(free_pars):
-        raise ValueError(kwargs.keys() + " not parameters of " + fn.__name__)
+        raise ValueError(str(kwargs.keys()) + " not parameters " + fn.__name__)
+    elif not np.any(free_pars):
+        raise ValueError("Fixing all parameters pointless")
     elif np.count_nonzero(free_pars) > len(fn_pars) - len(kwargs):
         raise ValueError("Some parameters not found in " + fn.__name__)
     
@@ -38,9 +40,8 @@ def fixed_parameters(fn, **kwargs):
         all_args[free_pars] = pars
         func_value = fn(x, *all_args)
         ## hack to maintain spe_functions underlying decorator
-        try:
+        if hasattr(fn, 'n_gaussians'):
             fixed_fn.n_gaussians = fn.n_gaussians
-        except AttributeError: pass
         return func_value
     
     ## Correct the doc string and signature for the new wrapped function
--- a/invisible_cities/core/fit_functions.py
+++ b/invisible_cities/core/fit_functions.py
@@ -4,14 +4,53 @@ A set of functions for data fitting.
 GML November 2016
 """
 
-import numpy  as np
-import pandas as pd
+import numpy   as np
+import pandas  as pd
+import inspect as insp
 import scipy.optimize
 import scipy.stats
+
+from functools import wraps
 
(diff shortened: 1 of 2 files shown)


jjgomezcadenas:
@andLaing This needs to be resolved, it works on old solid cities

invisible_cities/core/fit_functions.py
```diff
@@ -4,16 +4,55 @@
...
 
 
+def fixed_parameters(fn, **kwargs):
+    """
+    Wrapper to fix parameters of a function
+    which can then be used in curve_fit.
+    """
+    fn_sig    = insp.signature(fn)
+    fn_pars   = np.array(list(fn_sig.parameters))[1:]
+    par_gen   = (kwargs[k] if k in kwargs.keys() else np.nan for k in fn_pars)
```
gonzaponte: `kwargs` seems to behave like a dictionary, so you can also do `kwargs.get(k , np.nan)` instead of `kwargs[k] if k in kwargs.keys() else np.nan`

invisible_cities/core/fit_functions.py
```diff
@@ -4,16 +4,55 @@
...
+        raise ValueError("Some parameters not found in " + fn.__name__)
+    
+    @wraps(fn)
+    def fixed_fn(x, *pars):
+        all_args[free_pars] = pars
+        func_value = fn(x, *all_args)
+        ## hack to maintain spe_functions underlying decorator
+        try:
+            fixed_fn.n_gaussians = fn.n_gaussians
+        except AttributeError: pass
```
gonzaponte: I am not sure whether it is better
```
try:
    do_something_with_something_that_may_have_an_attribute
except AttributeError:
    pass
```
or
```
if hasattr(something_that_may_have_an_attribute, attribute):
    do_something
```

In this case I don't think it makes any difference, but it could in more complex cases, so I would go for the latter.

invisible_cities/core/fit_functions_test.py
```diff
@@ -289,6 +289,18 @@ def test_fit_with_errors(reduced):
...
+    x = np.arange(10.)
+    y = fitf.gauss(x, *pars)
+    e = 0.1 * y
+
+    fixed_mu = fitf.fixed_parameters(fitf.gauss, mu = pars[1])
+    seeds = np.array(pars[::2])
+    f = fitf.fit(fixed_mu, x, y, seeds * 1.2, sigma=e)
+    assert_allclose(f.values, seeds)
+
+
```
gonzaponte, line 315: Can I ask for a couple of simple tests that check that it explodes in the appropriate cases? Namely:
- Try to fix all parameters and check that raises an exception (the exact one you expect)
- Try to keep all parameters free, it should also raise an exception
- Try to send unknown parameters, it should raise something as well

Also, you can add a test previous to `test_fixed_parameters` that does some basic checks like ensuring that the number of parameters in the returned function is correct for a number of cases (`pytest.mark.parametrize` is useful here).

jjgomezcadenas:
@gonzaponte are u happy with this? Can we approve?

invisible_cities/core/fit_functions_test.py
gonzaponte: > Try to fix all parameters and check that raises an exception (the exact one you expect)

I guess this is the first case in `test_fix_wrong_parameters_raises_error`. Can you make this a separate test? It deserves it.

> Try to keep all parameters free, it should also raise an exception

This is not implemented, right?

These comments aside, I think the PR is ready.

andLaing:
Any chance you'll manage to take a shifty any time soon, @gonzaponte ?

It's not the most urgent PR but it's been there for a while now.

gonzaponte:
> >Try to fix all parameters and check that raises an exception (the exact one you expect)

> I guess this is the first case in test_fix_wrong_parameters_raises_error. Can you make this a separate test? It deserves it.

> >Try to keep all parameters free, it should also raise an exception

> This is not implemented, right?
> These comments aside, I think the PR is ready.

My last comments have not been addressed. They are not of the utmost importance but it would be nice to have them. I will approve it anyway and it is left for you to decide.

gonzaponte (approved):
Useful wrapper around the fitting function to fix certain parameters in a function. Bunch of tests added as well.

andLaing:
I thought I had implemented those tests, I'll check and update as appropriate

PR #550, opened 2018-10-19
Seeds and bounds for calibration fits

carmenromo:
This PR adds several functions to provide seeds and bounds for sensor calibration fits. For gain and sigma it provides either the option to take their values from previous runs in the database, or it looks for the pedestal and 1 pe peaks. Corresponding tests are also added.

This is an LFS version of #540.

### DIFF
--- a/invisible_cities/evm/ic_containers.py
+++ b/invisible_cities/evm/ic_containers.py
@@ -42,7 +42,7 @@ for name, attrs in (
         ('TriggerParams'  , 'trigger_channels min_number_channels charge height width'),
         ('PeakData'       , 'charge height width'),
         ('Measurement'    , 'value uncertainty'),
-        ('SensorParams'   , 'spectra, peak_range, min_bin_peak, max_bin_peak, half_peak_width, p1pe_seed, lim_ped'),
+        ('SensorParams'   , 'spectra peak_range min_bin_peak max_bin_peak half_peak_width p1pe_seed lim_ped'),
         ('PedestalParams' , 'gain gain_min gain_max sigma sigma_min sigma_max')):
     _add_namedtuple_in_this_module(name, attrs)
 
--- a/invisible_cities/evm/ic_containers.py
+++ b/invisible_cities/evm/ic_containers.py
@@ -42,7 +42,7 @@ for name, attrs in (
         ('TriggerParams'  , 'trigger_channels min_number_channels charge height width'),
         ('PeakData'       , 'charge height width'),
         ('Measurement'    , 'value uncertainty'),
-        ('SensorParams'   , 'spectra, peak_range, min_bin_peak, max_bin_peak, half_peak_width, lim_ped'),
+        ('SensorParams'   , 'spectra, peak_range, min_bin_peak, max_bin_peak, half_peak_width, p1pe_seed, lim_ped'),
         ('PedestalParams' , 'gain gain_min gain_max sigma sigma_min sigma_max')):
     _add_namedtuple_in_this_module(name, attrs)
 
--- a/invisible_cities/evm/ic_containers.py
+++ b/invisible_cities/evm/ic_containers.py
@@ -41,7 +41,9 @@ for name, attrs in (
         ('FitFunction'    , 'fn values errors chi2 pvalue cov'),
         ('TriggerParams'  , 'trigger_channels min_number_channels charge height width'),
         ('PeakData'       , 'charge height width'),
-        ('Measurement'    , 'value uncertainty')):
+        ('Measurement'    , 'value uncertainty'),
+        ('SensorParams'   , 'spectra, peak_range, min_bin_peak, max_bin_peak, half_peak_width, lim_ped'),
+        ('PedestalParams' , 'gain gain_min gain_max sigma sigma_min sigma_max')):
     _add_namedtuple_in_this_module(name, attrs)
 
 # Leave nothing but the namedtuple types in the namespace of this module
--- a/invisible_cities/reco/calib_functions.py
+++ b/invisible_cities/reco/calib_functions.py
@@ -13,6 +13,8 @@ from .. core.stat_functions    import poisson_sigma
 from .. core                   import fit_functions    as fitf
 from .. database               import load_db          as DB
 from .. types.ic_types         import AutoNameEnumBase
+from .. evm.ic_containers      import SensorParams
+from .. evm.ic_containers      import PedestalParams
 
 
 def bin_waveforms(waveforms, bins):
@@ -151,6 +153,7 @@ class SensorType(AutoNameEnumBase):
     SIPM = auto()
     PMT  = auto()
 
+
 def seeds_db(sensor_type, run_no, n_chann):
     """
     Take gain and sigma values of previous runs in the database
@@ -204,7 +207,7 @@ def sensor_values(sensor_type, n_chann, scaler, spec, bins, ped_vals):
         max_bin_peak    = 50
         half_peak_width = 10
         lim_ped         = 10000
-    return spectra, peak_range, min_bin_peak, max_bin_peak, half_peak_width, lim_ped
+    return SensorParams(spectra, peak_range, min_bin_peak, max_bin_peak, half_peak_width, lim_ped)
 
 
 def pedestal_values(ped_vals, lim_ped, ped_errs):
@@ -218,7 +221,7 @@ def pedestal_values(ped_vals, lim_ped, ped_errs):
     ped_sig_min  = max(0.001, ped_sig_seed - lim_ped * ped_errs[2])
     ped_sig_max  = ped_sig_seed + lim_ped * ped_errs[2]
 
-    return ped_seed, ped_sig_seed, ped_min, ped_max, ped_sig_min, ped_sig_max
+    return PedestalParams(ped_seed, ped_min, ped_max, ped_sig_seed, ped_sig_min, ped_sig_max)
 
 
 def seeds_and_bounds(sensor_type, run_no, n_chann, scaler, bins, spec, ped_vals,
@@ -260,25 +263,25 @@ def seeds_and_bounds(sensor_type, run_no, n_chann, scaler, bins, spec, ped_vals,
 
     norm_seed = spec.sum()
     gain_seed, gain_sigma_seed = seeds_db(sensor_type, run_no, n_chann)
-    spectra, p_range, min_b, max_b, hpw, lim_ped = sensor_values(sensor_type, n_chann,
-                                                                 scaler, spec, bins, ped_vals)
+    sens_values = sensor_values(sensor_type, n_chann, scaler, spec, bins, ped_vals)
 
     if not use_db_gain_seeds:
-        pDL  = find_peaks_cwt(spectra, p_range, min_snr=1, noise_perc=5)
-        p1pe = pDL[(bins[pDL]>min_b) & (bins[pDL]<max_b)]
+        pDL  = find_peaks_cwt(sens_values.spectra, sens_values.p_range, min_snr=1, noise_perc=5)
+        p1pe = pDL[(bins[pDL]>sens_values.min_b) & (bins[pDL]<sens_values.max_b)]
         if len(p1pe) == 0:
             try:
-                p1pe = np.argwhere(bins==(min_b+max_b)/2)[0][0]
+                p1pe = np.argwhere(bins==(sens_values.min_b+sens_values.max_b)/2)[0][0]
             except IndexError:
                 p1pe = len(bins)-1
         else:
-            p1pe = p1pe[spectra[p1pe].argmax()]
+            p1pe = p1pe[sens_values.spectra[p1pe].argmax()]
 
-        fgaus = fitf.fit(fitf.gauss, bins[p1pe-hpw:p1pe+hpw],
-                        spectra[p1pe-hpw:p1pe+hpw],
-                        seed=(spectra[p1pe], bins[p1pe], 7),
-                        sigma=np.sqrt(spectra[p1pe-hpw:p1pe+hpw]),
-                        bounds=[(0, -100, 0), (1e99, 100, 10000)])
+        fgaus = fitf.fit(fitf.gauss,
+                         bins[p1pe-sens_values.hpw:p1pe+sens_values.hpw],
+                         sens_values.spectra[p1pe-sens_values.hpw:p1pe+sens_values.hpw],
+                         seed   = (sens_values.spectra[p1pe], bins[p1pe], 7),
+                         sigma  = np.sqrt(sens_values.spectra[p1pe-sens_values.hpw:p1pe+sens_values.hpw]),
+                         bounds = [(0, -100, 0), (1e99, 100, 10000)])
         gain_seed = fgaus.values[1] - ped_vals[1]
 
         if fgaus.values[2] <= ped_vals[2]:
@@ -286,19 +289,20 @@ def seeds_and_bounds(sensor_type, run_no, n_chann, scaler, bins, spec, ped_vals,
         else:
             gain_sigma_seed = np.sqrt(fgaus.values[2]**2 - ped_vals[2]**2)
 
-    mu_seed = poisson_mu_seed(sensor_type, bins, spectra, ped_vals, scaler)
+    mu_seed = poisson_mu_seed(sensor_type, bins, sens_values.spectra, ped_vals, scaler)
     if mu_seed < 0: mu_seed = 0.001
 
-    sd0 = [norm_seed, mu_seed, gain_seed, gain_sigma_seed]
-    bd0 = [[0, 0, 0, 0.001], [1e10, 10000, 10000, 10000]]
+    sd0middle  = ()
+    bd00middle = ()
+    bd01middle = ()
 
     if 'gau' in func:
-        p_seed, p_sig_seed, p_min, p_max, p_sig_min, p_sig_max = pedestal_values(ped_vals,
-                                                                                 lim_ped, ped_errs)
-        sd0[2:2]    = p_seed, p_sig_seed
-        bd0[0][2:2] = p_min, p_sig_min
-        bd0[1][2:2] = p_max, p_sig_max
+        ped_values = pedestal_values(ped_vals, sens_values.lim_ped , ped_errs)
+        sd0middle  = (ped_values.gain        , ped_values.sigma    )
+        bd00middle = (ped_values.gain_min    , ped_values.gain_max )
+        bd01middle = (ped_values.sigma_min   , ped_values.sigma_max)
 
-    sd0 = tuple(sd0)
-    bd0 = [tuple(bd0[0]), tuple(bd0[1])]
-    return sd0, bd0
+    seed   = (norm_seed, mu_seed) + sd0middle  + (gain_seed, gain_sigma_seed)
+    bound1 = (0       ,       0)  + bd00middle + (0        ,           0.001)
+    bound2 = (1e10    ,   10000)  + bd01middle + (10000    ,           10000)
+    return seed, (bound1, bound2)
--- a/invisible_cities/reco/calib_functions.py
+++ b/invisible_cities/reco/calib_functions.py
@@ -186,7 +186,8 @@ def poisson_mu_seed(sensor_type, bins, spec, ped_vals, scaler):
     dscale = spec[bins<0].sum() / fitf.gauss(bins[bins<0], *ped_vals).sum()
     return fitf.fit(scaler,
                     bins[bins<0],
-                    spec[bins<0], (dscale)).values[0]
+                    spec[bins<0],
+                    (dscale)).values[0]
 
 
 def sensor_values(sensor_type, n_chann, scaler, spec, bins, ped_vals):
@@ -261,27 +262,35 @@ def seeds_and_bounds(sensor_type, run_no, n_chann, scaler, bins, spec, ped_vals,
         bd0 : sequence
         Minimum and maximum limits for the previous variables.
     """
-    
+
     norm_seed = spec.sum()
     gain_seed, gain_sigma_seed = seeds_db(sensor_type, run_no, n_chann)
     sens_values = sensor_values(sensor_type, n_chann, scaler, spec, bins, ped_vals)
 
+    spectra = sens_values.spectra
+    p_range = sens_values.peak_range
+    min_b   = sens_values.min_bin_peak
+    max_b   = sens_values.max_bin_peak
+    hpw     = sens_values.half_peak_width
+    p_seed  = sens_values.p1pe_seed
+    lim_ped = sens_values.lim_ped
+
     if not use_db_gain_seeds:
-        pDL  = find_peaks_cwt(sens_values.spectra, sens_values.p_range, min_snr=1, noise_perc=5)
-        p1pe = pDL[(bins[pDL]>sens_values.min_b) & (bins[pDL]<sens_values.max_b)]
+        pDL  = find_peaks_cwt(spectra, p_range, min_snr=1, noise_perc=5)
+        p1pe = pDL[(bins[pDL]>min_b) & (bins[pDL]<max_b)]
         if len(p1pe) == 0:
             try:
-                p1pe = np.argwhere(bins==(sens_values.min_b+sens_values.max_b)/2)[0][0]
+                p1pe = np.argwhere(bins==(min_b+max_b)/2)[0][0]
             except IndexError:
                 p1pe = len(bins)-1
         else:
-            p1pe = p1pe[sens_values.spectra[p1pe].argmax()]
+            p1pe = p1pe[spectra[p1pe].argmax()]
 
         fgaus = fitf.fit(fitf.gauss,
-                         bins[p1pe-sens_values.hpw:p1pe+sens_values.hpw],
-                         sens_values.spectra[p1pe-sens_values.hpw:p1pe+sens_values.hpw],
-                         seed   = (sens_values.spectra[p1pe], bins[p1pe], 7),
-                         sigma  = np.sqrt(sens_values.spectra[p1pe-sens_values.hpw:p1pe+sens_values.hpw]),
+                         bins[p1pe-hpw:p1pe+hpw],
+                         spectra[p1pe-hpw:p1pe+hpw],
+                         seed   = (spectra[p1pe], bins[p1pe], p_seed),
+                         sigma  = np.sqrt(spectra[p1pe-hpw:p1pe+hpw]),
                          bounds = [(0, -100, 0), (1e99, 100, 10000)])
         gain_seed = fgaus.values[1] - ped_vals[1]
 
@@ -290,7 +299,7 @@ def seeds_and_bounds(sensor_type, run_no, n_chann, scaler, bins, spec, ped_vals,
         else:
             gain_sigma_seed = np.sqrt(fgaus.values[2]**2 - ped_vals[2]**2)
 
-    mu_seed = poisson_mu_seed(sensor_type, bins, sens_values.spectra, ped_vals, scaler)
+    mu_seed = poisson_mu_seed(sensor_type, bins, spec, ped_vals, scaler)
     if mu_seed < 0: mu_seed = 0.001
 
     sd0middle  = ()
@@ -298,7 +307,7 @@ def seeds_and_bounds(sensor_type, run_no, n_chann, scaler, bins, spec, ped_vals,
     bd01middle = ()
 
     if 'gau' in func:
-        ped_values = pedestal_values(ped_vals, sens_values.lim_ped , ped_errs)
+        ped_values = pedestal_values(ped_vals, lim_ped             , ped_errs)
         sd0middle  = (ped_values.gain        , ped_values.sigma    )
         bd00middle = (ped_values.gain_min    , ped_values.gain_max )
         bd01middle = (ped_values.sigma_min   , ped_values.sigma_max)
--- a/invisible_cities/reco/calib_functions.py
+++ b/invisible_cities/reco/calib_functions.py
@@ -208,7 +208,8 @@ def sensor_values(sensor_type, n_chann, scaler, bins, spec, ped_vals):
         p1pe_seed       = 3
         lim_ped         = 10000
     elif sensor_type is SensorType.PMT:
-        scale           = spec[bins<0].sum() / fitf.gauss(bins[bins<0], *ped_vals).sum()
+        sel             = bins<0
+        scale           = spec[sel].sum() / fitf.gauss(bins[sel], *ped_vals).sum()
         spectra         = spec - fitf.gauss(bins, *ped_vals) * scale
         peak_range      = np.arange(10, 20)
         min_bin_peak    = 15
--- a/invisible_cities/reco/calib_functions.py
+++ b/invisible_cities/reco/calib_functions.py
@@ -245,7 +245,7 @@ def compute_seeds_from_spectrum(sens_values, bins, ped_vals):
     p_seed  = sens_values.p1pe_seed
 
     peaks_dark_led  = find_peaks_cwt(spectra, p_range, min_snr=1, noise_perc=5)
-    p1pe_samples    = peaks_dark_led[in_range(bins[peaks_dark_led], min_b, max_b)]
+    p1pe_samples    = peaks_dark_led[(bins[peaks_dark_led]>min_b) & (bins[peaks_dark_led]<max_b)]
     if len(p1pe_samples) == 0:
         try:
             p1pe_centroid = np.argwhere(bins==(min_b+max_b)/2)[0][0]
--- a/invisible_cities/reco/calib_functions.py
+++ b/invisible_cities/reco/calib_functions.py
@@ -199,6 +199,7 @@ def sensor_values(sensor_type, n_chann, scaler, spec, bins, ped_vals):
         min_bin_peak    = 10
         max_bin_peak    = 22
         half_peak_width = 5
+        p1pe_seed       = 3
         lim_ped         = 10000
     else:
         scale           = spec[bins<0].sum() / fitf.gauss(bins[bins<0], *ped_vals).sum()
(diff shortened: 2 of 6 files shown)


jacg:
@everyone If you're reopening an LFS version of a PR that was closed automatically by the switch-to-LFS of `nextic/master`, then please include the text *this is an LFS version of #XXX* in the PR description. I'll edit this one to make it comply.

gonzaponte:
That is funny...
@everyone points to https://github.com/everyone
captura de pantalla 2018-10-20 a las 10 22 20

jacg:
Hmm, I was mildly afraid that something like this might be the case. I'll have to drop the habit. At least it looks like nobody is listening.

jjgomezcadenas:
@carmenromo this needs to be updated and fixed

carmenromo:
This PR is ready to review @gonzaponte

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +139,180 @@ def copy_sensor_table(h5in, h5out):
...
             datasipm = dIn.root.Sensors.DataSiPM
             datasipm.copy(newparent=group)
+
+
+def dark_scaler(dark_spectrum):
+    """
+    A function to scale dark spectrum with mu value.
+    """
+    def scaled_spectrum(x, mu):
+        return np.exp(-mu) * dark_spectrum
```
gonzaponte, line 150: the first argument in scaled spectrum is not used. Also, maybe we can think of a test for this function.

invisible_cities/types/ic_types.py
```diff
@@ -1,4 +1,5 @@
 from argparse import Namespace
+from enum     import Enum, auto
```
gonzaponte: `auto` is not used

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +139,180 @@ def copy_sensor_table(h5in, h5out):
...
+    Take gain and sigma values of previous runs in the database
+    to use them as seeds.
+    """
+    if sensor_type == SensorType.SIPM:
+        gain_seed       = DB.DataSiPM(run_no).adc_to_pes.iloc[n_chann]
+        gain_sigma_seed = DB.DataSiPM(run_no).     Sigma.iloc[n_chann]
+    else:
+        gain_seed       = DB.DataPMT(run_no).adc_to_pes.iloc[n_chann]
+        gain_sigma_seed = DB.DataPMT(run_no).     Sigma.iloc[n_chann]
+    return gain_seed, gain_sigma_seed
```
gonzaponte, line 172: I think performing the comparison with  `is` is more appropriate than with `==`.

Also, use `elif sensor_type is SensorType.PMT` instead of `else`. `else` can be used to raise an exception when a non-valid value is used. This can be checked by a test!

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +139,180 @@ def copy_sensor_table(h5in, h5out):
...
+
+
+def poisson_mu_seed(sensor_type, bins, spec, ped_vals, scaler):
+    """
+    Calculate poisson mu using the scaler function.
+    """
+    if sensor_type == SensorType.SIPM:
```
gonzaponte: Again, `is` is better than `==`

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +139,180 @@ def copy_sensor_table(h5in, h5out):
...
+        sel    = (bins>=-5) & (bins<=5)
```
gonzaponte, line 180: Usually, intervals are taken as [a, b) (i.e. left-inclusive, right-exclusive). Check that you really want to use both inclusive. If not, replace with the function `in_range` from `core.core_functions`.

invisible_cities/evm/ic_containers.py
```diff
@@ -41,7 +41,9 @@ def _add_namedtuple_in_this_module(name, attribute_names):
         ('FitFunction'    , 'fn values errors chi2 pvalue cov'),
         ('TriggerParams'  , 'trigger_channels min_number_channels charge height width'),
         ('PeakData'       , 'charge height width'),
-        ('Measurement'    , 'value uncertainty')):
+        ('Measurement'    , 'value uncertainty'),
+        ('SensorParams'   , 'spectra, peak_range, min_bin_peak, max_bin_peak, half_peak_width, p1pe_seed, lim_ped'),
```
gonzaponte: remove commas between names.

Also what does the first `p` in `p1pe` mean?

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +139,180 @@ def copy_sensor_table(h5in, h5out):
...
+        gdist  = fitf.gauss(bins[sel], *ped_vals)
+        dscale = spec[sel].sum() / gdist.sum()
```
gonzaponte: maybe `spec` could have a better, more descriptive name? I can't figure out what it is.

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +139,180 @@ def copy_sensor_table(h5in, h5out):
...
+        errs   = poisson_sigma(spec[sel], default=1)
+        return fitf.fit(scaler,
+                        bins[sel],
+                        spec[sel],
+                        (dscale), sigma=errs).values[0]
```
gonzaponte: Why is `dscale` enclosed in parentheses?

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +139,180 @@ def copy_sensor_table(h5in, h5out):
...
+
+    dscale = spec[bins<0].sum() / fitf.gauss(bins[bins<0], *ped_vals).sum()
```
gonzaponte: define `bins < 0` once and use it later

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +139,180 @@ def copy_sensor_table(h5in, h5out):
...
+    return fitf.fit(scaler,
+                    bins[bins<0],
+                    spec[bins<0],
+                    (dscale)).values[0]
```
gonzaponte: `(dscale)` -> `dscale` unless there is a good reason for that

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +139,180 @@ def copy_sensor_table(h5in, h5out):
...
+    else:
+        scale           = spec[bins<0].sum() / fitf.gauss(bins[bins<0], *ped_vals).sum()
+        spectra         = spec - fitf.gauss(bins, *ped_vals) * scale
+        peak_range      = np.arange(10, 20)
+        min_bin_peak    = 15
+        max_bin_peak    = 50
+        half_peak_width = 10
+        p1pe_seed       = 7
+        lim_ped         = 10000
+    return SensorParams(spectra, peak_range, min_bin_peak, max_bin_peak, half_peak_width, p1pe_seed, lim_ped)
```
gonzaponte, line 223: `==` -> `is`
use `elif` for the PMT case and then `else` to raise an exception.

Are you sure we want these values hard-coded? Are not they susceptible to change in the future?

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +139,180 @@ def copy_sensor_table(h5in, h5out):
...
+
+    spectra = sens_values.spectra
+    p_range = sens_values.peak_range
+    min_b   = sens_values.min_bin_peak
+    max_b   = sens_values.max_bin_peak
+    hpw     = sens_values.half_peak_width
+    p_seed  = sens_values.p1pe_seed
+    lim_ped = sens_values.lim_ped
+
+    if not use_db_gain_seeds:
```
gonzaponte: everything within this `if` goes into a separate function

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +139,180 @@ def copy_sensor_table(h5in, h5out):
...
+        Returns
+        -------
+        sd0 : sequence
+        Seeds for normalization, mu, gain and sigma.
+        bd0 : sequence
+        Minimum and maximum limits for the previous variables.
+    """
+
+    norm_seed = spec.sum()
+    gain_seed, gain_sigma_seed = seeds_db(sensor_type, run_no, n_chann)
```
gonzaponte: this call is unnecessary in the case `use_db_gain_seeds = False`

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +139,180 @@ def copy_sensor_table(h5in, h5out):
...
+    p_range = sens_values.peak_range
+    min_b   = sens_values.min_bin_peak
+    max_b   = sens_values.max_bin_peak
+    hpw     = sens_values.half_peak_width
+    p_seed  = sens_values.p1pe_seed
+    lim_ped = sens_values.lim_ped
+
+    if not use_db_gain_seeds:
+        pDL  = find_peaks_cwt(spectra, p_range, min_snr=1, noise_perc=5)
+        p1pe = pDL[(bins[pDL]>min_b) & (bins[pDL]<max_b)]
```
gonzaponte: in this case you use exclusive on both ends. You sure? (just making sure this is not a mistake). If you meant to use `>=` and `<`  you can use `in_range`.

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +139,180 @@ def copy_sensor_table(h5in, h5out):
...
+    spectra = sens_values.spectra
+    p_range = sens_values.peak_range
+    min_b   = sens_values.min_bin_peak
+    max_b   = sens_values.max_bin_peak
+    hpw     = sens_values.half_peak_width
+    p_seed  = sens_values.p1pe_seed
+    lim_ped = sens_values.lim_ped
+
+    if not use_db_gain_seeds:
+        pDL  = find_peaks_cwt(spectra, p_range, min_snr=1, noise_perc=5)
```
gonzaponte: `pdL` could have a longer, meaningful name

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +139,180 @@ def copy_sensor_table(h5in, h5out):
...
+        p1pe = pDL[(bins[pDL]>min_b) & (bins[pDL]<max_b)]
+        if len(p1pe) == 0:
+            try:
+                p1pe = np.argwhere(bins==(min_b+max_b)/2)[0][0]
+            except IndexError:
+                p1pe = len(bins)-1
+        else:
+            p1pe = p1pe[spectra[p1pe].argmax()]
+
```
gonzaponte: you override the value of `p1pe`. It goes from an array to a single value. It would be more readable if you don't use the same variable.

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +139,180 @@ def copy_sensor_table(h5in, h5out):
...
+        fgaus = fitf.fit(fitf.gauss,
+                         bins[p1pe-hpw:p1pe+hpw],
+                         spectra[p1pe-hpw:p1pe+hpw],
+                         seed   = (spectra[p1pe], bins[p1pe], p_seed),
+                         sigma  = np.sqrt(spectra[p1pe-hpw:p1pe+hpw]),
+                         bounds = [(0, -100, 0), (1e99, 100, 10000)])
```
gonzaponte: I guess `1e99` is a way of saying "I don't want an upper limit`. That is set with `np.nan`.
On the other hand, are you sure you want hard-coded values?

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +139,180 @@ def copy_sensor_table(h5in, h5out):
...
+        gain_seed = fgaus.values[1] - ped_vals[1]
+
+        if fgaus.values[2] <= ped_vals[2]:
+            gain_sigma_seed = 0.5
+        else:
+            gain_sigma_seed = np.sqrt(fgaus.values[2]**2 - ped_vals[2]**2)
+
```
gonzaponte, line 318: This can be written as a one-liner as
```
gain_sigma_seed = np.sqrt(fgaus.values[2]**2 - ped_vals[2]**2) if fgaus.values[2] > ped_vals[2] else 0.5
```
(or the other way around, depending on which one is more relevant). If it is too long keep it as it is.

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +139,180 @@ def copy_sensor_table(h5in, h5out):
...
+    mu_seed = poisson_mu_seed(sensor_type, bins, spec, ped_vals, scaler)
+    if mu_seed < 0: mu_seed = 0.001
+
+    sd0middle  = ()
+    bd00middle = ()
+    bd01middle = ()
+
```
gonzaponte, line 325: better names?

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +139,180 @@ def copy_sensor_table(h5in, h5out):
...
+    if 'gau' in func:
+        ped_values = pedestal_values(ped_vals, lim_ped             , ped_errs)
+        sd0middle  = (ped_values.gain        , ped_values.sigma    )
+        bd00middle = (ped_values.gain_min    , ped_values.gain_max )
+        bd01middle = (ped_values.sigma_min   , ped_values.sigma_max)
+
+    seed   = (norm_seed, mu_seed) + sd0middle  + (gain_seed, gain_sigma_seed)
+    bound1 = (0       ,       0)  + bd00middle + (0        ,           0.001)
+    bound2 = (1e10    ,   10000)  + bd01middle + (10000    ,           10000)
+    return seed, (bound1, bound2)
```
gonzaponte, line 335: if you want to say "no upper limit" use `np.nan`. And again, are you sure you want hard-coded values?

invisible_cities/reco/calib_functions_test.py
```diff
@@ -168,4 +174,131 @@ def test_copy_sensor_table3(config_tmpdir):
...
-        
+
+
+@mark.parametrize('sensor_type     , run_number, n_channel, gain_seed, gain_sigma_seed',
+                  ((SensorType.SIPM,       6217,         1,   16.5622,         2.5),
+                   (SensorType.PMT ,       6217,         5,   24.9557,         9.55162)))
+def test_seeds_db(sensor_type, run_number, n_channel, gain_seed, gain_sigma_seed):
+    result = cf.seeds_db(sensor_type, run_number, n_channel)
+    assert result == (gain_seed, gain_sigma_seed)
+
```
gonzaponte, line 186: `run_number` here is not parametrized, it takes the same value, you can hard-code it.

invisible_cities/reco/calib_functions_test.py
```diff
@@ -168,4 +174,131 @@ def test_copy_sensor_table3(config_tmpdir):
...
+@mark.parametrize('     sensor_type,                    scaler,        mu',
+                  ((SensorType.SIPM, cf.dark_scaler(dark_sipm), 0.0698154),
+                   (SensorType.PMT , cf.dark_scaler(dark_pmt) , 0.0950066)))
+def test_poisson_mu_seed(sensor_type, scaler, mu):
+    bins     = np.array([-8,  -7,  -6,  -5,  -4,    -3,   -2,   -1,    0,    1,    2,    3,    4,   5,   6,   7])
+    spec     = np.array([28,  98,  28, 539, 1072, 1845, 2805, 3251, 3626, 3532, 3097, 2172, 1299, 665, 371, 174])
+    ped_vals = np.array([2.65181178e+04, 1.23743445e-01, 2.63794236e+00])
+
+    result   = cf.poisson_mu_seed(sensor_type, bins, spec, ped_vals, scaler)
+    np.testing.assert_approx_equal(result, mu)
```
gonzaponte: `assert_approx_equal` is imported. Remove `np.testing.`

invisible_cities/reco/calib_functions_test.py
```diff
@@ -168,4 +174,131 @@ def test_copy_sensor_table3(config_tmpdir):
...
+@mark.parametrize('     sensor_type, n_chann,                    scaler,    expected_range, min_b, max_b, half_width, p1pe_seed, lim_p',
+                  ((SensorType.SIPM,    1023, cf.dark_scaler(dark_sipm),  np.arange(4 ,20),    10,    22,          5,         3, 10000),
+                   (SensorType.PMT ,       0, cf.dark_scaler(dark_pmt) ,  np.arange(10,20),    15,    50,         10,         7, 10000)))
+def test_sensor_values(sensor_type, n_chann, scaler, expected_range, min_b, max_b, half_width, p1pe_seed, lim_p):
+    bins     = np.array([ -6,  -5,   -4,   -3,   -2,   -1,    0,    1,    2,    3,    4,   5,   6,   7])
+    spec     = np.array([ 28, 539, 1072, 1845, 2805, 3251, 3626, 3532, 3097, 2172, 1299, 665, 371, 174])
+    ped_vals = np.array([2.65181178e+04, 1.23743445e-01, 2.63794236e+00])
+    spectra, p_range, min_bin, max_bin, hpw, seed, lim_ped = cf.sensor_values(sensor_type, n_chann, scaler, spec, bins, ped_vals)
+
+    np.testing.assert_array_equal(p_range, expected_range)
```
gonzaponte: `assert_approx_equal` is imported. Remove `np.testing.`

invisible_cities/reco/calib_functions_test.py
```diff
@@ -168,4 +174,131 @@ def test_copy_sensor_table3(config_tmpdir):
...
+                  ((SensorType.SIPM,       6217,         1,   16.5622,         2.5),
+                   (SensorType.PMT ,       6217,         5,   24.9557,         9.55162)))
+def test_seeds_db(sensor_type, run_number, n_channel, gain_seed, gain_sigma_seed):
+    result = cf.seeds_db(sensor_type, run_number, n_channel)
+    assert result == (gain_seed, gain_sigma_seed)
+
+
+dark_sipm = np.array([612, 1142, 2054, 3037, 3593, 3769, 3777, 3319, 2321, 1298, 690])
+dark_pmt  = np.array([ 30,  107,  258,  612, 1142, 2054, 3037, 3593                 ])
+
```
gonzaponte, line 190: It seems these two variables are only used in the form of a scaler, so you can define
`sipm_scaler = cf.dark_scaler(dark_sipm)` and the equivalent for pmt and use them later instead of creating a scaler every time.

invisible_cities/reco/calib_functions_test.py
```diff
@@ -168,4 +174,131 @@ def test_copy_sensor_table3(config_tmpdir):
...
+
+@mark.parametrize('     sensor_type, n_chann,                    scaler,    expected_range, min_b, max_b, half_width, p1pe_seed, lim_p',
+                  ((SensorType.SIPM,    1023, cf.dark_scaler(dark_sipm),  np.arange(4 ,20),    10,    22,          5,         3, 10000),
+                   (SensorType.PMT ,       0, cf.dark_scaler(dark_pmt) ,  np.arange(10,20),    15,    50,         10,         7, 10000)))
+def test_sensor_values(sensor_type, n_chann, scaler, expected_range, min_b, max_b, half_width, p1pe_seed, lim_p):
+    bins     = np.array([ -6,  -5,   -4,   -3,   -2,   -1,    0,    1,    2,    3,    4,   5,   6,   7])
+    spec     = np.array([ 28, 539, 1072, 1845, 2805, 3251, 3626, 3532, 3097, 2172, 1299, 665, 371, 174])
+    ped_vals = np.array([2.65181178e+04, 1.23743445e-01, 2.63794236e+00])
+    spectra, p_range, min_bin, max_bin, hpw, seed, lim_ped = cf.sensor_values(sensor_type, n_chann, scaler, spec, bins, ped_vals)
```
gonzaponte: don't expand the output, use `whatever` and then `assert whatever.name == expected_value`, like you do in `test_pedestal_values`

invisible_cities/reco/calib_functions_test.py
```diff
@@ -168,4 +174,131 @@ def test_copy_sensor_table3(config_tmpdir):
...
+    specsL = np.array(getattr(h5in.root.HIST, sensor_spe)).sum(axis=0)
+    specsD = np.array(getattr(h5in.root.HIST, sensor_dark)).sum(axis=0)
+    bins   = np.array(getattr(h5in.root.HIST, sensor_bins))
+
+    min_stat = 10
+
+    for ich, (led, dar) in enumerate(zip(specsL, specsD)):
+        b1 = 0
+        b2 = len(dar)
+        if min_stat != 0:
```
gonzaponte: This `if` doesn't make sense as `min_stat` is fixed.

invisible_cities/reco/calib_functions_test.py
```diff
@@ -168,4 +174,131 @@ def test_copy_sensor_table3(config_tmpdir):
...
+    h5in    = tb.open_file(PATH_IN, 'r')
+    run_no  = get_run_number(h5in)
+
+    specsL = np.array(getattr(h5in.root.HIST, sensor_spe)).sum(axis=0)
+    specsD = np.array(getattr(h5in.root.HIST, sensor_dark)).sum(axis=0)
+    bins   = np.array(getattr(h5in.root.HIST, sensor_bins))
+
+    min_stat = 10
+
+    for ich, (led, dar) in enumerate(zip(specsL, specsD)):
```
gonzaponte, line 260: this `enumerate` doesn't seem useful

invisible_cities/reco/calib_functions_test.py
```diff
@@ -168,4 +174,131 @@ def test_copy_sensor_table3(config_tmpdir):
...
+                continue
+
+    assert run_no      == expected_run_no
+    assert len(specsL) == expected_n_sensors
+    assert len(specsL) == len(specsD)
+    assert len(bins)   == 349
+
+    assert not (b1 == 0)
+    assert not (b2 == 0)
+
```
gonzaponte, line 290: - asserts go inside the loop
- probably you want to check that `b2` is different from `len(dar)`
- use `!=` rather than `not ==`

invisible_cities/reco/calib_functions_test.py
```diff
@@ -168,4 +174,131 @@ def test_copy_sensor_table3(config_tmpdir):
...
+        b1 = 0
+        b2 = len(dar)
+        if min_stat != 0:
+            try:
+                valid_bins = np.argwhere(led>=min_stat)
+                b1 = valid_bins[ 0][0]
+                b2 = valid_bins[-1][0]
+            except IndexError:
+                continue
+
```
gonzaponte, line 269: Is the whole point to ensure that there is **not** an `IndexError`? If that is the case just leave this in the loop:
```
valid_bins = np.argwhere(led>=min_stat)
b1 = valid_bins[ 0][0]
b2 = valid_bins[-1][0]

(+ asserts)
```
If an exception is raised, the test will fail and will be more informative than the not equal to zero stuff.

invisible_cities/reco/calib_functions_test.py
```diff
@@ -168,4 +174,131 @@ def test_copy_sensor_table3(config_tmpdir):
...
+    assert run_no      == expected_run_no
+    assert len(specsL) == expected_n_sensors
+    assert len(specsL) == len(specsD)
+    assert len(bins)   == 349
```
gonzaponte: better define `expected_number_of_bins = 349` on the top of the function and use it here.

invisible_cities/reco/calib_functions_test.py
```diff
@@ -168,4 +174,131 @@ def test_copy_sensor_table3(config_tmpdir):
...
+    h5in    = tb.open_file(PATH_IN, 'r')
+    run_no  = get_run_number(h5in)
+
+    specsL = np.array(h5in.root.HIST.sipm_spe).sum(axis=0)
+    specsD = np.array(h5in.root.HIST.sipm_dark).sum(axis=0)
+    bins   = np.array(h5in.root.HIST.sipm_spe_bins)
+
+    min_stat = 10
+
+    for ich, (led, dar) in enumerate(zip(specsL, specsD)):
```
gonzaponte, line 303: same comments regarding the loop as in the previous test

invisible_cities/reco/calib_functions_test.py
```diff
@@ -168,4 +174,131 @@ def test_copy_sensor_table3(config_tmpdir):
...
+            try:
+                valid_bins = np.argwhere(led>=min_stat)
+                b1 = valid_bins[ 0][0]
+                b2 = valid_bins[-1][0]
+            except IndexError:
+                continue
+
+        pD = find_peaks_cwt(dar, np.arange(2, 20), min_snr=2)
+        if len(pD) == 0:
+            continue
```
gonzaponte, line 315: mmm, are you sure you want to skip if no peaks are found? shouldn't your data *always* contain some peak?

invisible_cities/reco/calib_functions_test.py
```diff
@@ -168,4 +174,131 @@ def test_copy_sensor_table3(config_tmpdir):
...
+
+        gb0     = [(0, -100, 0), (1e99, 100, 10000)]
```
gonzaponte: `np.nan` for no bound

invisible_cities/reco/calib_functions_test.py
```diff
@@ -168,4 +174,131 @@ def test_copy_sensor_table3(config_tmpdir):
...
+        sd0     = (dar.sum(), 0, 2)
+        sel     = np.arange(pD[0]-5, pD[0]+5)
+        errs    = poisson_sigma(dar[sel], default=0.1)
+        gfitRes = fitf.fit(fitf.gauss, bins[sel], dar[sel], sd0, sigma=errs, bounds=gb0)
+
+        ped_vals      = np.array([gfitRes.values[0], gfitRes.values[1], gfitRes.values[2]])
+        scaler_func   = cf.dark_scaler(dar[b1:b2][(bins[b1:b2]>=-5) & (bins[b1:b2]<=5)])
```
gonzaponte: too many operations in one line, split into many. Maybe define `indices = slice(b1, b2)`. Maybe use `in_range`.

invisible_cities/reco/calib_functions_test.py
```diff
@@ -168,4 +174,131 @@ def test_copy_sensor_table3(config_tmpdir):
...
+        seeds, bounds = cf.seeds_and_bounds(SensorType.SIPM, run_no, ich, scaler_func, bins[b1:b2],
+                                            led[b1:b2], ped_vals, gfitRes.errors, use_db_gain_seeds=False)
+        assert not (all(seeds) == 0)
```
gonzaponte: shouldn't you also check "exact" values for the seeds?

gonzaponte (changes requested):
Long review, but many are repeated messages and others are questions. Don't be intimidated by the number of comments :)

invisible_cities/reco/calib_functions.py
carmenromo: I know, it is like that because we want to obtain poisson mu from it

invisible_cities/types/ic_types.py
carmenromo: uops you're right!

invisible_cities/reco/calib_functions.py
andLaing: As far as I understand the x is required for a function to work with scipy.curve_fit (I'd have to study a bit more to see if I'm just being paranoid) and we preferred to use the fit_functions.fit method in all cases. We could simplify using scipy.leastsq but it would mean using a different wrapper for this seed fit. That being said, as far as I understand curve_fit uses leastsq so it wouldn't be too much of a problem

invisible_cities/reco/calib_functions.py
carmenromo: ok, do you think I should change it in all functions sensor_type appears? And the same for the addition of the raise ValueError?

gonzaponte: > do you think I should change it in all functions sensor_type appears? And the same for the addition of the raise ValueError?

Yes, I think I've commented in all occurrences

invisible_cities/reco/calib_functions.py
gonzaponte: Seems weird, but ok

invisible_cities/reco/calib_functions_test.py
carmenromo: Yes, but I realized that this apparts to the script that calls seeds_and_bound function, so I will delete this test, it must be included in a future PR.

invisible_cities/reco/calib_functions_test.py
carmenromo: I don't understand what do you mean here.

invisible_cities/reco/calib_functions_test.py
carmenromo: yes, but the problem is that there is a seed for each channel, so this is like 1800 values multiply per 4...

invisible_cities/evm/ic_containers.py
carmenromo: peak

invisible_cities/reco/calib_functions.py
carmenromo: Yes, they are susceptible to change in the future but where could I define then those parameters as namedtuple?

gonzaponte: I don't understand the connection between the fact that they are hardcoded and the namedtuple

invisible_cities/reco/calib_functions_test.py
gonzaponte: As in previous comments, if `1e99` means "I don't want to have an upper bound", use `np.nan` instead

invisible_cities/reco/calib_functions_test.py
gonzaponte: Mmm. How is it that `bounds` contains only four quantities and `seeds` can contain that many?

invisible_cities/evm/ic_containers.py
gonzaponte: I recommend using more explicit names for the variables, as it helps to understand the code. At some point, someone will need to read that code and having to decipher the meaning of a variable will not help. This applies to in particular to this namedtuple field, to `pDL` and to some other variables with extremely abbreviated names.

andLaing (commented):
Sorry I've been a bit unresponsive on this. On a coding level, I helped @carmenromo a bit with the development and it's based on some messy code of mine so I have very few useful comments other than:

Some of the docstrings need to be updated to document the changes made in the PR (Probably best to do it at the end)
The main seeds_and_bounds function seems a bit long to me, maybe it needs to be factorised a bit.

In a general sense the function works with my scripts and I get the same results so algorithmically it looks good.

carmenromo:
> Some of the docstrings need to be updated to document the changes made in the PR (Probably best to do it at the end)
The main seeds_and_bounds function seems a bit long to me, maybe it needs to be factorised a bit.

ok! Thanks! These days is Canfranc Gonzalo and I will work on this.

carmenromo:
By the way, Gonzalo agrees with me in calling the new cities for calibration: Phyllis (PMTs), Trude (SiPM) and Berenice (pdf). @andLaing, @jjgomezcadenas what do you think?

andLaing:
I thought Phyllis was already taken by the spectrum generation city for pmts?

In any case, I don't care. As long as we remember to update the shiftertools when we update the code in Canfranc

carmenromo:
> I thought Phyllis was already taken by the spectrum generation city for pmts?

Yes, it was.

> In any case, I don't care. As long as we remember to update the shiftertools when we update the code in Canfranc

Nice!

invisible_cities/reco/calib_functions.py
carmenromo: I checked if I wanted to use both inclusive and I don't

invisible_cities/reco/calib_functions.py
carmenromo: @andLaing do you know?

invisible_cities/reco/calib_functions.py
carmenromo: @andLaing here, 1e10 means infinite?

jjgomezcadenas:
@carmenromo, are we done here? If so, rebase.

carmenromo:
@jjgomezcadenas I need @gonzaponte review (or anyone, but he is familiar with the code)

jjgomezcadenas:
Doctor @gonzaponte, your help needed!

invisible_cities/reco/calib_functions.py
```diff
@@ -130,3 +140,195 @@ def copy_sensor_table(h5in, h5out):
...
+                        bins    [sel],
+                        spectrum[sel],
+                        dscale, sigma=errs).values[0]
+    elif sensor_type is SensorType.PMT:
+        dscale = spectrum[bins<0].sum() / fitf.gauss(bins[bins<0], *ped_vals).sum()
+        return fitf.fit(scaler,
+                        bins    [bins<0],
+                        spectrum[bins<0],
+                        dscale).values[0]
+    else:
```
gonzaponte, line 195: define `sel = bins < 0` and use the variable to avoid repeating the operation.

invisible_cities/reco/calib_functions.py
gonzaponte: any news on this?

gonzaponte (changes requested):
Just two minor issues + a rebase and this is ready.

I am not making emphasis on the hard-coded values in `sensor_values` intentionally. But they will probably need to be somehow taken from somewhere. Maybe from a function that has hard-coded values as a function of the run number, or even from the database.

invisible_cities/reco/calib_functions.py
carmenromo: `np.nan` gives error

gonzaponte: Yes, because it is `np.inf`, my mistake.

carmenromo: Yes, now it works :)

carmenromo:
Rebase done. Once tests finish this can be approved

gonzaponte (approved):

PR #553, opened 2018-10-20
Fix `test_mean_for_pmts_fee_is_unbiased`

gonzaponte:
While trying to fix this test I realized that we have a fishy trick in the code.
It was introduced in #489 and essentially ignores zeros in the waveforms passed to baseline-subtraction functions so the baseline calculation works for both full waveforms and zero-suppressed ones. That is fine for waveforms that are symmetric wrt zero, but not arbitrary waveforms.

I will push a couple of commits that show it a bit better.

### DIFF
--- a/invisible_cities/reco/calib_sensors_functions_test.py
+++ b/invisible_cities/reco/calib_sensors_functions_test.py
@@ -256,6 +256,17 @@ def test_subtract_mean_diff_minmax_remains_constant(square_pmt_and_sipm_waveform
     assert np.allclose(wf_diff_bls, wf_diff_original)
 
 
+@flaky(max_runs=10, min_passes=10)
+def test_subtract_mean_mean_is_zero(square_pmt_and_sipm_waveforms):
+    _, _, pmts_fee, _, _, _, _ = square_pmt_and_sipm_waveforms
+    pmts_bls = csf.subtract_mean(pmts_fee)
+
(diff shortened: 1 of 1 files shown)


andLaing:
I'm not sure I fully understand your reasoning. If anything I'd say that it wouldn't work for a waveform symmetric about zero?

Is there any situation in data where you see a negative effect? The baselines are ~40 ADC in SiPMs and ~2000 ADC in PMTs so they should be far enough from zero that the baseline is correctly calculated at least to within a reasonable error

gonzaponte:
No, I am not saying that the code works badly for MC or detector data (I am sure you checked that), I mean that it can produce unexpected results in custom-made waveforms, like those in the tests, because the function is more complex than what its name indicates.

That is my hypothesis, but I will have to take a more detailed look at it to confirm it.

jjgomezcadenas:
@andLaing, @gonzaponte, Travis is failing and the branch is out of date, we need this fixed asap.

gonzaponte:
I have just overridden the trick I mentioned earlier and the tests (the one failing before and the new one) passed 100/100 times.

I think this proves my hypothesis, but now we need a solution. I think adding a baseline to the waveforms used in the test will work and makes sense. This will use wfs that have the properties suitable for the test while making test wfs and real wfs look more alike.

jjgomezcadenas (approved):
A general improvement of various tests.

PR #554, opened 2018-10-20
Silence warnings

jacg:
An execution of our full test suite under the hard hypothesis profile
tended to produce an enormous amount of noise in the form of
warnings. Somehere among all this noise, there may be some interesting
or important signal, but any such signal is drowned out by all the
noise, which everyone has grown accustomed to ignoring.

We therefore introduce a policy of promoting all warnings to
errors, during test time. This forces us to actually look at the
warning and 

+ understand each warning as it appears and

+ decide what should be done about it, rather than igoring it

There are three main situations:

1. The warning points out a problem in our code, in which case we fix
   our code so that the warning goes away.

2. The warning emanates from some module that we use, in which
   case we have to wait for the module authors to deal with
   it (or submit a patch ourselves!) In the meantime, we silence
   this specific warning emanating from this specific point in a
   third party library.

3. It's a hypothesis deadline or timeout. In this case try to
   optimize the hypothesis test. Failing that, we increase the
   deadline and/or reduce `max_examples`.

This PR introduces the policy of upgrading warnings to errors at
test time, and then proceeds to deal with each warning that
appears.

As a consequence, the full test suite now runs under the hard
hypothesis profile without generating a single warning.

Warnings coming out of third party libraries are silenced in
`pytest.ini`. The list of ignored warings that appears therein
should be reviewed every time we bump dependency versions.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -8,8 +8,6 @@ settings.register_profile("hard"      , settings(max_examples = 1000))
 settings.register_profile("dev"       , settings(max_examples =   10))
 settings.register_profile("hard_nocov", settings(max_examples = 1000, use_coverage=False))
 settings.register_profile("dev_nocov" , settings(max_examples =   10, use_coverage=False))
-settings.register_profile("noshrink"  , settings(max_examples =   10,
-                                                 max_shrinks  =    0))
 settings.register_profile("debug"     , settings(max_examples =   10,
                                                 verbosity=Verbosity.verbose))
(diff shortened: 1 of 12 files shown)


jacg:
@jmbenlloch, can you make any sense out of the failure https://travis-ci.org/jacg/IC/jobs/444108481 ?

The commit with the same hash is passing these tests locally.

jjgomezcadenas (approved):
Approving this PR under some pressure to stabilize liquid cities. We are accepting a temporary hack out of necessity.

Issue #551, opened 2018-10-20
Spurious test failures


jacg:
Every now and then our Travis builds (or our local tests) fail spuriously. This sometimes wastes a lot of time and causes confusion. It's difficult to understand why these spurious failures occur, because it happens rarely, across different tests, and usually while we are in a hurry to do something else.

In order to gather information about these failures, so that we might be able to eliminate them, please add a comment to this issue every time you encounter such a spurious test failure.

Please mention the name of the test that failed, and information about exactly how it failed. If it happened in a Travis build, then provide a link to the build. If it happend on your local machine, then copy paste a relevant portion of the traceback.

Also, every time you add a failure, please edit this list, so that we keep a summary of the failures we have seen.

+ `test_mean_for_pmts_fee_is_unbiased`: 1

+ `test_inverse_cdf_hypothesis_generated`: 2 [fixed in #575]

+ `test_pmap_event_id_selection`: 1

+ `test_scaled_dark_pedestal_pedestal`: 1

+ `test_fill_kdst_var_1d`: 1

jacg:
`test_mean_for_pmts_fee_is_unbiased`

https://travis-ci.org/nextic/IC/jobs/443937686

jacg:
Hmm, restarting the job makes the original build which contains the failure, disappear. Does anybody know how whether Travis keeps the original build somewhere?

If not, we'll need a copy-paste rather than a link.

jacg:
=================================== FAILURES ===================================
____________________ test_inverse_cdf_hypothesis_generated _____________________
[gw3] linux -- Python 3.7.0 /home/travis/miniconda/envs/IC-3.7-2018-08-29/bin/python

    @given(valid_distributions(),
>          floats(min_value = 0.1, max_value = 0.9))
    def test_inverse_cdf_hypothesis_generated(distribution, percentile):

invisible_cities/core/random_sampling_test.py:140: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../miniconda/envs/IC-3.7-2018-08-29/lib/python3.7/site-packages/hypothesis/core.py:587: in execute
    result = self.test_runner(data, run)
../../../miniconda/envs/IC-3.7-2018-08-29/lib/python3.7/site-packages/hypothesis/executors.py:58: in default_new_style_executor
    return function(data)
[...]

PR #555, opened 2018-10-20
Fix random sampling tests

gonzaponte:
As pointed out in #551, the test `test_inverse_cdf_hypothesis_generated` failed from time to time.
The origin of the failure was the exact (`==`) comparison between floats. This has been replaced with an approximate comparison (`pytest.approx`). I also took the chance to improve the data generation algorithm for `hypothesis`.

### DIFF
--- a/invisible_cities/core/random_sampling_test.py
+++ b/invisible_cities/core/random_sampling_test.py
@@ -3,6 +3,7 @@ import numpy as np
 from flaky  import flaky
 from pytest import mark
 from pytest import fixture
+from pytest import approx
 
 from hypothesis            import given
 from hypothesis.strategies import composite
@@ -146,7 +147,7 @@ def test_inverse_cdf_hypothesis_generated(distribution, percentile):
             true_value = d
             break
     icdf = inverse_cdf(domain, cdf, percentile)
-    assert true_value == icdf
+    assert icdf == approx(true_value)
 
 
 @fixture(scope="module")


gonzaponte:
I would appreciate some input from @jacg here, if possible.

Take this piece of code:
```
def valid_domains(size):
    arrays = float_arrays(min_value = -100,
                          max_value = +100,
                          size      = size)
    sorted_arrays = arrays.map(np.sort)
    increasing_arrays = sorted_arrays.filter(lambda x: np.all(np.diff(x) > 0))
    return increasing_arrays
```
It generates float arrays (i.e., `np.array(dtype=float)`) of a given length, with values in a certain range. The values are sorted and then the sequence required to be monotically increasing.

The whole thing essentially requires that all the values of the array are different. That doesn't seem to me that much of a "big chunk of the search space", but it complains, so it must be. Is there a smarter way to produce the same data?

jacg:
It probably isn't a big chunk of the search space, but don't forget that hypothesis shrinks counterexamples that it finds, looking for simpler ones that still falsify the hypothesis, and it is thus almost guaranteed to come up with arrays with repeated values.

But even in the absence of counterexamples and shrinking, it's going to happen sooner or later, and you have to avoid it.

I suspect that there may well be a *smarter* way to produce what you want, but I'm too deep in 3 other things at the moment to be able to think about it.

Your problem is that too many examples are being rejected?

gonzaponte:
Yes. It says it can only generate 8 valid ones in 50 tries.

jacg:
Any chance of getting the fix for the exact comparison of floats right now? I've wasted many, many, many hours today on spurious test failures ... and this one has just popped up again, stalling all sorts of pipelines in my head.

gonzaponte:
Sure, I didn't think it was urgent, sorry.

jjgomezcadenas (approved):
A simple fix of a bug.

PR #557, opened 2018-10-20
Use pytest-instafail in manage.sh run_tests

jacg:
By default pytest reports test progress by printing a single character
for each test that runs, indicating whether it has passed, failed,
been skipped, etc.

If a test fails, we have to wait until the whole test suite completes
before finding out exactly which test failed, and how.

To mitigate this we use pytest's `-v` option, which displays progress
by dedicating a whole line to each test, and showing the full test
name. However, this produces very verbose output on our full test
suite, and we still have to wait for the whole suite to complete
before we see *how* a test failed, even if we know which one failed.

`pytest-xfail` is a pytest plugin which acts just like the default
progress report (one character per test) but shows a test's failure
report as soon as it fails.

### DIFF
--- a/manage.sh
+++ b/manage.sh
@@ -72,7 +72,7 @@ function install_conda {
     fi
 }
 
-CONDA_ENV_TAG=2018-10-18
+CONDA_ENV_TAG=2018-10-20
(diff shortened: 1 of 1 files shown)


jjgomezcadenas (approved):
Useful and relevant.

PR #558, opened 2018-10-20
Fix corona bug with masked sensors

gonzaponte:
This is an LFS version of #538.

### DIFF
--- a/invisible_cities/reco/xy_algorithm_test.py
+++ b/invisible_cities/reco/xy_algorithm_test.py
@@ -57,7 +57,8 @@ def toy_sipm_signal_and_inds():
 
 
 @given(positions_and_qs())
-def test_barycenter(p_q):
+@settings(max_examples=100)
+def test_barycenter_generic(p_q):
     pos, qs = p_q
     B  = barycenter(pos, qs)[0]
     assert np.allclose(B.posxy, np.average(pos, weights=qs, axis=0))
--- a/invisible_cities/reco/xy_algorithm_test.py
+++ b/invisible_cities/reco/xy_algorithm_test.py
@@ -1,4 +1,5 @@
-import numpy as np
+import numpy  as np
+import pandas as pd
 
 from pytest import fixture
 from pytest import mark
@@ -44,6 +45,25 @@ def toy_sipm_signal():
     pos = np.stack((xs, ys), axis=1)
     return pos, qs
 
+
+def _create_fake_datasipm(x, y, active):
+    assert len(x) == len(y) == len(active)
+    size = len(x)
+    sensorid   = np.arange(size, dtype=np.int  )
+    channelid  = np.arange(size, dtype=np.int  )
+    adc_to_pes = np.ones  (size, dtype=np.float)
+    sigma      = np.ones  (size, dtype=np.float)
+    active     =         active.astype(np.int  )
+    x          =              x.astype(np.float)
+    y          =              y.astype(np.float)
+
+    return pd.DataFrame(dict( SensorID  =   sensorid,
+                             ChannelID  =  channelid,
+                             Active     =     active,
+                             X          =          x,
+                             Y          =          y,
+                             adc_to_pes = adc_to_pes,
+                             Sigma      =      sigma))
 
 @fixture(scope="session")
 def toy_sipm_signal_and_inds():
--- a/invisible_cities/reco/xy_algorithm_test.py
+++ b/invisible_cities/reco/xy_algorithm_test.py
@@ -24,7 +24,6 @@ from .       xy_algorithms     import corona
 from .       xy_algorithms     import barycenter
 from .       xy_algorithms     import discard_sipms
 from .       xy_algorithms     import get_nearby_sipm_inds
-from .       xy_algorithms     import get_neighbours
 from .       xy_algorithms     import have_same_position_in_space
 from .       xy_algorithms     import is_masked
 
@@ -225,6 +224,7 @@ def test_get_nearby_sipm_inds():
         else       : assert np.sqrt((xs[i] - xc)**2 + (ys[i] - yc)**2) >  d
 
 
+@mark.skip("Function removed")
 def test_get_neighbours():
     pos = np.array([(35.5, 55.5)])
 
--- a/invisible_cities/reco/xy_algorithm_test.py
+++ b/invisible_cities/reco/xy_algorithm_test.py
@@ -17,6 +17,7 @@ from .. core.testing_utils     import float_arrays
 from .. core.system_of_units_c import units
 from .. core.exceptions        import SipmEmptyList
 from .. core.exceptions        import ClusterEmptyList
+from .. core.exceptions        import SipmEmptyListAboveQthr
 from .. core.exceptions        import SipmZeroCharge
 
 from .       xy_algorithms     import corona
@@ -187,6 +188,17 @@ def test_corona_Qlm_too_high_raises_ClusterEmptyList(toy_sipm_signal):
         corona(pos, qs,
                msipm          =      1,
                Qlm            =    Qlm,
+               new_lm_radius  = np.inf)
+
+
+def test_corona_Qthr_too_high_raises_SipmEmptyListAboveQthr(toy_sipm_signal):
+    pos, qs  = toy_sipm_signal
+    Qthr     = max(qs) * 1.1 * units.pes
+
+    with raises(SipmEmptyListAboveQthr):
+        corona(pos, qs,
+               msipm          =      1,
+               Qthr           =   Qthr,
                new_lm_radius  = np.inf)
 
 
--- a/invisible_cities/reco/xy_algorithm_test.py
+++ b/invisible_cities/reco/xy_algorithm_test.py
@@ -240,6 +240,7 @@ def test_corona_msipm(toy_sipm_signal, datasipm):
         corona(pos, qs, datasipm, msipm=2)
 
 
+@mark.skip
 @parametrize(' Qlm,    rmax, nclusters',
              ((4.9,      15,         2),
               (4.9, 1000000,         1)))
--- a/invisible_cities/reco/xy_algorithm_test.py
+++ b/invisible_cities/reco/xy_algorithm_test.py
@@ -185,7 +185,7 @@ def test_raises_sipm_zero_charge(algorithm):
         algorithm(np.array([[1, 2]]), np.array([0, 0]))
 
 
-def test_corona_barycenter_are_same_with_one_cluster(toy_sipm_signal, datasipm):
+def test_corona_barycenter_can_be_the_same_with_one_cluster(toy_sipm_signal, datasipm):
     pos, qs = toy_sipm_signal
     c_clusters = corona(pos, qs, datasipm,
                         new_lm_radius = 10 * units.m,
--- a/invisible_cities/reco/xy_algorithm_test.py
+++ b/invisible_cities/reco/xy_algorithm_test.py
@@ -42,8 +42,8 @@ def positions_and_qs(draw, min_value=1, max_value=100):
 
 @fixture(scope="session")
 def toy_sipm_signal():
-    xs  = np.array([65, -64]) * units.mm
-    ys  = np.array([63, -62]) * units.mm
+    xs  = np.array([65, -55]) * units.mm
+    ys  = np.array([65, -75]) * units.mm
     qs  = np.array([ 6,   5]) * units.pes
     pos = np.stack((xs, ys), axis=1)
     return pos, qs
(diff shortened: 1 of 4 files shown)


jjgomezcadenas:
@paolafer can you review this at your earliest convenience?

paolafer:
I'm aware also of this.

invisible_cities/reco/xy_algorithm_test.py
```diff
@@ -1,106 +1,218 @@
...
 
-def test_barycenter_raises_sipm_empty_list():
-    with raises(SipmEmptyList):
-        barycenter(np.array([]), None)
+@parametrize("     x         y        q     expected_xy".split(),
+             (( (-1, +1), ( 0,  0), (1, 1),   ( 0, 0)  ),
+              ( ( 0,  0), (-1, +1), (1, 1),   ( 0, 0)  ),
+              ( (-1, +1), (-1, +1), (1, 1),   ( 0, 0)  ),
+              ( (-4, +3), (+2, -5), (2, 5),   ( 1,-3)  )))
+def test_barycenter_simple_cases(x, y, q, expected_xy):
```
paolafer, line 147: Why do you need to add this specific case? Isn't the `test_barycenter_generic` test enough?

invisible_cities/reco/xy_algorithm_test.py
```diff
@@ -1,106 +1,218 @@
...
+    xy = np.stack((x, y), axis=1)
+    qs = np.array(q)
+    b  = barycenter(xy, qs)[0]
+    assert np.allclose(b.posxy, expected_xy)
 
 
-def test_barycenter_raises_sipm_zero_charge():
-    with raises(SipmZeroCharge):
-        barycenter(np.array([1,2]), np.array([0,0]))
+@given(float_arrays(size=8, min_value=0, max_value=100, mask=np.sum))
```
paolafer, line 154: What's the purpose of the 'np.sum' filter here? I think I'm not sure how the hypothesis strategy works ...
Also, if the purpose of this test is to ensure that only one cluster is returned, why not add it to the `test_barycenter_generic` test? I see that below you add also a `test_barycenter_single_cluster_generic`. As in the previous comment, why do you need a generic and a specific case?

invisible_cities/reco/xy_algorithm_test.py
```diff
@@ -1,106 +1,218 @@
...
 
-def test_corona_barycenter_are_same_with_one_cluster(toy_sipm_signal):
+
+@parametrize("algorithm", (barycenter, partial(corona, all_sipms=datasipm())))
+def test_raises_sipm_zero_charge(algorithm):
+    with raises(SipmZeroCharge):
+        algorithm(np.array([[1, 2]]), np.array([0, 0]))
+
+
+def test_corona_barycenter_are_same_with_one_cluster(toy_sipm_signal, datasipm):
```
paolafer: I know that this name was already there, but isn't it misleading? Corona and barycenter could give different results in the case of one cluster, it depends on the actual pattern of charge in the SiPMs.

invisible_cities/reco/xy_algorithm_test.py
```diff
@@ -109,58 +221,57 @@ def test_corona_min_threshold_Qthr():
...
-    k = 10000
-    xs = np.arange(k)
-    ys = xs + k
-    pos = np.stack((xs, ys), axis=1)
-    qs = xs + 2*k
-    i  = np.array([0, 5, 1000, 9999])
-    return k, i, pos, qs
+             ((4.9,      15,         2),
+              (4.9, 1000000,         1)))
+def test_corona_simple_examples(toy_sipm_signal, datasipm, Qlm, rmax, nclusters):
```
paolafer, line 247: Doesn't this test test exactly the same as `test_corona_multiple_clusters` and `test_corona_barycenter_are_same_with_one_cluster`? Do we need it?

invisible_cities/reco/xy_algorithm_test.py
```diff
@@ -174,23 +285,21 @@ def test_discard_sipms(toy_sipm_signal_and_inds):
...
     for i in range(len(xs)):
-        if i in sis:
-            assert np.sqrt((xs[i] - xc)**2 + (ys[i] - yc)**2) <= d
-        else:
-            assert np.sqrt((xs[i] - xc)**2 + (ys[i] - yc)**2) >  d
+        if i in sis: assert np.sqrt((xs[i] - xc)**2 + (ys[i] - yc)**2) <= d
+        else       : assert np.sqrt((xs[i] - xc)**2 + (ys[i] - yc)**2) >  d
 
 
+@mark.skip("Functions removed")
```
paolafer, line 301: Why do you not delete this test at all?

invisible_cities/reco/xy_algorithm_test.py
```diff
@@ -217,44 +327,112 @@ def test_is_masked():
     assert not is_masked(sipm_alive , pos_masked)
 
 
-def test_masked_channels():
+
+def test_count_masked_all_active(datasipm_all_active):
+    xy0 = np.array([0, 0], dtype=np.float)
+    is_masked = datasipm_all_active.Active.values
+
+    # All sipms are active in run number 0
```
paolafer: At this moment, I can see that run number 0 has some sensor marked as not active. I think that you want to say "run number 1", here.

invisible_cities/reco/xy_algorithm_test.py
```diff
@@ -217,44 +327,112 @@ def test_is_masked():
...
+                   (  1006,     15,                2),
+                   ( 26062,      5,                1),
+                   ( 26062,     15,                2)))
+def test_count_masked_near_masked(datasipm_5000, sipm_id, radius, expected_nmasked):
+    sipm_indx    = np.argwhere(datasipm_5000.SensorID.values == sipm_id)[0][0]
+    masked_sipm  = datasipm_5000.iloc[sipm_indx]
+    masked_xy    = np.array([masked_sipm.X, masked_sipm.Y])
+    is_masked    = datasipm_5000.Active.values
+
+    # small mear so the search point doesn't fall exactly at sipm position
```
paolafer:
```suggestion
    # small smear so the search point doesn't fall exactly at sipm position
```

invisible_cities/reco/xy_algorithm_test.py
```diff
@@ -217,44 +327,112 @@ def test_is_masked():
...
+                            msipm           = 6              ,
+                            Qthr            = 0   * units.pes,
+                            Qlm             = 4   * units.pes,
+                            new_lm_radius   = 1.5 * units.mm ,
+                            consider_masked = True)
 
     assert len(found_clusters) == expected_nclusters
+
+
+def test_corona_finds_masked_sipms_correclty(datasipm5x5):
```
paolafer:
```suggestion
def test_corona_finds_masked_sipms_correctly(datasipm5x5):
```

invisible_cities/reco/xy_algorithm_test.py
```diff
@@ -217,44 +327,112 @@ def test_is_masked():
...
+    # With the appropiate parameters, this should yield a single cluster
+    # with only 25 - 8 = 17 SiPM.
+    datasipm = datasipm5x5
+
+    # For visualization purposes we use o = np.nan and then remove
+    # those values
+    x       = np.nan
+    all_xys = np.stack([datasipm.X.values, datasipm.Y.values], axis=1)
+    all_qs  = np.array([[1, 2, 3, 4, 9], # create slight assymetry
```
paolafer:
```suggestion
    all_qs  = np.array([[1, 2, 3, 4, 9], # create slight asymmetry
```

invisible_cities/reco/xy_algorithm_test.py
```diff
@@ -217,44 +327,112 @@ def test_is_masked():
...
+                        [3, x, 9, x, 3], # fall exactly at the center
+                        [4, x, x, x, 2], # of the array
+                        [5, 4, 3, 2, 1]], dtype=np.float).flatten()
+
+    ok = ~np.isnan(all_qs)
+
+    c = corona(all_xys[ok], all_qs[ok], datasipm,
+               Qthr            =      0,
+               Qlm             =      6,
+                   lm_radius   =      0,
```
paolafer, line 433:
```suggestion
               lm_radius       =      0,
```

paolafer (changes requested):
An important bug has been addressed and fixed, regarding how we are taking into account masked SiPMs in the corona algorithm. Unfortunately, this fix comes at the price that the corona algorithm needs to be passed the SiPM database. I would like to make explicit what the `all_sipms` object is, in the documentation of the function (didn't we decide to use type hints, also at some point?).
A complete set of tests has been reviewed and extended to ensure the correct behaviour of the corona algorithm with masked SiPMs. I have asked for some clarifications, but, apart from that, I approve this PR.

invisible_cities/reco/xy_algorithm_test.py
gonzaponte: It is, but the simpler the example the easier it is to find the problem in case of a failure.

invisible_cities/reco/xy_algorithm_test.py
gonzaponte: > What's the purpose of the 'np.sum' filter here?

`np.sum` will return the sum of the values generated by the strategies. This will be treated as a boolean, i.e., `False if sum == 0` and `True` otherwise. This is to ensure the strategy doesn't generate an array filled with zeros.

> why not add it to the test_barycenter_generic test?

Simpler tests can help to pinpoint the error in case of failure If `test_barycenter_generic` fails, it can be for a number of reasons. If this one fails, it is because it is returning more than one cluster.

> As in the previous comment, why do you need a generic and a specific case?

The specific case helps to illustrate the workings of the function, the generic case tries to be more robust than just trying a few cases by hand.

invisible_cities/reco/xy_algorithm_test.py
gonzaponte: it should actually be `test_corona_barycenter_are_same_with_one_cluster_given_certain_corona_configuration`. Maybe we can think of something that is at the same time descriptive and short.

invisible_cities/reco/xy_algorithm_test.py
gonzaponte: They are not *exactly* the same, but similar. This one checks the number of clusters for two different configurations, while the other checks position and charge (and number of clusters), but just for one case. They are not completely orthogonal, but they do not overlap completely, either.

invisible_cities/reco/xy_algorithm_test.py
gonzaponte: It will be. There are multiple skipped tests in the code and I am thinking of a PR that removes all of them. I don't mind removing it here, directly.

invisible_cities/reco/xy_algorithm_test.py
gonzaponte: Yes, this is a typo, It is run number 1

invisible_cities/reco/xy_algorithm_test.py
gonzaponte: I liked how it aligned with `new_lm_radius`, but I don't have a strong opinion

invisible_cities/reco/xy_algorithm_test.py
paolafer: Mmmm, I still don't see why we need this test. These two configurations are arbitrary, the configuration used in the other two tests serves for the same purpose, right? Following your argument, it seems to me that we should add an infinite number of tests to check each configuration, doesn't it?

invisible_cities/reco/xy_algorithm_test.py
paolafer: Ah, ok, if there's another PR in sight, there's no need to remove it here.

invisible_cities/reco/xy_algorithm_test.py
paolafer: Ah, I thought it was a mistake, if you had a purpose there, just leave it as you like it :-).

invisible_cities/reco/xy_algorithm_test.py
paolafer: What about `test_corona_barycenter_can_be_the_same_with_one_cluster` :-D?

invisible_cities/reco/xy_algorithm_test.py
paolafer: Ok, I see the reasoning, it's fine with me!

invisible_cities/reco/xy_algorithm_test.py
gonzaponte: Nice

invisible_cities/reco/xy_algorithm_test.py
gonzaponte: Ok, I didn't read the code properly. I agree, this test is redundant. However, I would like to have a similar one, where we explore the search for multiple clusters with the same data under different conditions. But that may be something for a different PR.

paolafer (approved):

PR #559, opened 2018-10-21
Fix test mean pmts

jjgomezcadenas:
A very simply fix, which allows  an absolute precision of 10**-2 as opposed to the default (very strict) precision.


gonzaponte:
Why this PR? This test is supposed to be fixed in #553.
As discussed there, the test is (probably) not failing because of fluctuations, but rather because the function we are testing is doing something more complex than removing the mean of the wf. Thus, increasing the tolerance will only hide the problem instead of solving it.

gonzaponte (approved):
Temporarily skip tests + increase tolerance. OK. Desperately needed to move the liquid cities forward.

PR #562, opened 2018-10-21
Improve testing for the CLI and configuration side of liquid cities

gonzaponte:
This PR fixes a failure of the `event_range` function that didn't treat all cases properly.
A couple of tests have been introduced to demonstrate the misbehavior.
The implemented fix makes the tests pass.


jjgomezcadenas (approved):
Fixes an identified problem and adds a test. Good job.

jjgomezcadenas:
@gonzaponte you can merge!

jacg:
I don't see any changes in the command line parsing tests in core/config_test.py. These were originally written for the solid cities, and weren't adapted for the liquid ones. I started working on this with @jjgomezcadenas but we didn't have time to finish it. That work naturally belongs in this PR, so I think that approving and merging this PR in its current state would be premature.

gonzaponte:
Ok. Do you have your work pushed somewhere and/or can you push it to this PR? I can pick it where you left it (I am assuming you are out of the picture for a few days, at least).

jacg:
I ran out of time before I got around to pushing it. I'll get access to my computer sometime tomorrow, then I'll send you something.

jjgomezcadenas:
I see no reason not to approve a small PR that fixes an identified problem, including test. We can proceed later to identify further problems, but this one is neatly solved, so in my opinion there is nothing wrong or premature in approving it.

jacg:
I've push the WIP to this branch. I repeat the commit message here, for convenience:

`configure_liquid_test.py` is just a copy of `configure_test.py` in which
the solid cities are being replaced with liquid ones.

This got through the net during the liquification, probably because
all other city tests were right next to the city implementation, while
these ones are tucked away somewhere else, out of sight.

As the liquid cities are so much simpler, the liquid version of the
dummy city is completely trivial: it just returns whatever arguments
it gets passed.

These tests use our sample config files, and these configs inject some
parameters which no longer exist (such as `write_mc_info` in
`penthesilea`).

The sample config files are hierarchical, and IIRC, we decided to get
rid of that. So I think that we need an new `liquid_config` directory
with new sample config files which conform to and document our new
config file structures. These can then coexist with the solid ones
until they are removed in my `goodbye-solid-cities` branch, which is
almost ready; the only thing that's missing is the liquified version
of these tests.

While doing this I also noticed that there are comments in the city
decorator which talk about simplifying the fiddling we do with the
configuration which arrives. We didn't deal with it at the time
because of the old structure of the config file. We should probably
open an issue reminding us to do this at some point. Probably not
right away.

jacg:
I've push the WIP to this branch. I repeat the commit message here, for convenience:

`configure_liquid_test.py` is just a copy of `configure_test.py` in which
the solid cities are being replaced with liquid ones.

This got through the net during the liquification, probably because
all other city tests were right next to the city implementation, while
these ones are tucked away somewhere else, out of sight.

As the liquid cities are so much simpler, the liquid version of the
dummy city is completely trivial: it just returns whatever arguments
it gets passed.

These tests use our sample config files, and these configs inject some
parameters which no longer exist (such as `write_mc_info` in
`penthesilea`).

The sample config files are hierarchical, and IIRC, we decided to get
rid of that. So I think that we need an new `liquid_config` directory
with new sample config files which conform to and document our new
config file structures. These can then coexist with the solid ones
until they are removed in my `goodbye-solid-cities` branch, which is
almost ready; the only thing that's missing is the liquified version
of these tests.

While doing this I also noticed that there are comments in the city
decorator which talk about simplifying the fiddling we do with the
configuration which arrives. We didn't deal with it at the time
because of the old structure of the config file. We should probably
open an issue reminding us to do this at some point. Probably not
right away.

jacg:
I've pushed the WIP to this branch. I repeat the commit message here, for convenience:

`configure_liquid_test.py` is just a copy of `configure_test.py` in which
the solid cities are being replaced with liquid ones.

This got through the net during the liquification, probably because
all other city tests were right next to the city implementation, while
these ones are tucked away somewhere else, out of sight.

As the liquid cities are so much simpler, the liquid version of the
dummy city is completely trivial: it just returns whatever arguments
it gets passed.

These tests use our sample config files, and these configs inject some
parameters which no longer exist (such as `write_mc_info` in
`penthesilea`).

The sample config files are hierarchical, and IIRC, we decided to get
rid of that. So I think that we need an new `liquid_config` directory
with new sample config files which conform to and document our new
config file structures. These can then coexist with the solid ones
until they are removed in my `goodbye-solid-cities` branch, which is
almost ready; the only thing that's missing is the liquified version
of these tests.

While doing this I also noticed that there are comments in the city
decorator which talk about simplifying the fiddling we do with the
configuration which arrives. We didn't deal with it at the time
because of the old structure of the config file. We should probably
open an issue reminding us to do this at some point. Probably not
right away.

jacg:
I've pushed the WIP to this branch. I repeat the commit message here, for convenience:

`configure_liquid_test.py` is just a copy of `configure_test.py` in which
the solid cities are being replaced with liquid ones.

This got through the net during the liquification, probably because
all other city tests were right next to the city implementation, while
these ones are tucked away somewhere else, out of sight.

As the liquid cities are so much simpler, the liquid version of the
dummy city is completely trivial: it just returns whatever arguments
it gets passed.

These tests use our sample config files, and these configs inject some
parameters which no longer exist (such as `write_mc_info` in
`penthesilea`).

The sample config files are hierarchical, and IIRC, we decided to get
rid of that. So I think that we need an new `liquid_config` directory
with new sample config files which conform to and document our new
config file structures. These can then coexist with the solid ones
until they are removed in my `goodbye-solid-cities` branch, which is
almost ready; the only thing that's missing is the liquified version
of these tests.

While doing this I also noticed that there are comments in the city
decorator which talk about simplifying the fiddling we do with the
configuration which arrives. We didn't deal with it at the time
because of the old structure of the config file. We should probably
open an issue reminding us to do this at some point. Probably not
right away.

gonzaponte:
We've got your message repeated three times. I will take the liberty of deleting the repeats. I don't know if this was a problem with github, but I also had to refresh the page many (many) times for it to show updates.

gonzaponte:
We've got your message repeated three times. I will take the liberty of deleting the repeats. I don't know if this was a problem with github, but I also had to refresh the page many (many) times for it to show updates.

jacg:
GitHub had problems (as you can see on their status page), at first it was telling me I couldn't comment, then it showed me many copies, which it didn't let me delete. I still see many copies here, even after your deletion of them (assuming you succeeded).

jacg:
Looking good.

It's not entirely clear to me how this is passing tests with the solid city config files, which contain names which the liquid cities don't expect.

gonzaponte:
Mmm, not sure which names are those. The tests in `configure_liquid_test.py` are not calling cities (except for `test_config_penthesilea_counters`, which already uses the liquid config) and therefore we cannot predict the parameters, we can just make sure the parameters will be there in the output of  `configure`.

On a similar but maybe related topic, if a parameter is missing the city will complain, but if there are extra parameters it will not (I think).

jacg:
Ah, OK, I've found the liquid config files. it makes sense now.

> if a parameter is missing the city will complain, but if there are extra parameters it will not (I think).

It should complain in both cases: the cities have named parameters now, with no default values, and no `*args` or `**kwds`, so you have to pass exactly the correct set of arguments.

BTW, I think that we should force them to be passed as keyword arguments, i.e. add a `*, ` to the front of the parameter list of each city.

jacg:
Are `dorothea_kr_with_barycenter_low_threshold_mc_example.conf` and the like relevant to liquid cities, or should they disappear?

jacg:
EDIT: this was stuck somewhere for a few hours because of the GitHub problems. It was supposed to appear higher up!


I have pushed something. Please read the commit message for a description.

> I see no reason not to approve a small PR that fixes an identified problem, including test. We can proceed later to identify further problems, but this one is neatly solved, so in my opinion there is nothing wrong or premature in approving it.

There is no question if identifying further problems later:
The work is half-done, and this is known right now. Specifically, half of the tests of this feature have been left unliquified, and we already know that those tests reveal that the sample config files have been left unquified. This is understandable, because, unlike all other city tests, these (by virtue of being config parser tests shared by all the cities) are not right next to the city implementation, but hidden away elsewhere.

I am somewhat surprised that you are not aware of this, given that this is *exactly* the issue that you and I were working on 24 hours ago in the config tests. At that point we identified the work that is already done in this PR, as an inherent part of the work that we did not have time to complete. These two bits of work:

1. liquifying the parsing of `event-range`
2. testing it in the liquified command line parsing tests

form a unit. It makes no sense to put these in separate PRs.

This PR does step 1. It does not do step 2, therefore it is too early to approve and merge.

I appreciate that in the last few days we have been pushing PRs through hurriedly, in order to unblock progress on further work. The only thing that directly depends on this PR is the work that is missing from this PR, so there is no point in rushing it through half-done.

Once this PR is complete, my `goodbye-solid-cities` branch can be rebased on top of it, and merged, because the only thing missing in that branch is the *complete* implementation of this PR. This work does not belong in `goodbye-solid-cities` because the latter should start at a point where we have solid and liquid implementations which are equivalent. Without a *complete* implementation of this PR, these prerequisites are not satisfied.

In summary, it would be wrong to merge this PR in its current state.

The only question that remains is whether the CLI `event-range` tests can be liquefied without liquifying the sample config files. Ideally this should be done in 2 PRs, but it might be significantly easier to do it all in one go.

gonzaponte:
> It should complain in both cases: the cities have named parameters now, with no default values, and no *args or **kwds, so you have to pass exactly the correct set of arguments.

That's correct. I thought the implementation was different.

> BTW, I think that we should force them to be passed as keyword arguments, i.e. add a *, to the front of the parameter list of each city.

I agree. Although it wouldn't make a difference when running from the command line.

> Are dorothea_kr_with_barycenter_low_threshold_mc_example.conf and the like relevant to liquid cities, or should they disappear?

They are probably there for illustration purposes. It is not clear to me whether they are useful or not because, at the end of the day, the configurations are data-dependent, so it may be misleading. I think I would keep them, just in case we forget in the future how to build this type of configurations.

jacg:
> I think I would keep them, just in case we forget in the future how to build this type of configurations.

Are they compatible with liquid cities?

gonzaponte:
> Are they compatible with liquid cities?

No. But they are not far from being compatible. I can take care of that if you want.

gonzaponte:
Also, the config files themselves set input files that are not in the repository, so I would change that.

jacg:
I haven't looked at the code in detail, but it seems to do what it is supposed to. I have rebased `goodbye-solid-cities` on top of it, and am about to open a PR. It contains some problems, some of them relating to the config file issues discussed above. Feel free to fix them.

gonzaponte:
I've just reviewed the files again and I've changed my mind. Those named `dorothea_kr_with_corona_(low/high)_threshold_mc_example.conf` don't use the corona algorithm. I would delete these and make a PR to introduce proper ones, if needed. I don't think `dorothea_kr_with_barycenter_low_threshold_mc_example.conf` provides any (relevant) insight on the functionality of the barycenter algorithm, so I would be comfortable removing it. Finally, `dorothea_with_corona.conf` seems useful as it provides a *real* example of a configuration for the `corona` algorithm.

gonzaponte:
The config-file stuff seems orthogonal to this PR, should I open a new open with that?

jjgomezcadenas:
Yes, the config file can go into a separate PR.

gonzaponte:
I've restored the tests concerning the hierarchical configuration as discussed in #565. Is this PR ready, @jacg?

jacg:
> Is this PR ready, @jacg?

I don't think it would be appropriate to address minor details at this point. It seems to do what it's supposed to, specifically it provides what #565 needs. I vote for merging it.

gonzaponte:
Ok, I will merge it. I would appreciate knowing those minor points to address them in the future, provided you have the time to write them.

PR #563, opened 2018-10-21
Fix print that should be a warning

gonzaponte:
This is an LFS version of #536.

This might not be implemented as the liquid cities don't have this mechanism (which may be wrong).
The PR is open so the topic does not fall into oblivion.


jjgomezcadenas:
@gonzaponte can you please report on the status of this?

gonzaponte:
Unclear how to proceed. The original PR affected solid cities. With the implementation of the liquid cities, we've skipped a "safety check" that didn't seem to be appropriate. This check printed a message whenever we had problems reading a file and continued execution. However, the check also caught other scenarios and obscured potential problems.

I propose we close this PR and open a new one if we find ourselves again in a similar situation.

jjgomezcadenas:
Agreed, go ahead

PR #564, opened 2018-10-22
Safety check for liquid cities' sources

gonzaponte:
Introduce a safety check to the liquid cities sources to ensure all the accessible and relevant fields in the input file have the same length. This way we minimize the possibility of a mismatch among the data pushed through the pipes.


gonzaponte:
Do you have the time to take a look to this, @jacg?

jacg:
> Do you have the time to take a look to this, @jacg?

I'm afraid not. This would require me to string together more than 30s of careful, uniterrupted thought. Unfortunately I won't have that available to dedicate to IC for weeks to come.

jjgomezcadenas:
@gonzaponte this is a difficult one. Shall we work con peras to review it?

gonzaponte:
How & when? I’ll be on shift next week.

jjgomezcadenas:
OK, we can skype next week at a convenient time.

gonzaponte:
Ok, I'll be in touch and try to set a time depending on the work we have there.

jjgomezcadenas (approved):
Reviewed with peras, very good stuff, trademark of the house. Good job.

PR #565, opened 2018-10-22
Goodbye solid cities

jacg:
The solid cities are removed, and the liquid cities are put in their place. Thus, from now on, the liquid cities are not only the default cities but also the only cities.


jacg:
All tests now pass locally (but Travis isn't detecting the push, for some reason). The dorothea and penthesilea config files currently include `s12_selector.conf`. We were planning to remove the conf file inclusion mechanism. Do we still want to do this?

jjgomezcadenas:
Actually, in the case of a selector, the inclusion mechanism makes sense, so, I believe we should keep it precisely for this purpose.

jacg:
Two mysteries:

1. Tests failing on Travis Mac. Could a Mac owner please run the tests locally?

2. Travis is not picking up pushes. For example, at present the last commit in this PR is e83957675dc01458f85d645be932e702106ea298, but the Travis build (which has formally passed in this PR) is 7b526c21ce9b4d95c00f296463e73bb34c751023.

gonzaponte:
> Actually, in the case of a selector, the inclusion mechanism makes sense, so, I believe we should keep it precisely for this purpose.

If that is the case I will keep the tests related to hierarchical configuration in #562.

> Could a Mac owner please run the tests locally?

They pass on my machine

> Travis is not picking up pushes. 

That has been happening since yesterday's problems with github. Should be working now.

gonzaponte:
> 7b526c2.

This seems to be a merge commit?

jacg:
This needs to go on top of #562 and #567.

Apart from the FILE TRANSFER COMMIT, which needs squashing out, is there anything stopping #562 from being merged?

I'm out of touch with #567, how far is that from being approved?

As for this, we have Travis failures on Mac, even though all tests pass on real Macs. I propose that we don't let that slow us down. Travis also throws some hypothes warnings: likewise, I propose we don't let that stop us.

So, how far are we from merging these three PRs (#562, #567, #565)?

gonzaponte:
> So, how far are we from merging these three PRs (
> #562

I've just asked the same, I think it is ready. Unless you think we are missing something, we can go ahead and merge it.

> #567

Someone needs to review it. @jjgomezcadenas?

> #565 

I guess we need to make a mini-data production with solid and liquid to approve this. I will start to work on that asap.

gonzaponte (approved):
Good news, I've run over ~1000 events in MC and data and the liquid cities match exactly the output of the solid ones. Let's thank the solid cities for their service since the first days of the *invisible cities* and welcome our flowing metropolis.

jacg:
Are we waiting for #567, or shall I go ahead and merge this?

gonzaponte:
Go ahead. #567 is rebased on top of this one.

jacg:
BTW, reading your approval message ... have you got designs on the *poet in residence* post?

gonzaponte:
> have you got designs on the poet in residence post?

I am not quite sure whether I understood this correctly or not, but I just wanted to give you all my best JJ impression :)

PR #567, opened 2018-10-23
Review config files

gonzaponte:
Review config files:

- Remove the `drift_v` parameter from `s12_selector.conf`
- Introduce it in `dorothea.conf` and `penthesilea.conf`.
- Minor reword of `dorothea.conf`
- Define `barycenter.conf` and `corona.conf`
- Remove misleading config files


gonzaponte:
This is rebased on top of #565.

jacg:
I have rebased this on the merge of #565, but I'm not allowed to push. Have you enabled modification by reviewers?

gonzaponte:
That is weird... it is enabled by default and it is enabled now

gonzaponte:
Are you doing it form the old clone or with the one used for the migration? It seems one cannot push from the old one.

jacg:
I'm doing it from a worktree in a clone that I created *after* having had problems with LFS. It's a clone from which I've spun off many worktrees from which I have been pushing to all sorts of remote branches. Hmm, maybe all those remote branches have been on my fork or `nextic`, but not on another user's fork.

jacg:
Just tried it from a completely new clone. It worked.

gonzaponte:
Ok, mistery solved, but it seems a bit troubling...

jjgomezcadenas:
@paolafer can you please review this at your earliest convenience?

paolafer:
Yes, I'm aware of this.

paolafer (commented):
Nice job, needed to clean up the house a little bit.
I have a question regarding the drift velocity: at the moment, we are not using it at the level of hit reconstruction (penthesilea/dorothea), which, by the way, I consider misleading, since we produce hits with a property called Z, which is actually time. We change time to z in a following step, using the drift velocity from time evolution files (for data). If we decide to keep this way, I think we shouldn't mention the drift velocity in the config file at all at this stage, and keep it as an internal parameter set to 1. In a later stage, the city that makes the geometrical corrections could also make the transit from time to Z.
Alternatively, we could calculate Z here, but we would need to read the drift velocity somewhere.

gonzaponte:
I think it was introduced with the prospect of reading the drift_velocity from db in the future, but it was never implemented. We should add an issue with that so we don't forget about that and we can also move the discussion there, as it has nothing to do with this PR.

paolafer (commented):
Ok, I approve this PR and move the drift velocity discussion to an issue.

paolafer (approved):
I approve this PR.

gonzaponte:
I am waiting for #574. There will be conflicts and will solve them here.

PR #570, opened 2018-10-24
Fix diomira random seed

gonzaponte:
Allow the fixing of the random seed in `Diomira`.
Add a test to ensure this feature is not removed without notice.


jjgomezcadenas (approved):
Needed and straight forward

PR #571, opened 2018-10-24
Remove unnecessary raw data type parameter

gonzaponte:
The parameter `raw_data_type` is not needed, as each city reads always from the same type of data. It has been removed from all cities.

### DIFF
--- a/invisible_cities/cities/irene.py
+++ b/invisible_cities/cities/irene.py
@@ -47,8 +47,7 @@ from .  components import wf_from_files
 
 @city
 def irene(files_in, file_out, compression, event_range, print_mod, run_number,
-          n_baseline, raw_data_type,
-          n_mau, thr_mau, thr_sipm, thr_sipm_type,
+          n_baseline, n_mau, thr_mau, thr_sipm, thr_sipm_type,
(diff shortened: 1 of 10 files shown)


jjgomezcadenas:
Do you remember the original meaning of the parameter? I believe it was related with reading raw data or blr data, let's reflect a bit why it was there...

gonzaponte:
Well, as far as I know, *for the liquid cities* the parameter was there to use the same source to read MCRD wfs and RWF. However, since each city always reads the same type of wfs, it doesn't make sense as an external parameter. Perhaps we had the possibility of reading BLR wfs in the solid cities (not sure, I will take a look), but certainly not in the liquid cities. This can change, of course.

jjgomezcadenas:
Well, DIOMIRA read MCRD, doesn't she? Her role in life is to read MCRD and produce RWFs...

gonzaponte:
Exactly, `diomira` reads *always* MCRD and `irene` or `isidora` read *always* RWF, so it is not a parameter of the city, but a parameter for the reader, which can be fixed.

jjgomezcadenas (approved):
Parameter is not needed now, so it's OK to get rid of it.

PR #572, opened 2018-10-26
Upload database

carmenromo:
In this PR database is uploaded with new gains for SiPMs based on the calibrations from run 6358.

### DIFF
--- a/invisible_cities/database/localdb.sqlite3
+++ b/invisible_cities/database/localdb.sqlite3
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:0fcd78cdcc70257d02c3ca0390b553e8e5ff08176622d245b54cd8da966ce9be
-size 95653888
+oid sha256:1b2508140533736850ae6c2a0894e10a3984a23efd943cd4edf460ffbd858399
+size 95748096


carmenromo:
sipmrundifferenceplots

Here is the difference of the most important parameters with previous runs.

gonzaponte:
Two minor comments:
- I think it would be helpful to produce the plots with the same binning for both distributions
- Shouldn't the range of the mu distribution difference display also negative values?

jmbenlloch:
There were two failing tests checking the exact result produced by diomira and irene. Those results are dependant on the DB used. The run number was one to high (10000), meaning "the lastest one", but the meaning of that varies along time. I have fixed those run numbers to a particular value that gives the same calibration that was obtained at the time that those tests were created. Every new calibration will give slightly different results, so we have to be careful with this.

With my last commit all test should be passing. If that is the case, everything is ok from my side. @carmenromo you may want to address @gonzaponte 's comments, when everything is clear we can approve.

gonzaponte:
Good catch!

carmenromo:
sipmrundifferenceplots

The binning is fixed here

carmenromo:
> Shouldn't the range of the mu distribution difference display also negative values?

@andLaing do you know how can this be?

andLaing:
How do you make the histogram? Do you set bins or just use pyplot.hist?

If the second, it's just that in this case the one with more light is 'a' in 'a-b', if not, we could be chopping values.

In any case that plot is the least interesting since it's run dependent. It's useful to check that the run conditions are the same for two runs with the same LED settings but the gain and 1pe sigma are the parameters that we're uploading and the chi^2 checks that we're getting reasonable fits

carmenromo:
pmtrundifferenceplots

PMT plot comparison

carmenromo:
@jmbenlloch now you can approve, tests are fixed with #577

jmbenlloch (approved):
This PR includes the latest calibration for both SiPMs and PMTs and the issues with some tests has already been solved in another PR, therefore I approve.

PR #574, opened 2018-10-29
Add global S1 and S2 information to Penthesilea

mmkekic:
This naively adds the same kdst table as in dorothea. To be discussed:
1. Do we want to constrain penthesilea to only one S1? If not how do we handle multiple S1s?
2. The names of the groups (for now its RECO for hits and DST for overall peak information)
3 ...


msorel:
thanks @mmkekic !

1. DST (pointlike event) information in principle does allow for multiple S1s, right? I would keep it like that, one table row per (S1, S2) combination. In practice, it is probably always run with nS1=1 requirement anyway. 

As for RECO (hit info), I would not allow for duplication of hits for each S1 in the event, not to make file size too big. In case one wants to run penthesilea when multiple S1s are present, one possibility is that penthesilea does not crash but rather assign hit time/z information using first S1 in the event by convention.

2. The names RECO and DST are OK with me, to maintain convention up to now in dorothea and penthesilea. But I have no strong feelings here.

gonzaponte:
> DST (pointlike event) information in principle does allow for multiple S1s, right?

Not quite, we can run (it wouldn't crash) allowing multiple S1s, but we just take into consideration the first one. If we want to work with multiple S1s, then we need to either:
- Repeat all hits in each S2 peak for all possible S1 peaks.
- Allow the column Z to be a variable-length array, which I don't know how easy is to implement or if it is even possible with the current structure.
- Think of a different structure to store hits (not necessarily a crazy idea).

msorel:
Thanks @gonzaponte . While something to think about (there are analyses that would benefit from multiple S1s, for example all analyses involving internal Bi214), I do not think it is the most urgent to fix or develop right now. I'd rather have an analysis DST with event-level, hit and topological information soon (meaning days or weeks), even if it only works for a single S1/event. And we could rethink about the multiple S1 optimal approach in 2019

Just my two cents

gonzaponte:
I agree, allowing multiple S1s in Penthesilea would have a time scale of weeks rather than days.
Then this PR can focus and be limited to introducing global S1/S2 info in Penthesilea.

invisible_cities/cities/penthesilea.py
```diff
@@ -42,26 +43,34 @@
...
 
 @city
 def penthesilea(files_in, file_out, compression, event_range, print_mod, run_number,
                 drift_v, rebin,
                 s1_nmin, s1_nmax, s1_emin, s1_emax, s1_wmin, s1_wmax, s1_hmin, s1_hmax, s1_ethr,
                 s2_nmin, s2_nmax, s2_emin, s2_emax, s2_wmin, s2_wmax, s2_hmin, s2_hmax, s2_ethr, s2_nsipmmin, s2_nsipmmax,
-                qthr, qlm, lm_radius, new_lm_radius, msipm):
+                slice_reco_params  = dict(),
+                global_reco_params = dict()):
+    # global_reco_params are qth, qlm, lm_radius, new_lm_radius, msipm used for hits reconstruction
```
gonzaponte:
```suggestion
    # slice_reco_params are qth, qlm, lm_radius, new_lm_radius, msipm used for hits reconstruction
```

invisible_cities/cities/penthesilea_test.py
```diff
@@ -187,7 +207,7 @@ def test_penthesilea_exact_result(ICDATADIR, output_tmpdir):
     penthesilea(**conf)
 
     tables = (  "MC/extents", "MC/hits", "MC/particles", "MC/generators",
-              "RECO/Events")
+                "RECO/Events","DST/Events")
```
gonzaponte:
```suggestion
              "RECO/Events", "DST/Events")
```

gonzaponte (approved):
This PR adds global S1 and S2 info to Penthesilea. The chosen implementation simply introduces the Dorothea machinery within the Penthesilea pipeline. There is a caveat with this implementation: while Dorothea works fine for any number of S1s, Penthesilea does not and therefore only the 1 S1 configuration will produce the desired result. This will be addressed in a future PR.

The previous tests have been extended to include this new feature.

PR #577, opened 2018-10-30
Set proper run number exact result

gonzaponte:
Some tests used a high run number that is interpreted as "the most up-to-date values in the database". This has been changed so the tests do not depend on the status of the database. This is needed for #572 to pass the tests.


carmenromo (approved):
Reviewed with peras. Needed for #572.

PR #578, opened 2018-10-31
Address tests failing due to hypothesis health checks

gonzaponte:
Some tests have been failing recurrently because hypothesis' data generation is too slow, [as pointed out in](https://github.com/nextic/IC/issues/551#issuecomment-434364870) #551. Since we don't know how to improve data generation at present, we will suppress this specific health check in these tests.

This PR is going to remain open until all the tests of this kind are identified and fixed.

### DIFF
--- a/invisible_cities/reco/pmaps_functions_test.py
+++ b/invisible_cities/reco/pmaps_functions_test.py
@@ -1,6 +1,8 @@
 import numpy as np
 
 from hypothesis            import given
+from hypothesis            import settings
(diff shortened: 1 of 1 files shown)


gonzaponte:
It seems this is the only test that fails due to a HealthCheck. This can be reviewed.

jmbenlloch (approved):
Until we find a solution to generate the samples in a faster way, this is the only thing we can do to avoid infinite errors in travis.

PR #579, opened 2018-10-31
city esmeralda

mmkekic:
This is a suggestion for the new city structure. Some of, and not limited to, things to decide:
1. universal interface for correction maps 
2. filtering events - my idea was to pass all events that contain at least one non-NN hit
3. merging NN hits - guess we should stick to @paolafer  comment from  #520 , ie merging to same peak hits only. this can also be a configuration parameter.  
4. corrected hits output  format , notice that HitTable already contains fields for corrected energies. do we want to copy all information from penthesilea hits and add corrected hits or copy just non-NN hits?
5. other outputs of the city?


jjgomezcadenas:
Good stuff. Could you please use typehints in all functions? I would like to stick to typehints from now on (it really makes your life easier), and Esmeralda is a good place to start.

jahernando:
About the interface for correcting hits. This is a proposal, with two possibilities. Please @gonzaponte comment.

The idea is that the calibration can be done in one go for all hits in a pMaps.

1) `Ec, Qc = calibrate(E, X, Y, Z, Q, T)`
      input and output are numpy arrays.
      E  = PMT  pes
      Q = SiPM pes      (default None)
      T = time-stamp  (default None, actually there is no correction by time)
      it returns:
      Ec = corrected PMT pes (lifetime and geometry)
      Qc = corrected SiPMs pes (only if Q is provided)
      if Z is not provided (Z = None) it does only the geometrical calibration, not lifetime

2) `Ec = calibrateE (E,  X, Y, Z, T)`
    `Qc = calibrateQ(Q, X, Y, Z, T)`
     T by default is None 
     if Z is None, only geometrical correction is applied, not lifetime
      
Maybe the second is more elegant, and for the moment we are only calibrating E, while I suspect that we will need soon Q.

jahernando:
The structure of Esmeralda seems to me quite good, really. @mmkekic 

There are two algorithms issues that slightly worries me:
1) The NN-hits. They should be very few. But I am afraid that there could always be some NN-hits.
2) The Voxels. It looks to me could be again Hits, and store them as Hits, to avoid expanding our data-class and storing many data-types.

jjgomezcadenas:
I like option two. Is more modular.

PR #580, opened 2018-11-01
optimize  mc_info_writer last_row parameter update

mmkekic:
This  PR changes the update of last_row parameter in mc_info_writer, setting it to the value of the row number of the last written event, which speeds up writing of sparse filtered events

### DIFF
--- a/invisible_cities/io/mcinfo_io.py
+++ b/invisible_cities/io/mcinfo_io.py
@@ -125,7 +125,7 @@ class mc_info_writer:
         self.extent_table.flush()
 
         hits, particles, generators = read_mcinfo_evt(mctables, evt_number, self.last_row)
-        self.last_row += 1
+        self.last_row = iext + 1
 
         for h in hits:
             new_row = self.hit_table.row


msorel:
thanks @mmkekic , i could look at this next week. but perhaps @paolafer can already look at it tomorrow?

paolafer:
I find the solution to speed up the code, in case of filtered events, correct. The test suit that we have covers several cases of filtering, therefore it should be enough to ensure that the code keeps behaving correctly. However, @mmkekic, if you did some other, more complete test , which is not redundant with the ones we already have, to check that your results were correct, you could consider adding it to this PR.

mmkekic:
I checked that in my specific problems it gives the same result as before the change, but I didnt think of any conscise test that is not already included in the existing tests... @paolafer  if you have anything in mind I can try to work on it :)

paolafer:
Not really, @msorel was suggesting to do some further check, but nothing comes to my mind that isn't another variation of the existing ones. If you have a suggestion, Michel, I think that we should implement it as test, to strengthen the suit.

msorel:
OK. I will look at it in some detail on monday.

jjgomezcadenas:
A possible test, and very simple is like this: Write a bunch of events. with the old, non-opmitized function and time how long it takes. Repeat with the optimized function. Check that events are the same and also check that second timing is faster than first.

msorel:
i have checked this today. i am happy. in addition to passing all tests, i checked with another "real-life" example (within quotes because MC events,  of course, given the nature of this PR), along the lines of what suggested by @jjgomezcadenas: how MC truth is copied in a irene job of 200 background MC RWF events, with randomly-ordered (ie, non-increasing) evt_numbers. 

Given @mmkekic comment that this may work better on sparse filtered events, i have tweaked irene's configuration parameters to test both a high and low selection efficiency. I quote some numbers below.

- Before mc_info_writer_fix, high selection efficiency in irene
191 out of 200 events pass
time: real 0m39.258s  

- After mc_info_writer_fix, high selection efficiency in irene
191 out of 200 events pass, as above, and 191 passed events have same MC truth info as before fix
time: real 0m39.337s, compatible with time before fix within errors

- Before mc_info_writer_fix, low selection efficiency in irene
7 out of 200 events pass
time: real 0m25.416s

- After mc_info_writer_fix, low selection efficiency in irene
7 out of 200 events pass, as above, and 7 passed events have same MC truth info as before fix
time: real 0m26.877s, compatible with time before fix within errors

In summary, although in this particular application I cannot see any performance improvement, outputs are consistent and the code change actually makes sense. Hence, I will add a review now suggesting to approve this PR.

msorel (approved):
This PR optimises the way MC truth information is copied, avoiding to start an event loop from the first MC event in the input file for every MC event to be copied in the output. This particular event loop now starts from the last written event, as it should.

In addition to passing all tests, it has been checked on a variety of MC files, iincluding on events with randomly-ordered evt_numbers, that MC information is identical before and after the fix.

In terms of performance, I could not find a performance improvement in the particular application I tested (irene), but I have not seen a deterioration either. It is quite possible that on different MC input files and data processing streams a significant improvement can be found. 

To conclude, I would merge this as a potentially useful optimisation for MC processing.

PR #581, opened 2018-11-02
Filter empty pmaps

gonzaponte:
Add a filter for empty pmaps in `irene`. The lack of this filter creates a mismatch between the different tables in the file, producing output for mc tables or the event number table, but not for the pmaps table. This causes next stages of processing to fail.

### DIFF
--- a/invisible_cities/cities/irene.py
+++ b/invisible_cities/cities/irene.py
@@ -93,8 +93,10 @@ def irene(files_in, file_out, compression, event_range, print_mod, run_number,
 
     ### Define data filters
 
-    # Filter events with zero peaks
+    # Filter events without signal over threshold
     empty_indices   = fl.count_filter(check_nonempty_indices, args = ("s1_indices", "s2_indices"))
+    # Filter events with zero peaks
+    empty_pmaps     = fl.count_filter(check_empty_pmap      , args = "pmap")
 
     event_count_in  = fl.spy_count()
     event_count_out = fl.spy_count()
@@ -124,6 +126,7 @@ def irene(files_in, file_out, compression, event_range, print_mod, run_number,
                                 empty_indices.filter,
                                 sipm_rwf_to_cal,
                                 compute_pmap,
+                                empty_pmaps.filter,
                                 event_count_out.spy,
                                 fl.fork(write_pmap,
                                         write_mc,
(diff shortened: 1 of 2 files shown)


jacg:
I dunno, it all seems too easy with these liquid cities. I look at the diff, and it all makes sense immediately, without any effort. Where's the challenge? I think we need to go back to the solid ones. How can we maintain our self-respect if it's this easy?

gonzaponte:
to the solid cities? I vote we go back to ROOT.

jjgomezcadenas:
SOLID Root!!!

gonzaponte:
@jjgomezcadenas proposes to extend this PR by including a table with the filtered events. While working on it I realized that, despite not being a large number of changes, it touches many files and becomes a bit orthogonal to the original one. I propose we leave this one as it is and start another one *straight away* to implement the writing of the filtered event numbers. Do you agree?

paolafer:
I'm sorry, I'm missing something here. If an event is filtered, with the current code it shouldn't be written in the MC output table. Also, we have been processing penthesilea and dorothea after irene with no crash, what is failing?

gonzaponte:
> with the current code it shouldn't be written in the MC output table

The thing is that we were only filtering those events without signal over threshold and we never checked whether the resulting pmap (which includes some requirements on the peaks) is empty. Thus, an event with signal over threshold for which we don't find any peaks satisfying the requested conditions passes to the sink stage of the pipeline. When the `pmap` writer tries to write an empty pmap it just does nothing, but the other writers do.

> we have been processing penthesilea and dorothea after irene with no crash, what is failing?

With the liquid cities? Not exhaustively. This occurs because #564 (which checks that the input data makes sense) wasn't yet implemented.

paolafer:
Ah, ok, this is an issue of liquid cities, that wasn't there with the solid ones, is that right? I was just wondering why we didn't see the issue before...

gonzaponte:
> Ah, ok, this is an issue of liquid cities, that wasn't there with the solid ones, is that right? I was just wondering why we didn't see the issue before...

Not sure this wasn't an issue for the solid ones as well... I don't we have a test for that. The thing is that it doesn't affect "anything serious". It creates a mismatch between the different structures in the file, but we can read them anyway. With the liquid cities the problem becomes evident.

jjgomezcadenas:
Yes agreed.

gonzaponte:
Can you review? :)

jjgomezcadenas (approved):
As agreed we approve this needed PR and throw away PMAPS that do not pass the selection filter.

PR #575, opened 2018-10-30
Fix `test_inverse_cdf_hypothesis_generated`

gonzaponte:
#555 didn't fix all problems.
This will address [the problem](https://github.com/nextic/IC/issues/551#issuecomment-434222203) detailed in #551.

PR #582, opened 2018-11-02
Temporarily skip `test_inverse_cdf_hypothesis_generated`

gonzaponte:
While we find a fix for it (#575).


gonzaponte:
We skip this test to avoid frequent Travis failed builds. Since #575 is open, the chances of forgetting this test are minimal.

@jjgomezcadenas, please approve whenever you can.

jjgomezcadenas (approved):
Troublesome test momentarily skipped while investigating.

PR #583, opened 2018-11-03
Store filters results in output file

gonzaponte:
Following #581, we implement the addition of the result of data filters to file in such a way that we can always know where a given event has been rejected.


jjgomezcadenas (approved):
Implements a needed feature, which allow us to keep all events adding vectors of filters. Clearly coded in with an extensive test suite. Good job.

gonzaponte:
I will just extend the tests `test_{city}_exact_result` to include the check for these new tables and then merge.

PR #585, opened 2018-11-05
Add hits to voxels

paolafer:
This PR adds to the `Voxel` class the `hits` attribute, which contains the list of the hits that belong to a certain voxel. In the initialization of the class, this list is empty by default, since we can use voxels that are not related to any hits (for instance, voxels coming from RESET).
In the `voxelize_hits()` function of paolina, the calculation of which hits fall in which voxel has been added, and voxels are created with their hits.
Two tests have been added: the first one checks that the energy of the voxel is equal to the sum of the energies of the hits that fall inside the voxel. Another option would have been to calculate the energy of the voxels from the energy of the hits directly in the code, but I preferred to have it this way, to keep the check that the hits are assigned correctly.
The second test takes a specific case, where the hits fall on voxel boundaries, to check that they are assigned correctly.

### DIFF
--- a/invisible_cities/evm/event_model.py
+++ b/invisible_cities/evm/event_model.py
@@ -181,7 +181,7 @@ class MCHit(BHit):
 
 class Voxel(BHit):
     """Represents a Voxel"""
-    def __init__(self, x,y,z, E, size, hits):
+    def __init__(self, x,y,z, E, size, hits=[]):
         super().__init__(x,y,z, E)
         self._size = size
         self.hits = hits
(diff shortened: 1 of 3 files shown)


ausonandres (approved):
I approve this PR. Very useful and relevant.

paolafer:
I've added a new test, to check the exact equality of voxel hits and original hits.
I've also changed the voxel hits to be a `list`instead of a `numpy.array`, in the `Voxel` construction.

paolafer:
This PR has been approved by @ausonandres. Can we proceed with the merging?

PR #589, opened 2018-11-06
Fix Matplotlib version to 3.0.1

jmbenlloch:
There is a bug in matplotlib that makes it incompatible with python 3.7.1. https://github.com/matplotlib/matplotlib/issues/12601

Some Mac users got python 3.7.1 after running `source manage.sh install_and_check 3.7` and many test failed. So until the bug is fixed and we have a better way of controlling the exact python version we are using, we should fix it in the environment specification, which is what this PR does.

### DIFF
--- a/manage.sh
+++ b/manage.sh
@@ -72,7 +72,7 @@ function install_conda {
     fi
 }
 
-CONDA_ENV_TAG=2018-10-20
+CONDA_ENV_TAG=2018-11-07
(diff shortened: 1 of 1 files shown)


gonzaponte:
Matplotlib's [v3.0.1](https://github.com/matplotlib/matplotlib/releases/tag/v3.0.1) is supposed to fix that already.

jjgomezcadenas:
OK, so please, change the PR to a bump to 3.0.1

jjgomezcadenas (approved):
A needed upgrade, to avoid a matplotlib bug

PR #594, opened 2018-11-12
Rename calibration cities

carmenromo:
With this PR sipm and pdf calibration cities names are changed:
- Moriana becomes Trude
- Zemrude becomes Berenice


paolafer (approved):
This PR is quite straightforward, changing the name of two cities in all the places where they appeared in the code. I don't see why the proposal of changing phyllis for fillide was discharged some time ago by a grumpy @andLaing with the reason that it was too difficult to change it all over the code :-P :-P :-P ... 
Anyway, I approve this PR. Please, proceed with the rebase, @carmenromo.

carmenromo:
Thanks @bpalmeiro, I was doing the same but you were faster ;)

PR #475, opened 2018-06-05
Allow multiple s1 events to be stored in Dorothea

gonzaponte:
This PR adds the possibility of include multiple-S1 events in the kDST produced by Dorothea.
Useful to study new data.

PR #596, opened 2018-11-13
Fix Z-related assignments in case of an XYRecoFail

gonzaponte:
The assignment of the variables related to the longitudinal coordinate was not properly done in #475 as they don't take into account the possibility of multiple S1s.


jmbenlloch (approved):
Simple fix, approved.

PR #598, opened 2018-11-15
Add sipm hits table to penthesilea output

mmkekic:
@paolafer checked that by fixing some corona parameters one can achieve 1 hit per sipm table. Maybe @jahernando can check that his code gives the same result (or send me one example of files: pmap, output_hit_table and threshold value, so I can check it), and afterwards I can add the tests.

### DIFF
--- a/invisible_cities/cities/penthesilea.py
+++ b/invisible_cities/cities/penthesilea.py
@@ -51,27 +51,42 @@ def penthesilea(files_in, file_out, compression, event_range, print_mod, run_num
                 drift_v, rebin,
                 s1_nmin, s1_nmax, s1_emin, s1_emax, s1_wmin, s1_wmax, s1_hmin, s1_hmax, s1_ethr,
                 s2_nmin, s2_nmax, s2_emin, s2_emax, s2_wmin, s2_wmax, s2_hmin, s2_hmax, s2_ethr, s2_nsipmmin, s2_nsipmmax,
-                slice_reco_params  = dict(),
-                global_reco_params = dict()):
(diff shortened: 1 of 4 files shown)


mmkekic:
Since we have 2 hit tables we should rename it. The proposals for now are :
1) - HITS (for 1 hit per sipm ) and 
    -CHITS (for old-style corona reconstruction)
or 
2)  - CORONA_HITS
     - SINGLE_HITS  ;   that are slightly more informative

If nobody says anything I will randomly choose one of the 2 options

jjgomezcadenas:
Are we sure about the two tables? the more I think about it, the less I like it.

paolafer:
I'm a bit against SINGLE_HITS for one-hit-per-SiPM, since when I read it I thought  immediately to 1 hit per slice.
I propose something more like SIPM_HITS.

gonzaponte:
Useful and short names should be used, if possible. I vote in favor of sipm_hits, as proposed by @paolafer:
- not too long
- the meaning is obvious to me (but maybe not to others)

The other could be corona_hits, cluster_hits, slice_hits, or whatever is useful to remind the physicist that it comes from that specific type of reconstruction.

> Are we sure about the two tables? the more I think about it, the less I like it.

Plase explain, not sure what you mean or what the other option is :)

jahernando:
I am also with @gonzaponte, sipm_hits and cluster_hits, are in fact what they are.

mmkekic:
Hi, after talking to @jjgomezcadenas we agreed to keep only one table in the output since sipm_hits can be achieved by setting radius to 0 and msipm to 1 in corona.

PR #601, opened 2018-11-28
Upload database

carmenromo:
Database is uploaded with new gains for  SiPMs based on the calibrations from run 6477, as well as pdfs for MC. Channels 12005 and 12048 have been masked.


carmenromo:
gainvalues_r6506
sipmrundifferenceplots_

Plots with the absolute gain values for run 6506 and with the comparison with other stable runs.

jmbenlloch (approved):
Includes the latest calibration.

PR #603, opened 2018-12-01
Fix voxels with no hits

paolafer:
This PR fixes an issue with the way numpy builds histograms, which we use to create voxels.
@mmkekic found out that changing the precision of the hit positions to `float64` solved this particular problem. I've also added a test to show the problem in one particular event.

### DIFF
--- a/invisible_cities/database/test_data/test_voxels_with_no_hits.h5
+++ b/invisible_cities/database/test_data/test_voxels_with_no_hits.h5
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:2dc490dfcc55526d6634019bdb5da63e5503beb8a6940d8d9865485709dcae9b
+size 1772351
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -55,7 +55,7 @@ def voxelize_hits(hits             : Sequence[BHit],
     voxel_edges_lo -= eps
     voxel_edges_hi += eps
 
-    hit_positions = np.array([h.pos for h in hits])
+    hit_positions = np.array([h.pos for h in hits]).astype('float64')
     hit_energies  =          [h.E   for h in hits]
     E, edges = np.histogramdd(hit_positions,
                               bins    = number_of_voxels,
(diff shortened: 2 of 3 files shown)


jacg:
I suspect that increasing the precision merely shuffles the problem along a little bit. Eventually an event will come along which fails even with 64 bit precision. What then? Switch to 128 bit?

paolafer:
I don't quite understand why this is an issue of precision, since in the particular event I posted as example it didn't seem to me that hits were falling on the border between two voxels. Maybe @mmkekic can shed some light on the issue...

mmkekic:
The problem comes in this line in numpy histogramdd:
`on_edge = (sample[:, i] == edges[i][-1]) ` ;
it wrongly identifies the hit to be at the right most edge and shifts it to the previous bin. The problem comes from comparing sample (that is np.float32) with edges (np.float64) but I am not quite sure how does the comparison work. Here is a concrete example, when using == to compare the array with the scalar it gives True, while using scalar to scalar it gives False (the difference between 2 values is -3.0127011996228248e-12). 
`sample[:,0].dtype, edges[0].dtype`
(dtype('float32'), dtype('float64'))

`sample[-4,0]-edges[0][-1]`
-3.0127011996228248e-12

`[sample[-4,0]]==edges[0][-1]`
array([ True])

`sample[-4,0]==edges[0][-1]`
False

I believe it is safe to use histogramdd if we convert the data to float64  cause then all the arrays will have the same precision and we should not have round off inconsistency, but maybe someone more expert in numpy can comment.

jjgomezcadenas:
@paolafer, are u done here? If so, please rebase and let's review with peras.

paolafer:
Rebase done. Who is going to peer review this PR, @jjgomezcadenas ?

paolafer:
@jjgomezcadenas, we were supposed to peras-review this PR together, if I'm not wrong. When could we have it done?

jjgomezcadenas:
@paolafer please rebase, then I will review

paolafer:
Rebase done.

jjgomezcadenas (approved):
The problem seems identified and fixed, with a test demonstrating that it works.

jjgomezcadenas:
@paolafer PR approved. @carmenromo please merge.

PR #604, opened 2018-12-07
Calculate SiPM signal to noise

andLaing:
First version with the functions added to calculate
signal to noise for the sipms.

Currently only uses the mean of the appropriate
distribution as an estimator for dark current but
will be generalised to use other definitions.

Addition as an option in reconstruction cities
to be added.

One new test failing (commented out at moment)
because of float accuracy but passes in interactive.

### DIFF
--- a/invisible_cities/core/random_sampling.py
+++ b/invisible_cities/core/random_sampling.py
@@ -199,6 +199,8 @@ class NoiseSampler:
         pad_xbins = pad_pdfs(self.xbins)
 
         if dark_model == DarkModel.threshold:
+            pdfs = np.apply_along_axis(normalize_distribution, 1,
+                                       self.mask(pdfs))
             dark_pes = general_thresholds(pad_xbins, pdfs, 0.99)
             dark_pes[self.active.flatten() == 0] = 0
             return dark_pes
--- a/invisible_cities/core/random_sampling.py
+++ b/invisible_cities/core/random_sampling.py
@@ -1,6 +1,7 @@
 import numpy as np
 
 from functools import partial
+from functools import lru_cache
 
 from .. database import load_db as DB
 
@@ -30,6 +31,18 @@ def inverse_cdf(x, y, percentile):
     if not y.any():
         return np.inf
     return x[inverse_cdf_index(y, percentile)]
+
+
+def pad_pdfs(bins, spectra=None):
+    bin_diff = np.count_nonzero(bins <= 0) - np.count_nonzero(bins > 0)
+    pad = abs(min(bin_diff, 0)), abs(max(bin_diff, 0))
+    if spectra is None:
+        ## We just want to extend the bins
+        return np.pad(bins, pad, 'linear_ramp',
+                      end_values=(-bins[-1], -bins[0]))
+
+    return np.apply_along_axis(np.pad, 1, spectra, pad,
+                               'constant', constant_values=0)
 
 
 class NoiseSampler:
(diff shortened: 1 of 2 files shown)


jjgomezcadenas:
@andLaing shall I review this PR? If so confirm and rebase

andLaing:
Would be good.

There is one algorithmic point I'm not 100% sure on, if you don't notice it yourself I'd like to talk about it

jjgomezcadenas (commented):
```
def pad_pdfs(bins, spectra=None):
    bin_diff = np.count_nonzero(bins <= 0) - np.count_nonzero(bins > 0)
    pad = abs(min(bin_diff, 0)), abs(max(bin_diff, 0))
    if spectra is None:
        ## We just want to extend the bins
        return np.pad(bins, pad, 'linear_ramp',
                      end_values=(-bins[-1], -bins[0]))

    return np.apply_along_axis(np.pad, 1, spectra, pad,
                               'constant', constant_values=0)
```
This function is pretty cryptic to me. It would help a more descriptive name, a doc string and type hints.

jjgomezcadenas (commented):
Sorry, I was looking at an old version of the code, this contains doc string and type hints. 

I guess you can remove commented out code:

```
def pad_pdfs(bins, spectra=None):
    bin_diff = np.count_nonzero(bins <= 0) - np.count_nonzero(bins > 0)
    pad = abs(min(bin_diff, 0)), abs(max(bin_diff, 0))
    if spectra is None:
        ## We just want to extend the bins
        return np.pad(bins, pad, 'linear_ramp',
                      end_values=(-bins[-1], -bins[0]))

    return np.apply_along_axis(np.pad, 1, spectra, pad,
                               'constant', constant_values=0)
```

jjgomezcadenas (commented):
Please type-hint (and if relevant doc-string):

```
def normalize_distribution(y):
    ysum = np.sum(y)
    return y / ysum if ysum else y


def sample_discrete_distribution(x, y, size=1):
    if not y.any():
        return np.zeros(size)
    return np.random.choice(x, p=y, size=size)


def uniform_smearing(max_deviation, size=1):
    return np.random.uniform(-max_deviation,
                             +max_deviation,
size = size)
```

jjgomezcadenas (commented):
```
def normalize_distribution(y):
    ysum = np.sum(y)
    return y / ysum if ysum else y
```
this is normalization to the total sum (other normalization choices are available, such as nor to max..). should it be reflected in name of function?

jjgomezcadenas (commented):
```
def sample_discrete_distribution(x, y, size=1):
    if not y.any():
        return np.zeros(size)
    return np.random.choice(x, p=y, size=size)
```
Who is being sampled and from whom? A doc string and type hints will help here

jjgomezcadenas (commented):
```
def sample_discrete_distribution(x, y, size=1):
    if not y.any():
        return np.zeros(size)
    return np.random.choice(x, p=y, size=size)


def uniform_smearing(max_deviation, size=1):
    return np.random.uniform(-max_deviation,
                             +max_deviation,
                             size = size)
```
This two functions are anyways nothing but almost-trivial wrappers around numpy... is it worth? I believe that we could (should?) always expose numpy interface, since numpy is the facto our "python array standard"

jjgomezcadenas (commented):
```
def pad_pdfs(bins : np.array,
             spectra : np.array) -> Tuple[np.array]:
    """
    Pads the spectra to a range of
    bins = (-100, 100) according to
    the bin widths.
    """
    bin_width = np.round(np.diff(bins)[0], 5)
    pad = (int(abs(-100 - bins[0]) // bin_width),
           int(abs(100 - bins[-1]) // bin_width))

    bin_min     = np.round(bins[ 0] - pad[0] * bin_width, 5)
    bin_max     = np.round(bins[-1] + pad[1] * bin_width, 5)
    padded_bins = np.pad(bins, pad,      'linear_ramp',
                         end_values=(bin_min, bin_max))

    padded_spectra = np.apply_along_axis(np.pad, 1, spectra, pad,
'constant', constant_values=0)
```
Confused by the number "5", which appears hard-wired... what is it? Should we make clear its meaning?

jjgomezcadenas (commented):
```
class NoiseSampler:
    def __init__(self, detector, run_number, sample_size=1, smear=True):
        """Sample a histogram as if it was a PDF.
```
Add type hints, please...

jjgomezcadenas:
@andLaing I have left a number of (pretty straight-forward) comments from a first reading of the code. There is quite a lot of intelligent stuff in it and I like it. It is also extensively tested but would it be possible to prepare a demonstration NB? 

Also, please, rebase!

Thanks!

andLaing:
I'll get to this later on, probably.

The only general reply to your comments is that they are all, with one exception, about existing code which has nothing to do with the pull request.

jjgomezcadenas:
Goops, I was looking directly at the files involved in the PR, didn’t realise that some of the code was possibly already “in”. Sorry.

Enviado desde mi iPhone

jjgomezcadenas:
@andLaing thanks. Code reads better now. What about a demo Notebook?

andLaing:
notebook testing the signal_to_noise function in https://github.com/andLaing/scripts.git

called Signal_to_noise_check.ipynb

jjgomezcadenas:
Thanks, In the second round of plots the first panel (rebined threshold) looks pretty different left to right. Any obvious explanation?

Can you produce a plot of number of SiPMs vs S/N cut (and versus Raw Signal), to illustrate the effect of the cut?

andLaing:
> Thanks, In the second round of plots the first panel (rebined threshold) looks pretty different left to right. Any obvious explanation?

One's raw charge, the other's S/N as the titles say.

> Can you produce a plot of number of SiPMs vs S/N cut (and versus Raw Signal), to illustrate the effect of the cut?

There aren't any cuts

jjgomezcadenas:
>One's raw charge, the other's S/N as the titles say.

I had understood that but I was expecting similar patterns (like in the other panels). Why is this so different?

>There aren't any cuts
Yes, I am suggesting that you make a new plot as a function of a cut either in raw signal or in S/N.

andLaing:
> I had understood that but I was expecting similar patterns (like in the other panels). Why is this so different?

There is a typo in the non_zero selection for that set of plots. Sorry

andLaing:
Updated notebook

jjgomezcadenas:
Very nice. The last round of plots in the NB show very nicely the effect of the cuts in S/N.

jjgomezcadenas (approved):
Functions to compute SiPM s/N added to the extensive suite already available, with tests and demo NB. Good job.

jjgomezcadenas:
@carmenromo please rebase

carmenromo:
ok!

PR #593, opened 2018-11-08
Fix peak width

gonzaponte:
This is a fix to #584.

PR #607, opened 2018-12-17
Threshold rebin

andLaing:
A new rebinning function based on a minimum
number of photo-electrons in the sum of the
pmts.

Shouldn't be approved before #593 since there
is likely clashes.


jjgomezcadenas:
@andLaing do you want me to review this PR? If so, please rebase

andLaing:
That'd be great. just pushed a rebased version

andLaing:
Now rebased on the last merge with #593 with all clashes removed

jjgomezcadenas (approved):
A new method to rebin S2 based in charge threshold, rather than in a fix time interval. It offers the appealing advantage of virtually eliminating the zero-charge slices, which appear in fix-time-bin methods. The method is extensively tested.

PR #609, opened 2019-01-11
Eliminate not-used import.

paolafer:
I've eliminated a matplotlib import, which wasn't used, because it caused a segmentation fault with a particular IC environment in the majorana machine.


jmbenlloch (approved):
Matplotlib sometimes can cause issues in headless machines, as has happened to @paolafer while trying to run penthesilea. By removing this import we avoid that situation in the future.

PR #610, opened 2019-01-11
Fix arguments in sensor functions

carmenromo:
While using `sensor_functions` I realized that there were two minor argument hints and bugs that nevertheless confused me. The first one refers to the hint of an argument and the second one is an unused argument. This PR fixes both.


andLaing (approved):
Simple changes cleaning up some things left over by accident from old PRs

PR #612, opened 2019-01-18
Esmeralda city resurrecting  

mmkekic:
The proposed structure for the city


jahernando:
Hi Marija,

First, the structure of the City looks good to me! good job! Here there are some comments related with names and classes, nothing serious, the only worry is about the need of passing the kdsts. 

There is the need to define a better name for the classes that represent the persistency of the data, for example; *track_info_in_dsk*, *hit_info_in_disk*, *event_summary_in_disk*, or something like that.

I have an issue with *hits_and_kdsts_from_file*. What functions in the pipes will take the *kdst*? In fact I want to open the question of what we need from the kdst. I will say that only the S1 information (for the moment is stored in the kdst). The rest of the variables we want to store in *event_summary_in_disk* we will be computed from the *TrackCollection* (Do we have such a class?) and in *HitCollection*. 

I will propose that hits_and_kdsts_from_file will be more like hits_and_event_from_file, and yield:
i) a HitCollection ii) run_info class iii) S1-info class (just a simple container, S1e, S1time, number S1s). To be discussed. The test will need to be changed accordingly.

A question of taste, the name of *extract_track_blob_info* should be more  better something like *reconstruct_tracks_and_blobs*.

I am reluctant with the fact the *extract_track_blob_info* is modifying the class *HitCollection*, to set the Paolinas energy of the hit and the ID of the track (something else?). I am not sure what is the elegant solution.

mmkekic:
About the names - I completely agree, they are not very elegant and should be changed. 
About the classes where we store track info and event summary, I am not sure if we should make it a dataclass (that has a nice type hint) or a simple pandas DataFrame (that is easy to manipulate). I vote for a dataclass because it is more transparent.
I also agree that we only need S1 information that is stored in kdst table of the file. We can probably use pytables read_where, but since the kdst table is anyway small maybe is better to read it whole using IC read_hdst function. 
Now, about modifying hits table, I am not sure what is the other option if we want to have only one hits table in the output file that contains information about tracks. We should return a copy of hits table, but it still means that extract_track_blob_info will have to output both HitCollection and track_summary. 
Even uglier is that the function has to threshold the hits and the operations of correction and threshold dont necessary commute, so the hits should be corrected again inside the function using uncorrected energies... Lets discuss this on tomorrows update meeting :)

jjgomezcadenas:
1. About the names, I propose something along the line: track_persistency, hits_persistency and the like. 

2. Concerning the track info and event summary.  I think the crucial question is how those objects will be manipulated down the line. It appears to me that they are bona fide DataFrames. I like very much type hints, but I like better the added functionality of the DF.

3. Agree, the correction of hits is an ugly mess. We need to discuss it.

paolafer:
In the current version of the code I'm applying the z-width correction *track by track* and not over all the hits of an event. I think that this makes more sense, but the factor provided by Josh is extracted using all the hits of the events. Moreover, we are currently having different factors for different energies, which couldn't be used in the current code. I think we should discuss how to extract and to apply this correction.

mmkekic:
Looking at the writers I realized it is not trivial to combine pytables tables and pandas dataframes in the same file, because the way the flow is done the file we write to is kept open during the dataflow and pandas hdf writer can not write to pytables.open_file directly, so seems that we do have to transform pandas dataframes to pytable tables

jjgomezcadenas:
>pytables.open_file directly, so seems that we do have to transform pandas dataframes to pytable tables

This should not be a problem, we can always transform then back to DF when we read them, no?

mmkekic:
> > pytables.open_file directly, so seems that we do have to transform pandas dataframes to pytable tables
> 
> This should not be a problem, we can always transform then back to DF when we read them, no?

Sure, load_dst  from dst_io already does that

jjgomezcadenas:
So, no problem

Enviado desde mi iPhone

jahernando:
in *Add track sorting by energy*, you pass a on-the-fly function to compute the energy of the track, but I guess that this function can be used in other parts of the code or the analysis, better to include that function in a module, with a clear name `track_sum_voxels_energy(track)`

jahernando:
*Rename `track_blob_info_extractor`* OK

*Rename and reorganize config params in esmeralda tests* Ok But for my own understanding what does the `same_peak` parameter for hit-correction configuration mean?

*Rename and reorganize config params in esmeralda config.* OK

*Rename and reorganize config parameters in esmeralda city input.* OK. There are some code lines to long to be displayed, please put code in different lines to make it more readable.

*Add exact track comparison in esmeralda_test.* Does this test of equal Esmeralda output depend on the DB. Imagine that the DB for this data changes that could affect the results, but maybe Esmeralda is still working fine. Should this test include a fix DB (maybe already does it).

*Add fixtures for esmeralda tests.*  OK

mmkekic:
> _Rename and reorganize config params in esmeralda tests_ Ok But for my own understanding what does the `same_peak` parameter for hit-correction configuration mean?

It is the parameter when merging NN hits, it means that we can only associate the energy of NN hits to the closest hit from the same peak. Maybe we should change the name in config params?

 
> _Add exact track comparison in esmeralda_test._ Does this test of equal Esmeralda output depend on the DB. Imagine that the DB for this data changes that could affect the results, but maybe Esmeralda is still working fine. Should this test include a fix DB (maybe already does it).

In principle we dont use the database, but we do read information from the correction map itself : the minimum/maximum x,y that we apply the corrections to, that will obviously depend on the size of the detector but for now this information is stored inside the maps themselves.

invisible_cities/cities/esmeralda.py
```diff
@@ -188,10 +188,10 @@ def make_event_summary(event_number : int, timestamp : int, topology_info : pd.D
     S2q0 = np.sum(kdst.S2q.values)
     S2qc = max(-1, sum([h.Qc for h in paolina_hits.hits]))
 
-    x_avg = sum([h.X*h.E for h in paolina_hits.hits])
-    y_avg = sum([h.Y*h.E for h in paolina_hits.hits])
-    z_avg = sum([h.Z*h.E for h in paolina_hits.hits])
-    r_avg = sum([(h.X**2 + h.Y**2)**0.5*h.E for h in paolina_hits.hits])
+    x_avg = sum([h.X*h.Ec for h in paolina_hits.hits])
```
mmkekic, line 303: @gondiaz pointed out that coordinates average was wrongly calculated (using h.E rather than  h.Ec), so this commit fixes this bug.

jahernando:
I just went through the esmeralda files. 

The structure of the city in `emeralda.py` is quite clear, the functions have meaningful names, and the flow is logic and clear. The code in the functions is also quite clear and readable. 

I will suggest to export some 'functions on tracks' to a module, to compute the total energy, and the average position. These functions could be useful for other clients and then we avoid to repeat code elsewhere.

There is some test `filter_events` and `out_of_map_hits` that have some very different thresholds values as output, why is that?

In the extraction of information from paulina, there are some functions: `find_extrema`, `blob_centers` and `blob_energy_and_hits` that seems to me related. Somehow, the output of find_extrema should be arguments of blob_centers. Maybe some re-organization can be in place there. But these changes are out of the scope of Esmeralda that can be approved as it is.

Finally, I see that the track information has no information about what is the S2 from which that trac was constructed. This information can be obtained from the zmin-zmax of the track. I wonder if this was discussed or not. 

In summary, this city is beautifully coded, and we can proceed with the merging and update of IC.

jahernando (approved):
I approve the current this Esmeralda branch.

The functions and the structure of the city are quite clear. This city is elegantly coded.

There are several test, in particular the output comparison with a references output of tracks, to ensure the correctness of the city in the future.

mmkekic:
> There is some test `filter_events` and `out_of_map_hits` that have some very different thresholds values as output, why is that?

They are manually  tuned to filter out events, it is very rare (if happens at all) that none of the hits are above typical values we use (30 pes). out_of_map hits are using toy hits input that contain hits at R>200 mm, showing that paolina filters out those events while they are still contained in RECO/Events table.

> I will suggest to export some 'functions on tracks' to a module, to compute the total energy, and the average position. These functions could be useful for other clients and then we avoid to repeat code elsewhere.

Im not sure this is necessary cause we dont have average event position or energy as a separate function, since those things are very trivial to calculate. But no strong preferences from my side... 

> In the extraction of information from paulina, there are some functions: `find_extrema`, `blob_centers` and `blob_energy_and_hits` that seems to me related. Somehow, the output of find_extrema should be arguments of blob_centers. Maybe some re-organization can be in place there. But these changes are out of the scope of Esmeralda that can be approved as it is.

Agree, this should probably be another PR

> Finally, I see that the track information has no information about what is the S2 from which that trac was constructed. This information can be obtained from the zmin-zmax of the track. I wonder if this was discussed or not.

I dont think we ever discussed this.

I also modified load_hits to read Ec, Qc and track_id from hits tables.

PR #613, opened 2019-01-18
Fix in configure.py to be able to run mypy.

jmunozv:
The baheviour of configure.py is the same than before but, by splitting the double assignment into two different lines, the error raised by mypy when checking types of fanal, is fixed.


jmbenlloch (approved):
This change will fix `mypy` in nextic/FANAL and may allow us to run it also in IC as a regular check at some point.

PR #614, opened 2019-01-23
Removed error from PR #593 which saved test data

andLaing:
Fixes error of leaving the test data save in irene_test
and penthesilea_test which had been added to update
test data.


jmbenlloch (approved):
Important fix for an error that was overwriting files within IC.

PR #615, opened 2019-01-24
Drop voxels

paolafer:
In this PR a couple of functions are added to paolina, which allow one to eliminate voxels from a track, with energy lower than a threshold, and redistribute their energy among the closest voxels.
The function `drop_end_point_voxels` modifies the `voxels` argument.
I would require a review from @ausonandres, because he knows the algorithm, but he hasn't taken part in this particular implementation, and @mmkekic for an outsider view.

### DIFF
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -254,7 +254,7 @@ def drop_end_point_voxels(voxels: Sequence[Voxel], energy_threshold: float, min_
 
         ### add voxel energy to hit and to voxel, separately
         min_hit.energy += the_vox.E
-        min_v.energy += the_vox.E
+        min_v.energy   += the_vox.E
 
 
     mod_voxels = copy.deepcopy(voxels)
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -258,7 +258,8 @@ def drop_voxel(voxels: Sequence, the_vox: Voxel):
 
 
 def drop_end_point_voxels(voxels: Sequence[Voxel], energy_threshold: float, min_vxls: int):
-    """Eliminate voxels at the end-points of a track, if their energy is lower than a threshold"""
+    """Eliminate voxels at the end-points of a track, recursively,
+       if their energy is lower than a threshold"""
     while True:
         modified_voxels = 0
         trks = make_track_graphs(voxels)
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -236,7 +236,7 @@ def drop_end_point_voxels(voxels: Sequence[Voxel], energy_threshold: float, min_
         voxels.remove(the_vox)
 
         pos = [h.pos for h in the_vox.hits]
-        qs = [h.E for h in the_vox.hits]
+        qs  = [h.E for h in the_vox.hits]
         bary_pos = np.average(pos, weights=qs, axis=0)
 
         ### find hit with minimum distance, only among neighbours
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -9,6 +9,7 @@ from networkx                   import Graph
 from .. evm.event_model         import Voxel
 from .. core.exceptions         import NoHits
 from .. core.exceptions         import NoVoxels
+from .. core.exceptions         import VoxelNotInList
 from .. evm.event_model         import BHit
 from .. evm.event_model         import Voxel
 from .. evm.event_model         import Track
@@ -219,3 +220,61 @@ def make_tracks(evt_number       : float,
         track = Track(voxels_from_track_graph(trk), blobs)
         tc.tracks.append(track)
     return tc
+
+
+def drop_voxel(voxels: Sequence, the_vox: Voxel):
+    """Eliminate an individual voxel from a set of voxels and give its energy to the hit
+       that is closest to the barycenter of the eliminated voxel hits, provided that it
+       belongs to a neighbour voxel."""
+    ### remove voxel from list of voxels
+    try:
+        voxels.remove(the_vox)
+    except ValueError:
+        raise VoxelNotInList
+
+    if len(voxels) == 0:
+        return
+
+    pos = [h.pos for h in the_vox.hits]
+    qs = [h.E for h in the_vox.hits]
+    bary_pos = np.average(pos, weights=qs, axis=0)
+
+    ### find hit with minimum distance, only among neighbours
+    min_dist = 1e+06
+    min_hit = voxels[0].hits[0]
+    min_v = voxels[0]
+    for v in voxels:
+        if neighbours(the_vox, v):
+            for hh in v.hits:
+                dist = np.linalg.norm(bary_pos - hh.pos)
+                if dist < min_dist:
+                    min_dist = dist
+                    min_hit = hh
+                    min_v = v
+
+    ### add voxel energy to hit and to voxel, separately
+    min_hit.energy += the_vox.E
+    min_v.energy += the_vox.E
+
+
+def drop_end_point_voxels(voxels: Sequence[Voxel], energy_threshold: float, min_vxls: int):
+    """Eliminate voxels at the end-points of a track, if their energy is lower than a threshold"""
+    while True:
+        modified_voxels = 0
+        trks = make_track_graphs(voxels)
+        vxl_size = voxels[0].size
+
+        for t in trks:
+            if len(t.nodes()) < min_vxls:
+                continue
+
+            extr1, extr2 = find_extrema(t)
+            if extr1.E < energy_threshold:
+                modified_voxels += 1
+                drop_voxel(voxels, extr1)
+            if extr2.E < energy_threshold:
+                modified_voxels += 1
+                drop_voxel(voxels, extr2)
+
+        if modified_voxels == 0:
+            break
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -236,7 +236,7 @@ def drop_end_point_voxels(voxels: Sequence[Voxel], energy_threshold: float, min_
         voxels.remove(the_vox)
 
         pos = [h.pos for h in the_vox.hits]
-        qs  = [h.E for h in the_vox.hits]
+        qs  = [h.E   for h in the_vox.hits]
         bary_pos = np.average(pos, weights=qs, axis=0)
 
         ### find hit with minimum distance, only among neighbours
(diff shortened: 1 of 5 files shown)


mmkekic:
@jjgomezcadenas , @jacg ,  @gonzaponte  what is the preferred way of manipulating a mutable object (in this case list) - to return a copy or to change the original object?

jjgomezcadenas:
I would not change the object unless I really had to. No need to introduce change effects. I would return a new object.

jacg:
As we are mixing paradigms, there is no clear answer to this. But I think that in IC we are leaning more and more towards functional, side-effect free programming wherever practically possible, because we recognize that it does make understanding, testing, avoiding bugs and debugging the code much easier.

As a default position, I would therefore leave the original untouched, and return a new object, unless an overriding argument comes along in any specific case.

invisible_cities/reco/paolina_functions_test.py
```diff
@@ -469,4 +475,47 @@ def test_voxel_is_dropped(hits, requested_voxel_dimensions):
...
     assert len(voxels) == n_starting_voxels - 1
+
+
+@given(bunch_of_hits, box_sizes, min_n_of_voxels)
+def test_total_energy_is_conserved(hits, requested_voxel_dimensions, min_voxels):
+
+    tot_initial_energy = sum(h.E for h in hits)
+    voxels = voxelize_hits(hits, requested_voxel_dimensions, strict_voxel_size=False)
+    energies = [v.E for v in voxels]
+    e_thr = random.uniform(min(energies), max(energies))
```
jacg: Generating random data inside Hypothesis properties (and keeping Hypothesis out of the loop) makes test failures irreproducible and therefore destroys Hypothesis' ability to shrink counterexamples, and present you with the simplest failing case. You really don't want to do this! Use `hypothesis.strategies.random_module` or `hypothesis.strategies.randoms`.

invisible_cities/reco/paolina_functions.py
```diff
@@ -222,60 +223,60 @@ def make_tracks(evt_number       : float,
...
+                        min_dist = dist
+                        min_hit = hh
+                        min_v = v
+
+        ### add voxel energy to hit and to voxel, separately
+        min_hit.energy += the_vox.E
+        min_v.energy += the_vox.E
+
+
+    mod_voxels = copy.deepcopy(voxels)
```
jacg, line 268: Why does a shallow copy not suffice?

invisible_cities/reco/paolina_functions_test.py
```diff
@@ -469,4 +475,47 @@ def test_voxel_is_dropped(hits, requested_voxel_dimensions):
     drop_voxel(voxels, voxel_to_drop)
 
     assert len(voxels) == n_starting_voxels - 1
+
+
+@given(bunch_of_hits, box_sizes, min_n_of_voxels)
+def test_total_energy_is_conserved(hits, requested_voxel_dimensions, min_voxels):
```
jacg: This test name should contain 'drop' somewhere.

invisible_cities/reco/paolina_functions.py
```diff
@@ -219,3 +220,61 @@ def make_tracks(evt_number       : float,
...
+    try:
+        voxels.remove(the_vox)
+    except ValueError:
+        raise VoxelNotInList
+
+    if len(voxels) == 0:
+        return
+
+    pos = [h.pos for h in the_vox.hits]
+    qs = [h.E for h in the_vox.hits]
```
jacg: Please align the listcomps.

invisible_cities/reco/paolina_functions.py
```diff
@@ -219,3 +220,61 @@ def make_tracks(evt_number       : float,
...
+            for hh in v.hits:
+                dist = np.linalg.norm(bary_pos - hh.pos)
+                if dist < min_dist:
+                    min_dist = dist
+                    min_hit = hh
+                    min_v = v
+
+    ### add voxel energy to hit and to voxel, separately
+    min_hit.energy += the_vox.E
+    min_v.energy += the_vox.E
```
jacg: Please align these two statements, starting at the first dot.

invisible_cities/reco/paolina_functions.py
mmkekic: Because voxel contains hits and she is changing energies of the hits, so shallow copy wont protect us from keeping the hits of the original voxel list the same. On the other hand I dont quite like this new style, I think that drop voxels should be a function as it was in the original proposal.

jacg: I suspected it had something to do with mutating the voxels, but then I was surprised to find that no tests fail if a shallow copy is performed instead.

Please provide a test that demonstrates the need for `deepcopy`, otherwise someone in the future will remove it as superfluous. Although if *none* of our extensive tests spot the problem ...

mmkekic: Yes, my suggestion is to assert that the sum of hit energies for all hits inside the track before and after dropping voxels doesnt change. That fails with the shallow copy.
I would also suggest to make E  of a voxel a property of voxel that is being calculated based on hits contained inside, that way the voxel energy will be correctly updated when the hits energies inside the voxel change.

paolafer: 1. We can add a test that ensures that the original list of hits is not changed by the function.
2. The E property of a voxel was originally kept independent on the sum of the hit energies on purpose as a check that the voxelization algorithm (`voxelize_hits`) was performing correctly. I would like to keep this check, in case the voxelization algorithm is changed, but maybe there's another way to do that.

jjgomezcadenas: > We can add a test that ensures that the original list of hits is not changed by the function.

We could do this by construction, passing a tuple rather than a list.

> The E property of a voxel was originally kept independent on the sum of the hit energies on purpose as a check that the voxelization algorithm (voxelize_hits) was performing correctly. I would like to keep this check, in case the voxelization algorithm is changed, but maybe there's another way to do that.

Noting prevents us from adding a second number, say E_voxel, E_sumHits

invisible_cities/reco/paolina_functions_test.py
paolafer: Done.

invisible_cities/reco/paolina_functions.py
paolafer: I put the `drop_voxel` function inside the `drop_end_point_voxels`function, in order to keep using a mutable object. I think that this is a case where mutable objects are useful, since it is a recursive algorithm (it keeps dropping voxels until a condition is reached). The alternative is to make a lot of useless copies, I think. Since there is no current user case for the `drop_voxel` function alone, I think we can implement it this way, and if in the future somebody asks for it, we can change the code.

paolafer: As for the need of a `deepcopy`: I'm using it because we decided that functions in IC shouldn't have side effects, therefore we don't want the function to modify its argument. I will add a test to ensure that the function does not modify the input voxels, but I cannot figure out a test that demonstrates the need for not modifying the argument.

jacg: The question was not why you copy, the question was why the copy needs to be *deep*. The answer is twofold:

1. The elements are mutable. For immutable elements you only ever need a shallow copy.

2. The elements are mutated. If they were never mutated, you could get away with a shallow copy.

> I put the `drop_voxel` function inside the `drop_end_point_voxels` function, in order to keep using a mutable object.

More precisely: in order not to violate the IC principle that functions should not modify their inputs. `drop_voxel` mutates its input, so by making it an *internal* implementation detail of `drop_end_point_voxels`, nobody can call it and be nastily surprised.

Let's introduce a bit of nomenclature: functions which mutate their inputs are called *destructive*, while ones which do not, are called *non-destructive*. So, we want IC functions to be non-destructive wherever pragmatically possible. But what if we do end up needing to write a destructive function?

I really like the Scheme approach to this: any function whose name ends in an exclamation mark (`!`) is destructive. Frequently one finds two different versions of the same function (for example `foo!` and `foo`) one destructive and one non-destructive. The former exists because it is more efficient than the latter, the latter exists because it is safer than the former. The universal and concise conventional meaning of `!` minimizes nasty surprises.

Unfortunately `!` is not a valid constituent character in Python, so we can't adopt that exact convention. But perhaps we should consider formalizing some such convention in IC.

Anyway, I think that making `drop_voxel` internal to `drop_end_point_voxels` is a decent idea. If we discover the need to pull it out, we should use whatever convention we agree on, to make it clear that it is a destructive function.

invisible_cities/reco/paolina_functions.py
paolafer: Done.

invisible_cities/reco/paolina_functions.py
paolafer: Done.

invisible_cities/reco/paolina_functions_test.py
paolafer: Done.

invisible_cities/reco/paolina_functions.py
```diff
@@ -236,7 +236,7 @@ def drop_voxel(voxels: Sequence[Voxel], the_vox: Voxel):
         voxels.remove(the_vox)
 
         pos = [h.pos for h in the_vox.hits]
-        qs = [h.E for h in the_vox.hits]
+        qs  = [h.E for h in the_vox.hits]
```
jacg: The contents are still misaligned!

paolafer: You're right, I must be alignment blind or something :-D.

invisible_cities/reco/paolina_functions.py
jjgomezcadenas: BTW, Julia copies the same notation from Scheme! We could implement the convention
_D to mark destructive functions. But I also like the solution proposed by@paolafer!

jacg: > Julia copies the same notation from Scheme! 

Unfortunately Julia is repeatedly failing to copy the similar `?` notation for predicates.

`_D` is probably the best we're going to come up with in Python. It's not bad: not too long but slightly ugly, and the ugliness is a good thing, as we want to feel a little discomfort when using destructive functions.

paolafer:
I see what you mean, @mmkekic and you're right. I've changed the check on the number of voxels in the track inside the `drop_voxel` function, and I've added a test to show the good behaviour, using your counterexample.

invisible_cities/reco/paolina_functions_test.py
```diff
@@ -443,4 +451,89 @@ def test_contiguity(proximity, contiguity, are_neighbours):
...
+
+    energies = [v.E for v in voxels]
+    e_thr = min(energies) + fraction_zero_one * (max(energies) - min(energies))
+    mod_voxels = drop_end_point_voxels(voxels, e_thr, min_voxels)
+    tot_final_energy = sum(v.E for v in mod_voxels)
+    final_trks = make_track_graphs(mod_voxels)
+    final_trk_energies = [sum(vox.E for vox in t.nodes()) for t in final_trks]
+
+    assert tot_initial_energy == approx(tot_final_energy)
+    assert np.allclose(ini_trk_energies, final_trk_energies)
```
mmkekic, line 471: the test failed because the track order is not the same, you can either sort the energies when comparing them, or sort the tracks.

ausonandres (approved):
I've tested the functions and they work properly. So I approve this PR.

PR #616, opened 2019-01-25
Add some more names to system-of-units

jjgomezcadenas:
A few names have been added to the system of units file.


jmbenlloch (approved):
Simple change adding some more unit definitions.

PR #618, opened 2019-01-29
Multi-database compatibility

jmbenlloch:
In order to generalize IC to be compatible with different detectors the main issue is the database that is being read. The database defines the number of channels, positions, gains, masked channels, etc.

This PR adds the option to download different databases from the server and modifies all the functions that reads from the DB. Up to now, these functions only needed the run number, now they will need the database and the run number.

Cities has also been modified to read the detector name from the config file (so far `new` and `demoppdb`).

For this PR I have started from @neuslopez's work in https://github.com/neuslopez/IC/tree/DBdetectors and fixed all the failing tests.


jmbenlloch:
I have added a new commit that sets NEW as the default DB in case it is not defined in the config files. In this way, old config files will continue working.

jjgomezcadenas (approved):
This is a major upgrade of the code, which allows IC to serve any detector configuration specified in the DB. Although the changes in the code are minimal, they need to be propagated to a large numbers of files. This is a book example of a change that would result in a nightmare in the absence of tests. With the extensive test suite in hand, however, it was "straight forward" to adapt the code. "Straight forward" here is a poetic licence. It is quite a bit of a job. All tests pass, giving us very robust confidence that no major disaster will strike. The default configuration is still "NEW", so the change should be transparent for "normal operations". 

Good job.

neuslopez:
Thanks so much guys!!


El dc, 30 gen 2019 a les 18:18 jjgomezcadenas <notifications@github.com> va
escriure:

PR #621, opened 2019-02-03
Hits functions

mmkekic:
This PR adds functions for hits list manipulation - merge_NN_hits and  threshold_hits.

### DIFF
--- a/invisible_cities/reco/hits_functions.py
+++ b/invisible_cities/reco/hits_functions.py
@@ -1,0 +1,37 @@
+import numpy  as np
+from itertools   import compress
+from copy        import deepcopy
+from typing      import List
+from .. cities.components import split_energy
+from .. evm  import event_model as evm
+
+from .. types.ic_types      import NN
+from .. types.ic_types      import xy
+
+def merge_NN_hits(hits : List[evm.Hit], same_peak : bool = True) -> List[evm.Hit]: 
+    """ Returns a list of the hits where the  energies of NN hits are distributed to the closest hits such that the added energy is proportional to 
+    the hit energy. If all the hits were NN the function returns empty list. """
+    nn_hits     = [h for h in hits if h.Q==NN]
+    non_nn_hits = [deepcopy(h) for h in hits if h.Q!=NN]
+    passed = len(non_nn_hits)>0
+    if not passed:
+        return []
+    hits_to_correct=[]
+    for nn_h in nn_hits:
+        peak_num = nn_h.npeak
+        if same_peak:
+            hits_to_merge = [h for h in non_nn_hits if h.npeak==peak_num]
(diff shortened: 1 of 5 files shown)


invisible_cities/reco/hits_functions.py
```diff
@@ -0,0 +1,58 @@
...
+        peak_num = nn_h.npeak
+        if same_peak:
+            hits_to_merge = [h for h in non_nn_hits if h.npeak==peak_num]
+        else:
+            hits_to_merge = non_nn_hits
+        try:
+            z_closest  = min(abs(h.Z-nn_h.Z) for h in hits_to_merge)
+        except ValueError:
+            continue
+        h_closest = [h for h in hits_to_merge if abs(h.Z-nn_h.Z)==z_closest]
```
jerenner, line 34: Could the second loop through hits be avoided by simultaneously finding the closest Z and the index of this hit, using something like:
```
z_closest, ihit = min((abs(h.Z-nn_h.Z), ihit) for (ihit, h) in enumerate(hits_to_merge))
```

mmkekic: The problem is that min outputs one element that satisfy the condition and we can have multiple minimal distance hits...

jerenner (approved):
Functions to:
- merge NN hits into the closest (in z) neighboring non-NN hits
- remove all hits in a list with charge below a given threshold, adding their energies to hits in the same z-slice.

Various tests are included for each function.

invisible_cities/reco/hits_functions.py
jerenner: I see now - `h_closest` ends up being a list of all those hits - ok then, looks good as is.

mmkekic:
Hi all, whats the status of this? @paolafer , @ausonandres  are you also happy with those functions? They always act on hit.E (changing hit.energy) but as @jjgomezcadenas mentioned we dont ever want to redistribute the corrected energies anyway... Comments please

ausonandres:
I've been looking at the functions, especially at the `merge_NN_hits` one, checking with some test events that the energy is conserved before and after the merging, and it is well shared with the closest hits, so I like it.

paolafer:
This PR has been reviewed by two people and approved, can we rebase and merge, please?

carmenromo:
@jerenner can you review the latest commits? I can't merge...

jerenner (approved):
Function `split_energy` moved - ok

carmenromo:
Before merging this PR I think the failing test should be resolved

mmkekic:
yeah... it is fixed in the PR of blobs redefinition so we can just wait till that one is merged and rebase

carmenromo:
ok!

PR #623, opened 2019-02-05
Added rebin_method option to penthesilea

andLaing:
Adds the possibility to rebin PMaps
using the new threshold method using
a configuration option.

stride option as default for backwards
compatibility.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -410,7 +410,7 @@ def split_energy(total_e, clusters):
     return total_e * qs / np.sum(qs)
 
 
-def hit_builder(dbfile, run_number, drift_v, reco, rebin_slices):
+def hit_builder(dbfile, run_number, drift_v, reco, rebin_slices, rebin_method):
     datasipm = load_db.DataSiPM(dbfile, run_number)
     sipm_xs  = datasipm.X.values
     sipm_ys  = datasipm.Y.values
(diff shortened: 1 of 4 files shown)


andLaing:
Is this an acceptable way of setting an Enum from config? @jacg @gonzaponte ??

jjgomezcadenas (approved):
This is the continuation of previous PR. Straight forward implementation in Penthesilea. Please rebase and merge

gonzaponte:
> Is this an acceptable way of setting an Enum from config? @jacg @gonzaponte ??

Sorry for the late response, I've just read it. Right now we cannot use them. The easy, messy thing to do (not permanently!) is to take a string in the config and use `getattr(enum_class, field_name)` in the city. We may have already a couple of examples in the calibration cities.

I've been wanting to add support for symbols in the config file for a while but I haven't had the time.
If someone wants to address this, I think it is as easy as importing all symbols in `core/configure.py`, but since they are dispersed everywhere, it is a bit messy. This would be a bit more elegant if we had the `symbols.py`package, but we've never actually created it.

andLaing:
ok, for now I just used the class as a dictionary enum_class[field_name] as it was what I found doing some googling. It's not as robust probably as getattr probably but since the standard method is there as a default value for when the config doesn't set it we shouldn't have problems in the short term.

PR #624, opened 2019-02-07
Updates on database code for different detectors

jmbenlloch:
To be compatible with different detectors, the code must be able of reading different schemas for the same tables. For instance, `PMTFELowFrequencyNoise` has one column per FE box, so this is different for different detectors. So now the DB downloader will get the table definition from MySQL server instead of being hardcoded in `download.py`. This change also simplifies a lot the code to clone databases.

New tests has been added to check this ability of getting table schemas from the server.

Tests for `load_db` are now parametrized and will run for DEMOPP, NEW and NEXT100.

NEXT100 database has been updated to include all the tables needed by IC, in particular the ones related to the BLR and sensor calibrations. The values included are averages of NEW values. For example, PMTs gains are set to 24.75, SiPM gains to 17.25, etc. This PR adds the sqlite3 copy of NEXT100 DB too.


jjgomezcadenas (approved):
An important development, that allows generalization of IC code to NEXT-100, DEMO and in general any configuration specified via the DB. Good job!

jjgomezcadenas:
@carmenromo please merge

PR #625, opened 2019-02-08
Fix bug in cal functions related to detector type

carmenromo:
This PR fix a bug that remained after PR of compatibility for different detector was approved. In the function seeds_db there was not the type of detector argument.


jjgomezcadenas:
@jmbenlloch please review this PR

jmbenlloch (approved):
I have split one commit into two separate commits and changed the test to use a fixture instead of hardwriting the detector name.

Besides that, this commit fixes properly a bug in a calibration function, approved.

PR #626, opened 2019-02-11
Change blob definition

paolafer:
In this PR, I have changed the definition of blob, according to the latest studies performed by @ausonandres and myself. Now each blob centre is calculated as the barycenter of the hits that fall within each one of the end-point voxels. Moreover, the energies of the individual hits are used to build the energy of the blob, instead of the energy of the voxels. Last, but not least, only the hits of the voxels that are at a distance-along-the-track shorter than a fixed distance are considered to be included in the blob. The final requirement for a hit to be included is that its euclidean distance from the centre of the blob is shorter than the blob radius. Tests have been changed following the change in the definition and one more has been added. 
I would like Javi Muñoz to be a reviewer of this PR, I'll add him as soon as he approves the request to be a collaborator.

### DIFF
--- a/invisible_cities/evm/event_model.py
+++ b/invisible_cities/evm/event_model.py
@@ -292,18 +292,18 @@ class VoxelCollection:
         return self.__str__()
 
 
-class Blob(VoxelCollection):
-    """A Blob is a collection of Voxels which adds the energy of the blob. """
-    def __init__(self, a: Voxel, voxels : List[Voxel]) ->None:
-        super().__init__(voxels)
-        self.seed = a
+class Blob():
+    """A Blob is a collection of Hits which adds the energy of the blob. """
+    def __init__(self, a: Tuple[float], hits : List[BHit], radius : float) ->None:
+        super().__init__(hits)
+        self.seed   = a
+        self.radius = radius
 
     def __str__(self):
-        s =  """Blob: (number of voxels = {})\n,
-                voxels = {} \n
+        s =  """Blob: (hits = {} \n
                 seed   = {} \n
                 blob energy = {}
-        """.format(self.number_of_voxels, self.voxels, self.seed, self.E)
+        """.format(self.hits, self.seed, sum(h.E for h in hits))
 
         return  s
 
--- a/invisible_cities/evm/event_model.py
+++ b/invisible_cities/evm/event_model.py
@@ -294,9 +294,9 @@ class VoxelCollection:
 
 class Blob():
     """A Blob is a collection of Hits with a seed and a radius. """
-    def __init__(self, a: Tuple[float], hits : List[BHit], radius : float) ->None:
+    def __init__(self, seed: Tuple[float, float, float], hits : List[BHit], radius : float) ->None:
         super().__init__(hits)
-        self.seed   = a
+        self.seed   = seed
         self.hits   = hits
         self.energy = sum(h.E for h in hits)
         self.radius = radius
--- a/invisible_cities/evm/event_model.py
+++ b/invisible_cities/evm/event_model.py
@@ -293,17 +293,20 @@ class VoxelCollection:
 
 
 class Blob():
-    """A Blob is a collection of Hits which adds the energy of the blob. """
+    """A Blob is a collection of Hits with a seed and a radius. """
     def __init__(self, a: Tuple[float], hits : List[BHit], radius : float) ->None:
         super().__init__(hits)
         self.seed   = a
+        self.hits   = hits
+        self.energy = sum(h.E for h in hits)
         self.radius = radius
 
     def __str__(self):
         s =  """Blob: (hits = {} \n
                 seed   = {} \n
-                blob energy = {}
-        """.format(self.hits, self.seed, sum(h.E for h in hits))
+                blob energy = {} \n
+                blob radius = {}
+        """.format(self.hits, self.seed, self.energy, self.radius)
 
         return  s
 
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -230,6 +230,11 @@ def blob_energies(track_graph : Graph, radius : float) -> Tuple[float, float]:
     Ea = sum(h.E for h in ha)
     Eb = sum(h.E for h in hb)
 
+    # Consider the case where voxels are built without associated hits
+    if len(ha) == 0 and len(hb) == 0 :
+        Ea = voxel_energy_within_radius(distances[a], radius)
+        Eb = voxel_energy_within_radius(distances[b], radius)
+
     return (Eb, Ea) if Ea < Eb else (Ea, Eb)
 
 
@@ -243,6 +248,11 @@ def blob_energies_and_hits(track_graph : Graph, radius : float) -> Tuple[float, 
     Ea = sum(h.E for h in ha)
     Eb = sum(h.E for h in hb)
 
+    # Consider the case where voxels are built without associated hits
+    if len(ha) == 0 and len(hb) == 0 :
+        Ea = voxel_energy_within_radius(distances[a], radius)
+        Eb = voxel_energy_within_radius(distances[b], radius)
+
     return (Eb, Ea, hb, ha) if Ea < Eb else (Ea, Eb, ha, hb)
 
 
@@ -255,6 +265,11 @@ def blob_centres(track_graph : Graph, radius : float) -> Tuple[Tuple[float], Tup
 
     Ea = sum(h.E for h in ha)
     Eb = sum(h.E for h in hb)
+
+    # Consider the case where voxels are built without associated hits
+    if len(ha) == 0 and len(hb) == 0 :
+        Ea = voxel_energy_within_radius(distances[a], radius)
+        Eb = voxel_energy_within_radius(distances[b], radius)
 
     centre_of_blob_a = blob_centre(a)
     centre_of_blob_b = blob_centre(b)
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -221,7 +221,7 @@ def hits_in_blob(track_graph : Graph, radius : float, extreme: Voxel) -> Sequenc
     return blob_hits
 
 
-def blob_energies_hits_and_centres(track_graph : Graph, radius : float) -> Tuple[float, float, Sequence[Tuple[float, float, float]], Sequence[Tuple[float, float, float]], Tuple[float, float, float], Tuple[float, float, float]]:
+def blob_energies_hits_and_centres(track_graph : Graph, radius : float) -> Tuple[float, float, Sequence[BHit], Sequence[BHit], Tuple[float, float, float], Tuple[float, float, float]]:
     """Return the energies, the hits and the positions of the blobs.
        For each pair of observables, the one of the blob of largest energy is returned first."""
     distances = shortest_paths(track_graph)
@@ -251,7 +251,7 @@ def blob_energies(track_graph : Graph, radius : float) -> Tuple[float, float]:
     return E1, E2
 
 
-def blob_energies_and_hits(track_graph : Graph, radius : float) -> Tuple[float, float, Sequence[Tuple[float, float, float]], Sequence[Tuple[float, float, float]]]:
+def blob_energies_and_hits(track_graph : Graph, radius : float) -> Tuple[float, float, Sequence[BHit], Sequence[BHit]]:
     """Return the energies and the hits around the extrema of the track.
        The largest energy is returned first, as well as its hits."""
     E1, E2, h1, h2, _, _ = blob_energies_hits_and_centres(track_graph, radius)
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -220,64 +220,50 @@ def hits_in_blob(track_graph : Graph, radius : float, extreme: Voxel) -> Sequenc
     return blob_hits
 
 
+def blob_energies_hits_and_centres(track_graph : Graph, radius : float) -> Tuple[float, float, Sequence[Tuple[float, float, float]], Sequence[Tuple[float, float, float]], Tuple[float, float, float], Tuple[float, float, float]]:
+    """Return the energies, the hits and the positions of the blobs.
+       For each pair of observables, the one of the blob of largest energy is returned first."""
+    distances = shortest_paths(track_graph)
+    a, b, _   = find_extrema_and_length(distances)
+    ha = hits_in_blob(track_graph, radius, a)
+    hb = hits_in_blob(track_graph, radius, b)
+
+    Ea = sum(h.E for h in ha)
+    Eb = sum(h.E for h in hb)
+
+    # Consider the case where voxels are built without associated hits
+    if len(ha) == 0 and len(hb) == 0 :
+        Ea = voxel_energy_within_radius(distances[a], radius)
+        Eb = voxel_energy_within_radius(distances[b], radius)
+
+    ca = blob_centre(a)
+    cb = blob_centre(b)
+
+    return (Eb, Ea, hb, ha, cb, ca) if Ea < Eb else (Ea, Eb, ha, hb, ca, cb)
+
+
 def blob_energies(track_graph : Graph, radius : float) -> Tuple[float, float]:
     """Return the energies around the extrema of the track.
        The largest energy is returned first."""
-    distances = shortest_paths(track_graph)
-    a, b, _   = find_extrema_and_length(distances)
-    ha = hits_in_blob(track_graph, radius, a)
-    hb = hits_in_blob(track_graph, radius, b)
+    E1, E2, _, _, _, _ = blob_energies_hits_and_centres(track_graph, radius)
 
-    Ea = sum(h.E for h in ha)
-    Eb = sum(h.E for h in hb)
-
-    # Consider the case where voxels are built without associated hits
-    if len(ha) == 0 and len(hb) == 0 :
-        Ea = voxel_energy_within_radius(distances[a], radius)
-        Eb = voxel_energy_within_radius(distances[b], radius)
-
-    return (Eb, Ea) if Ea < Eb else (Ea, Eb)
+    return E1, E2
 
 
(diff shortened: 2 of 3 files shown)


invisible_cities/evm/event_model.py
```diff
@@ -292,18 +292,21 @@ def __repr__(self):
...
 
 
-class Blob(VoxelCollection):
-    """A Blob is a collection of Voxels which adds the energy of the blob. """
-    def __init__(self, a: Voxel, voxels : List[Voxel]) ->None:
-        super().__init__(voxels)
-        self.seed = a
+class Blob():
+    """A Blob is a collection of Hits with a seed and a radius. """
+    def __init__(self, a: Tuple[float], hits : List[BHit], radius : float) ->None:
```
jmunozv: I would use the name 'seed' instead of 'a' for the parameter ...
```suggestion
def __init__(self, seed: Tuple[float], hits : List[BHit], radius : float) ->None:
```
Anyway, whats is the meaning of the 'seed' ? What does it represent ??
If it is the blob centre, shouldn't we call it 'centre' ?? And type-hint it as Tuple[float, float, float] ??

invisible_cities/reco/paolina_functions.py
```diff
@@ -176,31 +176,105 @@ def length(track: Graph) -> float:
     return length
 
 
-def energy_within_radius(distances : Dict[Voxel, Dict[Voxel, float]], radius : float) -> float:
+def voxel_energy_within_radius(distances : Dict[Voxel, float], radius : float) -> float:
```
jmunozv: I would use this name for the functions. I think is clearer, and it would be consistent with other function names like:   voxels_within_radius(...).

```suggestion
def voxels_energy_within_radius(distances : Dict[Voxel, float], radius : float) -> float:
```

invisible_cities/reco/paolina_functions.py
```diff
@@ -176,31 +176,105 @@ def length(track: Graph) -> float:
...
     return [v for (v, d) in distances.items() if d < radius]
 
 
+def blob_centre(voxel: Voxel) -> Tuple[float]:
+    """Calculate the blob position, starting from the end-point voxel."""
+    positions = [h.pos for h in voxel.hits]
+    energies  = [h.E   for h in voxel.hits]
+    if sum(energies):
+        bary_pos = np.average(positions, weights=energies, axis=0)
+    else:
```
jmunozv, line 195: In which situation could we have the situation of trying to get the centre of a voxel with no energy in its hits ??
Is a voxel without hits possible ??
Is a hit without energy possible ??

invisible_cities/reco/paolina_functions.py
```diff
@@ -176,31 +176,105 @@ def length(track: Graph) -> float:
...
+            for h in v.hits:
+                hit_distance = np.linalg.norm(blob_pos - h.pos)
+                if hit_distance < radius:
+                    blob_hits.append(h)
+
+    return blob_hits
+
+
 def blob_energies(track_graph : Graph, radius : float) -> Tuple[float, float]:
     """Return the energies around the extrema of the track."""
```
jmunozv: Specify in the explanation which of the returned blob energies is the highest one.

invisible_cities/reco/paolina_functions.py
```diff
@@ -176,31 +176,105 @@ def length(track: Graph) -> float:
...
+
+    # Consider the case where voxels are built without associated hits
+    if len(ha) == 0 and len(hb) == 0 :
+        Ea = voxel_energy_within_radius(distances[a], radius)
+        Eb = voxel_energy_within_radius(distances[b], radius)
+
+    return (Eb, Ea) if Ea < Eb else (Ea, Eb)
+
+
+def blob_energies_and_hits(track_graph : Graph, radius : float) -> Tuple[float, float, Tuple[Tuple[float], Tuple[float]]]:
```
jmunozv: I think I wouldn't make two different functions to provide E, or E + hits. I would implement just the one returning all the info.

invisible_cities/reco/paolina_functions.py
```diff
@@ -176,31 +176,105 @@ def length(track: Graph) -> float:
...
+    if len(ha) == 0 and len(hb) == 0 :
+        Ea = voxel_energy_within_radius(distances[a], radius)
+        Eb = voxel_energy_within_radius(distances[b], radius)
 
+    return (Eb, Ea, hb, ha) if Ea < Eb else (Ea, Eb, ha, hb)
 
-def compute_blobs(track_graph : Graph, radius : float) -> Tuple[Voxel, Voxel, List[Voxel], List[Voxel]]:
-    """Return the blobs (list of voxels) around the extrema of the track."""
+
+def blob_centres(track_graph : Graph, radius : float) -> Tuple[Tuple[float], Tuple[float]]:
```
jmunozv: Something similar with this function. I would include the information of the blob centres into the same function that returns blob energies and hits.
Now there are 3 different functions that do basically the same computing, but they provide:
1st version: blob hits
2nd version: blob hits and energies
3rd version: blob centres
If someone needs for their analysis the energy and the centre of blobs, he would need to make the same computation twice.

jmunozv (changes requested):
May be some of my comments come from my ignorance of the new implementation ...

invisible_cities/evm/event_model.py
paolafer: The name `seed` was decided last year in the sprint week of January 2018, if I remember correctly, when we reviewed the event model. I think that we decided to keep a generic name because it doesn't need to be the centre of a sphere (we can change the definition of blob at some point, as we already did, and not change the `Blob`class). However, it can be changed if we want to.
I have changed the type hint and the name of the argument for the seed.

invisible_cities/reco/paolina_functions.py
paolafer: Here I think that `voxel_energy` is a correct way to write it, but English native speakers can confirm.

invisible_cities/reco/paolina_functions.py
paolafer: > Is a voxel without hits possible ??

Yes.

> Is a hit without energy possible ??

No.

A voxel without hits is possible if one builds voxels directly without passing through a collection of hits. We have several tests that use voxels without hits just because the property to be tested has to do with the graph algorithms and not with the content of the voxels. I think that the paolina functions can be left generic enough as to be able to deal with 3D pixels (the voxels) with just one position and an energy.

invisible_cities/reco/paolina_functions.py
paolafer: Done.

invisible_cities/reco/paolina_functions.py
paolafer: See below.

invisible_cities/reco/paolina_functions.py
paolafer: I have followed what was already implemented by @jacg. Maybe he can gives us some insight on why he decided to write several functions.

jacg: I'm having problems contacting the LFS server, so I am unable to check out this branch, so it's not clear to me what the relationship is between `blob_centres` (the function we seem to be discussing) and any function I wrote. The only thing that I wrote in that file that contains `centre` in its name is `centre` in `voxelize_hits` If you are alluding to its similarity to `sizes` this is because

+ Each function does exactly one thing clearly and concisely, and is therefore trivial to implement and understand.
+ Each function has a trivial and easy to understand interface.
+ Each function has a name that clearly and concisely describes what it does.
+ These functions were originally written inside `voxelize_hits`, as implementation details of the latter: and were made as simple as possible, because they were written in order to make the code self-documenting, and were never intended for use by anyone outside of the function which they helped to implement.

But as I cannot check out this branch, I cannot be entirely sure that this is relevant to the question being asked.

paolafer: Hi Jacek. We're talking about the functions `blob_energies`, `blob_energies_and_hits` and `blob_centres`, which are all new functions. I decided not to write just one big function which returns 6 objects (blob energies, hits and centres), because it looked to me that it was clearer to separate functions, and I was following what you did with `find extrema`, `length` and `find extrema_and_length`, but it's possible that I completely missed your point there.

invisible_cities/reco/paolina_functions.py
paolafer: I have changed the name of the function to `energy_of_voxels_within_radius`, under reviewer's suggestion.

invisible_cities/reco/paolina_functions.py
paolafer: I have written one function that returns all the quantities (blob energies, hits and centres) and three different functions that call this big one, to extract only what is needed.

msorel:
Very nice work, @paolafer !

Quick question. Is type hint for what following function returns correct?

`def blob_energies_hits_and_centres(track_graph : Graph, radius : float) -> Tuple[float, float, Sequence[Tuple[float, float, float]], Sequence[Tuple[float, float, float]], Tuple[float, float, float], Tuple[float, float, `float]]:`


I wonder if 

`Sequence[Tuple[float, float, float]], Sequence[Tuple[float, float, float]]` 

in there should be replaced by 

`Sequence[BHit], Sequence[BHit]`

since this is what the function `hits_in_blob` used in `blob_energies_hits_and_centres` returns. The hit has more to it than its 3D position, right?

invisible_cities/reco/paolina_functions.py
msorel: @paolafer, I am confused by the possibility of having voxels without hits, now. Your comment here seems to suggest that we allow for this possibility just to simplify testing of paolina functions? 

It seems strange to me to make a paolina function more general (hence, more complicated) than necessary only to make its testing simpler, if we do not have a real use-case for this generalization.

I mean, code like 

```
if sum(energies):
        bary_pos = np.average(positions, weights=energies, axis=0)
    # Consider the case where voxels are built without associated hits
    else:
        bary_pos = voxel.pos
```

in `blob_centre`

and

```
# Consider the case where voxels are built without associated hits
    if len(ha) == 0 and len(hb) == 0 :
        Ea = energy_of_voxels_within_radius(distances[a], radius)
        Eb = energy_of_voxels_within_radius(distances[b], radius)
```

in `blob_energies_hits_and_centres`

is there only to simplify testing? Isn't it better to keep functions simple, at the cost of making test functions more complicated?

I understand we need to test these paolina functions not only for reconstructed hits, but also for MC truth hits, as this is a real use-case we have when performing fast analyses. Fine. But this is what `test_voxels_with_no_hits` does already, I think. And despite its name, since in this test function voxels do have hits. So, fast analyses are also covered by the hits-in-voxel use case.

Let me know what you think about always requiring voxels to have non-zero collections of hits (reconstructed or MC truth ones), to simplify code.

paolafer:
Yes, Michel, you're right, I have changed the type hint as you suggested.

invisible_cities/reco/paolina_functions.py
paolafer: It's true that right now there's not a NEXT user case for voxels that don't come from a set of hits. However, until very short time ago, we have worked with voxels with no associated hits at all (a voxel didn't know anything about the hits that fell within its boundaries) and paolina functions are very general (you just need a set of nodes with a proximity criterium). In my opinion it is good that the code can keep dealing with this kind of objects, and be more general, even if the code itself looks more complicated. The code is transparent to the user, who doesn't need to know about this possibility, unless voxels without hits are being used.

On the other hand, the `test_voxels_with_no_hits` function was placed there to show a problem related to the precision used in the position of the hits. We had to change it to `float64`, otherwise voxels without hits were created. However, with the current version of the code, no voxel can be created without hits, if voxels come from the `voxelize_hits` function. It can happen only if one creates directly a list of `Voxel` objects with a position and an energy and no associated hits.

msorel: OK, you convinced me. OK to leave as is. I have no further comments, good job!

msorel (approved):
A clear improvement in the way we construct blob candidates at the track extremes, now using finer 3D hit information. Also introducing a voxel energy threshold to disregard weak voxels at track extremes, and only accounting  for hits along paolina tracks to mitigate problem with overlapping blobs. Thoroughly optimised in real-life data and MC analyses for the past several months by @paolafer and @ausonandres , and with a complete test suite in IC. Good job!

invisible_cities/reco/paolina_functions.py
jacg: OK, I see what you are talking about now.

 The bodies of these three functions (`blob_energies`, `blob_energies_and_hits` and `blob_centres`) are *identical* in their first 9 lines, and only differ in the 1,1 or 3 lines that follow.

This copy'n'paste code reuse is, of course, very wrong, and *must* be removed.

Please write a single function which performs the first 9 common lines. Once this is done, there are a few obvious approaches to consider:

1. Have this function return the values it calculated: `((a, ha, Ea), (b, hb, Eb))`, where the `a`s refer to the higher-energy blob.
2. Parametrize this function with a function which extracts whatever combination of that information the user wants.

To be clear, here are some code samples that show precisely what I mean:

     # Do we already have a Blob type somewhere?
    Blob = namedtuple('Blob', 'voxel, hits, energy')
    
    # Option 1 (Just return the voxels, hits and energies calculated)
    def blob_whatever(track_graph : Graph, radius : float) -> Tuple[Blob, Blob]:
        distances = shortest_paths(track_graph)
        a, b, _   = find_extrema_and_length(distances)
        ha = hits_in_blob(track_graph, radius, a)
        hb = hits_in_blob(track_graph, radius, b)
    
        Ea = sum(h.E for h in ha)
        Eb = sum(h.E for h in hb)
    
        # Consider the case where voxels are built without associated hits
        if len(ha) == 0 and len(hb) == 0 :
            Ea = voxel_energy_within_radius(distances[a], radius)
            Eb = voxel_energy_within_radius(distances[b], radius)
    
        if Ea < Eb:
            a,ha,Ea, b,hb,Eb = b,hb,Eb, a,ha,Ea 
    
        return Blob(a,ha,Ea), Blob(b,hb,Eb)
    
    
    # Option 2 (pass in a function which determines what should be returned)
    def blob_whatever(track_graph : Graph, radius : float, xxx: Callable[[Blob, Blob], T]) -> T:
        distances = shortest_paths(track_graph)
        a, b, _   = find_extrema_and_length(distances)
        ha = hits_in_blob(track_graph, radius, a)
        hb = hits_in_blob(track_graph, radius, b)
    
        Ea = sum(h.E for h in ha)
        Eb = sum(h.E for h in hb)
    
        # Consider the case where voxels are built without associated hits
        if len(ha) == 0 and len(hb) == 0 :
            Ea = voxel_energy_within_radius(distances[a], radius)
            Eb = voxel_energy_within_radius(distances[b], radius)
    
        if Ea < Eb:
            a,ha,Ea, b,hb,Eb = b,hb,Eb, a,ha,Ea 
    
        return xxx(Blob(a,ha,Ea), Blob(b,hb,Eb))
    
    # With option 1, blob_energies could be implemented thus:
    def blob_energies(track_graph : Graph, radius : float) -> Tuple[float, float]:
        a,b = blob_whatever(track_graph, radius)
        return a.energy, b.energy
    
    # With option 2, blob_energies could be implemented thus:
    def blob_energies(track_graph : Graph, radius : float) -> Tuple[float, float]:
        return blob_whatever(track_graph, radius, lambda a,b: a.energy, b.energy)
    
    # Or, given that you are unlikely to want to extract different bits of
    # information off the two blobs, perhaps the following variation on
    # the second theme would be better
    
    def blob_whatever(track_graph : Graph, radius : float, xxx: Callable[[Blob], T]) -> Tuple[T,T]:
        distances = shortest_paths(track_graph)
        a, b, _   = find_extrema_and_length(distances)
        ha = hits_in_blob(track_graph, radius, a)
        hb = hits_in_blob(track_graph, radius, b)
    
        Ea = sum(h.E for h in ha)
        Eb = sum(h.E for h in hb)
    
        # Consider the case where voxels are built without associated hits
        if len(ha) == 0 and len(hb) == 0 :
            Ea = voxel_energy_within_radius(distances[a], radius)
            Eb = voxel_energy_within_radius(distances[b], radius)
    
        if Ea < Eb:
            a,ha,Ea, b,hb,Eb = b,hb,Eb, a,ha,Ea 
    
        return xxx(Blob(a,ha,Ea)), xxx(Blob(b,hb,Eb))
    
    # in which case blob_energies could look like this:
    def blob_energies(track_graph : Graph, radius : float) -> Tuple[float, float]:
        return blob_whatever(track_graph, radius, lambda blob: blob.energy)
    
I have a strong preference for option 2, but if too many people on the team find it more confusing than option 1, then perhaps you should go with option 1. (Or one could say that if too many people are confused by option 2, then we should educate them until they find it me most natural thing to do!)

Whatever you choose, the copy'n'paste code reuse has to go as it violates the **DRY** principle: **D**o not **R**epeat **Y**ourself.

Also, don't forget that

    lambda blob: blob.energy
    lambda blob: blob.energy, blob.voxel

can be written as 

    attrgetter('energy')
    attrgetter('energy', 'voxel')

paolafer: Thanks Jacek! I already changed the code following the DRY principle, without passing through the `Blob`class, which we have, by the way. I just wrote three different functions that call the same one, so the code is not repeated. Do you think it is fine?

jmunozv:
Good job made to deal with blobs made of hits, with a test suite in IC.

bpalmeiro:
Please, @jmunozv , approve the changes, do not close the PR!

jmunozv (approved):

invisible_cities/reco/paolina_functions.py
jacg: This is much better.

I'm not too enthusiastic about `blob_energies_and_hits` returning 4 essentially anonymous values, but it's not obvious how to fix that without complicating things too much.

The end of `blob_energies_hits_and_centres` could benefit from having the two alternative return values being aligned one above the other, to facilitate seeing what the difference is. The cleanest way to spell it is probably:

    if Ea < Eb: return (Eb, Ea, hb, ha, cb, ca) 
    else      : return (Ea, Eb, ha, hb, ca, cb)

Actually, the highest-energy-blob-first condition is probably expressed more clearly like this:

    if Eb > Ea: return (Eb, Ea, hb, ha, cb, ca) 
    else      : return (Ea, Eb, ha, hb, ca, cb)

PR #627, opened 2019-02-11
reintroduced sensor info in calibration output

andLaing:
sensor information copied from raw waveform
files to output for:

berenice (sipm pdfs)
phyllis  (pmt pe_to_adc)
trude    (sipm pe_to_adc)

### DIFF
--- a/invisible_cities/cities/berenice.py
+++ b/invisible_cities/cities/berenice.py
@@ -7,7 +7,7 @@ From Βερενíκη, Ancient Macedonian form of the Attic Greek, Φερενí�
 
 This city produces the spectrum of dark noise of the SiPMs. This is
 achieved by binning either the pe or ADC content of each raw WF sample.
-Some slices will have only electronic noise while others will contain 
+Some slices will have only electronic noise while others will contain
 one or more dark counts. The resulting spectra give a representation
 of the SiPM charge in the absence of external light above detector
 ambient conditions.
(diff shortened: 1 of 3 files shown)


andLaing:
Neus needs this for start-up tests on demo++ so if a quick turn around is possible it'd be preferred

andLaing:
Neus needs this asap for demo, @jmbenlloch , @gonzaponte , @jjgomezcadenas , por favor!!

gonzaponte (approved):
Looks good.

Seems weird that `cf.copy_sensor_table` takes a *filename* and a *file*, but otherwise is ok.

jacg:
> Neus needs this asap 

If she needs it to produce some results, then she can do that witout it being approved and merged, simply by checking out this branch.

gonzaponte:
Please, add a test to each city to ensure the table is written (can be done in some other PR if you are in a hurry)

andLaing:
> > Neus needs this asap
> 
> If she needs it to produce some results, then she can do that witout it being approved and merged, simply by checking out this branch.

It's for the demo frontend installation, I'm not sure they want to be making multiple branches there but in the short term maybe they can just do that

andLaing:
> Looks good.
> 
> Seems weird that `cf.copy_sensor_table` takes a _filename_ and a _file_, but otherwise is ok.

I think this was due to some requirement/bothersome structure in the old 'solid' cities. I just reinstated the use of the function in these cities since it was something that was relatively urgent. I'll look to see if it can be improved when I add the tests.

PR #628, opened 2019-02-12
Fix possible bug with exception in building clusters.

ausonandres:
The function `build_pointlike_event` of `cities.components` has been changed since there were two loops like this one:  _(for _ in evt.nS1)_, where evt.nS1 is an int type. Now it appears to work properly, so is this change correct?


jmbenlloch (approved):
I have added a new test to check that failure and explained to @ausonandres how to write those kind of tests. Everything is working now, so approved.

PR #630, opened 2019-02-14
Fix bug when reading hit time

paolafer:
In this PR I fix a bug in the way we read the hits. A multiplicative `1e-3` factor was used, without taking into account that the same factor is already used when the hit is written (see `build_hits` function in `cities/components.py`), to convert milliseconds (coming from raw data) to seconds. A test has been added to ensure that the time we read is in seconds. I have chosen as a check range of dates 2008-2050; on one hand, 2008 is the year NEXT started and, on the other hand, if we are still using IC in 2050 it will not be me the one who will debug the code ;-).


jjgomezcadenas (approved):
Good bug catch, with test. Bravo.

PR #631, opened 2019-02-22
Variable hit energy

paolafer:
With this PR, we add the possibility of using a different attribute than `hit.E` to build voxels and to perform the other operations related to paolina functions. This is needed since with the new Esmeralda city we will have raw energy, corrected energy... for the same hit.

### DIFF
--- a/invisible_cities/evm/event_model.py
+++ b/invisible_cities/evm/event_model.py
@@ -192,10 +192,10 @@ class MCHit(BHit):
 
 class Voxel(BHit):
     """Represents a Voxel"""
-    def __init__(self, x,y,z, E, size, hits=[], e_type : HitEnergy = HitEnergy.energy.value):
+    def __init__(self, x,y,z, E, size, hits=None, e_type : HitEnergy = HitEnergy.energy.value):
         super().__init__(x,y,z, E)
         self._size  = size
-        self.hits   = hits
+        self.hits   = hits if hits is not None else []
         self.e_type = e_type
 
     @property
--- a/invisible_cities/evm/event_model.py
+++ b/invisible_cities/evm/event_model.py
@@ -200,7 +200,7 @@ class Voxel(BHit):
     def size(self): return self._size
 
     @property
-    def Ehits(self): return sum(getattr(h, self.e_type) for h in hits)
+    def Ehits(self): return sum(getattr(h, self.e_type) for h in self.hits)
 
     @property
     def Etype(self): return self.e_type
--- a/invisible_cities/evm/event_model.py
+++ b/invisible_cities/evm/event_model.py
@@ -192,14 +192,18 @@ class Voxel(BHit):
     """Represents a Voxel"""
     def __init__(self, x,y,z, E, size, hits=[], e_type : HitEnergy = HitEnergy.E.value):
         super().__init__(x,y,z, E)
-        self._size = size
-        self.hits = hits
+        self._size  = size
+        self.hits   = hits
+        self.e_type = e_type
 
     @property
     def size(self): return self._size
 
     @property
-    def Ehits(self): return sum(getattr(h, e_type) for h in hits)
+    def Ehits(self): return sum(getattr(h, self.e_type) for h in hits)
+
+    @property
+    def Etype(self): return self.e_type
 
 
 class Cluster(BHit):
@@ -312,6 +316,10 @@ class Blob():
         self.hits   = hits
         self.energy = sum(getattr(h, e_type) for h in hits)
         self.radius = radius
+        self.e_type = e_type
+
+    @property
+    def Etype(self): return self.e_type
 
     def __str__(self):
         s =  """Blob: (hits = {} \n
--- a/invisible_cities/evm/event_model.py
+++ b/invisible_cities/evm/event_model.py
@@ -3,6 +3,8 @@
 import tables as tb
 import numpy  as np
 
+from enum import Enum
+
 from networkx                  import Graph
 from .. types.ic_types         import NN
 from .. types.ic_types         import minmax
@@ -109,6 +111,13 @@ class MCParticle:
     __repr__ =  __str__
 
 
+class HitEnergy(Enum):
+    E        = 'E'
+    Ec       = 'Ec'
+    energy   = 'energy'
+    energy_c = 'energy_c'
+
+
 class BHit:
     """Base class representing a hit"""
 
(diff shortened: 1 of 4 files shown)


invisible_cities/evm/event_model.py
```diff
@@ -109,6 +111,13 @@ def __str__(self):
     __repr__ =  __str__
 
 
+class HitEnergy(Enum):
```
mmkekic: Use AutoNameEnumBase in ic_types that does exactly this, also energy and E are in principle the same attribute, why we need both of them?

paolafer: Ok, I wasn't aware of the existence of such a class. I agree that we don't need both `energy`and `E`, but this belongs to a different discussion, which we started some time ago, about conventions in the event data model and where to use properties or not (I cannot find an open issue, I don't remember if we opened one at all).

mmkekic: I dont think we opened one, but yes, it is related to that discussion so maybe we should open an issue?

mmkekic (changes requested):
Can you add some new test that uses different hits attributes?

jacg:
I am perplexed by the title of https://github.com/nextic/IC/pull/631/commits/48d60ac2d87fab2fb50743a03ac8e624bcb5c9. I don't see how the title ("Change default value from property to attribute to modify it") relates to what happens in the commit.

Now that my eyes having been drawn the code that is touched in the commit, I notice that the `value` of an enum symbol is being taken. **THIS IS WRONG!**

The whole point of these enums is to create sets of symbols used to describe a well-defined and limited set of possibilities, in order to avoid the very bad habit that has evolved in python of using strings in such situations. This habit evolved largely because there was no convenient symbol type in Python. Enums mitigate this, and we have decided to use Enums as symbols in IC.

The `str` type is populated by arbitrary sequences of characters. If you have a parameter whose acceptable values are, for example, `ON` and `OFF` you want to create a set of *symbols* containing exactly those two values and nothing else. Strings are bad, because, in addition to `'ON'` and `'OFF'` they also admit the values `'banana'`, `'green'`, `**_%2   !'` and an infinity of other nonsensical values.

So, in this case we would create a type populated by exactly two values, like this:

    class State(AutoNameEnumBase):
        ON  = auto()
        OFF = auto()

`State.ON` and `State.OFF` are the two values we are interested in, in this context. No other values are acceptable. Their `value` attributes are a completely irrelevant and annoying implementation detail of enums-as-symbols in Python. The symbol itself **is** the value we are interested in!

Indeed, the type annotation `e_type : HitEnergy = HitEnergy.energy.value` is a lie, because `type(HitEnergy.energy.value) is str`, while `type(HitEnergy.energy) is HitEnergy`.
 
By taking the `value` of the symbol, you are extracting a string from the symbol, thereby defeating the whole point of having the symbols: you are reverting to the use of strings, only with an extra layer of complexity complicating the issue.

Moral of the story: don't take the `value` of enums, use the enum members themselves as the values.

(A minor secondary benefit is that symbols are compared by identity rather than value, so symbol comparison is cheaper than string comparison: to discover that the symbols `SecretPassage1` and `SecretPassage2` are different, all that needs to be done is to check their locations in memory; to discover that the strings `"SecretPassage1"` and `"SecretPassage2"` are different you need to iterate through 14 elements of each string before finding the mismatch. By extracting the value from the symbol, you're throwing away this benefit too.)

paolafer:
@mmkekic, I've added a test using the new attributes.
@jacg, I've modified the code following your instructions.

invisible_cities/evm/event_model.py
```diff
@@ -192,14 +192,18 @@ class Voxel(BHit):
...
+        self.e_type = e_type
 
     @property
     def size(self): return self._size
 
     @property
-    def Ehits(self): return sum(getattr(h, e_type) for h in hits)
+    def Ehits(self): return sum(getattr(h, self.e_type) for h in hits)
+
+    @property
```
jacg, line 207: What is the purpose of this property? There already is a public attribute containing exactly the same information. Why the duplication?

paolafer: If your question is what is the difference between `Ehits`and `energy`, the answer is that the voxel energy is set in the voxelization algorithm, coming from the numpy histogram function, and, although it should be the same as the sum of the hit energies, I wanted to keep it separated as a mean of checking that the numpy histogram function keeps doing its job.

jacg: Sorry, I don't understand how to use GitHub's code comment feature: I click on the line I want to comment on, and it highlights some other lines nearby!

I was talking about this:

    def Etype(self): return self.e_type

With this, `x.Etype` and `x.e_type` are 100% synonymous (though the latter is settable, while the former is not). What is the point of this duplication?

(The same happens in `Cluster` too.)

paolafer: Ok, this is an issue we've been talking about for some time, and we need to decide a unique way of using properties and attribute in the IC event model (right now it's confusing). However, I think that this issue deserves another PR.

mmkekic (approved):

invisible_cities/reco/paolina_functions.py
```diff
@@ -187,8 +188,8 @@ def voxels_within_radius(distances : Dict[Voxel, float],
 
 def blob_centre(voxel: Voxel) -> Tuple[float, float, float]:
     """Calculate the blob position, starting from the end-point voxel."""
-    positions = [h.pos for h in voxel.hits]
-    energies  = [h.E   for h in voxel.hits]
+    positions = [h.pos              for h in voxel.hits]
+    energies  = [getattr(h, voxel.Etype) for h in voxel.hits]
```
mmkekic: here you need to put voxel.Etype.value to get actual string right?

invisible_cities/reco/paolina_functions.py
```diff
@@ -229,8 +232,10 @@ def blob_energies_hits_and_centres(track_graph : Graph, radius : float) -> Tuple
     ha = hits_in_blob(track_graph, radius, a)
     hb = hits_in_blob(track_graph, radius, b)
 
-    Ea = sum(h.E for h in ha)
-    Eb = sum(h.E for h in hb)
+    voxels = list(track_graph.nodes())
+    e_type = voxels[0].Etype
```
mmkekic: here again you need to extract the string

PR #632, opened 2019-03-01
Added tests for calibration city Sensor table copy

andLaing:
copy_sensor_table function updated to be
more like the other functions in the data
flow as requested in previous PR

### DIFF
--- a/invisible_cities/reco/calib_functions.py
+++ b/invisible_cities/reco/calib_functions.py
@@ -131,13 +131,12 @@ def copy_sensor_table(h5in_name : str,
     with tb.open_file(h5in_name) as dIn:
         try:
             sensor_info = dIn.root.Sensors
-        except tb.exceptions.NoSuchNodeError:
-            sensor_info = None
-
-        if sensor_info:
             h5out.copy_node(sensor_info,
                             newparent = h5out.root,
                             recursive = True)
+        except tb.exceptions.NoSuchNodeError:
+            sensor_info = None
+
 
 
 def dark_scaler(dark_spectrum):
--- a/invisible_cities/reco/calib_functions.py
+++ b/invisible_cities/reco/calib_functions.py
@@ -8,14 +8,14 @@ import tables as tb
 from scipy.signal import find_peaks_cwt
 from enum         import auto
 
-from .. core.system_of_units_c import units
-from .. core.core_functions    import in_range
-from .. core.stat_functions    import poisson_sigma
-from .. core                   import fit_functions    as fitf
-from .. database               import load_db          as DB
+from .. core.system_of_units_c import            units
+from .. core.core_functions    import         in_range
+from .. core.stat_functions    import    poisson_sigma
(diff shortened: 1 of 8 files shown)


gonzaponte:
You changed the implementation of the function and added it to more cities, while the PR is supposed to be about tests :).

There is just one thing I am not completely happy with. The new implementation splits the job into two parts. I don't see the need for that, is there any? What concerns me here is that the input file remains open during the city execution and it will be opened again when the city runs. I would prefer a single call that does it all if there is nothing against it.

The tests look good. Please also clean up the commented-out code.

andLaing:
> You changed the implementation of the function and added it to more cities, while the PR is supposed to be about tests :).

I didn't add it to any other cities, it was already in all three of the calibration cities. Just had to change how it was accessed due to the change in function.

> There is just one thing I am not completely happy with. The new implementation splits the job into two parts. I don't see the need for that, is there any? What concerns me here is that the input file remains open during the city execution and it will be opened again when the city runs. I would prefer a single call that does it all if there is nothing against it.

There was a complaint about the original function taking __filename__, __file__ which is the information available, this is the only way I could find to change that, I agree it's not ideal. Would you prefer I changed it back to how it was before?

gonzaponte:
> I didn't add it to any other cities, it was already in all three of the calibration cities. Just had to change how it was accessed due to the change in function.

You are right. My mistake.

> There was a complaint about the original function taking filename, file which is the information available, this is the only way I could find to change that, I agree it's not ideal. Would you prefer I changed it back to how it was before?

It wasn't that much of a complain, but rather an observation. I think the previous version looks better, even if it seems weird. I do like a lot the simplification of the function using the `copy_node` call. This can be kept, right?

andLaing:
> I do like a lot the simplification of the function using the copy_node call. This can be kept, right?

yup, on it.

invisible_cities/reco/calib_functions.py
```diff
@@ -125,21 +125,19 @@ def valid_integral_limits(sample_width, n_integrals, integral_start, integral_wi
...
+    with tb.open_file(h5in_name) as dIn:
+        try:
+            sensor_info = dIn.root.Sensors
+        except tb.exceptions.NoSuchNodeError:
+            sensor_info = None
+
+        if sensor_info:
+            h5out.copy_node(sensor_info,
+                            newparent = h5out.root,
+                            recursive = True)
```
gonzaponte, line 140: You can "merge" the `try` and `if` statements as:

```
with tb.open_file(h5in_name) as dIn:
    try:
        sensor_info = dIn.root.Sensors
        h5out.copy_node(sensor_info,
                        newparent = h5out.root,
                        recursive = True)
        pass
```

I would like a info message in the except part, but I don't know how to test it... I think we have had this problem already.

gonzaponte (approved):
Good! The test suite has been extended to account for this new functionality. The implementation of the relevant function has also been improved.

PR #633, opened 2019-03-04
Update DEMOPP database

jmbenlloch:
This PR updates `DEMOPP` database that has now the proper sensor id's and positions.

It also includes a small change in manage.sh that allows to download only one DB. If no option is given to `bash manage.sh download_test_db` all databases will be downloaded (`NEWDB`, `DEMOPPDB`, `NEXT100DB`).

With this change it is possible to download only one database: `bash manage.sh download_test_db DEMOPPDB`


carmenromo (approved):
This PR fixes some problems related to DEMO++'s db that we saw last week. Thanks JM! Approved!

PR #636, opened 2019-03-07
store pandas as pytable

mmkekic:
the function that transforms pandas dataframe to pytable, reading column names and types directly from dataframe; some tests are still missing and in particularly what to do with string object of dataframes

### DIFF
--- a/invisible_cities/core/testing_utils.py
+++ b/invisible_cities/core/testing_utils.py
@@ -101,7 +101,6 @@ def _compare_dataframes(assertion, df1, df2, check_types=True, **kwargs):
         col2 = df2[col]
         if check_types:
             assert col1.dtype == col2.dtype
-        print(col, col1.values - col2.values)
         assertion(col1.values, col2.values, **kwargs)
 
 
--- a/invisible_cities/io/dst_io.py
+++ b/invisible_cities/io/dst_io.py
@@ -5,8 +5,9 @@ from tables import NoSuchNodeError
 from tables import HDF5ExtError
 import warnings
 
-from .. reco import tbl_functions as tbl
+from .. core.exceptions    import TableMismatch
 from .  table_io           import make_table
+
 
 def load_dst(filename, group, node):
     """load a kdst if filename, group and node correctly found"""
@@ -27,10 +28,6 @@ def load_dsts(dst_list, group, node):
     dsts = [load_dst(filename, group, node) for filename in dst_list]
     return pd.concat(dsts)
 
-class TableMismatchError(Exception):
-    def __init__(self):
-        s  = 'The table and dataframe dont match! '
-        Exception.__init__(self, s)
 
 def _store_pandas_as_tables(h5out, df, group_name, table_name, compression='ZLIB4', descriptive_string=None, str_col_length=32):
     if len(df) == 0:
@@ -59,7 +56,7 @@ def _store_pandas_as_tables(h5out, df, group_name, table_name, compression='ZLIB
                            description = descriptive_string,
                            compression = compression)
     if not np.array_equal(df.columns,table.colnames):
-        raise TableMismatchError
+        raise TableMismatch
     for indx in df.index:
         tablerow = table.row
         for colname in table.colnames:
--- a/invisible_cities/io/dst_io.py
+++ b/invisible_cities/io/dst_io.py
@@ -42,13 +42,11 @@ def _make_tabledef(column_types:pd.Series,str_col_length:int=32)->dict:
 def _store_pandas_as_tables(h5out:tb.file.File, df:pd.DataFrame, group_name:str, table_name:str, compression:str='ZLIB4', descriptive_string:[str]="", str_col_length:int=32)->None:
     if len(df) == 0:
         warnings.warn(f'dataframe is empty', UserWarning)
-    if '/'+group_name not in h5out:
+    if group_name not in h5out.root:
         group = h5out.create_group(h5out.root,group_name)
-    else:
-        group = getattr(h5out.root,group_name)
-    if table_name in group:
-        table=getattr(group,table_name)
-    else:
+
+    group = getattr(h5out.root,group_name)
+    if table_name not in group:
         tabledef=_make_tabledef(df.dtypes)
         table = make_table(h5out,
                            group       = group_name,
@@ -57,6 +55,7 @@ def _store_pandas_as_tables(h5out:tb.file.File, df:pd.DataFrame, group_name:str,
                            description = descriptive_string,
                            compression = compression)
 
+    table=getattr(group,table_name)
     if not np.array_equal(df.columns,table.colnames):
         raise TableMismatch(f'dataframe differs from already existing table structure')
     for indx in df.index:
--- a/invisible_cities/io/dst_io.py
+++ b/invisible_cities/io/dst_io.py
@@ -1,10 +1,12 @@
 import tables as tb
 import pandas as pd
+import numpy  as np
 from tables import NoSuchNodeError
 from tables import HDF5ExtError
 import warnings
 
-
+from .. reco import tbl_functions as tbl
+from .  table_io           import make_table
 
 def load_dst(filename, group, node):
     """load a kdst if filename, group and node correctly found"""
@@ -24,3 +26,43 @@ def load_dst(filename, group, node):
 def load_dsts(dst_list, group, node):
     dsts = [load_dst(filename, group, node) for filename in dst_list]
     return pd.concat(dsts)
+
+class TableMismatchError(Exception):
+    def __init__(self):
+        s  = 'The table and dataframe dont match! '
+        Exception.__init__(self, s)
+
+def _store_pandas_as_tables(h5out, df, group_name, table_name, compression='ZLIB4', descriptive_string=None, str_col_length=32):
+    if len(df) == 0:
+        warnings.warn(f' dataframe is empty', UserWarning)
+    if '/'+group_name not in h5out:
+        group = h5out.create_group(h5out.root,group_name)
+    else:
+        group = getattr(h5out.root,group_name)
+    if table_name in group:
+        table=getattr(group,table_name)
+    else:
+        tabledef={}
+        for indx, colname in enumerate(df.columns):
+            coltype = df[colname].dtype.name
+            try:
+                tabledef[colname] = tb.Col.from_type(coltype, pos = indx)
+            except ValueError:
+                tabledef[colname] = tb.StringCol(32, pos = indx)
+
+        if descriptive_string is None:
+            descriptive_string = ''
+        table = make_table(h5out,
+                           group       = group_name,
+                           name        = table_name,
+                           fformat     = tabledef,
+                           description = descriptive_string,
+                           compression = compression)
+    if not np.array_equal(df.columns,table.colnames):
+        raise TableMismatchError
+    for indx in df.index:
+        tablerow = table.row
+        for colname in table.colnames:
+            tablerow[colname] = df.at[indx,colname]
+        tablerow.append()
+    table.flush()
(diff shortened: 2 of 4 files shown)


invisible_cities/core/testing_utils.py
```diff
@@ -101,7 +101,7 @@ def _compare_dataframes(assertion, df1, df2, check_types=True, **kwargs):
         col2 = df2[col]
         if check_types:
             assert col1.dtype == col2.dtype
-        print(col, col1.values - col2.values)
+        #print(col, col1.values - col2.values)
```
gonzaponte: Remove instead of commenting the line

invisible_cities/io/dst_io.py
```diff
@@ -1,10 +1,12 @@
 import tables as tb
 import pandas as pd
+import numpy  as np
 from tables import NoSuchNodeError
 from tables import HDF5ExtError
 import warnings
 
-
+from .. reco import tbl_functions as tbl
+from .  table_io           import make_table
```
gonzaponte, line 12: Review alignment and spaces in the whole PR.

invisible_cities/io/dst_io.py
```diff
@@ -24,3 +26,43 @@ def load_dst(filename, group, node):
 def load_dsts(dst_list, group, node):
     dsts = [load_dst(filename, group, node) for filename in dst_list]
     return pd.concat(dsts)
+
+class TableMismatchError(Exception):
+    def __init__(self):
+        s  = 'The table and dataframe dont match! '
+        Exception.__init__(self, s)
+
```
gonzaponte, line 41: This class should go in `core/exceptions.py`

gonzaponte: The other IC exceptions don't have a predefined message. Since it is used only in one place you may want to move the message to the place where the exception is actually raised.

invisible_cities/io/dst_io.py
```diff
@@ -24,3 +26,43 @@ def load_dst(filename, group, node):
...
+def _store_pandas_as_tables(h5out, df, group_name, table_name, compression='ZLIB4', descriptive_string=None, str_col_length=32):
+    if len(df) == 0:
+        warnings.warn(f' dataframe is empty', UserWarning)
```
gonzaponte: Just a warning? Are you convinced about it?

invisible_cities/io/dst_io.py
```diff
@@ -24,3 +26,43 @@ def load_dst(filename, group, node):
...
+    if table_name in group:
+        table=getattr(group,table_name)
+    else:
+        tabledef={}
+        for indx, colname in enumerate(df.columns):
+            coltype = df[colname].dtype.name
+            try:
+                tabledef[colname] = tb.Col.from_type(coltype, pos = indx)
+            except ValueError:
+                tabledef[colname] = tb.StringCol(32, pos = indx)
```
gonzaponte: I guess here goes the `str_col_length` variable.

invisible_cities/io/dst_io.py
```diff
@@ -24,3 +26,43 @@ def load_dst(filename, group, node):
...
+
+        if descriptive_string is None:
+            descriptive_string = ''
+        table = make_table(h5out,
+                           group       = group_name,
+                           name        = table_name,
+                           fformat     = tabledef,
+                           description = descriptive_string,
+                           compression = compression)
```
gonzaponte: I think the whole `else` section should be a separate function.

invisible_cities/io/dst_io.py
```diff
@@ -24,3 +26,43 @@ def load_dst(filename, group, node):
...
+        tabledef={}
+        for indx, colname in enumerate(df.columns):
+            coltype = df[colname].dtype.name
+            try:
+                tabledef[colname] = tb.Col.from_type(coltype, pos = indx)
+            except ValueError:
+                tabledef[colname] = tb.StringCol(32, pos = indx)
+
+        if descriptive_string is None:
+            descriptive_string = ''
```
gonzaponte: the argument `descriptive_string` could be `""` directly instead of `None` and `if None set it to ""`

invisible_cities/io/dst_io_test.py
```diff
@@ -59,3 +68,37 @@ def test_load_dsts_warns_if_not_existing_file(ICDATADIR):
...
+
+df3 = data_frames(index=range_indexes(min_size=1),columns=[column('int_value', dtype=int),column('float_val',dtype='float')])
+@given(df1 = df, df2=df3)
+def  test_store_pandas_as_tables_raises_exception(config_tmpdir, df1, df2):
+    filename = config_tmpdir+'dataframe_to_table_exact.h5'
+    group_name ='test_group'
+    table_name = 'table_name_2'
+    with tb.open_file(filename,'w') as h5out:
+        _store_pandas_as_tables(h5out, df1, group_name,table_name)
+        assert_raises(TableMismatchError, _store_pandas_as_tables,h5out, df2, group_name,table_name)
```
gonzaponte: You don't need to define three different strategies, I think. Define it once and use it everywhere. Also, use a more verbose name than `df`.

gonzaponte: I read it too fast. `df3` is a different strategy, so it stays.

gonzaponte (changes requested):
First round of comments. I did it pretty quickly, so there may be some things that escaped the eye :)

invisible_cities/io/dst_io.py
mmkekic: I tried but I am not sure I did it right, I think I still dont get the alignment style we use

invisible_cities/io/dst_io.py
mmkekic: I moved it to exceptions.py and removed message, I think the name is self explanatory

invisible_cities/io/dst_io.py
mmkekic: I dont want to program to crash if some of the dataframes to be added to the table are empty (ie when writing in the dataflow event-per-event)

invisible_cities/io/dst_io.py
mmkekic: what do you mean by this? i dont think we should take out the 'make_table' function since i dont see where else would it be needed

invisible_cities/io/dst_io.py
mmkekic: Probably in this case it doesnt matter but I prefer not to have mutable keyword arguments

mmkekic:
I am still not sure what to do in case of str type in dataframes since the casting between pandas and pytables types of str is not automatic

invisible_cities/io/dst_io.py
jacg: Use alignment to highlight symmetries and patterns. Anything that makes it easier for the reader to spot things that are similar or belong together, and anything that helps differences to stand out.

In the case of imports, the most obvious thing that always needs to be done is to align the import keywords: `from`, `import` and `as`. Align the names which are being bound, which may require a nanosecond's thought in the absence/precesce of `as`. For example:

    from vegetable import aubergine as eggplant
    from fruit     import banana    as platano
    from fruit     import              orange

(If you are reading the above in your mail client, who knows what alignment you are seeing: Make sure you read it in a fixed-width font!)

Aligning `orange` with `aubergine` and `banana` would be OK, as it belongs in both the column of components being imported and in the column of names being bound. (Strictly speaking this means that we should write `from fruit import orange as orange`, but I think that would be taking it too far.)

In the case of names which share a common pattern at the end or in the middle, align the common parts:

    from a import            a_thingy
    from a import         bbbb_thingy
    from a import cccccccccccc_thingy

in order to highlight that we are importing 3 thingies.

mmkekic: Thanks for the reply! What happens when I have unrelated imports ? As:
```
import warnings
from .. core.exceptions    import TableMismatch
from .  table_io           import make_table
from typing                import Optional
```

What should i try to match in this example?

jacg: The first line goes in a separate block from the rest. The last 3 have the same structure consisting of 5 columns:

1. `from`
2. relative path
3. module
4. `import`
5. component

Align each column. (The last line has an empty 2nd column.)

invisible_cities/io/dst_io.py
```diff
@@ -59,7 +56,7 @@ def _store_pandas_as_tables(h5out, df, group_name, table_name, compression='ZLIB
                            description = descriptive_string,
                            compression = compression)
     if not np.array_equal(df.columns,table.colnames):
-        raise TableMismatchError
+        raise TableMismatch
```
gonzaponte: Wouldn't hurt to have a explicit error message telling you that the dataframe changed shape from one call to another

invisible_cities/io/dst_io_test.py
```diff
@@ -101,4 +102,4 @@ def  test_store_pandas_as_tables_raises_exception(config_tmpdir, df1, df2):
     table_name = 'table_name_2'
     with tb.open_file(filename,'w') as h5out:
         _store_pandas_as_tables(h5out, df1, group_name,table_name)
-        assert_raises(TableMismatchError, _store_pandas_as_tables,h5out, df2, group_name,table_name)
+        assert_raises(TableMismatch, _store_pandas_as_tables,h5out, df2, group_name,table_name)
```
gonzaponte: To maintain a homogeneous style in IC, I would like this to be rewritten using a `with raises(TableMismatch)`, which I think is also more readable.

invisible_cities/io/dst_io_test.py
```diff
@@ -94,8 +93,8 @@ def  test_store_pandas_as_tables_2df(config_tmpdir, df1, df2):
     df_read = load_dst(filename, group_name, table_name)
     assert_dataframes_close(df_read, pd.concat([df1, df2]).reset_index(drop=True),False, rtol=1e-5)
 
-df3 = data_frames(index=range_indexes(min_size=1),columns=[column('int_value', dtype=int),column('float_val',dtype='float')])
-@given(df1 = df, df2=df3)
+dataframe_diff = data_frames(index=range_indexes(min_size=1),columns=[column('int_value', dtype=int),column('float_val',dtype='float')])
+@given(df1 = dataframe, df2=dataframe_diff)
 def  test_store_pandas_as_tables_raises_exception(config_tmpdir, df1, df2):
     filename = config_tmpdir+'dataframe_to_table_exact.h5'
     group_name ='test_group'
```
gonzaponte: review spaces

gonzaponte (commented):
Second round

invisible_cities/io/dst_io.py
gonzaponte: strings are not mutable :)

invisible_cities/io/dst_io.py
gonzaponte: not that is needed elsewhere, but would be more readable. Like `else: create_the_table`. The other function could be defined with an leading underscore so it is not exported

invisible_cities/io/dst_io.py
```diff
@@ -24,3 +28,39 @@ def load_dst(filename, group, node):
...
+    if table_name in group:
+        table=getattr(group,table_name)
+    else:
+        tabledef={}
+        for indx, colname in enumerate(df.columns):
+            coltype = df[colname].dtype.name
+            try:
+                tabledef[colname] = tb.Col.from_type(coltype, pos = indx)
+            except ValueError:
+                tabledef[colname] = tb.StringCol(str_col_length, pos = indx)
```
gonzaponte: I am not sure if `try/except` is optimal here. In principle, you could have a ValueError raised for a different reason. Also, I think I would like this to be a separate function, but I don't have a strong opinion.

invisible_cities/io/dst_io.py
mmkekic: oooops, true, my fault

invisible_cities/io/dst_io.py
```diff
@@ -24,3 +28,39 @@ def load_dst(filename, group, node):
 def load_dsts(dst_list, group, node):
     dsts = [load_dst(filename, group, node) for filename in dst_list]
     return pd.concat(dsts)
+
+def _make_tabledef(column_types:pd.Series,str_col_length:int=32)->dict:
```
```suggestion
def _make_tabledef(column_types:pd.Series, str_col_length:int=32) -> dict:
```

gonzaponte (approved):

PR #637, opened 2019-03-08
Blob class fix

paolafer:
In this PR I fix a couple of bugs in the `Blob` class definition, which had escaped previous reviews because there was no test for this class. Therefore, I've also added a test for the `make_tracks`function, which is the only function that uses the `Blob` class, and which was lacking its own test.

### DIFF
--- a/invisible_cities/evm/event_model.py
+++ b/invisible_cities/evm/event_model.py
@@ -313,10 +313,9 @@ class Blob():
                        hits : List[BHit],
                        radius : float,
                        e_type : HitEnergy = HitEnergy.energy) ->None:
-        super().__init__(hits)
         self.seed   = seed
         self.hits   = hits
-        self.energy = sum(getattr(h, e_type) for h in hits)
+        self.energy = sum(getattr(h, e_type.value) for h in hits)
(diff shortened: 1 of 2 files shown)


paolafer:
For some reason, a commit that has nothing to do with this PR (444b942) has sneaked in.
Could some expert have a look at it and possibly remove it? Thanks!

jmbenlloch:
> For some reason, a commit that has nothing to do with this PR (444b942) has sneaked in.
Could some expert have a look at it and possibly remove it? Thanks!

Done! You could do an interactive rebase, reorder the commits leaving 444b942 as the last one, and the reset your branch to the previous one. Since there are no conflicts you can do that reordering without any problem.

paolafer:
I think I managed to remove the wrong commit, but maybe @jmbenlloch was faster than me :-). I'm not sure what has happened now :-).

jjgomezcadenas (approved):
Good catch again, and a book example of the need of tests for functions.

jjgomezcadenas:
@paolafer please rebase then @bpalmeiro or @carmenromo pleaase merge

PR #638, opened 2019-03-08
Fix bug in the computation of xy position

paolafer:
This PR fixes a bug in the `compute_xy_position` function. The function was reading value from the database corresponding to run number 0. Now, the function receives the correct run number, as read from the input files. A test has been added, which fails if the function doesn't take into account the given run number.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -410,7 +410,7 @@ def hit_builder(dbfile, run_number, drift_v, reco, rebin_slices, rebin_method):
     sipm_ys  = datasipm.Y.values
     sipm_xys = np.stack((sipm_xs, sipm_ys), axis=1)
 
-    baricenter = partial(corona,
+    barycenter = partial(corona,
                          all_sipms      =  datasipm,
                          Qthr           =  0 * units.pes,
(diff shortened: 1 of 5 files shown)


jjgomezcadenas (approved):
A good catch, with a demonstration test. Good job

jjgomezcadenas:
@carmenromo @bpalmeiro please merge

paolafer:
The failing test is `test_scaled_dark_pedestal_pedestal`, which is unrelated to this PR. Something similar happened to me a few days ago with `test_rebin_peak_threshold` and @gonzaponte found out that it was an accuracy problem.

jjgomezcadenas:
Please comment the test out until it is fixed by the relevant party.

jacg:
> comment the test out

Grrrrr! **NEVER** comment out tests! We are using a testing framework with a rich set of features, including ones that cater for this sort of situation.

Use some sort of mark to skip the test conditionally. Preferably put some expiry date (or some other smarter condition) on the skip, so that it doesn't get forgotten forever.

jjgomezcadenas:
Sorry, this is what I meant... skip test... I wrote in a haste those unprecise.

Enviado desde mi iPhone

PR #640, opened 2019-03-11
Allow penthesilea to run with 0 S1

Aretno:
**Status**: Not final but may need some discussion.

This is a PR to allow Penthesilea and Dorothea to run with events with no S1 signals. 

Dorothea already allowed us to run for 0 S1s. In this case, the S1 peak parameters were set to NaN as well as the Z. The Z could easily changed by 0 at the analysis level with this simple line:

```
kdst['Z'] = np.where(kdst.nS1 == 0, 0.0, kdst.Z)
```
Therefore, since dorothea works efficiently I would probably not change the code and reassign the Z at the analysis level although at the test level, the default S1 values can be troublesome, as I'll explain later.

In the case of Penthesilea, the drift time of the S2s will be calculated relative to the time of the first time bin of the first S2 (so first slice of the first S2 will have a drift time equal to 0, the rest will be equal to the time difference between their time position and the first slice). The change to do this is minimal and easy (already uploaded).

However, when adding an specific code to test penthesilea with 0 S1s, I've noticed an Assertion Error that I'm currently unsure about how to handle. The values compared are actually equal but the Assertion Error jumps because there are NaN values in the arrays that are compared. 

This happens in the comparison of the stored S1 information. As I said before, in the absence of S1, the code initializes everything to NaN values, for analysis there is no problem with this but at the test level, it seems that two NaNs are not equal. To fix this I can think of three options:

1) Change the default values of S1s to have a different value and not NaN (fast solution although I suspect it may require changing several things in some tests)
2) Write a piece of code that handles the 0 S1 case as an specific case (more work and less elegant but maybe the "rightest" solution?)
3) Adapt the assertion code to handle the NaN comparison and not result in an error.

I already wrote point 3 (and is also in the uploaded code) so you could take a look at it. I'm not sure if giving an exception to the AssertionError is the most sensitive thing to do although it should fail if it's not equal (besides NaN cases).  Anyway, this is the status as of now, how do you think this should be approached?

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -430,8 +430,12 @@ def hit_builder(dbfile, run_number, drift_v, reco, rebin_slices, rebin_method):
         # take events with exactly one s1. Otherwise, the
         # convention is to take the first peak in the S1 object
         # as reference.
-        first_s1 = np.where(selector_output.s1_peaks)[0][0]
-        s1_t     = pmap.s1s[first_s1].time_at_max_energy
+        if len(selector_output.s1_peaks) > 0:
+            first_s1 = np.where(selector_output.s1_peaks)[0][0]
+            s1_t     = pmap.s1s[first_s1].time_at_max_energy
+        else:
+            first_s2 = np.where(selector_output.s2_peaks)[0][0]
+            s1_t     = pmap.s2s[first_s2].times[0]
 
         # here hits are computed for each peak and each slice.
         # In case of an exception, a hit is still created with a NN cluster.
--- a/invisible_cities/cities/penthesilea_test.py
+++ b/invisible_cities/cities/penthesilea_test.py
@@ -124,7 +124,7 @@ def test_penthesilea_threshold_rebin(ICDATADIR, output_tmpdir):
 
     conf        = configure('dummy invisible_cities/config/penthesilea.conf'.split())
     rebin_thresh = 4000
-    
+
     conf.update(dict(run_number      =        -6340,
                      files_in        =      file_in,
                      file_out        =     file_out,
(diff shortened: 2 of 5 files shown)


mmkekic:
Considering the solution of comparing nan values inside the tables I think the best option is to modify  assert_tables_equality since new city Esmeralda will be outputting nans as well. 
The reason it gives an error is because np.nan !=np.nan but numpy testing functions are able to handle nans. However, it doesnt work on numpy structured arrays and it seems its a numpy issue already: https://github.com/numpy/numpy/issues/8192. 
The only escape I see is to loop over columns of numpy arrays (similar as your solution, just use it from the beginning and not in the case of AssertationError and make sure not to call it if it is not a structured numpy array) and compare using numpy testing, I propose this solution:
```
def assert_tables_equality(got_table, expected_table):
    table_got      =      got_table[:]
    table_expected = expected_table[:]
    assert len(table_got) == len(table_expected)
    assert len(table_got.dtype) == len(table_expected.dtype)

    if table_got.dtype.names is not None:
        for col_name in table_got.dtype.names:
            got = table_got[col_name]
            expected = table_got[col_name]
            assert type(got) == type(expected)
            try:
                assert_allclose(got, expected)
            except TypeError:
                assert_equal(got, expected)                
    else:
        try:
            assert_allclose(got_table, expected_table)
        except TypeError:
            assert_equal(got_table, expected_table)
```
Maybe the original authors of the function can comment? 
@gonzaponte  or @jmbenlloch ?

jmbenlloch:
> Maybe the original authors of the function can comment?

I'm not the original author, but I think what @mmkekic is proposing looks fine. Although it is a rather technical problem with some subtleties (the nans, the structured arrays...). So I think it would be very useful to have some test function for `assert_tables_equality` checking those edge cases. Those test will also serve as documentation for the function.

jmbenlloch (approved):
This PR adds a new capability for penthesilea and includes new tests tu ensure everything is working ok.

PR #642, opened 2019-03-15
Fix drop voxel function

paolafer:
This PR fixes (and add a test to show the incorrect behaviour) a problem in the `drop_voxel` function, when a hit energy different than the default one was used to build the voxels. While a `Hit` object has several energies (raw, corrected...), a `Voxel` object just has one, and it has to be accessed through the `voxel.energy` attribute.
Additionally, another bug in the check of the length of the neighbour list has been found and fixed.

### DIFF
--- a/invisible_cities/reco/paolina_functions_test.py
+++ b/invisible_cities/reco/paolina_functions_test.py
@@ -664,13 +664,6 @@ def test_paolina_functions_with_voxels_without_associated_hits(blob_radius, min_
         assert np.allclose(blob_centre(b), b.pos)
 
 
-@given(bunch_of_corrected_hits(), box_sizes, radius)
-def test_voxelize_hits_with_hit_energy_different_from_default_value(hits, requested_voxel_dimensions, blob_radius):
-    with raises(AttributeError):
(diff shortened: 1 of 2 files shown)


invisible_cities/reco/paolina_functions_test.py
```diff
@@ -701,6 +702,11 @@ def test_paolina_functions_with_hit_energy_different_from_default_value(requeste
 
     assert not np.allclose(energies, energies_c)
 
+    energies_c = [v.E for v in voxels_c]
+    e_thr = min(energies_c) + fraction_zero_one * (max(energies_c) - min(energies_c))
+    # Test that this function doesn't fail
+    mod_voxels_c = drop_end_point_voxels(voxels_c, e_thr, min_vxls=0)
+
```
mmkekic, line 686: maybe also show that the energy of voxels are equal to the sum of corresponding hit energies depending on energy_type?

mmkekic (approved):
This PR fixes the bug in drop_voxels function and adds appropriate tests to show the desired behavior.

carmenromo:
@mmkekic can I merge?

mmkekic:
From my part - yes :)

PR #643, opened 2019-03-18
Change hit attribute

mmkekic:
This simple PR removes El, Ql and Zc from Hit and Cluster class and adds track_id to Hit class since it will be needed in Esmeralda.
What is not clear is if we want to keep the energies as properties since in paolina_functions they are modified and is a bit messy to keep track of which attributes are changeable and which arent, thoughts @paolafer , @jjgomezcadenas, @jerenner

### DIFF
--- a/invisible_cities/cities/penthesilea_test.py
+++ b/invisible_cities/cities/penthesilea_test.py
@@ -8,6 +8,7 @@ from .. core.core_functions    import in_range
 from .. core.system_of_units_c import units
 from .. core.testing_utils     import assert_dataframes_close
 from .. core.testing_utils     import assert_tables_equality
+from .. core.testing_utils     import assert_MChit_equality
 from .. core.configure         import configure
 from .. core.configure         import all as all_events
 from .. io                     import dst_io as dio
@@ -172,7 +173,7 @@ def test_penthesilea_true_hits_are_correct(KrMC_true_hits, config_tmpdir):
         penthesilea_hits = penthesilea_evts[evt_no]
 
         assert len(penthesilea_hits) == len(true_hits)
-        assert all(p_hit == t_hit for p_hit, t_hit in zip(penthesilea_hits, true_hits))
+        (assert_MChit_equality(p_hit, t_hit) for p_hit, t_hit in zip(penthesilea_hits, true_hits))
 
 
 @mark.skip("This scenario is not possible in liquid cities")
--- a/invisible_cities/cities/penthesilea_test.py
+++ b/invisible_cities/cities/penthesilea_test.py
@@ -16,6 +16,9 @@ from .. io.mcinfo_io           import load_mchits
 from .  penthesilea            import penthesilea
 
 
+#in order not to fail direct comparation tests when changing hit attribute we compare only the columns that penthesilea is using
+columns = ['event', 'time', 'npeak', 'Xpeak', 'Ypeak', 'nsipm', 'X', 'Y', 'Xrms', 'Yrms', 'Z', 'Q', 'E']
+
 def test_penthesilea_KrMC(KrMC_pmaps_filename, KrMC_hdst, KrMC_kdst, config_tmpdir):
     PATH_IN   = KrMC_pmaps_filename
     PATH_OUT  = os.path.join(config_tmpdir,'Kr_HDST.h5')
@@ -39,10 +42,11 @@ def test_penthesilea_KrMC(KrMC_pmaps_filename, KrMC_hdst, KrMC_kdst, config_tmpd
     df_penthesilea_dst  = dio.load_dst(PATH_OUT , 'DST' , 'Events')
     assert len(set(df_penthesilea_dst .event)) == cnt.events_out
     assert len(set(df_penthesilea_reco.event)) == cnt.events_out
-    assert_dataframes_close(df_penthesilea_reco, DF_TRUE_RECO,
-                            check_types=False)
-    assert_dataframes_close(df_penthesilea_dst , DF_TRUE_DST ,
-                            check_types=False, rtol=1e-4)
+
+    assert_dataframes_close(df_penthesilea_reco[columns], DF_TRUE_RECO[columns],
+                            check_types=False           , rtol=1e-4            )
+    assert_dataframes_close(df_penthesilea_dst          , DF_TRUE_DST          ,
+                            check_types=False           , rtol=1e-4            )
 
 
 def test_penthesilea_filter_events(config_tmpdir, Kr_pmaps_run4628_filename):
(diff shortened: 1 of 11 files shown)


paolafer:
I agree with the removals of unused attributes, since they add noise to classes. I also promote the suppression of properties for this kind of attributes.

jacg:
What is the issue with properties? There shouldn't really be much discussion on this topic. If it is something that you might read and write, without any extra functionality associated with those reads and writes (logging, caching, constraint between degrees of freedom, etc.) then it should simply be a plain attribute. If it should be read-only, or the reads and writes require extra functionality, then it has to be a property.

My point is that I don't see scope for opinion here: Either it needs to be a property, or it doesn't. If an attribute is good enough, make it an attribute. If an attribute is not good enough, make it a property.

As long as it works, the client doesn't care; the implementor should pick the simplest implementation that works.

mmkekic:
I guess the reason we need properties is inside BHit class where there is __eq__ method (and __hash__) relying on XYZ and E property of the class. However, the energy of the hit (E) has to be changed in paolina_functions so the hash of the class instance changes, guess this is something we don't want?
Further on, other attributes were also added as properties even though they get changed all the time (example: a corrected energy that changes from the default value to a calculated one)...
What is a workaround?

jacg:
Hmm, I'm looking at `BHit`.

The `.xyz` attribute is a tuple, with a read-only synonym `.XYZ`. I don't like this synonym. It just complicates the interface. I think it should go. But that's just an opinion.

There is a `.energy` attribute, with a read-only synonym `.E`. Again, I'd prefer there to be only one of these.

There is also `.pos` which is the `.xyz` values stuck into a numpy array. I can imagine that this might be convenient for the client. Whether this convenience outweighs the noise, I am not in a position to say.

There are `.X`, `.Y` and `.Z` which give read-only access to the components of `.xyz`. Same again: this could be convenient, but might not be worth it.

So far, I reckon that `.XYZ` and `.E` should be removed (though the latter might be kept in preference to `.energy`.

Now we come to `__eq__` and `__hash__`. Why do we have these? Where do we use them? (I have a vague recollection that we might have wanted to make these hashable, in order to put them in sets, in order to simplify some search algorithm, or some elimination of duplicates.)

`__eq__` is a bit dodgy, because it uses `==` on floats: I trust that we all understand the shortcomings of this. `__hash__` is only really necessary if `BHit`s are used as dictionary keys (or `set` members), but you state that paolina functions mutate these objects, so they aren't sensibly usable as keys or set members. Consequently, my guess would be that we should probably remove `__eq__` and `__hash__`.

> I guess the reason we need properties is inside `BHit` class where there is `__eq__` method (and `__hash__`) relying on `XYZ` and `E` property of the class.

Nope! That's just an implementation detail, and those `__eq__` and `__hash__` methods could perfectly well be implemented without those properties. The data accessed by those properties are accessible without them.

I propose you remove `__eq__` and `__hash__` and see if any tests break. If not, just get rid of them. I think that they are misleading and cannot be trusted because:

+ `__eq__` performs zero-tolerance comparison on floats
+ `__hash__` cannot be relied on in the places where it is needed (sets, dicts) because the object is mutable.

Alternatively, try to go functional and see whether you can replace the whole object, rather than mutating it, at the places where paolina functions need to modify the energy.

As for the redundancy in the rest of the interface, you have to weigh up whether the complexity and noise are justified by the convenience to the client. I would suggest to keep

+ `.E`
+ `.xyz`
+ (possibly) `.x`, `.y` and `.z`. I (and PEP 8) prefer lowercase, but I vaguely recall that you have uppercase in some interfaces outside of IC, which might justify uppercase here.
+ (possibly) `.pos`.

That's to say, we might tolerate *some* redundancy for the sake of convenience, but please don't go overboard with the redundancy.

mmkekic:
Thanks for looking at this in details! 
I think __hash__ should be removed since it is not used anywhere anyway, but __eq__ is used in some tests, maybe instead of == we can give some tolerance? 
About removing redundant attributes, maybe people using Hit class in the analysis can give their preferences of which to keep? @jjgomezcadenas , @paolafer , @jerenner , @ausonandres , @bpalmeiro , @Aretno 
But the conclusion is that the read-only attributes should go away?

jacg:
> the conclusion is that the read-only attributes should go away?

I see three possible conclusions:

1. Store energy and position as plain public attributes. Maybe add some convenience-redundancy with properties.
2. Same as 1. but make position private and expose it with a read-only property. Energy cannot be read-only because paolia functions modify it.
3. Same as 2. but make energy private and expose it with a read-only property. Make paolina functions replace the original object rather than mutating it, whenever the energy needs to be modified.

Why this range of options? 1 is the default Python anything-goes approach. 3 is motivated by the immutable-data functional approach, that helps to eliminate all sorts of nasty-to-track down bugs. 2 is a compromise between 1 and 3.

I'd be inclined to go with 3, unless it excessively complicates the implementation of paolina functions at the point where they mutate the energy.

jacg:
> eq is used in some tests, maybe instead of == we can give some tolerance?

If `__eq__` is only used in the tests, then I'd get rid of it from the interface of the class. We shouldn't be tempting users to use unreliable features.

paolafer:
I would prefer 1. or 2., unless somebody volunteers to change the paolina functions accordingly.

mmkekic:
I think the replacing of original object becomes complicated in functions dealing with voxels, because changing the energy of one hit inside voxels we will have to replace the whole voxel object  (it inherits from BHit), taking care that the corresponding hit has been replaced in new voxel.hits list.
The particularly complicated place to do so is in drop_voxels function (PR #615) where changes happen in a loop.
Therefore I also think option 3 is outside of the reach for now. What I don't get is what is advantage of 2 over 1?

jacg:
I would not be surprised if 3 introduced more complications than it is worth, into paolina functions. Unfortunately I don't have the time to check it.

> What I don't get is what is advantage of 2 over 1?

If the position should never change after object creation, then the interface shouldn't provide the means to do it. By removing this ability from the interface (though not removing it altogether), 2 slightly reduces the scope for bugs creeping into the code. In Python this particular issue far less significant than in many other languages, so I really don't have a strong opinion on it.

jerenner:
I would be ok with options 1 or 2, option 2 being preferred if we really never need to change the position for any reason.

jerenner (approved):
Removes several unused hits attributes

carmenromo:
Before merging resolve the conflicts please

mmkekic:
No no, this is not ready for being approved at all!! I have to remove read-only energies and replace the tests files, please @carmenromo dont try to merge before I say something (good thing there were conflicts!)

carmenromo:
ok

mmkekic:
For now I decided to only change .E to attribute (replacing .energy); and the same for .Ec. There is a lot of redundancy with coordinates but since most of the attributes and properties are being used in the code somewhere I am not sure it is worth the effort of going through it now.
Also, there was a but in load_hits; there was a square missing for rms attributes

PR #644, opened 2019-03-21
Update system of units adding some more short names for units.

jjgomezcadenas:


jmbenlloch (approved):
Straightforward change to add new units.

PR #647, opened 2019-04-01
Charge type in threshold hits

mmkekic:
Adds option to correct hits based on Qc, leaving default option to be Q

### DIFF
--- a/invisible_cities/reco/hits_functions_test.py
+++ b/invisible_cities/reco/hits_functions_test.py
@@ -2,6 +2,7 @@ import os
 import numpy                     as     np
 from   numpy.testing             import assert_allclose
 from   numpy.testing             import assert_almost_equal
+from   numpy.testing             import assert_raises
 from   .. core.configure         import configure
 from   .. evm.event_model        import Hit
 from   .. evm.event_model        import Cluster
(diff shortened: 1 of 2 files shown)


invisible_cities/reco/hits_functions_test.py
```diff
@@ -108,26 +113,34 @@ def test_threshold_hits_with_penthesilea(config_tmpdir, Kr_pmaps_run4628_filenam
...
 
+
 @given(list_of_hits(), floats())
-def test_threshold_hits_energy_conserved(hits,th):
-    hits_thresh = threshold_hits(hits,th)
-    assert_almost_equal(sum((h.E for h in hits)), sum((h.E for h in hits_thresh)))
+def test_threshold_hits_energy_conserved(hits, th):
+    hits_thresh   = threshold_hits(hits, th)
+    assert_almost_equal(sum((h.E for h in hits)), sum((h.E for h in hits_thresh  )))
+    hits_thresh_c = threshold_hits(hits, th)
```
paolafer: Are these two variables `hits_thresh` and `hits_thresh_c` intended to be equal? It looks to me that they are, since they are both the result of `threshold_hits(hits, th)` with the `uncorrected` argument being False. I'm probably missing the meaning of this test.

mmkekic: Because it indeed has no meaning, I corrected hits_thresh_c to be on corrected hits

paolafer (approved):
This PR adds the possibility of applying the threshold on the hit corrected charge instead of the raw charge. The PR is simple, with a set of tests and I approve it.

carmenromo:
Can I merge?

mmkekic:
Let me just squash the last commit

carmenromo:
ok

PR #622, opened 2019-02-04
Functions to read and apply corrections using new dataframe format

mmkekic:
The relevant functions to read map and apply geometric and lifetime
corrections are added.

PR #649, opened 2019-04-04
Add dt to z and pes to Kr scale converters

bpalmeiro:
This PR is meant to add some functionalities left in #622, as the converters from drift time to z and from pes to KeV. Also fix some minor issues that escaped the review.


invisible_cities/reco/corrections_new.py
```diff
@@ -200,10 +200,91 @@ def lt_xy_corrections(X : np.array, Y : np.array, Z : np.array, maps : ASectorMa
...
+    if norm_strat is norm_strategy.max:
+        norm_value = amap_max(map_e0).e0
+    elif norm_strat is norm_strategy.mean:
+        norm_value = np.mean(np.mean(maps.e0))
+    elif norm_strat is norm_strategy.custom:
+        if norm_value is None:
+            raise MissingArgumentError
+    else:
+        raise ValueError
+    kr_energy  = 41.5575
```
mmkekic: Add units here

invisible_cities/reco/corrections_new.py
```diff
@@ -229,11 +310,19 @@ def apply_all_correction_single_maps(map_e0     : ASectorMap,
         raise MissingArgumentError
         pass
 
+    if norm_strat is norm_strategy.max:
+        norm_value = amap_max(map_e0).e0
+    elif norm_strat is norm_strategy.mean:
+        norm_value = np.mean(np.mean(maps.e0))
```
mmkekic: I dont see a reason to have separate dataclass for max values, while not for mean, so either we add amap_mean or  remove amap_max (and FitMapValue) (Im more for second option)

invisible_cities/reco/corrections_new.py
```diff
@@ -229,11 +310,19 @@ def apply_all_correction_single_maps(map_e0     : ASectorMap,
...
+    elif norm_strat is norm_strategy.custom:
+        if norm_value is None:
+            raise ValueError
+    else:
+        raise ValueError
+
```
mmkekic, line 281: The exceptions should contain more explanations;  as norm_value has to be float; and norm_strategy not recognized

invisible_cities/reco/corrections_new.py
```diff
@@ -200,10 +200,91 @@ def lt_xy_corrections(X : np.array, Y : np.array, Z : np.array, maps : ASectorMa
...
+    return df_to_z_converter
+
+
+class norm_strategy(AutoNameEnumBase):
+    mean   = auto()
+    max    = auto()
+    custom = auto()
+
+def get_pes_to_kev_factor(map_e0    : ASectorMap,
+                         norm_strat: norm_strategy   = norm_strategy.max,
```
mmkekic: check alignment

mmkekic (commented):
Also in 
```
apply_all_correction_single_maps(map_e0     : ASectorMap,
                                 map_lt     : ASectorMap,
                                 map_te     : Optional[ASectorMap]  = None,
                                 apply_temp : bool                  = True,
                                 norm_strat : norm_strategy         = norm_strategy.max,
                                 norm_value : Optional[float]       = None,
                                 kev_output : bool                  = False

```
there is a check that raises exception if map_te is not given and apply_temp is true. However in case map_te is given but does not contain t_evol map (this can happen wince t_evol is optional inside ASectorMap), the exception  MissingArgumentError will not be raised and the code will crash later with some non-intuitive errors.

invisible_cities/reco/corrections_new.py
bpalmeiro: If units are used, this will be translating energies to the official unit in IC, that is MeV. Then, I guess that this function shouldn't be pes_to[whatever_unit] but it will always be refired to the official one, am I right?
What do the final users think? @paolafer @ausonandres

paolafer: If the function's name is `pes_to_keV`, it's correct to keep it as it is, as Brais says. If we multiply the number by the units, the function should be called `pes_to_MeV`.
Personally, I'm a keV fan :-).

invisible_cities/reco/corrections_new.py
```diff
@@ -229,6 +229,46 @@ class norm_strategy(AutoNameEnumBase):
     max    = auto()
     custom = auto()
 
+def get_pes_to_kev_converter(map_e0    : ASectorMap,
```
paolafer: I would use the capital V for `keV`.

invisible_cities/reco/corrections_new.py
```diff
@@ -229,6 +229,46 @@ class norm_strategy(AutoNameEnumBase):
...
+    elif norm_strat is norm_strategy.custom:
+        if norm_value is None:
+            raise MissingArgumentError
+    else:
+        raise ValueError
+
+    kr_energy  = 41.5575
+    norm_value = kr_energy / norm_value
+
+    def pes_to_kev_converter(energy):
```
paolafer: Same here, capital V.

invisible_cities/reco/corrections_new.py
mmkekic: I think it is a bit dangerous not to have units. This function will be used for example in esmeralda where a user has to define the threshold voxel energy, and if we dont have units here it means that the user can not specify units in the threshold energy parameter neither right?

paolafer: Yeah, you've got a point there. I'm afraid we need to be clear with the units.

mmkekic: i think we should try not to have explicit unit in the function name, maybe something like to krypton_energy_scale or similar

paolafer: Fine with me.

invisible_cities/reco/corrections_new.py
```diff
@@ -229,41 +325,66 @@ def apply_all_correction_single_maps(map_e0     : ASectorMap,
...
-            temp_correct_lt = lambda t : time_coefs_corr(t,
-                                                         evol_table.ts,
-                                                         evol_table['lt'],
-                                                         evol_table.ltu)
-            e0evol_vs_t     = temp_correct_e0
-            ltevol_vs_t     = temp_correct_lt
+        try:
+            assert map_te.t_evol is not None
+        except:
+            raise TimeEvolutionTableMissing
```
mmkekic: Id try to catch both the exceptions, 
```
except (AttributeError , AssertationError):
raise ...
```

mmkekic: And remove 
```
if apply_temp and map_te is None:
        raise MissingArgumentError
        pass
```

jacg:
>        try:
>            assert map_te.t_evol is not None
>        except:
>            raise TimeEvolutionTableMissing

First of all, a general point about the third line of the above sample: please do not write `except` immediately followed by a colon. Exceptions are messages. The `except` keyword should be followed by the type of message that you are expecting, understand, and are prepared to handle. By writing `except:` you are saying "I don't care what the message is, I know that *this* is the correct way to react to it". This is patently absurd!

Now, something specific about the whole four lines of code shown above: isn't this just a very perverse way of spelling 

    if map_te.t_evol is not None:
        raise TimeEvolutionTableMissing

?

Or is there some richer semantics covered by the missing exception specification? I initially guessed that you were handling `AssertionError` but now I wonder whether you are reacting to `AttributeError` ... or maybe both? But the point is that, you are making me guess and wonder, and the code shouldn't do that: it should be clear.

These four lines are an abomination and need to rewritten in some more sensible way.

jacg:
Ah, I see that @mmkekic has already made this point. Ignore me. I should go to sleep.

invisible_cities/reco/corrections_new.py
```diff
@@ -200,10 +184,111 @@ def lt_xy_corrections(X : np.array, Y : np.array, Z : np.array, maps : ASectorMa
...
+    ----------
+    map_te : AsectorMap
+        Correction map with time evolution of some kdst parameters.
+
+    Returns
+    -------
+    A function that returns z converted array for a given drift time input
+    array.
+    """
+
```
mmkekic, line 190: Maybe we can raise exception if map_te.t_evol is None (which is the case of MC)

invisible_cities/reco/corrections_new.py
```diff
@@ -200,10 +184,111 @@ def lt_xy_corrections(X : np.array, Y : np.array, Z : np.array, maps : ASectorMa
...
+            raise ValueError(s)
+    else:
+        s  = "None of the current available normalization"
+        s += " strategies was selected"
+        raise ValueError(s)
+
+    return norm_value
+
+def get_pes_to_krscale_factor(map_e0    : ASectorMap,
+                              norm_strat: norm_strategy   = norm_strategy.max,
```
mmkekic: I think kripton scale normalization should just be one more option in norm_strategy, no need for a separate function

invisible_cities/reco/corrections_new.py
bpalmeiro: I agree, but maybe mc should have the table in order to converge data and mc analysis.

invisible_cities/reco/corrections_new_test.py
```diff
@@ -183,23 +170,89 @@ def test_time_coefs_corr(map_filename, time):
...
+def test_get_normalization_factor_custom_norm(correction_map_filename, custom_vaule):
+    map_e  = read_maps(correction_map_filename)
+    factor = get_normalization_factor(map_e,
+                                      norm_strategy.custom,
+                                      custom_vaule)
+    norm   = custom_vaule
+    assert factor == norm
+
+def test_get_normalization_factor_krscale(correction_map_filename):
+    map_e     = read_maps(correction_map_filename)
```
mmkekic, line 229: review indentation

invisible_cities/reco/corrections_new_test.py
```diff
@@ -183,23 +170,89 @@ def test_time_coefs_corr(map_filename, time):
...
+    assert_raises(TimeEvolutionTableMissing,
+                  apply_all_correction_single_maps,
+                  map_e, map_e, None)
+
+def test_apply_all_correction_single_maps_raises_exception_when_invalid_map(map_filename_MC):
+    map_e = read_maps(map_filename_MC)
+    assert_raises(TimeEvolutionTableMissing,
                   apply_all_correction_single_maps,
-                  maps.e0, maps.lt, None, True)
+                  map_e, map_e, map_e)
```
mmkekic: Add apply_temp = True so that it is obvious why the test fails (ie now it looks like it will always fail with MC map)

mmkekic (approved):
This PR adds  time to Z conversion and normalization to the krypton keV scale, fixes old bugs and improves structure of the code in general. Good job!

PR #650, opened 2019-04-10
Allow for 0 S1s after penthesilea S1 selection.

Aretno:
Previous code (PR #640 ) only worked if there where 0 S1s at the pmap level, not
after the S1 selection in penthesilea. This commit adapts the code to
that possibility which is the intended behavior.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -430,7 +430,7 @@ def hit_builder(dbfile, run_number, drift_v, reco, rebin_slices, rebin_method):
         # take events with exactly one s1. Otherwise, the
         # convention is to take the first peak in the S1 object
         # as reference.
-        if len(selector_output.s1_peaks) > 0:
+        if np.any(selector_output.s1_peaks):
             first_s1 = np.where(selector_output.s1_peaks)[0][0]
             s1_t     = pmap.s1s[first_s1].time_at_max_energy
         else:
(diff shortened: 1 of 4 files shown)


jmbenlloch:
I'm sorry but I don't really understand how the change in the code works. If none of the previous tests were failing with this bug it would be interesting to add a test for it. Can you do it @Aretno?

jacg:
> If none of the previous tests were failing with this bug it would be interesting to add a test for it.

This is a very important principle: if you think you have found a bug, prove it by writing a test. Add it to the test suite, to prevent regression.

mmkekic:
Not related directly to this bug, but we probably also want to fix nonsense number for s1_peak in kdst output file in case of 0 s1s. The default value for that number is -1 (set in KrEvent class), however when saving it to h5 file since we use unsigned integers it outputs 65535. If we want to keep -1 in the output table then change  KrTable tb.is_description class in nh5.

mmkekic:
And related to the particular part of the code, that has nothing to do with the actual 0 S1, seems its a bug in case no s1 peak passes the filter. So a test where, for example, a very high threshold for s1 peak energy is given will fail.

Aretno:
Yup, the failure will happen exactly as @mmkekic is saying. I'll add a test where penthesilea is run with a high S1 threshold to the PR as soon as I can.

jmbenlloch (approved):
This PR fixes a bug including a new test that shows it. Good job!

PR #654, opened 2019-05-01
Faster function to load MC hits

paolafer:
In this PR I changed the function called by `load_mchits()`, which loads the true MC hits of a range of events in a file. The previous function  loads all the true information (including particles) and returns only hits, while this function loads only the hits. The time is reduced to ~75% of the original time, for a file of ~8000 events of 511-keV gamma interactions, therefore I thought that this new implementation could be worth it. I haven't added any test, because we have already a test for the `load_mchits()` function, which uses internally the new function. This change should be transparent to the user.

### DIFF
--- a/invisible_cities/io/mcinfo_io.py
+++ b/invisible_cities/io/mcinfo_io.py
@@ -292,7 +292,7 @@ def compute_mchits_dict(mcevents:Mapping[int, Mapping[int, MCParticle]]) -> Mapp
 
 
 def read_mchit_evt (mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
-                     event_number: int, last_row=0) -> ([tb.Table], [tb.Table], [tb.Table]):
+                     event_number: int, last_row=0) -> [tb.Table]:
     h5extents    = mctables[0]
     h5hits       = mctables[1]
 
--- a/invisible_cities/io/mcinfo_io.py
+++ b/invisible_cities/io/mcinfo_io.py
@@ -235,7 +235,7 @@ def read_mcinfo_evt (mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
 
 def read_mcinfo(h5f, event_range=(0, int(1e9))) -> Mapping[int, Mapping[int, Sequence[MCParticle]]]:
     mc_info = tbl.get_mc_info(h5f)
-    
+
     h5extents = mc_info.extents
 
     events_in_file = len(h5extents)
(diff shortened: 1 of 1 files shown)


invisible_cities/io/mcinfo_io.py
```diff
@@ -292,6 +291,65 @@ def compute_mchits_dict(mcevents:Mapping[int, Mapping[int, MCParticle]]) -> Mapp
     return mchits_dict
 
 
+def read_mchit_evt (mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
+                     event_number: int, last_row=0) -> ([tb.Table], [tb.Table], [tb.Table]):
```
mmkekic: The return type of the function is just one tb.Table

mmkekic (commented):
Since most of the code in both function is just a copy from `read_mcinfo_evt` and `read_mcinfo`, isnt it better to have a flag in those function as `return_only_hits = False`  and make a minimal change to continue the loop if particle and generator information is not needed?

invisible_cities/io/mcinfo_io.py
paolafer: Ok!

paolafer:
Ok, I added a flag to the `read_mcinfo_evt` function. Tests are running fine.

invisible_cities/io/mcinfo_io.py
```diff
@@ -209,16 +209,20 @@ def read_mcinfo_evt (mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
                 previous_row = h5extents[iext-1]
 
                 ihit         = int(previous_row['last_hit']) + 1
-                ipart        = int(previous_row['last_particle']) + 1
+                if not return_only_hits:
+                    ipart        = int(previous_row['last_particle']) + 1
 
             ihit_end  = this_row['last_hit']
-            ipart_end = this_row['last_particle']
+            if not return_only_hits:
```
mmkekic: Cant we lower  `ipart_end = this_row['last_particle']`  and put it after `if return_only_hits: break`

invisible_cities/io/mcinfo_io.py
```diff
@@ -187,8 +187,8 @@ def load_mcsensor_response(file_name: str,
         return read_mcsns_response(h5in, event_range)
 
 
-def read_mcinfo_evt (mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
-                     event_number: int, last_row=0) -> ([tb.Table], [tb.Table], [tb.Table]):
+def read_mcinfo_evt (mctables: (tb.Table, tb.Table, tb.Table, tb.Table), event_number: int, last_row=0,
+                     return_only_hits=False) -> ([tb.Table], [tb.Table], [tb.Table]):
```
mmkekic: can you add `bool` for the return_only_hits type?

mmkekic (commented):
I think after this changes this PR can be approved

invisible_cities/io/mcinfo_io.py
paolafer: Done.

invisible_cities/io/mcinfo_io.py
paolafer: Yes.

mmkekic (approved):
This PR adds a more efficient way to read mchit info only from MC tables, with a minimal change to existing functions. Good job :)

PR #655, opened 2019-05-09
Fix drop voxels

mmkekic:
This PR shows that drop_voxel depends on voxels order when there are 2 voxels at the same distance from the dropped one and fix the drop voxels function such that it redistributes energy to all the voxels that are at the minimum distance from the dropped one.


jahernando:
Hi Marija,

The *drop_end_point_voxels* functions is quite readable and well written.

I got confuse by the conde under the condition *(len(pos)==0)*, where there are removing a 'voxel', but inside the code use the *bar_pos* variable that has not been defined or computed. Can you have a look and clarify the code there? To me the condition *(len(pos) == 0)* should never occur right?

The rest of the changes are quite straight-forward. 

The rest respect the ordering is quite clear.

Clarify the issue commented above and we proceed to approve.

mmkekic:
Hi, I agree there is a typo that should be fixed. The whole len(pos)==0 is in case there are voxels with no associated hits (in the previous version of voxel class it didnt have hits attribute). @paolafer can you remind us why we still want to keep this possibility? 
If we decide to keep the possibility then we should add tests as well.

paolafer:
The `len(pos)==0` condition is needed because we want to deal also with voxels that are defined without starting from any hits. The voxel object is quite useful by itself and we decided to keep this possibility in one of the past PRs.

mmkekic:
> I got confuse by the conde under the condition _(len(pos)==0)_, where there are removing a 'voxel', but inside the code use the _bar_pos_ variable that has not been defined or computed. Can you have a look and clarify the code there? To me the condition _(len(pos) == 0)_ should never occur right?

As @paolafer  said this is the case when voxels are build without associated  hits. There was a typo in this case and no test failed since the fixture for testing with voxels_without_associated_hits had all the energy the same so no voxel was being dropped. I changed the energies to ensure the test fails and then corrected the typo.

mmkekic:
@jahernando , @paolafer  who is reviewing this?

paolafer:
I thought that @jahernando was the assigned reviewer...

invisible_cities/reco/paolina_functions.py
```diff
@@ -321,38 +321,32 @@ def drop_voxel(voxels: Sequence[Voxel], the_vox: Voxel) -> int:
...
-            for v in voxels:
-                if neighbours(the_vox, v):
-                    dist = np.linalg.norm(the_vox.pos - v.pos)
-                    if dist < min_dist:
-                        min_dist = dist
-                        min_v = v
-
-            ### add voxel energy to voxel
-            min_v.E += the_vox.E
+            min_dist  = min([np.linalg.norm(the_vox.pos-v.pos) for v in voxels  if neighbours(the_vox, v)])
```
gonzaponte: no need for `[]`

invisible_cities/reco/paolina_functions.py
```diff
@@ -321,38 +321,32 @@ def drop_voxel(voxels: Sequence[Voxel], the_vox: Voxel) -> int:
...
+            min_v     = [v for v in voxels  if (neighbours(the_vox, v) and np.isclose(np.linalg.norm(the_vox.pos-v.pos), min_dist))]
+
```
gonzaponte: I think you can first define `neighbour_voxels = [v for v in voxels  if neighbours(the_vox, v)]` and then use it for `min_dist` and for `min_v`. This would reduce looping twice over the voxels that you already know they are not close.

gonzaponte: Also maybe you can define `def dist_to_the_vox(vox): return np.linalg(the_vox.pos - vox.pos)` to improve the readability of the code

invisible_cities/reco/paolina_functions.py
```diff
@@ -321,38 +321,32 @@ def drop_voxel(voxels: Sequence[Voxel], the_vox: Voxel) -> int:
...
-                        min_dist = dist
-                        min_hit = hh
-                        min_v = v
-
-        ### add voxel energy to hit and to voxel, separately
-        setattr(min_hit, e_type, getattr(min_hit, e_type) + the_vox.E)
-        min_v.E += the_vox.E
+
+        min_dist    = min([np.linalg.norm(bary_pos-hh.pos) for v in voxels for hh in v.hits if neighbours(the_vox, v)])
+        min_h_v  = [(h, v) for v in voxels for h in v.hits if ((neighbours(the_vox, v) and np.isclose(np.linalg.norm(bary_pos-h.pos), min_dist)))]
```
gonzaponte: same comments as before apply here

gonzaponte: Another possibility is something like:
```
# this will order them by distance
dist_hit_vox = sorted((dist_to_bar(h), h, v) for v in voxels for h in v.hits)
closest = list(itertools.takewhile(lambda x: isclose(x, dist_hit_vox[0][0]), dist_hit_vox))
for (d, h, v) in closest:
    # do stuff
```

Not necessarily better, but something to explore.

invisible_cities/reco/paolina_functions_test.py
```diff
@@ -565,6 +565,25 @@ def test_tracks_with_dropped_voxels(ICDATADIR):
...
+    hits = all_hits[evt_number].hits
+    voxels = voxelize_hits(hits, vox_size, strict_voxel_size=False)
+    ini_trks = make_track_graphs(voxels)
+    initial_n_of_tracks = len(ini_trks)
+    ini_energies = [sum(vox.E for vox in t.nodes()) for t in ini_trks]
+    ini_n_voxels = np.array([len(t.nodes()) for t in ini_trks])
+    mod_voxels   = drop_end_point_voxels(sorted(voxels, key = lambda v:v.E, reverse = False), e_thr, min_voxels)
+    mod_voxels_r = drop_end_point_voxels(sorted(voxels, key = lambda v:v.E, reverse = True ), e_thr, min_voxels)
+    for v1, v2 in zip(sorted(mod_voxels, key = lambda v:v.E), sorted(mod_voxels_r, key = lambda v:v.E)):
+        assert np.isclose(v1.E, v2.E)
```
gonzaponte: please align stuff

invisible_cities/reco/paolina_functions_test.py
```diff
@@ -565,6 +565,25 @@ def test_tracks_with_dropped_voxels(ICDATADIR):
     assert np.allclose(ini_energies, energies)
     assert np.all(ini_n_voxels - n_voxels == expected_diff_n_voxels)
 
+def test_drop_voxels_deterministic(ICDATADIR):
+    hit_file = os.path.join(ICDATADIR, 'tracks_0000_6803_trigger2_v0.9.9_20190111_krth1600.h5')
+    evt_number = 19
+    e_thr = 5867.92
+    min_voxels = 3
+    size = 15.
+    vox_size = np.array([size,size,size],dtype=np.float16)
```
gonzaponte: spaces after commas

gonzaponte (commented):
good effort! some suggestions about the implementation. The test looks good.

gonzaponte (approved):
Neat fix to drop voxels that shares the energy among many hits if they are at the same distance as the closest one.

Good job!

invisible_cities/reco/paolina_functions.py
gonzaponte: Discarded because the author doesn't think improves readability

invisible_cities/reco/paolina_functions.py
gonzaponte: Discarded because the author doesn't think it adds anything to the code

PR #656, opened 2019-05-27
Penthesilea sn option

andLaing:
Adds the option to use signal to noise
instead of raw charge for the reconstruction
in penthesilea.

Also improves a couple of functions based
on advice from @mmkekic


invisible_cities/cities/components.py
```diff
@@ -17,38 +17,41 @@
...
+from .. evm    .event_model       import                    MCInfo
+from .. evm    .pmaps             import                SiPMCharge
+from .. core   .system_of_units_c import                     units
+from .. core   .exceptions        import                XYRecoFail
+from .. core   .exceptions        import              NoInputFiles
+from .. core   .exceptions        import              NoOutputFile
+from .. core   .exceptions        import InvalidInputFileStructure
+from .. core   .configure         import                EventRange
+from .. core   .configure         import          event_range_help
+from .. core   .random_sampling   import              NoiseSampler as siN
```
gonzaponte: can you pick a more descriptive name for this import? It is a bit cryptic.. even if it doesn't match the other ones, it is not used very frequently, so even the full name would be fine

invisible_cities/cities/components.py
```diff
@@ -421,7 +427,7 @@ def build_pointlike_event(pmap, selector_output, event_number, timestamp):
             evt.S2t.append(peak.time_at_max_energy)
 
             xys = sipm_xys[peak.sipms.ids           ]
-            qs  =          peak.sipms.sum_over_times
+            qs  = peak.sipm_charge_array(sipm_noise, charge_type, True)
```
gonzaponte: use `single_point=True` instead of only `True` for readability

invisible_cities/cities/components.py
```diff
@@ -493,25 +503,28 @@ def build_hits(pmap, selector_output, event_number, timestamp):
 
             peak = pmf.rebin_peak(peak, rebin_slices, rebin_method)
 
-            xys     = sipm_xys[peak.sipms.ids           ]
-            qs      =          peak.sipms.sum_over_times
+            xys  = sipm_xys[peak.sipms.ids]
+            qs   = peak.sipm_charge_array(sipm_noise, charge_type, True)
```
gonzaponte: `single_point=True`

invisible_cities/cities/components.py
```diff
@@ -493,25 +503,28 @@ def build_hits(pmap, selector_output, event_number, timestamp):
...
             try              : cluster = barycenter(xys, qs)[0]
             except XYRecoFail: xy_peak = xy(NN, NN)
             else             : xy_peak = xy(cluster.X, cluster.Y)
 
-            for slice_no, t_slice in enumerate(peak.times):
+            sipm_charge = peak.sipm_charge_array(sipm_noise ,
+                                                 charge_type)
```
gonzaponte: `single_point=False`

invisible_cities/core/random_sampling_test.py
```diff
@@ -320,3 +321,13 @@ def test_noise_sampler_signal_to_noise(noise_sampler,
 
     assert len(signal_to_noise) == len(ids)
     assert np.allclose(np.round(signal_to_noise, 2), expected)
+
+
+def test_signal_to_noise_zero_sample_raises_error(noise_sampler):
+
+    noise_sampler, *_ = noise_sampler
```
gonzaponte, line 328: this line (and the fixture the test takes) do nothing, right? if so they can be removed :)

invisible_cities/evm/pmaps.py
```diff
@@ -2,8 +2,16 @@
...
+from enum import auto
+
 from .. core.system_of_units_c import units
 from .. core.core_functions    import weighted_mean_and_std
+from .. types.ic_types         import      AutoNameEnumBase
+
+
+class SiPMCharge(AutoNameEnumBase):
+    raw = auto()
+    sn  = auto()
```
gonzaponte: I would prefer a more verbose name for `sn`. maybe even `signal_to_noise`

invisible_cities/evm/pmaps.py
```diff
@@ -98,7 +106,31 @@ def _check_valid_input(self, times, pmts, sipms):
             raise ValueError(msg)
 
 class S1(_Peak): pass
-class S2(_Peak): pass
+class S2(_Peak):
+
+    def sipm_charge_array(self, noise_func, charge_type,
+                          single_point=False):
+
+        if charge_type == SiPMCharge.raw:
```
gonzaponte:
```suggestion
        if charge_type is SiPMCharge.raw:
```

invisible_cities/evm/pmaps.py
```diff
@@ -98,7 +106,31 @@ def _check_valid_input(self, times, pmts, sipms):
...
+        ## future.
+        if single_point:
+            charges    = self.sipms.sum_over_times
+            sample_wid = int(np.ceil(np.round(self.width) / units.mus))
+            return noise_func(self.sipms.ids, charges, sample_wid)
+
+        sample_widths = np.ceil(np.round(self.bin_widths) / units.mus)
+        mapping       = map(noise_func,
+                            np.repeat(self.sipms.ids[None, :],
+                                      len(self.bin_widths), 0),
```
gonzaponte: I think assigning this `repeat` call to a variable with a meaningful name will make it more readable

gonzaponte: Incidentally, this `[None, :]` means creating a new axis? If so use `[np.newaxis]` better.

invisible_cities/evm/pmaps_test.py
```diff
@@ -45,7 +52,7 @@ def sensor_responses(draw, n_samples=None, subtype=None, ids=None):
 
 @composite
 def peaks(draw, subtype=None, pmt_ids=None, with_sipms=True):
-    nsamples      = draw(integers(1, 50))
+    nsamples      = draw(integers(1, 20))
```
gonzaponte, line 56: was it taking too long?

invisible_cities/evm/pmaps_test.py
```diff
@@ -305,3 +315,58 @@ def test_PMap_s2s(pmps):
     assert len(pmp.s2s) == len(s2s)
     for kept_s2, true_s2 in zip(pmp.s2s, s2s):
         assert_Peak_equality(kept_s2, true_s2)
+
+
+
+@fixture(scope='module')
+def noise_func():
```
gonzaponte: `signal_to_noise_6400` would be a better name

invisible_cities/evm/pmaps_test.py
```diff
@@ -305,3 +315,58 @@ def test_PMap_s2s(pmps):
...
+    return NoiseSampler('new', 6400).signal_to_noise
+
+
+@fixture(scope='module')
+def s2_peak():
+    times      = np.arange(0, 20 * units.mus, 1 * units.mus)
```
gonzaponte:
```suggestion
    times      = np.arange(20) * units.mus
```

invisible_cities/evm/pmaps_test.py
```diff
@@ -305,3 +315,58 @@ def test_PMap_s2s(pmps):
...
+    bin_widths = np.full(len(times), 1 * units.mus)
```
gonzaponte:
```suggestion
    bin_widths = np.full_like(times, units.mus)
```

invisible_cities/evm/pmaps_test.py
```diff
@@ -305,3 +315,58 @@ def test_PMap_s2s(pmps):
...
+
+    pmt_ids  =    DB.DataPMT ('new', 6400).SensorID.values
+    n_sipms = len(DB.DataSiPM('new', 6400).SensorID.values)
+    sipm_ids = np.arange(n_sipms)
```
gonzaponte:
```suggestion
    sipm_ids = DB.DataSiPM('new', 6400).index.values
```

removing the previous line alse

gonzaponte: It can be also used for PMTs, but it doesn't matter

invisible_cities/evm/pmaps_test.py
```diff
@@ -305,3 +315,58 @@ def test_PMap_s2s(pmps):
...
+
+    sipms = SiPMResponses(sipm_ids,
+                          np.random.uniform(0, 10,
+                                            (len(sipm_ids), len(times))))
+
+    return S2(times, bin_widths, pmts, sipms)
+
+
+@mark.parametrize("charge_type",
+                  (SiPMCharge.raw, SiPMCharge.sn))
```
gonzaponte: `@mark.parametrize("charge_type", SiPMCharge)` is enough

invisible_cities/evm/pmaps_test.py
```diff
@@ -305,3 +315,58 @@ def test_PMap_s2s(pmps):
...
+def test_sipm_charge_array(charge_type,
+                           s2_peak    ,
+                           noise_func ):
+    charge_arr = s2_peak.sipm_charge_array(noise_func ,
+                                           charge_type,
+                                           False      )
```
gonzaponte: `single_point=False`

invisible_cities/evm/pmaps_test.py
```diff
@@ -305,3 +315,58 @@ def test_PMap_s2s(pmps):
...
+
+    all_wf = s2_peak.sipms.all_waveforms
+    assert np.array(charge_arr).shape[0] == all_wf.shape[1]
+    assert np.array(charge_arr).shape[1] == all_wf.shape[0]
```
gonzaponte: do you need these `np.array` conversion? If not remove it, if you do maybe convert it in the definition to remove a bit of noise here.

gonzaponte: also you could just do `assert charge_arr.shape == all_wf.T.shape`

invisible_cities/evm/pmaps_test.py
```diff
@@ -305,3 +315,58 @@ def test_PMap_s2s(pmps):
...
+    assert np.count_nonzero(charge_arr) == np.count_nonzero(all_wf)
```
gonzaponte: Shouldn't they be exactly the same values? If so add an assertion for the values using `np.allclose`

invisible_cities/evm/pmaps_test.py
```diff
@@ -305,3 +315,58 @@ def test_PMap_s2s(pmps):
...
+
+
+@mark.parametrize("charge_type",
+                  (SiPMCharge.raw, SiPMCharge.sn))
```
gonzaponte: `@mark.parametrize("charge_type", SiPMCharge)`

invisible_cities/evm/pmaps_test.py
```diff
@@ -305,3 +315,58 @@ def test_PMap_s2s(pmps):
...
+def test_sipm_charge_array_single(charge_type,
+                                  s2_peak    ,
+                                  noise_func ):
+    charge_arr = s2_peak.sipm_charge_array(noise_func ,
+                                           charge_type,
+                                           True       )
```
gonzaponte: `single_point=True`

invisible_cities/evm/pmaps_test.py
```diff
@@ -305,3 +315,58 @@ def test_PMap_s2s(pmps):
...
+
+    assert charge_arr.shape == s2_peak.sipms.ids.shape
+    orig_zeros = np.count_nonzero(s2_peak.sipms.sum_over_times)
+    assert np.count_nonzero(charge_arr) == orig_zeros
```
gonzaponte: Same comment about the value comparison.

gonzaponte (commented):
Some small comments mostly about the readability of the code. Has @andLaing been seduced by the dark side?

invisible_cities/core/random_sampling_test.py
andLaing: They already have the NoiseSampler class defined for use so I don't have to do NoiseSampler(dbnew, run_number, nsamples, smear) every time I use it

gonzaponte: I'm sorry, I didn't see the call to `noise_sampler.signal_to_noise` when I read the test. I don't know how I've missed it!

invisible_cities/evm/pmaps_test.py
andLaing: I think so, although I can't remember details any more. In any case I think at 20 it tests everything well enough. If you disagree I can have another look to see where I can improve on speed

invisible_cities/evm/pmaps_test.py
andLaing: I need the np.array because of the way I wrote the function which might be an indication of its suboptimal style. In raw mode it just returns the transpose (essentially) so it wouldn't be necessary but in signal_to_noise mode it's a tuple basically because it seemed like a waste to convert to an array for the uses it gets in the reconstruction city (it just gets looped over)

it returns tuple(map()) which would have to be np.array(tuple(map())) (as far as I know) to avoid this line in the test but would add another operation when running over data. An extra operation seemed more acceptable in a test than in the other case

invisible_cities/evm/pmaps_test.py
gonzaponte: You are right. I don't see any reason for the quality test to drop because of that.

invisible_cities/evm/pmaps_test.py
andLaing: Again only in raw mode. Otherwise its s/N

invisible_cities/evm/pmaps_test.py
gonzaponte: Yes, I agree. I was asking that, in that case, you move the conversion to a previous line (in the test) so it doesn't add noise to the assertion. That may be me being too picky, but I like the assertions to be as clean and clear as possible, as they are usually the lines that define what a test aims to do.

invisible_cities/evm/pmaps_test.py
gonzaponte: Mm ok. I think I would add some test that checks the output values. Is it too difficult?

andLaing: Looking into it

andLaing: I've put something together but I'm not sure I'm completely happy.

Let me know what you think

gonzaponte: Selecting a channel and a slice is because the array is too big, right? 20 number is not that much, but I guess it's ok. Besides, I would like those numbers to have more decimals to decrease the tolerance, but if it takes too long you can leave it as is. Finally, I would include the expected values thing inside the `s2_peak` fixture cause they are dependent on each other. Instead of 1 and 2, use something more descriptive like single_point or whatever you find appropriate (I am not very familiar with the nomenclature in this case)

andLaing: Will combine with S2_peak as you say.

On the shapes, 20 is the number of slices, there are 1792 channels so in the single case the returned tuple has shape (1792,) and for the array (20, 1792). Since I'm comparing to hand written values a slice seemed sufficient to check but...

gonzaponte: Yes, I agree on the channel, I was referring to the number of slices.

gonzaponte: wait, I thought we agreed on only one channel, why did you add more?

andLaing: I compare the 20 slices, 6 channels (one more than I had before).

gonzaponte: Ah! I think I misunderstood the shapes. I thought the 5 things (now 6) were the time slices... sorry about that. Anyway, I am not bothering you anymore. If you want to reduce the amount of numbers to check to make it less ugly feel free to do so.

andLaing: If this level of ugly is ok with the bosses, I'm happy not to touch it again!

gonzaponte:
Please rebase so I can approve.

gonzaponte (approved):
This PR adds the option of using signal to noise ratios in the reconstruction. It adds a number of tests that ensure the results are correct. The raw option is still the default one. Good job!

PR #657, opened 2019-05-29
Upload database

carmenromo:
NEW Database has been uploaded with the gains for SiPMs and PMTs as well as the pdfs based on the calibrations from run 7146.

sipmRunDifferencePlots

pmtRunDifferencePlots


andLaing:
Do we know why run 7172 is different?

carmenromo:
> Do we know why run 7172 is different?

No, I don't know, the fit has been performed correctly and the steps applied to the LED gave enough light.

carmenromo:
We have decided to upload the DB with the values from runs 7168, 7169, 7171 because all of them have similar values between them and with the previous runs. So the values from run 7172 will be ignored.

pmtRunDifferencePlots

jmbenlloch (approved):
This PR includes the latest version of the calibration and passes all the tests.

PR #658, opened 2019-06-26
Added protection for rebin bin time

andLaing:
Only uses positive charge for
weighting to avoid non-sequential
bins.


gonzaponte (approved):
Perfect! Good way to get rid of the negative weights that were messing up the order of the time bins.

gonzaponte:
Are you sure about this last commit? I don't think there is a huge difference in any case, but at first glance, the previous option seems less invasive, meaning that you only do something when it is actually needed.

andLaing:
Yes, sorry. I realised after the push that it was wrong and undid it. We're seeing some fail paths though due to an issue with the deconv and/or the selection of the peaks. I'mm looking at protections but I think it's a wider issue that should be addressed so I'll put an issue

andLaing:
I'm going to close this pull request as we found some failures in high statistics testing.

The changes plus additional protections are revisited in PR #683

PR #659, opened 2019-07-03
Run cities on empty files

### LINKED ISSUE #653: Irene crashes when RWF file is empty
irene crashes if no waveforms in input file, generating an empty output file which in turn cannot be processed by penthesilea/dorothea. In v0.9.9 (current production version), irene just does not create a pmap file when the input RWF file is empty (note this is a common situation in trigger2 data production as several binary files do not contain trigger2 data). This should be fixed before we promote current IC version to a "data production" release.

Traceback below:

  File "/software/IC-dev//bin/city", line 21, in <module>
    city_function(**configure(args))
  File "/software/IC-dev/invisible_cities/cities/components.py", line 86, in proxy
    result = city_function(**vars(conf))
  File "/software/IC-dev/invisible_cities/cities/irene.py", line 151, in irene
    full_pmap  = empty_pmaps    .future))
  File "/software/IC-dev/invisible_cities/dataflow/dataflow.py", line 253, in push
    for item in source:
  File "/software/IC-dev/invisible_cities/cities/components.py", line 224, in wf_from_files
    event_info  = get_event_info  (h5in)
  File "/software/IC-dev/invisible_cities/cities/components.py", line 199, in get_event_info
    return h5in.root.Run.events
  File "/software/miniconda3_dev/envs/IC-3.7-2018-11-14/lib/python3.7/site-packages/tables/group.py", line 840, in __getattr__
    return self._f_get_child(name)
  File "/software/miniconda3_dev/envs/IC-3.7-2018-11-14/lib/python3.7/site-packages/tables/group.py", line 712, in _f_get_child
    self._g_check_has_child(childname)
  File "/software/miniconda3_dev/envs/IC-3.7-2018-11-14/lib/python3.7/site-packages/tables/group.py", line 399, in _g_check_has_child
    % (self._v_pathname, name))
tables.exceptions.NoSuchNodeError: group ``/`` does not have a child named ``Run``

jmbenlloch:
Some of you must have been at 47.62 astronomical units from the black hole in M87 galaxy* ... For me it's like a week has already passed since our last software coordination meeting :stuck_out_tongue_winking_eye:

This PR includes new tests to check all cities run in production can handle empty files as input. There are two kinds of empty files: one without any structure, just an empty HDF5 file and the rest of them has the proper structure for each city but no events inside. The updates needed to run over those files without breaking IC are already included. This PR closes #653.

*In your frame of reference 4.5 hours has already passed while in mine it's been a whole week. The caveat in this approximation is that our M87-like black hole is a non-rotating one with spherical symmetry  and I'm at an infinite distance from it, where I'm not affected by its gravitational field. [This](https://en.wikipedia.org/wiki/Gravitational_time_dilation#Outside_a_non-rotating_sphere) is the approximation used.

### DIFF
--- a/invisible_cities/cities/cities_test.py
+++ b/invisible_cities/cities/cities_test.py
@@ -1,0 +1,26 @@
+import os
+import tables as tb
+
+from importlib import import_module
+from pytest import mark
+from .. core.configure      import configure
+
+cities = "irene dorothea penthesilea esmeralda".split()
+
+@mark.parametrize("city", cities)
+def test_city_empty_input_file(config_tmpdir, ICDATADIR, city):
+    # All cities run in Canfranc must run on an empty file
+    # without raising any exception
+
+    PATH_IN  = os.path.join(ICDATADIR    , 'empty_file.h5')
+    PATH_OUT = os.path.join(config_tmpdir, 'empty_output.h5')
+
+    config_file = 'dummy invisible_cities/config/{}.conf'.format(city)
+    conf = configure(config_file.split())
+    conf.update(dict(files_in      = PATH_IN,
+                     file_out      = PATH_OUT))
+
+    module_name   = f'invisible_cities.cities.{city}'
+    city_function = getattr(import_module(module_name), city)
+
+    city_function(**conf)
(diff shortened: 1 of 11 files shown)


jjgomezcadenas (approved):
Earth to Andromeda: We have corrected for the time-shift difference that delays our inertial reference frames in two weeks and can therefore approve this brilliant PR. Thanks for the light-speed work.

PR #660, opened 2019-07-03
Upload database

carmenromo:
NEW Database has been uploaded with the gains for SiPMs and PMTs as well as the pdfs based on the calibrations from run 7235.

sipmRunDifferencePlots

pmt_gain_comparison


jmbenlloch (approved):
I have checked the database and new calibrations are added both for PMTs and SiPMs. All tests pass, so approved!

PR #662, opened 2019-07-16
Upload database to fix masked channel

carmenromo:
With this PR DB is uploaded in order to unmask PMT channel in neutrinos and in Canfranc.


jmbenlloch (approved):
This PR fixes an error en the local copy of the database. Approved.

PR #663, opened 2019-07-31
Upload database

carmenromo:
NEW Database has been uploaded with the gains for SiPMs and PMTs as well as the pdfs based on the calibrations from run 7389. SiPM 25043 has shown a suspicious behaviour. For the moment we won't mask it, but will keep an eye on it.

sipmRunDifferencePlots

pmtRunDifferencePlots


jmbenlloch (approved):
Latest calibration added and all tests passed. Approved!

PR #664, opened 2019-08-01
Increase string length for creator process, according to nexus.

paolafer:
This is an extremely simple PR, in which I change the string length to store the name of the creator process of an MC particle, following the change implemented in nexus. There are hadronic processes whose names are longer than 20 characters, which was the length previously set.

### DIFF
--- a/invisible_cities/evm/nh5.py
+++ b/invisible_cities/evm/nh5.py
@@ -76,16 +76,16 @@ class MCHitInfo(tb.IsDescription):
 class MCParticleInfo(tb.IsDescription):
     """Stores the simulated particles as metadata using Pytables.
     """
-    particle_indx  = tb.  Int16Col(    pos=0)
-    particle_name  = tb. StringCol(20, pos=1)
-    primary        = tb.  Int16Col(    pos=2)
-    mother_indx    = tb.  Int16Col(    pos=3)
-    initial_vertex = tb.Float32Col(    pos=4, shape=4)
-    final_vertex   = tb.Float32Col(    pos=5, shape=4)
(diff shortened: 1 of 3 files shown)


invisible_cities/evm/nh5.py
```diff
@@ -86,7 +86,7 @@ class MCParticleInfo(tb.IsDescription):
     final_volume   = tb. StringCol(20, pos=7)
     momentum       = tb.Float32Col(    pos=8, shape=3)
     kin_energy     = tb.Float32Col(    pos=9)
-    creator_proc   = tb. StringCol(20, pos=10)
+    creator_proc   = tb. StringCol(100, pos=10)
```
andLaing, line 89: Make it all lined up and pretty, we don't want to upset people

paolafer:
Mmm, right... The test that is failing reads a file with `creator_process` being a string of length 20 and writes one with length 100. The assertion that the contents are the same fails. Any idea of how to overcome this problem, without changing the input file?
An option would be to change the test (I know, very scary...) and test the equality of another table, for instance, `hits`, instead of `particles` (the choice of `particles` is arbitrary, I think that the purpose of this part of the test is to ensure that MC info is well written, and a table is chosen to this aim).

paolafer:
In the end, what I've done is to create a new test file just for this test, following @jmunozv's advice. The tests pass, I think that it is a really straightforward PR and can be approved and merged (maybe by @bpalmeiro?)

andLaing (approved):
Simple change that makes a more diverse set of data labels possible

paolafer:
Could it be merged? @bpalmeiro, for instance?

PR #667, opened 2019-09-18
Tools for PSF creation

Aretno:
Functions and tests for PSF creation starting from penthesilea Kr hits (1 hit = 1 sipm).


Aretno:
Closed for the moment as it's mixing beersheba commits with PSF commits (which should be independent). I probably fucked somewhere so I'll look into it and fix it.

PR #666, opened 2019-09-04
Add Beersheba - the deconvolution city

Aretno:
First version of the deconvolution liquid city, named Beersheba (as it could have any other name). 

Flow of the city:

- Read esmeralda hits 
- Cut them based on their absolute SiPM Q and redistribute the energy of those cutted out proportionally to the rest of the track.
- Cut SiPMs that are isolated from the rest of the signal, that means that they do not have any neighbour with detected light. Again, redistribute the energy of the cut out hits.
- Apply deconvolution according to the configuration. At the moment applies deconvolution to each slice, in 2D.
- Assign the total energy of the peak to the deconvolved track. The assignment is weighted by the charge density of each deconvolved hit. 

Deconvolution features:
 - Signal can be interpolated or not before applying deconvolution.
 - Either uncorrected or corrected energy can be assigned.
 - Can use a one-step deconvolution (all EL and diffusion effects
 integrated within one PSF) or two-steps (first apply only-EL PSF
 deconvolution, then a gaussian-PSF deconvolution based on the drift
 of the hits).
 - Stopping criteria based on whatever is achieved first:
 maximum number of iterations or relative change between iterations threshold.

Things missing: tests and finish the implementation/check of 3D deconvolution. Aside from that everything works just fine.

PSF creation, a must for this, I think it should be outside IC, similar to KrCalib (or even inside?). The code for the PSF is mostly ready with some minor detail that is bugging me. In any case, PSF creation requires an especific data processing so this should be addressed in terms of the production chain.

PR #668, opened 2019-09-18
Add tools for PSF creation.

Aretno:
Functions and tests for PSF creation starting from penthesilea Kr hits (1 hit = 1 sipm).

For the PSF, empty bins surrounding the charged SiPMs are added (as zero's are also relevant). The PSF is calculated using the relative position between SiPMs and barycenter. In the case of the Z dimension, the relative positions is between the slice position and the barycenter in the Z dimension.


invisible_cities/evm/nh5.py
```diff
@@ -247,3 +247,14 @@ class VoxelsTable(tb.IsDescription):
...
+
+class PSFfactors(tb.IsDescription):
+    xr     = tb.Float32Col(pos=0)
+    yr     = tb.Float32Col(pos=1)
+    zr     = tb.Float32Col(pos=2)
+    x      = tb.Float32Col(pos=3)
+    y      = tb.Float32Col(pos=4)
+    z      = tb.Float32Col(pos=5)
+    factor = tb.Float32Col(pos=6)
+    nevt   = tb. UInt32Col(pos=7)
```
gonzaponte: `pytables` recomends using smaller types first. If it doesn't disturb your code make `nevt` the first column

invisible_cities/io/kdst_io.py
```diff
@@ -57,3 +58,26 @@ def xy_lifetime_writer(hdf5_file, * ,
...
+                for k, z in enumerate(zr):
+                    row["xr"    ] = x
+                    row["yr"    ] = y
+                    row["zr"    ] = z
+                    row["x"     ] = xp
+                    row["y"     ] = yp
+                    row["z"     ] = zp
+                    row["factor"] = fr[i,j,k]
+                    row["nevt"  ] = nr[i,j,k]
+                    row.append()
```
gonzaponte, line 82: Mmm. this is clearly a homogeneous fixed-size data structure made of just a couple of arrays with 3 axes, which calls immediately for saving it as an `array` rather than a `table`, but on the other hand you would need to store `x`, `y` and `z` separately and combine them later somehow.
On the other hand, I am not sure there is a clear benefit of implementing that option for two reasons:
- You would need a custom reader
- The persistent and transient representations would be identical which, despite not being necessary, it's nice to have.

At the same time, this last point makes me think whether the transient representation is optimal. I would have to really dive into this code and into #666 to understand that which, of course, I can't do. So, it comes down to me asking: does the current `DataFrame` representation cause you any troubles? Any weird, untidy, messy pieces of code you are *forced* to do because of the data structure you've chosen?

invisible_cities/io/kdst_io_test.py
```diff
@@ -77,3 +78,31 @@ def test_xy_writer(config_tmpdir, corr_toy_data, writer):
...
     assert_allclose(N, dst.nevt       .values)
+
+
+def test_psf_writer(config_tmpdir):
+    output_file = os.path.join(config_tmpdir, "test_psf.h5")
+
+    xr, yr, zr = np.linspace(-50, 50, 101), np.linspace(-20, 20, 101), np.linspace(20, 30, 11)
+    x, y, z    = 3, 4, 5
+    factors    = np.linspace(0, 1, 101*101*11).reshape(101, 101, 11)
+    nevt       = np.linspace(0, 101*101*11 -1, 101*101*11).reshape(101, 101, 11)
```
gonzaponte: define the number of dimensions of the factors array as variables. This makes the code less prone to typing errors while simplifying the life of whoever has to deal with this piece of code in the future

gonzaponte: for `nevt` `np.arange` is probably more intuitive than `np.linspace`

invisible_cities/io/kdst_io_test.py
```diff
@@ -77,3 +78,31 @@ def test_xy_writer(config_tmpdir, corr_toy_data, writer):
...
+    xx, yy, zz = np.meshgrid(xr, yr, zr, indexing='ij')
+
+    assert_allclose(xx               .flatten() , psf.xr    .values)
+    assert_allclose(yy               .flatten() , psf.yr    .values)
+    assert_allclose(zz               .flatten() , psf.zr    .values)
+    assert_allclose(factors          .flatten() , psf.factor.values)
+    assert_allclose(nevt             .flatten() , psf.nevt  .values)
+    assert_allclose([x] * len(factors.flatten()), psf.x     .values)
+    assert_allclose([y] * len(factors.flatten()), psf.y     .values)
+    assert_allclose([z] * len(factors.flatten()), psf.z     .values)
```
gonzaponte: `[x]  * len(factors.flatten())` -> `np.full(factors.size, x)`

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,133 @@
...
+    ----------
+    psf         : Point-spread function.
+    entries     : Number of entries per bin in the PSF.:
+    bin_centers : Bin centers of the PSF.
+    """
+
+    entries, edges = np.histogramdd(pos, nbins, range=ranges, normed=False)
+    sumC   , edges = np.histogramdd(pos, nbins, range=ranges, normed=False, weights=charge)
+    with np.errstate(divide='ignore', invalid='ignore'):
+        psf = np.nan_to_num(sumC/entries)
```
gonzaponte:
```suggestion
        psf = np.nan_to_num(sumC / entries)
```

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,133 @@
...
+
+    return psf, entries, [shift_to_bin_centers(edge) for edge in edges]
```
gonzaponte: I think for someone reading this, it's better to have
```
centers = [shift_to_blabla...]
return pdf, entries, centers
```

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,133 @@
...
+        ----------
+        df        : groupby by event and npeak.
+        varMean   : variable to be averaged.
+        varWeight : variable to be uses as the weight.
+
+        Returns
+        ----------
+        df        : dst with the weighted average.
+        """
+        df[varMean + 'peak'] =  (df[varMean]*df[varWeight]).sum()/df[varWeight].sum()
```
gonzaponte:
```suggestion
        mean, weight = df.loc[:, (varMean, varWeight)].values.T
        df[varMean + 'peak'] =  np.average(mean, weights=weight)
```

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,133 @@
...
+        return(df)
```
gonzaponte: The function modifies the object, no need to return it

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,133 @@
...
+        Parameters
+        ----------
+        df  : groupby by event and npeak.
+        var : dimensions to be considered.
+
+        Returns
+        ----------
+        df  : dst with empty sipm hits.
+        """
+        distance = (np.diff(ranges)/2).flatten()
```
gonzaponte: This can be defined outside this function, it doesn't depend on any of it's arguments. On the other hand I would also define these two functions outside the main one so they can be tested independently

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,133 @@
...
+        means    = [int(df[f'{v}peak'].mean()) for v in var[:len(ranges)]]
+        means    = [mean - mean%sampleWidth[i]  for i, mean in enumerate(   means)]
+        varrange = [[means[i] - d, means[i] + d + sampleWidth[i]]                   for i,    d in enumerate(distance)]
+        allbins  = [int(np.diff(rang)/sampleWidth[i])              for i, rang in enumerate(varrange)]
+        Hs, edges   = np.histogramdd(tuple(df[v].values for v in var[:len(ranges)]), bins=allbins, range=varrange, normed=False, weights=df['Q'])
+        interPoints = np.meshgrid(*(shift_to_bin_centers(edge) for edge in edges), indexing='ij')
+        if len(ranges) < 3:
+            interPoints.append(np.array([df['Z'].min()] * len(interPoints[0].flatten())))
```
gonzaponte: This is a bit hard to read. I think I will need some explanation of what you do here

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,133 @@
...
+
+        pd_dict3 = {}
+        pd_dict3['Q'] = Hs.flatten()
+        pd_dict3['NormQ'] = Hs.flatten() / Hs.sum()
+        pd_dict3['E'] = pd_dict3['NormQ'] * df['E'].sum()
+        pd_dict3['nsipm'] = [len(df)] * len(Hs.flatten())
+        pd_dict4 = {k : [df[k].min()] * len(interPoints[0].flatten()) for k in ['event', 'time', 'npeak']}
+        pd_dict  = {**pd_dict1, **pd_dict2, **pd_dict3, **pd_dict4}
+
+        return pd.DataFrame(pd_dict).sort_values(['event', 'npeak', 'E'], ascending=[1, 1, 0]).reindex(list(df.columns).append('NormQ'), axis=1)
```
gonzaponte:
```suggestion
        df_out = pd.DataFrame(pd_dict)
        df_out.sort_values(['event', 'npeak', 'E'], ascending=[True, True, False], inplace=True)
        columns = list(df.columns) + ["NormQ"]
        return df_out.reindex(columns, axis=1)
```

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,133 @@
...
+
+    groupedDST = dsts.groupby(['event', 'npeak'], as_index=False)
+    if len(ranges) >= 3:
+        hdst          = groupedDST.apply(AddVariableWeightedMean, 'Z', 'E')
+        hdst          = hdst.groupby(['event', 'npeak'], as_index=False).apply(AddEmptySensors, ['X', 'Y', 'Z']).reset_index(drop=True)
+        hdst['RelZ']  = hdst.Z - hdst.Zpeak
+    else:
+        hdst          = groupedDST.apply(AddEmptySensors, ['X', 'Y', 'Z']).reset_index(drop=True)
+        #hdst          = dsts
+        #print(hdst)
```
gonzaponte: remove commented out code

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,133 @@
...
+        hdst['Zpeak'] = [hdst.Z.min()] * len(hdst)
+        hdst['RelZ']  = [0           ] * len(hdst)
```
gonzaponte:
```
hdst['Zpeak'] = np.full(len(hdst), hdst.Z.min())
hdst['RelZ'] = np.zeros(len(hdst))
```

Not really necessary, but I think it's useful to get people used to less-known `numpy` functions

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,133 @@
...
+
+    hdst._is_copy = False
```
gonzaponte: use `hdst.loc[:, column]` to modify the dataframe instead .

invisible_cities/reco/psf_functions_test.py
```diff
@@ -0,0 +1,34 @@
...
+def test_hdst_PSF_processing(ICDATADIR):
+    PATH_IN        = os.path.join(ICDATADIR, "exact_Kr_tracks_with_MC.h5")
+    PATH_TEST      = os.path.join(ICDATADIR, "exact_Kr_tracks_with_MC_psf.h5")
+
+    hdst           = load_dst(PATH_IN, 'RECO', 'Events')
+    hdst_psf       = pd.read_hdf(PATH_TEST)
+
+    hdst_processed = hdst_PSF_processing(hdst, [[-50, 50], [-50, 50]], [10, 10])
+
+    assert hdst_psf.equals(hdst_processed)
```
gonzaponte: I think `assert_dataframes_close` from `testing_utils` would be better

gonzaponte (commented):
As for the other PR, this is not a complete review, I'm just trying to provide an outsider's (pedantic) view of this.

Also for this PR, review the naming of variables. In this one, you mix `lower_case` with `camelCase` and `CamelCase`. The last one is particularly problematic because it clashes with the actual naming convention we use: `lower_case` for variables, `CamelCase` for classes/types.

Overall there is only one "major" (for a very loose definition of major) concern and one function I don't quite understand. The rest are minor details.

invisible_cities/reco/psf_functions.py
jacg: Chaining?

gonzaponte: You mean chaining the input df with the new column?

jacg: I'm not saying anything about this function specifically, rather generally addressing your comment

> The function modifies the object, no need to return it

In general, even if a function mutates its input, you may want to return it anyway, in order to allow chaining operations on it. Probably most useful (well, prettiest) when applied to member functions rather than free functions. Compare:

````python
weapon = make_weapon()
weapon.safety_off()
weapon.fire()
smoking_gun = weapon.safety_on()
````

with 

````python
smoking_gun = make_weapon().safety_off().fire().safety_on()
````

With free functions it would be

````python
weapon = make_weapon()
safety_off(weapon)
fire(weapon)
smoking_gun = safety_on(weapon)
````
and

````python
smoking_gun = safety_on(fire(safety_off(make_weapon())
````

(Can't let that last one go by without a nod to the Haskell spellings of this:
````haskell
smoking_gun = safety_on (fire (safety_off make_weapon)) -- direct translation (spaces not necessary, but conventional)

smoking_gun = (safety_on . fire . safety_off) make_weapon -- function composition (spaces optional ... and in the following ones too)

smoking_gun = safety_on . fire . safety_off $ make_weapon -- function composition + application operator

smoking_gun = safety_on $ fire $ safety_off make_weapon -- application operator (essentially: avoid writing brackets)

smoking_gun = make_weapon & safety_off & fire & safety_on -- reverse application operator: piping
````
)

Comment (gonzaponte) [reply invisible_cities/reco/psf_functions.py]: I agree it can be useful, but my intuition when a function returns something tells me that it is a new object. Since we are already building functions with side effects (which in this case i think is safe because the scope of the object is rather well defined), I would like it at least to be obvious which is not from this `smoking_gun` example. That of course doesn't happen in Haskell.

Actually i just realized that in the review I used the word *need*. I actually meant that I would rather not do so for the aforementioned reason.

Comment (jacg) [reply invisible_cities/reco/psf_functions.py]: Yup, it's a question of style and expectations. Your desire to stick to a certain style and associate certain expectations with it, is a perfectly valid choice with which I have a lot of sympathy. I'm merely pointing out that it's not the only way to do things, and that not everyone will share your expectations, at least not without documentation / education / agreement.

Continuing the Haskell asides: what is cool about Haskell is that it ensures that, in these sorts of cases, there can be no surprises; the type system tells you exactly what to expect, and ensures that the expectations will not by violated. In plain code it is guaranteed that nothing will be mutated, but by expressing your ideas in some monad of your choosing you switch to a different set of expectations and have them enforced by the language.

Comment (gonzaponte) [reply invisible_cities/reco/psf_functions.py]: >  a perfectly valid choice with which I have a lot of sympathy

Well, I hope, because it is your fault that I have that approach :laughing: 

> not everyone will share your expectations

Yes, that is a good point and I should defenitely keep it in mind :+1: 

> ensures that the expectations will not by violated.

Indeed. And I've found that more robust and easier to debug!

> but by expressing your ideas in some monad of your choosing you switch to a different set of expectations and have them enforced by the language.

I am still to understand monads... And I think I am still faraway from doing so.

Comment (jacg) [reply invisible_cities/reco/psf_functions.py]: > a good point and I should defenitely keep it in mind 

Good, as it's my central point: we have built a certain culture in IC, which implicitly brings certain expectations with it. We need to put at least a little effort into maintaiting the culture and expectations, for our design choices to continue to make sense to collaborators as time passes. So your review comments like the one from which I started this digression are important to make, and I'm prodding it because I think that it is important that we remind ourselves why we use this style, and also to raise awareness that it is a choice we have made, rather than being god-given.

> And I've found that more robust and easier to debug!

Should I infer from this that you've been doing some Haskell on the side recently?

> I am still to understand monads... And I think I am still faraway from doing so.

In many ways they are less difficult than the fuss around them suggests. I'd be happy to help, if you want, but this is not the place for it.

Comment (gonzaponte) [reply invisible_cities/reco/psf_functions.py]: >  it is important that we remind ourselves why we use this style, and also to raise awareness that it is a choice we have made, rather than being god-given.

Noted! thanks!

> Should I infer from this that you've been doing some Haskell on the side recently?

Sporadically. I had two periods of time in which I dedicated a few hours a week to learn.  The progress in the second attempt was way more successful than in the first, so my hope is that when I try again it will be even better.

> I'd be happy to help, if you want

Thanks a lot. I *will* definitely take the offer. Maybe this will trigger the third Haskell campaign!

**Comment by jjgomezcadenas** [comment]
Hereby I declare that if IC moves to Haskell it WON’T be my fault....

**Comment by jacg** [comment]
> Hereby I declare that if IC moves to Haskell it WON’T be my fault....

:-)

Haskell for IC (or HEP) would be a disaster. There are too many complexities, too much experimentation in the language (no serious work is done without using a number of language extensions which range from "why the hell isn't this in the standard" to "here be dragons"), the error messages are (all too often) impenetrable. And the build system situation is deep tragedy. You should not ask people whose primary interest is not programming, to do serious work in Haskell.

OTOH, within Haskell there are many things which would be absolutely marvellous for HEP. If you could get those without the complications and distractions, then it would be a huge boon. This has been done for front-end web development with Elm. Elm is a simplified Haskell derivative that compiles to JavaScript and has outstandingly clear and helpful error messages and a build story that won't drive you insane. It gives web developers many of the advantages of Haskell without any of the problems. A language like that, but with a different set of power-vs-simplicity tradeoffs, a set more appropriate for HEP, could do wonders.

I'm not aware of a viable candidate, so I'm afraid that (amongst many other disadvantages) we'll have to keep using Python's shitty type annotations instead of a robust, informative and expressive static type system built into the language; and continue to distinguish pure from impure code without the help of the type system and thus have to rely on @gonzaponte's review comments to remind us that 

+ the distinction exists
+ we are making it, by conventions which
     - most of us are probably not actively aware of
     - aren't reliable

(We'll also have to continue to compose our components (functions!) using a very clunky syntax. I tried to mitigate this by giving you dataflow, but it's still extremely sad compared to how it could look if we had Haskell syntax for this purpose, but that's another story.)

File: invisible_cities/io/kdst_io.py (hunk 2 above)
Comment (Aretno) [reply invisible_cities/io/kdst_io.py:82]: Mmm... not that I can think of. `x`, `y` and `z` are just used on the PSF selection
```
`            psf = psfs[(psfs.z == find_nearest(psfs.z,                 z)) &
                       (psfs.x == find_nearest(psfs.x, df.Xpeak.unique())) &
                       (psfs.y == find_nearest(psfs.y, df.Ypeak.unique())) ]`
```

and the dataframe is then just used to get the values and shape them accordingly:
`psf_deco      = psf.factor.values.reshape(tuple(psf[var].nunique() for var in varName[:len(data)]))`

The selection I feel it's optimal using dataframes  (at least user-friendly when reading the code), the second one may be more messy.

File: invisible_cities/reco/psf_functions.py (hunk 10 above)
Comment (Aretno) [reply invisible_cities/reco/psf_functions.py]: If I recall correctly:

Line 93-94: Get the mean position of the event and shift it always downwards (I've had problems to properly center everything if that changes). Knowing that the SiPM position is at  10*n + 5 mm I could have tried to be smart and move it upwards when the mean is above 10*n + 5 mm. However, since for future detectors, the pitch won't be the same, I felt inclined to just use a fixed decission which is to go downwards. But it's kind of arbitrary. I just want it to always be, relatively speaking, on the same place. In any case, it does not make a difference as long as the range, for adding empty SiPMs later, is large enough.

Line 95: an extra sampleWidth is added to one side, this is to compensate for the systematic downward choice made before. 

The rest is just plotting the SiPMs charge in a grid with bin centers equal to the sensors position, so I can later create a dataframe with the values of the histogram (filled for the sensors that have charge, 0 when the SiPM was not in the DST). This is as fast as I could think of for adding the 0s. I think I could have used pad but I would still have to put the charge SiPMs in a given grid and I think the histogram option would still be the easiest way to do so.

**Comment by Aretno** [comment]
Notebook explaining the contents of the PR: https://github.com/Aretno/Notebooks/blob/master/Tools%20for%20PSF%20creation%20auxiliar%20Notebook.ipynb

Also, I modified the code to make it a bit more compact but no big changes were made.

File: invisible_cities/io/kdst_io.py [hunk 16]
```diff
@@ -57,3 +58,26 @@ def xy_lifetime_writer(hdf5_file, * ,
                      description  = "XY-dependent lifetime values",
                      compression  = compression)
 
+def psf_writer(hdf5_file, **kwargs):
+    psf_table = make_table(hdf5_file,
+                           group       = "PSF",
+                           name        = "PSFs",
```
gonzaponte, line 64: Maybe a better name for either the group or the node? `/PSF/PSFs` seems a bit redundant.

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,141 @@
...
+               charge : np.ndarray,
+               nbins  : int,
+               ranges : List[List[float]]
+               ) -> Tuple[np.ndarray, np.ndarray, List[np.ndarray]] :
+    """
+    Computes the point-spread (PSF) function of a given dataset.
+
+    Parameters
+    ----------
+    pos    : Hits relative position. Only tested for 2D.
```
gonzaponte: If this works only for 2D you should add some protection in the function. Raise an exception if someone tries to use it in 3D.

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,141 @@
...
+
+    Returns
+    ----------
+    psf         : Point-spread function.
+    entries     : Number of entries per bin in the PSF.:
+    bin_centers : Bin centers of the PSF.
+    """
+
+    entries, edges = np.histogramdd(pos, nbins, range=ranges, normed=False)
+    sumC   , edges = np.histogramdd(pos, nbins, range=ranges, normed=False, weights=charge)
```
gonzaponte: Some things:
1. I guess the idea is to extend it to 3D in the future? If not why not use `np.histogram2d` instead of `np.histogramdd`?
2. You may want to generalize it to variable-sized bins. In that case, you may want to switch from `nbins, ranges` to simply `bin_edges`
3. `normed=False` is the default behavior. Is any reason to say it explicitely?

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,141 @@
...
+    with np.errstate(divide='ignore', invalid='ignore'):
+        psf = np.nan_to_num(sumC / entries)
+
+    centers = [shift_to_bin_centers(edge) for edge in edges]
+
+    return psf, entries, centers
```
gonzaponte, line 39: You may want to return `np.asarray(centers)`?

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,141 @@
...
+    grouped hits DST 'df' (grouped using groupby, by event id).
+
+    Parameters
+    ----------
+    df        : dataframe (groupby by event and npeak to do it peak by peak)
+    varMean   : variable to be averaged.
+    varWeight : variable to be uses as the weight.
+    """
+    mean, weight = df.loc[:, (varMean, varWeight)].values.T
+    df.loc[:, varMean + 'peak'] = np.average(mean, weights=weight)
```
gonzaponte: This hardcoded `"peak"` string makes me think that the function has a very specific use case, but its name indicates that it is a bit more generic than that. Particularly I would like the function to *just* compute the weighted average and the addition to happen outside. However, in that case I don't see a huge benefit of having that function at all.
If you opt for doing the column addition within the function, add a new argument that is the new column name so at least it's generic enough.

P.S. we may need to eventually create a `dataframe_functions` module so everybody uses the same tested functions. But that's another story.

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,141 @@
...
+        psf = np.nan_to_num(sumC / entries)
+
+    centers = [shift_to_bin_centers(edge) for edge in edges]
+
+    return psf, entries, centers
+
+
+def add_variable_weighted_mean(df        : pd.DataFrame,
+                               varMean   : str,
+                               varWeight : str
```
gonzaponte: `var_mean` and `var_weight`

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,141 @@
...
+                                      database    : pd.DataFrame
+                                      ) -> pd.DataFrame :
+    """
+    Adds empty sensors to the hDST
+
+    Parameters
+    ----------
+    df  : dataframe (groupby by event and npeak to do it peak by peak)
+    var : dimensions to be considered.
+
```
gonzaponte, line 76: Two arguments are not described

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,141 @@
...
+    Returns
+    ----------
+    df  : dst with empty sipm hits.
```
gonzaponte:
```suggestion
    df_out  : dst with empty sipm hits.
```

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,141 @@
...
+                                      var         : List[str],
+                                      ranges      : List[List[float]],
+                                      database    : pd.DataFrame
+                                      ) -> pd.DataFrame :
+    """
+    Adds empty sensors to the hDST
+
+    Parameters
+    ----------
+    df  : dataframe (groupby by event and npeak to do it peak by peak)
```
gonzaponte: The parenthesis is misleading. I understood that it is a `DataFrameGroupBy` object, but what you mean is that the dataframe contains only one peak of one event.

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,141 @@
...
+    variables.extend([f'{v}peak' for v in var])
+
+    for v in variables:
+        pd_dict[v] = np.full(len(sensors), df.loc[:,v].unique())
+    pd_dict['X'       ] = sensors.X
+    pd_dict['Y'       ] = sensors.Y
+    pd_dict['Z'       ] = np.full(len(sensors), df.loc[:,    'Z'].min())
+    for col in df.columns.values:
+        if col not in pd_dict.keys():
+            pd_dict[col]  = fill_dummy
```
gonzaponte: You don't need to create arrays yourself. If any of your keys is an array, all the other variables will be broadcasted to that shape. So the following should give you the same result:

```
pd_dict['X'       ] = sensors.X
pd_dict['Y'       ] = sensors.Y
pd_dict['Z'       ] = df.Z.min()
for v in variables:
    pd_dict[v] = df.loc[:,v].unique()[0]
for col in df.columns.values:
    if col not in pd_dict.keys():
        pd_dict[col]  = 0 # or whatever value you want to set
```

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,141 @@
...
+
+    df2 = pd.DataFrame(pd_dict)
+    df_out = df.merge(df2, on=list(df), how='outer')
+    df_out.drop_duplicates(subset=var, inplace=True, keep='first')
+    df_out['NormQ'] = df_out.Q/df_out.Q.sum()
+    df_out['nsipm'] = np.full(len(df_out), len(df))
```
gonzaponte:
```suggestion
    df_out['nsipm'] = len(df)
```

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,141 @@
...
+    pd_dict['Y'       ] = sensors.Y
+    pd_dict['Z'       ] = np.full(len(sensors), df.loc[:,    'Z'].min())
+    for col in df.columns.values:
+        if col not in pd_dict.keys():
+            pd_dict[col]  = fill_dummy
+
+    df2 = pd.DataFrame(pd_dict)
+    df_out = df.merge(df2, on=list(df), how='outer')
+    df_out.drop_duplicates(subset=var, inplace=True, keep='first')
+    df_out['NormQ'] = df_out.Q/df_out.Q.sum()
```
gonzaponte:
```suggestion
    df_out['NormQ'] = df_out.Q / df_out.Q.sum()
```

invisible_cities/reco/psf_functions.py
gonzaponte: Ok, I've just seen how it is used. Go for the second option, adding the new column name as a parameter.

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,141 @@
...
+    groupedDST = dsts.groupby(['event', 'npeak'], as_index=False)
+    sipm_db    = load_db.DataSiPM(detector_db, run_number)
+    if len(ranges) >= 3:
+        hdst          = groupedDST.apply(add_variable_weighted_mean, 'Z', 'E')
+        hdst          = hdst.groupby(['event', 'npeak'], as_index=False).apply(add_empty_sensors_and_normalize_q, ['X', 'Y', 'Z'], ranges, sipm_db).reset_index(drop=True)
+        hdst['RelZ']  = hdst.Z - hdst.Zpeak
+    else:
+        hdst          = groupedDST.apply(add_empty_sensors_and_normalize_q, ['X', 'Y'], ranges, sipm_db).reset_index(drop=True)
+        hdst['Zpeak'] = np.full (len(hdst), hdst.Z.min())
+        hdst['RelZ' ] = np.zeros(len(hdst))
```
gonzaponte: Because `add_variable_weighted_mean` modifies the object in place (it adds one column), `hdst` is the same object as `dsts` (needs to be checked, but it should be like this). This may be a bit misleading, but I am not sure of what can work as an alternative...

In any case try to break up those long lines so it is a bit more readable.

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    groupedDST = dsts.groupby(['event', 'npeak'], as_index=False)
+    sipm_db    = load_db.DataSiPM(detector_db, run_number)
+    if len(ranges) >= 3:
+        hdst          = groupedDST.apply(add_variable_weighted_mean, 'Z', 'E')
+        hdst          = hdst.groupby(['event', 'npeak'], as_index=False).apply(add_empty_sensors_and_normalize_q, ['X', 'Y', 'Z'], ranges, sipm_db).reset_index(drop=True)
+        hdst['RelZ']  = hdst.Z - hdst.Zpeak
+    else:
+        hdst          = groupedDST.apply(add_empty_sensors_and_normalize_q, ['X', 'Y'], ranges, sipm_db).reset_index(drop=True)
+        hdst['Zpeak'] = np.full (len(hdst), hdst.Z.min())
```
gonzaponte:
```suggestion
        hdst['Zpeak'] = hdst.Z.min()
```

invisible_cities/reco/psf_functions.py
gonzaponte:
```suggestion
        hdst['RelZ' ] = 0
```

invisible_cities/reco/psf_functions_test.py
```diff
@@ -0,0 +1,72 @@
...
+
+    add_variable_weighted_mean(hdst, 'X', 'E')
+    add_variable_weighted_mean(hdst, 'Y', 'E')
+
+    assert np.allclose(x_mean, hdst.Xpeak.unique())
+    assert np.allclose(y_mean, hdst.Ypeak.unique())
+
+    hdst_psf  = pd.read_hdf(PATH_TEST)
+
+    assert_dataframes_close(hdst_psf, hdst)
```
gonzaponte: Why do you read this file? I see that you do this in other tests as well

invisible_cities/reco/psf_functions_test.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def test_add_empty_sensors_and_normalize_q(ICDATADIR):
+    PATH_IN   = os.path.join(ICDATADIR, "exact_Kr_tracks_with_MC.h5")
+    PATH_TEST = os.path.join(ICDATADIR, "exact_Kr_tracks_with_MC_psf_empty_sensors.h5")
+
+    hdst           = load_dst(PATH_IN, 'RECO', 'Events')
+    group          = hdst.groupby(['event'], as_index=False)
+    sipm_db        = load_db.DataSiPM('new', 0)
+    hdst_processed = group.apply(add_empty_sensors_and_normalize_q, ['X', 'Y'], [[-50, 50], [-50, 50]], sipm_db).reset_index(drop=True)
```
gonzaponte: I think for tests it's important to specify what does each argument mean, so I thing you should define
```
coordinates/columns/whatever = ["X", "Y"]
ranges = [-50, 50], [-50, 50]
``` 
(or similar) and then pass these variables as arguments.
Also, try to break up this long line.

invisible_cities/reco/psf_functions_test.py
```diff
@@ -0,0 +1,72 @@
...
+    hdst_psf       = pd.read_hdf(PATH_TEST)
+
+    assert hdst_processed.NormQ.sum() == 1.0 * hdst.event.nunique()
```
gonzaponte: Maybe it is more explanatory to check that in each event the sum of the normalized charge is 1?

invisible_cities/reco/psf_functions_test.py
```diff
@@ -0,0 +1,72 @@
...
+    assert hdst_processed.Q.sum()     == hdst.Q.sum()
+    assert_dataframes_close(hdst_psf, hdst_processed)
+
+
+def test_hdst_psf_processing(ICDATADIR):
+    PATH_IN        = os.path.join(ICDATADIR, "exact_Kr_tracks_with_MC.h5")
+    PATH_TEST      = os.path.join(ICDATADIR, "exact_Kr_tracks_with_MC_psf.h5")
+
+    hdst           = load_dst(PATH_IN, 'RECO', 'Events')
+    hdst_processed = hdst_psf_processing(hdst, [[-50, 50], [-50, 50]], 'new', 0)
```
gonzaponte: same comment about naming arguments.  you can also pass them by keyword instead of defining new variables like
```
hdst_psf_processing(hdst,
                    ranges      = [[-50, 50], [-50, 50]],
                    detector_db = 'new'                 ,
                    run_number  = 0                     )
```

invisible_cities/reco/psf_functions_test.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def test_create_psf(ICDATADIR):
+    PATH_IN        = os.path.join(ICDATADIR, "exact_Kr_tracks_with_MC_psf.h5")
+    PATH_TEST      = os.path.join(ICDATADIR, "test_psf.npz")
+
+    hdst           = pd.read_hdf(PATH_IN)
+    psf            = np.load(PATH_TEST)
+
+    psf_val, entries, binss = create_psf((hdst.RelX, hdst.RelY), hdst.NormQ, [100, 100], [[-50, 50], [-50, 50]])
```
gonzaponte: same comment about naming arguments

gonzaponte (commented):
Second round, almost there. Check for unused variables and imports with `pyflakes`.

invisible_cities/reco/psf_functions.py
Aretno: 1.- Yeah, the idea is to eventually do it in 3D (and actually it runs, but since the behaviour at all steps is not fully understood yet, for the moment I'm just promoting it as 2D, but wouldn't want to break the 3d option)
2.-  Will do.
3.- No reason at all, probably when testing the code I tried with and without normalization.

invisible_cities/io/kdst_io.py
Aretno: Any suggestions? I don't really care much.

gonzaponte: I don't have much better ideas...

invisible_cities/reco/psf_functions.py
Aretno: Mmmm, is that the case? Since I'm using it with apply (called from a groupby object) I was doubtful about what to put here but I guess you're right. I'll modify it.

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,141 @@
...
+    mean_name  : Name of the column containing the average.
+    """
+    mean, weight = df.loc[:, (var_mean, var_weight)].values.T
+    df.loc[:, mean_name] = np.average(mean, weights=weight)
+
+
+def add_empty_sensors_and_normalize_q(df          : pd.DataFrame,
+                                      var         : List[str],
+                                      ranges      : List[List[float]],
+                                      database    : pd.DataFrame
```
gonzaponte:
```suggestion
def add_empty_sensors_and_normalize_q(df       : pd.DataFrame,
                                      var      : List[str],
                                      ranges   : List[List[float]],
                                      database : pd.DataFrame
```

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,141 @@
...
+    df_out.drop_duplicates(subset=var, inplace=True, keep='first')
+    df_out['NormQ'] = df_out.Q / df_out.Q.sum()
+    df_out['nsipm'] = len(df)
+
+    return df_out
+
+
+def hdst_psf_processing(dsts        : pd.DataFrame,
+                        ranges      : List[List[float]],
+                        database    : pd.DataFrame
```
gonzaponte:
```suggestion
def hdst_psf_processing(dsts     : pd.DataFrame,
                        ranges   : List[List[float]],
                        database : pd.DataFrame
```

invisible_cities/reco/psf_functions.py
```diff
@@ -0,0 +1,141 @@
...
+                        ) -> pd.DataFrame :
+    """
+    Adds the necessary info to a hits DST to create the PSF, namely the relative position and the normalized Q.
+
+    Parameters
+    ----------
+    dsts        : hits (1 SiPM per hit).
```
gonzaponte: Specify which type of df (event, peak, slice...)

invisible_cities/reco/psf_functions_test.py
```diff
@@ -0,0 +1,84 @@
...
+    group          = hdst.groupby(['event'], as_index=False)
+    hdst_processed = group.apply(add_empty_sensors_and_normalize_q    ,
+                                 var      = ['X', 'Y']                ,
+                                 ranges   = [[-50, 50], [-50, 50]]    ,
+                                 database = load_db.DataSiPM('new', 0))
+    hdst_processed.reset_index(inplace=True, drop=True)
+
+    assert np.allclose(hdst_processed.groupby('event').NormQ.sum().values, 1.0)
+    assert hdst_processed.E.sum()     == hdst.E.sum()
+    assert hdst_processed.Q.sum()     == hdst.Q.sum()
```
gonzaponte:
```suggestion
    assert np. isclose(hdst_processed.E.sum(), hdst.E.sum())
    assert np. isclose(hdst_processed.Q.sum(), hdst.Q.sum())
```

invisible_cities/reco/psf_functions_test.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def test_add_empty_sensors_and_normalize_q_file(ICDATADIR):
+    PATH_IN        = os.path.join(ICDATADIR, "exact_Kr_tracks_with_MC.h5")
+    PATH_TEST      = os.path.join(ICDATADIR, "exact_Kr_tracks_with_MC_psf_empty_sensors.h5")
+    hdst           = load_dst(PATH_IN, 'RECO', 'Events')
+    group          = hdst.groupby(['event'], as_index=False)
```
gonzaponte, line 51:
```suggestion
    group          = hdst.groupby('event', as_index=False)
```

invisible_cities/reco/psf_functions_test.py
```diff
@@ -0,0 +1,84 @@
...
+
+    hdst           = pd.read_hdf(PATH_IN)
+    psf            = np.load(PATH_TEST)
+
+    bin_edges = [np.linspace(-50, 50, 101) for i in range(2)]
+    psf_val, entries, binss = create_psf((hdst.RelX, hdst.RelY), hdst.NormQ, bin_edges)
+
+    np.testing.assert_allclose(psf['psf'    ], psf_val)
+    np.testing.assert_allclose(psf['entries'], entries)
+    np.testing.assert_allclose(psf['bins'   ],   binss)
```
gonzaponte:
```suggestion
    assert np.allclose(psf['psf'    ], psf_val)
    assert np.allclose(psf['entries'], entries)
    assert np.allclose(psf['bins'   ],   binss)
```

It's not wrong. It's just to have a more uniform style

gonzaponte: Actually I've just noticed that other files use `assert_allclose` as well. At least this keeps the style within the file. We should establish a style at some point.

invisible_cities/io/kdst_io.py
gonzaponte: It seems to be irrelevant as the file contains only this info and it is loaded through `pd.read_hdf`

gonzaponte (commented):
I think after these cosmetical changes we are good to go.

gonzaponte (approved):
This PR adds the relevant functions for PSFs, an input needed for the future city Beersheba (#666). The code adds the functions to create the PSF and the functions for IO operations. All of that is nicely implemented, well tested and with some nice protections against bad inputs.

The author has also provided a notebook with a demonstration of the code:
https://github.com/Aretno/Notebooks/blob/master/Tools%20for%20PSF%20creation%20auxiliar%20Notebook.ipynb

Great work!

PR #670, opened 2019-10-16
Upload database

carmenromo:
NEW Database has been uploaded with the gains for SiPMs and PMTs as well as the pdfs based on the calibrations from run 7572. We also have seen that SiPM 21051 is dead since run 7563, so we have masked it.

sipmRunDifferencePlots

pmtRunDifferencePlots


jmbenlloch (approved):
Latest calibration correctly included and all tests passed. Approved!

PR #674, opened 2019-11-08
Add Ep attribute to hit type

mmkekic:
This PR modifies Hit class and corresponding paolina functions

### DIFF
--- a/invisible_cities/cities/esmeralda_test.py
+++ b/invisible_cities/cities/esmeralda_test.py
@@ -13,7 +13,6 @@ from .. io                     import dst_io      as dio
 from .  esmeralda              import esmeralda
 from .. core.testing_utils     import assert_dataframes_close
 from .. core.testing_utils     import assert_tables_equality
-from .. reco.dst_functions     import load_dst
 
 
 def test_esmeralda_contains_all_tables(KrMC_hdst_filename, correction_map_MC_filename, config_tmpdir):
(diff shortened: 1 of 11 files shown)


invisible_cities/evm/event_model.py
```diff
@@ -375,6 +377,7 @@ def store(self, table):
             row["Qc"      ] = hit .Qc
             row["Ec"      ] = hit .Ec
             row["track_id"] = hit .track_id
+            row["Ep"      ] = hit .Ep
```
mmkekic: maybe change positions of track_id and Ep?

gonzaponte: ok

invisible_cities/cities/esmeralda_test.py
```diff
@@ -308,4 +308,4 @@ def test_esmeralda_exact_result_all_events(ICDATADIR, KrMC_hdst_filename, correc
             for table in tables:
                 got      = getattr(     output_file.root, table)
                 expected = getattr(true_output_file.root, table)
-                assert_tables_equality(got, expected)
+                assert_tables_equality(got, expected, 1 if table in ("RECO/Events", "PAOLINA/Events") else 0)
```
mmkekic: Rather than changing the test file here maybe we can temporarily use dataframes comparison excluding the Ep column since #675 will be changing the test file again ?

gonzaponte: Can we do that for #675 only? Or do you think that dirty fix is too unreadable?

mmkekic: I think the fix would be more readable if you use dataframe comparison for both esmeralda and penthesilea test, rather than changing testing_utils
but beside that I am nor sure its worth it to change esmeralda files since they will be changed again in #675. Though since it is already in git lfs I guess it doesnt matter

mmkekic:
This PR adds Ep attribute to the Hit type, which will be used in all tracking functions in esmeralda. It changes corresponding test files keeping commits that show that the only differences are in the added Ep column. Good job!

gonzaponte (approved):
This PR was open by @mmkekic, but assigned to @gonzaponte. Since @mmkekic cannot officially approve it, I will do so on her name, but she was the reviewer.

PR #675, opened 2019-11-08
Modify hit table output of Esmeralda

mmkekic:
This PR modifies Esmeralda to output complete list of hits with both corrected energy and paolina energy (redistributed energy of dropped voxels)


mmkekic:
- [x] create_extract_track_blob_info:
     - [x]  Set Ep to Ec value before voxelizing
     - [x]  Fix energy_type to Ep in all paolina functions  calls
     - [x]  Change Ep to nan in hits from dropped voxels
     - [x]  Add hits from dropped voxels to track_hitc hit collection
- [x] make_event_summary:
     - [x] Calculate all variables using full list of hits
     - [x] Rename columns in event summary

- [x]  Rename and reorganize output tables

invisible_cities/cities/esmeralda.py
```diff
@@ -46,6 +46,27 @@
 from .. io. event_filter_io  import  event_filter_writer
 from .. io.          dst_io  import _store_pandas_as_tables
 
+types_dict_summary = {'event'     : np.int32  , 'evt_energy' : np.float64, 'evt_charge'    : np.float64,
+                      'evt_ntrks' : int       , 'evt_nhits'  : int       , 'evt_x_avg'     : np.float64,
```
gonzaponte: not to be picky, but, `np.int`? :yum:

invisible_cities/cities/esmeralda.py
```diff
@@ -107,6 +128,18 @@ def threshold_and_correct_hits(hitc : evm.HitCollection) -> evm.HitCollection:
         return new_hitc
     return threshold_and_correct_hits
 
+def copy_Ec_to_Ep_hit_attribute_(hitc : evm.HitCollection) -> evm.HitCollection:
+    """
+    The functions copies values of Ec attributes into Ep attributes. Takes as input HitCollection and returns a copy.
+
```
gonzaponte: empty line

invisible_cities/cities/esmeralda.py
```diff
@@ -107,6 +128,18 @@ def threshold_and_correct_hits(hitc : evm.HitCollection) -> evm.HitCollection:
...
+    """
+    mod_hits = []
+    for hit in hitc.hits:
+        hit = evm.Hit(hit.npeak, evm.Cluster(hit.Q, xy(hit.X, hit.Y), hit.var, hit.nsipm),
+                      hit.Z, hit.E, xy(hit.Xpeak, hit.Ypeak), s2_energy_c=hit.Ec, Ep =hit.Ec)
```
gonzaponte: Do you think it is worth to define a `copy_hit` function somewhere where you do this operation? I did it for something I already forgot, but I think it could be a good idea.

gonzaponte: In any case, since you are copying, it would be good to copy all attributes (at least `track_id` is missing)

invisible_cities/cities/esmeralda.py
```diff
@@ -140,11 +173,11 @@ def track_blob_info_creator_extractor(vox_size         : [float, float, float],
     def create_extract_track_blob_info(hitc):
         #track_hits is a new Hitcollection object that contains hits belonging to tracks, and hits that couldnt be corrected
         track_hitc = evm.HitCollection(hitc.event, hitc.time)
-        out_of_map = any(np.isnan([h.Ec for h in hitc.hits]))
+        out_of_map = any(np.isnan([h.Ep for h in hitc.hits]))
```
gonzaponte:
```suggestion
        out_of_map = np.any(np.isnan([h.Ep for h in hitc.hits]))
```

invisible_cities/cities/esmeralda.py
```diff
@@ -46,6 +46,27 @@
...
+                     'y_max'           : np.float64, 'z_max'         : np.float64, 'r_max'       : np.float64,
+                     'x_ave'           : np.float64, 'y_ave'         : np.float64, 'z_ave'       : np.float64,
+                     'r_ave'           : np.float64, 'extreme1_x'    : np.float64, 'extreme1_y'  : np.float64,
+                     'extreme1_z'      : np.float64, 'extreme2_x'    : np.float64, 'extreme2_y'  : np.float64,
+                     'extreme2_z'      : np.float64, 'blob1_x'       : np.float64, 'blob1_y'     : np.float64,
+                     'blob1_z'         : np.float64, 'blob2_x'       : np.float64, 'blob2_y'     : np.float64,
+                     'blob2_z'         : np.float64, 'eblob1'        : np.float64, 'eblob2'      : np.float64,
+                     'vox_size_x'      : np.float64, 'vox_size_y'    : np.float64, 'vox_size_z'  : np.float64,
+                     'ovlp_blob_energy': np.float64}
+
```
gonzaponte: Now that this is defined,  you can use it for the definition of the empty dataframes used in the `create_extract_track_blob_info` and `make_event_summary`

invisible_cities/cities/esmeralda.py
```diff
@@ -251,48 +280,39 @@ def make_event_summary(event_number  : int              ,
...
     #change dtype of columns to match type of variables
-    types_dict = dict(zip(es.columns, [type(x) for x in list_of_vars]))
-    es = es.apply(lambda x : x.astype(types_dict[x.name]))
-
+    es = es.apply(lambda x : x.astype(types_dict_summary[x.name]))
     return es
 
 
-def track_writer(h5out, compression='ZLIB4', group_name='PAOLINA', table_name='Tracks', descriptive_string='Track information', str_col_length=32):
+def track_writer(h5out, compression='ZLIB4', group_name='Tracking', table_name='Tracks', descriptive_string='Track information', str_col_length=32):
```
gonzaponte, line 301: Can we have problems with backwards-compatibility?

invisible_cities/cities/esmeralda_test.py
```diff
@@ -166,71 +166,8 @@ def test_esmeralda_tracks_exact(data_hdst, esmeralda_tracks, correction_map_file
...
-    tables_in = ( "MC/extents"  , "MC/hits"       , "MC/particles"  , "MC/generators",
-                  "DST/Events"  , "Run/events"    , "Run/runInfo"                    )
-    with tb.open_file(file_in)  as true_output_file:
-        with tb.open_file(file_out) as  output_file:
-            for table in tables_in:
-                got      = getattr(     output_file.root, table)
-                expected = getattr(true_output_file.root, table)
-                assert_tables_equality(got, expected)
+    df_summary  = dio.load_dst(PATH_OUT, 'Summary', 'Events')
+    assert all(df_summary[df_summary.event.isin(diff_events)].loc[:,'evt_out_of_map'])
```
gonzaponte:
```suggestion
    assert np.all(df_summary.loc[df_summary.event.isin(diff_events), 'evt_out_of_map'])
```

invisible_cities/cities/esmeralda_test.py
```diff
@@ -298,14 +235,108 @@ def test_esmeralda_exact_result_all_events(ICDATADIR, KrMC_hdst_filename, correc
...
+    df_summary_new  = dio.load_dst(file_out, 'Summary', 'Events' )
+
+    #new_summary will have evt_ before column names, so lets rename it
+    df_summary_new.columns = df_summary_old.columns
+    assert_dataframes_close(df_summary_old, df_summary_new, check_types=False)
+
+    #compare Hits table without newly added Ep
+    df_hits_paolina_new = dio.load_dst(file_out, 'CHITS'  , 'highTh')
+    df_hits_paolina_old = dio.load_dst(true_out, 'PAOLINA', 'Events')
+    columns = df_hits_paolina_old.columns.drop('Ep')
```
gonzaponte: Isn't it correct to compare `Ep` to `Ec` in this case? If so, it would be better than dropping the column

invisible_cities/cities/esmeralda_test.py
```diff
@@ -298,14 +235,108 @@ def test_esmeralda_exact_result_all_events(ICDATADIR, KrMC_hdst_filename, correc
...
+    assert_dataframes_close(df_hits_paolina_new[columns], df_hits_paolina_new[columns])
+
+    #Ep should be set to Ec since voxel dropping threshold is 0
+    np.testing.assert_array_equal(df_hits_paolina_new.Ec, df_hits_paolina_new.Ep)
+
+
+#test showing that all events that pass charge threshold are contained in hits output
+def test_esmeralda_bug_duplicate_hits(data_hdst, esmeralda_tracks, correction_map_filename, config_tmpdir):
```
gonzaponte: `esmeralda_tracks` not used

invisible_cities/cities/esmeralda_test.py
```diff
@@ -298,14 +235,108 @@ def test_esmeralda_exact_result_all_events(ICDATADIR, KrMC_hdst_filename, correc
...
+                         energy_threshold         = 20 * units.keV         ,
+                         min_voxels               = 2                      ,
+                         blob_radius              = 21 * units.mm)        ))
+    cnt = esmeralda(**conf)
+
+    df_tracks           =  dio.load_dst(PATH_OUT, 'Tracking', 'Tracks')
+    df_phits            =  dio.load_dst(PATH_OUT,   'CHITS' , 'highTh')
+    df_in_hits          =  dio.load_dst(PATH_IN ,    'RECO' , 'Events')
+
+    num_pass_th_in_hits = sum(df_in_hits.Q >= th_p)
```
gonzaponte, line 310:
```suggestion
    num_pass_th_in_hits = np.nansum(df_in_hits.Q >= th_p)
```
I suggest this change to avoid a warning by numpy. I don't know if the hit charges can be nan

invisible_cities/cities/esmeralda_test.py
```diff
@@ -298,14 +235,108 @@ def test_esmeralda_exact_result_all_events(ICDATADIR, KrMC_hdst_filename, correc
...
+    num_pass_th_p_hits  = len(df_phits)
+    assert num_pass_th_in_hits == num_pass_th_p_hits
+
+
+    num_paolina_hits   = sum(df_phits.Ep>0)
```
gonzaponte:
```suggestion
    num_paolina_hits   = np.nansum(df_phits.Ep > 0)
```
why `> 0`?

gonzaponte (commented):
Minor corrections and some questions. Looks good!

invisible_cities/cities/esmeralda.py
mmkekic: I dont think we ever copy Ec to Ep explicitly, and I am not sure copying track_id is necessary, is also somehow misleading, a reader might believe that track_id already has a meaningful value at this point of the code which is not true

invisible_cities/cities/esmeralda.py
mmkekic: ok

invisible_cities/cities/esmeralda.py
mmkekic: ok

invisible_cities/cities/esmeralda.py
mmkekic: oook

invisible_cities/cities/esmeralda.py
mmkekic: yeah, but there is no way to avoid that. the data will have to be reprocessed and the users will have to adapt their scripts to the new output...

invisible_cities/cities/esmeralda_test.py
mmkekic: ok

invisible_cities/cities/esmeralda_test.py
mmkekic: I am dropping 'Ep' column from the column index of dataframe that doesnt even have that column... so ignore it

invisible_cities/cities/esmeralda_test.py
mmkekic: if they are nan something went wrong so maybe is better to keep the sum and make the code complain?

gonzaponte: Indeed

invisible_cities/cities/esmeralda_test.py
mmkekic: fixed

mmkekic:
@gonzaponte  I think I addressed all your comments. When you are happy with the changes I would replace the exact test file in esmeralda and remove the old test.

jahernando:
There is a name replacement of RECO and PAULINA hits for low-threshold and high-threshold, but the high-threshold, corrected me if I am wrong, contains some specific information of the Paulina algorithm.  I think the previous naming was more indicative for the client. The LowThreshold was in fact the 'default' definition of reconstructed hits. 

There is some work to revert the naming.  As most-likely we will move to a separation of pipes in Esmeralda, we can continue with the proposed naming and reconsidered later.

mmkekic:
I agree that the naming is not trivial, high threshold hits should exist unrelated to paolina but now it does contain some information from paolina functions. I am not sure the best way to store the information. 
Some clients only care about corrected hits that pass the high threshold, unrelated to tracks. And indeed, the highTh hit table contains all the hits with charge above given threshold, where 'paolina' hits will be only the ones with track_id>=0 (and, equally, Ep different than nan). 
Maybe @paolafer  has something to say?

invisible_cities/cities/esmeralda_test.py
```diff
@@ -138,99 +137,35 @@ def test_esmeralda_tracks_exact(data_hdst, esmeralda_tracks, correction_map_file
...
-    tables_in = ( "MC/extents"  , "MC/hits"       , "MC/particles"  , "MC/generators",
-                  "DST/Events"  , "Run/events"    , "Run/runInfo"                    )
-    with tb.open_file(file_in)  as true_output_file:
-        with tb.open_file(file_out) as  output_file:
-            for table in tables_in:
-                got      = getattr(     output_file.root, table)
-                expected = getattr(true_output_file.root, table)
-                assert_tables_equality(got, expected)
+    df_summary  = dio.load_dst(PATH_OUT, 'Summary', 'Events')
+    assert np.all(df_summary[df_summary.event.isin(diff_events)].loc[:,'evt_out_of_map'])
```
gonzaponte:
```suggestion
    assert np.all(df_summary.loc[df_summary.event.isin(diff_events), 'evt_out_of_map'])
```

invisible_cities/cities/esmeralda_test.py
gonzaponte: don't ignore me :stuck_out_tongue:

invisible_cities/cities/esmeralda_test.py
```diff
@@ -280,32 +215,125 @@ def test_esmeralda_exact_result_all_events(ICDATADIR, KrMC_hdst_filename, correc
...
+    df_phits            =  dio.load_dst(PATH_OUT,   'CHITS' , 'highTh')
+    df_in_hits          =  dio.load_dst(PATH_IN ,    'RECO' , 'Events')
+
+    num_pass_th_in_hits = sum(df_in_hits.Q >= th_p)
+    num_pass_th_p_hits  = len(df_phits)
+    assert num_pass_th_in_hits == num_pass_th_p_hits
+
+
+    num_paolina_hits   = sum(np.isfinite(df_phits.Ep))
+    assert num_paolina_hits <= num_pass_th_p_hits
```
gonzaponte, line 316: maybe add a comment explaining why this is the case

gonzaponte (commented):
This is essentially ready, there are some cosmetics left to do, but nothing too important. Apply, rebase and will approve.

invisible_cities/cities/esmeralda.py
gonzaponte: I agree, but I still think it is better to have something more general implemented. Anyway this is not for this PR.

gonzaponte (approved):
This PR fixes Paolina energy to `Ep` in esmeralda and uses all hits to compute statistics. The output includes all corrected hits above threshold. Besides the city has been refactored a bit leading to a more readable scheme. Finally, there has been a renaming of some nodes in the hdf5 output and the config file. Good effort!

PR #676, opened 2019-11-08
DataFrame based readers for MC data

andLaing:
New readers based on the FANAL readers
which speed up input for nexus saved
hdf5 data.

For a small file with 3 events, 395 hits, 212 particles:
load_mchits takes ~23.2 ms, new function: ~14 ms
load_mcparticles takes ~78 ms, new function ~25 ms

A larger file with 1000 evt, ~12000 hits:
load_mchits takes ~700 ms, new function ~25 ms

### DIFF
--- a/invisible_cities/io/mcinfo_io.py
+++ b/invisible_cities/io/mcinfo_io.py
@@ -342,11 +342,12 @@ def load_mcsensor_response_df(file_name : str,
                     how         = 'left')
     sns.evt_number.fillna(method='bfill', inplace = True)
 
-    sns['times'] = sns[sns.sensor_id.isin(pmt_ids)].time_bin * pmt_bin
-    sns.times.fillna(sns.time_bin * sipm_bin, inplace = True)
+    sns['time'] = sns[sns.sensor_id.isin(pmt_ids)].time_bin * pmt_bin
+    sns.time.fillna(sns.time_bin * sipm_bin, inplace = True)
 
     sns.evt_number = sns.evt_number.astype(int)
-    sns.set_index(['evt_number', 'sensor_id', 'time_bin'], inplace = True)
+    sns.rename(columns = {'evt_number': 'event_id'}, inplace = True)
+    sns.set_index(['event_id', 'sensor_id', 'time_bin'], inplace = True)
 
     return extents.evt_number.unique(), pmt_bin, sipm_bin, sns
 
--- a/invisible_cities/io/mcinfo_io.py
+++ b/invisible_cities/io/mcinfo_io.py
@@ -1,6 +1,7 @@
 
 import tables as tb
 import numpy  as np
+import pandas as pd
 
 from .. reco            import tbl_functions as tbl
 from .. core            import system_of_units as units
@@ -15,8 +16,11 @@ from .. evm.nh5 import MCExtentInfo
 from .. evm.nh5 import MCHitInfo
 from .. evm.nh5 import MCParticleInfo
 
+from .. database import load_db as DB
+
 from typing import Mapping
 from typing import Sequence
+from typing import Tuple
 
 # use Mapping (duck type) rather than dict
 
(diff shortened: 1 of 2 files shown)


invisible_cities/io/mcinfo_io.py
```diff
@@ -173,20 +177,180 @@ def load_mchits(file_name: str,
...
+
+    # Generating hits DataFrame
+    hits = pd.DataFrame({'hit_id'      : hits_tb.col('hit_indx'),
+                         'particle_id' : hits_tb.col('particle_indx'),
+                         'label'       : hits_tb.col('label').astype('U13'),
+                         'time'        : hits_tb.col('hit_time'),
+                         'x'           : hits_tb.col('hit_position')[:, 0],
+                         'y'           : hits_tb.col('hit_position')[:, 1],
+                         'z'           : hits_tb.col('hit_position')[:, 2],
+                         'E'           : hits_tb.col('hit_energy')})
```
paolafer: As talked earlier with Javi, let's use the same name for columns as the ones that will be implemented in nexus. Therefore, please, change `E` with `energy`. Also, the order of the columns in the table should is the following in nexus: `event_id`, `x`, `y`, `z`, `time`, `energy`, `label`, `particle_id`, `hit_id`. The order is not that relevant, though.

invisible_cities/io/mcinfo_io.py
```diff
@@ -173,20 +177,180 @@ def load_mchits(file_name: str,
...
+               Name of the file to be read
+    """
+    # Loading data extents table: event structure
+    extents = pd.read_hdf(iFileName, 'MC/extents')
+
+    with tb.open_file(iFileName ,mode='r') as iFile:
+        p_tb = iFile.root.MC.particles
+
+        # Generating parts DataFrame
+        parts = pd.DataFrame({'pid'     : p_tb.col('particle_indx'),
```
paolafer: As above: the name of the columns should be the following: `event_id`, `particle_id`, `name`, `primary`, `mother_id`, `initial_x`, `initial_y`, `initial_z`, `initial_t`, `final_x`, `final_y`,  `final_z`,  `final_t`, `initial_volume`, `final_volume`, `initial_momentum_x`, `initial_momentum_y`, `initial_momentum_z`, `kin_energy`, `creator_proc`.

invisible_cities/io/mcinfo_io.py
```diff
@@ -173,20 +177,180 @@ def load_mchits(file_name: str,
...
+                    how         = 'left')
+    sns.evt_number.fillna(method='bfill', inplace = True)
+
+    sns['times'] = sns[sns.sensor_id.isin(pmt_ids)].time_bin * pmt_bin
+    sns.times.fillna(sns.time_bin * sipm_bin, inplace = True)
+
+    sns.evt_number = sns.evt_number.astype(int)
+    sns.set_index(['evt_number', 'sensor_id', 'time_bin'], inplace = True)
+
+    return extents.evt_number.unique(), pmt_bin, sipm_bin, sns
```
paolafer, line 362: Why does this function return also the list of event numbers and the other two don't?

invisible_cities/io/mcinfo_io.py
```diff
@@ -173,20 +177,180 @@ def load_mchits(file_name: str,
...
+                    left_index  =   True,
+                    right_index =   True,
+                    how         = 'left')
+    sns.evt_number.fillna(method='bfill', inplace = True)
+
+    sns['times'] = sns[sns.sensor_id.isin(pmt_ids)].time_bin * pmt_bin
+    sns.times.fillna(sns.time_bin * sipm_bin, inplace = True)
+
+    sns.evt_number = sns.evt_number.astype(int)
+    sns.set_index(['evt_number', 'sensor_id', 'time_bin'], inplace = True)
```
paolafer: Here I would also use `event_id`, as in the other tables.

invisible_cities/io/mcinfo_io.py
```diff
@@ -173,20 +177,180 @@ def load_mchits(file_name: str,
...
+
+
 def load_mcparticles(file_name: str,
                      event_range=(0, int(1e9))) -> Mapping[int, Mapping[int, MCParticle]]:
 
     with tb.open_file(file_name, mode='r') as h5in:
         return read_mcinfo(h5in, event_range)
 
 
+def load_mcparticles_df(iFileName: str) -> pd.DataFrame:
```
paolafer: Why don't you use two different functions as you do with `hits`? One that takes as an argument the file name and the other one that takes as an argument the result of opening the table. I would expect that if they're useful for one table, they could be for the other one, too.

invisible_cities/io/mcinfo_io.py
andLaing: More a usage case than anything. The current function here is used in the new detsim where it was useful to have this information. The others are only used in FANAL so far where the extra info wasn't necessary

invisible_cities/io/mcinfo_io.py
andLaing: Sorted in the most recent push

invisible_cities/io/mcinfo_io.py
andLaing: done

invisible_cities/io/mcinfo_io.py
andLaing: done

invisible_cities/io/mcinfo_io.py
jmunozv: As physicist, I like symmetries ... although a user case, it is a user case.

jmunozv (commented):
I'm not sure about naming convention inside IC.
I mean, I don't know if load_mchits_df(...) fulfills them or not; anf if any of the following options would be better:
load_mc_hits_df
load_mcHits_df   (the one I like most)
load_MCHits_df   (the one closer to IC convention ??)

jjgomezcadenas:
small_function()   Is OK

Enviado desde mi iPhone

invisible_cities/io/mcinfo_io.py
andLaing: I added a function of this type for load_mcparticles_df but not for the sensors version.

Since the sensor info is read directly using pandas.read_hdf a tables.File object is never created so it seemed redundant to have another version

andLaing:
I changed a few of the names to make them a bit more self consistent.

load_mchits_fromstr(str) -> load_mchits_df
load_mchits_df(tb.File, pd.DataFrame) -> read_mchits_df

jmunozv (approved):
This MR extends the possibilities of reading nexus files ala FANAL, generating DataFrames and by-passing the GATE event data model.
It's much more faster and less memory consuming than GATE versions.
It passes all the tests suite.
I approve it.

PR #678, opened 2019-11-11
Return dropped voxels

gonzaponte:
This PR will modify the `drop_end_point_voxels` to return both the modified voxels and the dropped ones so they can be given to `esmeralda` and write them to the output file.

### DIFF
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -307,15 +307,6 @@ def drop_end_point_voxels(voxels: Sequence[Voxel], energy_threshold: float, min_
         """Eliminate an individual voxel from a set of voxels and give its energy to the hit
            that is closest to the barycenter of the eliminated voxel hits, provided that it
            belongs to a neighbour voxel."""
-
-        ### be sure that the voxel to be eliminated has at least one neighbour
-        ### beyond itself
-        the_neighbours = (neighbours(the_vox, v) for v in voxels)
-        if sum(the_neighbours) <= 1:
-            return 0
-
-        ### remove voxel from list of voxels
-        voxels.remove(the_vox)
         the_neighbour_voxels = [v for v in voxels if neighbours(the_vox, v)]
 
         pos = [h.pos              for h in the_vox.hits]
@@ -330,13 +321,11 @@ def drop_end_point_voxels(voxels: Sequence[Voxel], energy_threshold: float, min_
             sum_en_v = sum(v.E for v in min_v)
             for v in min_v:
                 v.E += the_vox.E/sum_en_v * v.E
-
-            return 1
+            return
 
         bary_pos = np.average(pos, weights=qs, axis=0)
 
         ### find hit with minimum distance, only among neighbours
-
         min_dist = min(np.linalg.norm(bary_pos-hh.pos) for v in the_neighbour_voxels for hh in v.hits)
         min_h_v  = [(h, v) for v in the_neighbour_voxels for h in v.hits if np.isclose(np.linalg.norm(bary_pos-h.pos), min_dist)]
         min_hs   = set(h for (h,v) in min_h_v)
@@ -350,30 +339,32 @@ def drop_end_point_voxels(voxels: Sequence[Voxel], energy_threshold: float, min_
         for v in min_vs:
             v.E += the_vox.E/sum_en_v * v.E
 
-        return 1
 
-    mod_voxels = copy.deepcopy(voxels)
+    mod_voxels     = copy.deepcopy(voxels)
+    dropped_voxels = []
 
-    while True:
-        n_modified_voxels = 0
+    modified = True
+    while modified:
+        modified = False
         trks = make_track_graphs(mod_voxels)
 
         for t in trks:
             if len(t.nodes()) < min_vxls:
                 continue
 
-            extr1, extr2 = find_extrema(t)
-            if extr1.E < energy_threshold:
-                n_mod = drop_voxel(mod_voxels, extr1)
-                n_modified_voxels += n_mod
-            if extr2.E < energy_threshold:
-                n_mod = drop_voxel(mod_voxels, extr2)
-                n_modified_voxels += n_mod
+            for extreme in find_extrema(t):
+                if extreme.E < energy_threshold:
+                    ### be sure that the voxel to be eliminated has at least one neighbour
+                    ### beyond itself
+                    n_neighbours = sum(neighbours(extreme, v) for v in mod_voxels)
+                    if n_neighbours > 1:
+                        mod_voxels    .remove(extreme)
+                        dropped_voxels.append(extreme)
+                        drop_voxel(mod_voxels, extreme)
+                        modified = True
 
-        if n_modified_voxels == 0:
-            break
+    return mod_voxels, dropped_voxels
 
-    return mod_voxels
 
 def get_track_energy(track):
     return sum([vox.Ehits for vox in track.nodes()])
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -339,6 +339,10 @@ def drop_end_point_voxels(voxels: Sequence[Voxel], energy_threshold: float, min_
         for v in min_vs:
             v.E += the_vox.E/sum_en_v * v.E
 
+    def nan_energy(voxel):
+        voxel.E = np.nan
+        for hit in voxel.hits:
+            setattr(hit, e_type, np.nan)
 
     mod_voxels     = copy.deepcopy(voxels)
     dropped_voxels = []
@@ -361,6 +365,7 @@ def drop_end_point_voxels(voxels: Sequence[Voxel], energy_threshold: float, min_
                         mod_voxels    .remove(extreme)
                         dropped_voxels.append(extreme)
                         drop_voxel(mod_voxels, extreme)
+                        nan_energy(extreme)
                         modified = True
 
     return mod_voxels, dropped_voxels
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -337,7 +337,7 @@ def drop_end_point_voxels(voxels: Sequence[Voxel], energy_threshold: float, min_
         for h in min_hs:
             setattr(h, e_type, getattr(h, e_type) + getattr(h, e_type) * the_vox.E/sum_en_h)
         for v in min_vs:
-            v.E += the_vox.E/sum_en_v * v.E
+            v.E = sum(getattr(h, e_type) for h in v.hits)
 
     def nan_energy(voxel):
         voxel.E = np.nan
--- a/invisible_cities/reco/paolina_functions_test.py
+++ b/invisible_cities/reco/paolina_functions_test.py
@@ -501,7 +501,7 @@ def test_energy_is_conserved_with_dropped_voxels(hits, requested_voxel_dimension
 
     energies = [v.E for v in voxels]
     e_thr = min(energies) + fraction_zero_one * (max(energies) - min(energies))
-    mod_voxels = drop_end_point_voxels(voxels, e_thr, min_voxels)
+    mod_voxels, _ = drop_end_point_voxels(voxels, e_thr, min_voxels)
     tot_final_energy = sum(v.E for v in mod_voxels)
     final_trks = make_track_graphs(mod_voxels)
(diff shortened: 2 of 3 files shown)


invisible_cities/reco/paolina_functions.py
```diff
@@ -350,30 +339,37 @@ def drop_voxel(voxels: Sequence[Voxel], the_vox: Voxel) -> int:
         for v in min_vs:
             v.E += the_vox.E/sum_en_v * v.E
 
-        return 1
+    def nan_energy(voxel):
+        voxel.E = np.nan
+        for hit in voxel.hits:
+            hit.Ep = np.nan
```
mmkekic: Here the nan energy of the hit should be voxel.Etype

invisible_cities/reco/paolina_functions_test.py
```diff
@@ -511,6 +511,18 @@ def test_energy_is_conserved_with_dropped_voxels(hits, requested_voxel_dimension
...
+@given(bunch_of_corrected_hits(), box_sizes, min_n_of_voxels, fraction_zero_one)
+def test_dropped_voxels_have_nan_energy(hits, requested_voxel_dimensions, min_voxels, fraction_zero_one):
+    voxels            = voxelize_hits(hits, requested_voxel_dimensions, strict_voxel_size=False)
+    energies          = [v.E for v in voxels]
+    e_thr             = min(energies) + fraction_zero_one * (max(energies) - min(energies))
+    _, dropped_voxels = drop_end_point_voxels(voxels, e_thr, min_voxels)
+    for voxel in dropped_voxels:
+        assert np.isnan(voxel.E)
+        for hit in voxel.hits:
+            assert np.isnan(hit.Ep)
```
mmkekic: Here as well the nan should voxels.Etype

invisible_cities/reco/paolina_functions_test.py
```diff
@@ -502,6 +511,18 @@ def test_energy_is_conserved_with_dropped_voxels(hits, requested_voxel_dimension
...
+
```
mmkekic: Here as well voxels.Etype

invisible_cities/reco/paolina_functions.py
gonzaponte: Ok to all comments like this

invisible_cities/reco/paolina_functions_test.py
```diff
@@ -511,6 +512,69 @@ def test_energy_is_conserved_with_dropped_voxels(hits, requested_voxel_dimension
...
+        if v1.X != v2.X: return -1 if v1.X < v2.X else +1
+        if v1.Y != v2.Y: return -1 if v1.Y < v2.Y else +1
+        if v1.Z != v2.Z: return -1 if v1.Z < v2.Z else +1
+        return 0
+
+    def energy_from_hits(voxel, e_type):
+        return [getattr(hit, e_type) for hit in voxel.hits]
+
+    voxels     = voxelize_hits(hits, requested_voxel_dimensions, strict_voxel_size=False, energy_type=energy_type)
+    voxels     = sorted(voxels, key=cmp_to_key(compare_voxels))
```
mmkekic: I think is clearer to use something like this
`key=lambda x: (x.X, x.Y, x.Z)`
rather than `cmp_to_key`

invisible_cities/reco/paolina_functions_test.py
```diff
@@ -575,32 +639,59 @@ def test_tracks_with_dropped_voxels(ICDATADIR):
...
+              Voxel( 5,  5, 0, 1.5, size=5, e_type=HitEnergy.Ep, hits = [make_hit( 5,  5, 0, 0.9), make_hit( 5,  8, 0, 0.6)]),
+              Voxel(10,  0, 0, 2.0, size=5, e_type=HitEnergy.Ep, hits = [make_hit(10,  5, 0, 1.2), make_hit(11,  5, 0, 0.8)]),
+              Voxel(15,  0, 0, 2.5, size=5, e_type=HitEnergy.Ep, hits = [make_hit(15,  0, 0, 1.8), make_hit(11,  0, 0, 0.7)]),
+              Voxel(20,  0, 0, 3.0, size=5, e_type=HitEnergy.Ep, hits = [make_hit(20,  0, 0, 1.5), make_hit(11,  0, 0, 1.5)])]
+
+    modified_voxels, _ = drop_end_point_voxels(voxels, energy_threshold = 0.5, min_vxls = 1)
+    for v in modified_voxels:
+        assert np.isclose(v.E, sum(h.Ep for h in v.hits))
+
+
```
mmkekic, line 703: I'm fine with having this explicit voxels example that clearly demonstrates the bug  but maybe we could add a general hypothesis one as well? Something like 
```
@mark.parametrize("energy_type", HitEnergy)
@given(hits                       = bunch_of_corrected_hits(),
       requested_voxel_dimensions = box_sizes,
       min_voxels                 = min_n_of_voxels,
       fraction_zero_one          = fraction_zero_one)
def test_drop_voxels_voxel_energy_is_sum_of_hits_general(hits, requested_voxel_dimensions, min_voxels, fraction_zero_one, energy_type):
    voxels            = voxelize_hits(hits, requested_voxel_dimensions, strict_voxel_size=False, energy_type=energy_type)
    energies = [v.E for v in voxels]
    e_thr = min(energies) + fraction_zero_one * (max(energies) - min(energies))
    mod_voxels, _ = drop_end_point_voxels(voxels, e_thr, min_voxels)
    for v in mod_voxels:
        assert np.isclose(v.E, sum(h.Ep for h in v.hits))

```

gonzaponte: Sure

invisible_cities/reco/paolina_functions_test.py
gonzaponte: That's probably more clever, yes.

jacg: You should tink of `cmp_to_key` as a tool for delaying how to think about one particular aspect of porting python2 code to python3. Given that you're not porting python2, just ignore the existence of `cmp_to_key`.

I can imagine some pathological solutions where `cmp_to_key` might actually be a better solution than writing a genuine `key` function, but to tree-level we can ignore those.

mmkekic (approved):
This PR modifies the output of `drop_end_point_voxels` to output both dropped and modified voxels, needed for city esmeralda. It also shows and fixes a nasty bug introduced in #655. Thanks @gonzaponte !

invisible_cities/reco/paolina_functions_test.py
gonzaponte: Thanks for the tip. I wasn't sure how to compare voxels and I looked around and found `cmp_to_key`. Wasn't the most elegant solution, but it was reasonably simple.

jacg: Oh, I think it's very *elegant*! Just think about how you would have to implement `cmp_to_key` and it will make you smile. But it will needlessly slow down your sorts in most cases.

jacg: Heh! And I hadn't spotted the implementation of `compare_voxels` until now. That's quite an effort!

gonzaponte: What do you mean by "how you would have to implement `cmp_to_key`?"

jacg: Imagine that you've been given the job of implementing `cmp_to_key`. How would you do it?

gonzaponte: That question has frustrated me a lot. I haven't been able to make it without defining a class that implements the comparison operations. I don't know how to catch the call to the object method that performs the comparison so I can check if it is `operator.eq`, `operator.lt` or `operator.gt`, which was my initial attempt.
My final attempt was to implement something in the lines of defining a class that implements `__eq__`, `__lt__` and `__gt__` comparing `cmp`'s output to 0:

```
def cmp_to_key(cmp):
    class key:
        def __init__(self, thing): self.thing = thing
        def __lt__  (self, other): return cmp(self.thing, other.thing) <  0
        def __eq__  (self, other): return cmp(self.thing, other.thing) == 0
        def __gt__  (self, other): return cmp(self.thing, other.thing) >  0
    return key
```

I would really like to see an implementation based on comparing with `operator.something`.

jacg: > That question has frustrated me a lot.

I'm sorry. That wasn't the intention.

> I haven't been able to make it without defining a class that implements the comparison operations.

Which is the obvious solution that comes to my mind. And I find it rather neat. The repetition in each of the methods is bloody tedious, but the idea that you have a type which adopts / adapts your `cmp` is rather neat.

Off the top of my head I'm not sure whether you have implemented enough of the comparison methods (`__le__` etc.) I think that they may have default implementations in terms of the others, and you might have implemented a valid minimal set, but I would have to check.

> I would really like to see an implementation based on comparing with `operator.something`.

Heh, yes, that would be `cmp` and `.__cmp__` which were removed in Python 3, as a consequence of the migration to rich comparisons.

The implementation is noisy, because Python decided to make a mess by conflating a bunch of things into a single protocol:

1. Equality (which can exist without ordering)
2. Ordering
3. Numpy wanting comparison operators to return something other than `bool`.

The last one is an interesting example of the "practicality beats purity" part of the Zen of Python. It seems pragmatic in Numpy at first blush, but I'm not entirely convinced that this outweighs the problems it causes in Numpy and elsewhere. It's exactly this sort of stuff that (justyfiably) gives overloading a bad name.

Haskell is far less cavalier with sacrificing purity for pragmatism. Consequently, in Haskell, you would only have to implement the `Ord` class, just one of its `compare` or `=>` methods will do as a minimum, and you'd be done. And sticking to purity is starting to pay dividents. Numpy would just have to come up with new operators rather hijacking existing ones and changing their universally expected meaning. This is, of course, mitigated in Haskell by the fact that anybody can create a new operator (they are just funny function names) while in Python only the language implementors can create new operators (they are part of the syntax). This is purity paying dividents once again: operators as a special and sacred construct are a dirty and impure idea.

But anyway, if you ignore the boilerplate, I find the idea behind the implementation of `cmp_to_key` rather nice: When I first saw that it had been added to `functools` my initial reaction was "but that's impossible", followed by "but it exists, so how would I write it ... oooh, that's cool!"

I guess that it's the transition from assuming that something is impossible to the realization that it's quite easy, which tickled me. So I like `cmp_to_key` for that reason.

gonzaponte: > I'm sorry. That wasn't the intention.

Nothing dramatic, I may have overstated it.

Actually this: "in Python only the language implementors can create new operators" was my frustration. I felt a lack of control over the operations being made. At the same time, I was thinking exactly this: " in Haskell, you would only have to implement the Ord class,  just one of its compare or => methods will do as a minimum" and I couldn't believe there was not a workaround in Python.

> Which is the obvious solution that comes to my mind

Yeah, the idea came vaguely to my mind, but I was a bit obfuscated with not using classes (I was convinced it *had to* be possible to do it without them).

> And I find it rather neat. 

It is! Actually, when I finished the code it wasn't as bad as I thought initially.

> Off the top of my head I'm not sure whether you have implemented enough of the comparison methods

Me neither, I had been googling a bit about that, but I didn't find the answer (I think I've read it in a python book ("Learning Python"?), but I am not sure). It should be possible to do it by defining a `Comparable` metaclass (type?) that defines every comparison in terms of the others. Any class deriving from it would only need to define two of the relations, I think.

> in Haskell by the fact that anybody can create a new operator

I guess this is not something specific from Haskell. I think Julia has it too, which is more Python-like.

> I guess that it's the transition from assuming that something is impossible to the realization that it's quite easy, which tickled me. So I like cmp_to_key for that reason.

I didn't have the "prejudice" that it was impossible, but I did like that the final result wasn't monstrous. Overall I think I've learned a lot about programming paradigms and the guts of a programming language, so thanks for the exercise.

jacg: >> In Haskell [...] anybody can create an operator

> I guess this is not something specific from Haskell

Absolutely not! But given that I was already using Haskell as an example of a language with different design choces ...

> I was a bit obfuscated with not using classes (I was convinced it had to be possible to do it without them).

Well, it *might* be possible, but I struggle to see where the scope for doing it is. A key function has to return one object per element of the iterable, and those objects will have to support comparison operators which respect the semantics of the comparison function fed to to `cmp_to_key` ... in a fully general way. This points to the implementation we've been discussing, but I'm not sure that it proves that it can't be done without writing a class (or some morally equivalent Python trickery).

> Overall I think I've learned a lot about programming paradigms and the guts of a programming language, so thanks for the exercise.

Excellent!

PR #679, opened 2019-11-22
Generalize in range

gonzaponte:
The `in_range` function has been always working to select data in a left-closed, right-open interval (`[ )` ). This PR generalizes it adding two keyword arguments to select which type of interval to use in each of the bounds.

Some tests are added to verify this new functionality.

### DIFF
--- a/invisible_cities/core/core_functions.py
+++ b/invisible_cities/core/core_functions.py
@@ -63,7 +63,7 @@ def relative_difference(x, y, *, norm_mode=NormMode.first):
 
 def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed=False):
     """
-    Find values in range [minval, maxval).
+    Find values from minval to maxval.
 
     Parameters
     ---------
@@ -73,16 +73,16 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
         Range minimum. Defaults to -inf.
     maxval : int or float, optional
         Range maximum. Defaults to +inf.
-    left   : {"open", "close"}
-        Use close (default) or open  interval on the lower bound
-    right  : {"open", "close"}
-        Use open  (default) or close interval on the upper bound
+    left_closed   : bool
+        Closed semi-interval if True (default); open if false.
+    right_closed  : {"open", "close"}
+        Closed semi-interval if True; open if false (default).
 
     Returns
     -------
     selection : np.ndarray
         Boolean array with the same dimension as the input. Contains True
-        for those values of data in the input range and False for the others.
+        for those values of data within the input range and False for the rest.
     """
     lower_bound = data >= minval if left_closed  else data > minval
     upper_bound = data <= maxval if right_closed else data < maxval
--- a/invisible_cities/core/core_functions.py
+++ b/invisible_cities/core/core_functions.py
@@ -59,7 +59,7 @@ def relative_difference(x, y, *, norm_mode=NormMode.first):
         raise TypeError(f"Unrecognized relative difference option: {norm_mode}")
 
 
-def in_range(data, minval=-np.inf, maxval=np.inf):
+def in_range(data, minval=-np.inf, maxval=np.inf, left="closed", right="open"):
     """
     Find values in range [minval, maxval).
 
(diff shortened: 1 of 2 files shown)


invisible_cities/core/core_functions.py
```diff
@@ -59,7 +59,7 @@ def relative_difference(x, y, *, norm_mode=NormMode.first):
         raise TypeError(f"Unrecognized relative difference option: {norm_mode}")
 
 
-def in_range(data, minval=-np.inf, maxval=np.inf):
+def in_range(data, minval=-np.inf, maxval=np.inf, left="close", right="open"):
```
andLaing: I'd prefer closed rather than close (which means near really). Also maybe an enum?

invisible_cities/core/core_functions_test.py
```diff
@@ -72,6 +80,32 @@ def test_in_range_right_shape(data):
     assert core.in_range(data, -1., 1.).shape == data.shape
 
 
+@given(sorted_unique_arrays)
+def test_in_range_left_open_interval(data):
+    output =core.in_range(data, minval=data[0], left="open")
```
andLaing: missing space post =

invisible_cities/core/core_functions_test.py
```diff
@@ -72,6 +80,32 @@ def test_in_range_right_shape(data):
...
+    output = core.in_range(data, maxval=data[-1], right="open")
+    assert not output[-1]
+    assert all(output[:-1])
+
+
+@given(sorted_unique_arrays)
+def test_in_range_left_close_interval(data):
+    output =core.in_range(data, minval=data[0], left="close")
+    assert all(output)
+
```
andLaing, line 104: Another missing space.

Also, this is the default behaviour and should have an open right side, shouldn't it? I understood this should mean the [-1] entry is false?

invisible_cities/core/core_functions_test.py
```diff
@@ -72,6 +80,32 @@ def test_in_range_right_shape(data):
...
+
+@given(sorted_unique_arrays)
+def test_in_range_right_close_interval(data):
+    output = core.in_range(data, maxval=data[-1], right="close")
+    assert all(output)
+
```
andLaing, line 111: This tests the same result as above despite being a different condition which is a bit confusing. I suppose it's because of the default -np.inf for the lower limit but it seems counter intuitive.

invisible_cities/core/core_functions.py
gonzaponte: Indeed, it should be "closed". I thought of an enum, but I think it can make it uncomfortable to use since this function is not so much used in the cities, but externally. On the other hand this might be a lazy-person argument... I am open to change it.

invisible_cities/core/core_functions_test.py
gonzaponte: It is the default behavior with left-closed and right-open. However the right doesn't matter because it's infinite (default value), therefore all elements must be within range. Also it should be the same as `all(core.in_range(data, minval=data[0], maxval=data[-1], left="close", right="close")`

invisible_cities/core/core_functions_test.py
gonzaponte: Yes, maybe I should specify the default value in the test for clarity.

invisible_cities/core/core_functions.py
gonzaponte: I've pushed a version using an enum to compare and make a decision on this.

andLaing: I like that now you have a test to show the valid values in the Enum case so I'd use this version as long as the other developers don't find it too clunky.

andLaing (approved):
Simple changes to a useful function to make it more general. Well tested an clearly developed.

invisible_cities/core/core_functions.py
mmkekic: I think I'd prefer something like `left=True, right=False`, that way the user will not have to import enum class and is somehow consistent with numpy/pandas style?

jahernando: I agree with @mmkekic to have as arguments: left_closed = True, right_closed = False

Issue #661, opened 2019-07-04
Version testing


Aretno:
Hi everybody,

This topic is to discuss the software version tests (additional to the usual test functions) that we want to implement to make sure we do not end in a similar situation as the one we are currently in.

After scratching our heads for a while, @mmkekic and I would like to propose this scheme:

1. After each PR is **approved and before merging**, an automatic script will run in majorana cluster. The script will process a big dataset (1000k files of Kr, and a full run of RunV high energy data) starting from rwf (after decoding). This will be run twice, once with the current head of IC and once with the changes of the PR.

2. The script will compare all the values of the output by comparing the tables themselves, if ok, we all can be happy. If the output is not the same, then a series of plots and numbers that compare the data, at the level (and posterior ones) where the mismatch occurs, will be created.

3. A report will be written with the result of the comparison, if needed due to a comparison fail, the report will include a path to the plots in pdf, the numbers of relevance (like survival rates and the alike). This report and plots should be automatically uploaded to git and we should keep storage of them through git-lfs.

4. The reviewer shall check the report, if it fails, the differences should be understood and explained (by the developer) before merging the PR.

In addition to this, we would also like to add a new label to git, something like 'data change'. This is to be added to a PR that is known to change the data. This way we can quickly identify at which PRs there has been a change in the processing output in case, for some reason, something smells fishy again.

We already talked with @jmbenlloch about whether this kind of scheme is possible and he is pretty sure we can do it. Not easy, but doable. But not in a two week period as this would be a major work.

I understand that this will be a more general testing thing and that the current situation will be handled by Gonzalo and Alberto so we don't strictly need this full machinery for the current situation but for future reassurance.

Details will be worked out as we work on this but, as a general idea, this is it. Suggestions, doubts, etc.?

msorel:
Hi, sounds like a very complete and ambitious plan to me! My only comment: lets not test only on Kr and trigger2 data, bit also on corresponding MC files (detsim output, probably). So, running on four datasets

jjgomezcadenas:
Ambitious indeed. I see the risk that we are aiming too high and it takes too long. I would like that you guys propose a simpler prototype as a first step that can be quickly implemented and tested (e.g, 2 weeks),THEN we can go for the bigger thing.

Enviado desde mi iPhone

Aretno:
[...]

PR #680, opened 2019-11-22
Fix hypothesis health checks

gonzaponte:
Every now and then the travis build fails because of data generation being too slow. This is a combination of two things:
- `data_frames` strategy is out of our hands
- the large number of columns increases execution time significantly and can't/shouldn't be reduced

This PR suppresses the HealthCheck for those tests using this strategy.

### DIFF
--- a/invisible_cities/reco/monitor_functions_test.py
+++ b/invisible_cities/reco/monitor_functions_test.py
@@ -5,6 +5,7 @@ import numpy as np
 
 from hypothesis              import given
 from hypothesis              import settings
+from hypothesis              import HealthCheck
 from hypothesis.extra.pandas import columns, data_frames
 from hypothesis.strategies   import floats
 
@@ -417,7 +418,7 @@ kdst_variables = ['nS2', 'S1w'  , 'S1h', 'S1e', 'S1t', 'S2w', 'S2h', 'S2e', 'S2q
 
 
 @given(data_frames(columns=columns(kdst_variables, elements=floats(allow_nan=False))))
-@settings(deadline=None)
+@settings(deadline=None, suppress_health_check=(HealthCheck.too_slow,))
 def test_fill_kdst_var_1d(kdst):
     var_dict = defaultdict(list)
     monf.fill_kdst_var_1d (kdst, var_dict)
@@ -430,7 +431,7 @@ def test_fill_kdst_var_1d(kdst):
 
 
 @given(data_frames(columns=columns(kdst_variables, elements=floats(allow_nan=False))))
-@settings(deadline=None)
+@settings(deadline=None, suppress_health_check=(HealthCheck.too_slow,))
 def test_fill_kdst_var_2d(kdst):
     var_dict = defaultdict(list)
     monf.fill_kdst_var_1d (kdst, var_dict)
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,4 +21,4 @@ install:
   - source manage.sh work_in_python_version_no_tests ${IC_PYTHON_VERSION}
 
 script:
-  - HYPOTHESIS_PROFILE=travis-ci bash manage.sh run_tests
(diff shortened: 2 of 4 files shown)


invisible_cities/reco/monitor_functions_test.py
```diff
@@ -417,7 +406,7 @@ def test_kdst_bins():
 
 
 @given(data_frames(columns=columns(kdst_variables, elements=floats(allow_nan=False))))
-@settings(deadline=None)
+@settings(deadline=None, suppress_health_check=(HealthCheck.too_slow,))
```
mmkekic, line 421: Maybe we can constrain the  index range? 
`index=range_indexes(min_size=1, max_size=10)`
But I agree that it still requires HealrCheck to be suppressed

Aretno:
Actually, since this happens in monitor_functions and those now live outside IC we could just erase the files and forget about their existance.

invisible_cities/reco/monitor_functions_test.py
gonzaponte: Sounds good

gonzaponte:
What do you mean they live outside IC?

Aretno:
Ah, this: https://github.com/nextic/Olivia

So, the kdst functions are no longer used, while the pmaps and rwf all are living as clients instead of inside IC.

gonzaponte:
Ah, if these functions are zombies, kill 'em all! (but keep the tests in the other repo! :))

mmkekic:
Shell we then remove the files altogether?  @Aretno , @gonzaponte  whats the status of this?

gonzaponte:
@jacg, can I ask for your advice here?

Hypothesis is throwing a ton of these messages when running the Travis build (this doesn't happen on my machine):

```
HypothesisDeprecationWarning: Test took XXX ms to run. In future the default deadline
setting will be 200ms, which will make this an error. You can set deadline to an
explicit value of e.g. 300 to turn tests slower than this into an error, or you can
set it to None to disable this check entirely.
```

so I attempted to raise the deadline for the specific tests that were throwing the warning. However, after doing so, more and more tests keep popping up with this warning. This makes me think that there are huge fluctuations in the time it takes Travis to run a given test and therefore it doesn't make sense to set this deadline on a test-by-test basis.

Do you recommend disabling the deadline when running in Travis, maybe setting a deadline on the global test-suite execution time? Do you have any other suggestions on how to address this?

jacg:
Hmm. I don't really have any words of wisdom on this. It would be helpful to understand what is going on with Travis. I would try disabling the deadlines on Travis, and see how long the tests are taking. If the tests exceed the deadlines relatively modestly, then it might just be best to (take on some technical debt, and) keep them switched off for now, thus putting off the need to understand what is causing this. If the tests turn out to be unacceptably slow, you'll probably have to dig deeper to find out why this is happening. If it gets to that, it might be worthwhile trying it locally in a Travis Docker image.

gonzaponte:
In the end, I have realized that part of the problem was that the Travis config had hardcoded the number of processors to use (thanks to @mmkekic for her input). This value was higher than the actual number of available processors.

Nonetheless, some tests were still hitting the hypothesis deadline, which will become an error in the future and therefore must be fixed. Execution times vary from machine to machine and Travis seems to be particularly slow. Thus, I have created a new hypothesis profile specifically for Travis with a slightly higher deadline. I have checked and now we don't get those warnings anymore.

We still get a warning about a global timeout for hypothesis, but this feature will disappear in a future release and therefore is not a concern.

As far as I am concern, this PR is ready for review.

mmkekic (approved):
This PR registers a new hypothesis profile for travis-ci with extended deadline and auto number of processes. Hopefully this will resolve occasional Travis failures due to hypothesis health checks. Also once (if ever) we setup our own CI server ( #661, miguelsimon/IC#3 ) this will not be an issue any more.

PR #681, opened 2019-11-26
Add too-many-hits filter esmeralda

mmkekic:
There are some nonphysical events that have more than 10s of thousands hits. This PR makes Esmeralda not run tracking functions on events that have too many hits.

### DIFF
--- a/invisible_cities/cities/esmeralda.py
+++ b/invisible_cities/cities/esmeralda.py
@@ -172,7 +172,7 @@ def track_blob_info_creator_extractor(vox_size         : [float, float, float],
     """
     def create_extract_track_blob_info(hitc):
         df = pd.DataFrame(columns=list(types_dict_tracks.keys()))
-        if len(hitc.hits)>max_num_hits:
+        if len(hitc.hits) > max_num_hits:
             return df, hitc, True
         #track_hits is a new Hitcollection object that contains hits belonging to tracks, and hits that couldnt be corrected
         track_hitc = evm.HitCollection(hitc.event, hitc.time)
@@ -185,7 +185,7 @@ def track_blob_info_creator_extractor(vox_size         : [float, float, float],
             hitc      = evm.HitCollection(hitc.event, hitc.time)
             hitc.hits = hits_without_nan
 
-        if len(hitc.hits)>0:
+        if len(hitc.hits) > 0:
             voxels           = plf.voxelize_hits(hitc.hits, vox_size, strict_vox_size, evm.HitEnergy.Ep)
             (    mod_voxels,
              dropped_voxels) = plf.drop_end_point_voxels(voxels, energy_threshold, min_voxels)
--- a/invisible_cities/cities/esmeralda.py
+++ b/invisible_cities/cities/esmeralda.py
@@ -145,7 +145,8 @@ def track_blob_info_creator_extractor(vox_size         : [float, float, float],
                                       strict_vox_size  : bool                 ,
                                       energy_threshold : float                ,
                                       min_voxels       : int                  ,
-                                      blob_radius      : float
+                                      blob_radius      : float                ,
+                                      max_num_hits     : int
                                      ) -> Callable:
     """
     For a given paolina parameters returns a function that extract tracks / blob information from a HitCollection.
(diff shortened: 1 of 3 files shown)


invisible_cities/cities/esmeralda.py
```diff
@@ -170,6 +171,9 @@ def track_blob_info_creator_extractor(vox_size         : [float, float, float],
     A function that from a given HitCollection returns a pandas DataFrame with per track information.
     """
     def create_extract_track_blob_info(hitc):
+        df = pd.DataFrame(columns=list(types_dict_tracks.keys()))
+        if len(hitc.hits)>max_num_hits:
```
gonzaponte:
```suggestion
        if len(hitc.hits) > max_num_hits:
```

invisible_cities/cities/esmeralda_test.py
```diff
@@ -317,3 +324,50 @@ def test_esmeralda_all_hits_after_drop_voxels(data_hdst, esmeralda_tracks, corre
 
     #check that the sum of Ep and Ec energies is the same
     assert np.nansum(df_phits.Ec) == np.nansum(df_phits.Ep)
+
+def test_esmeralda_filters_long_events(data_hdst, esmeralda_tracks, correction_map_filename, config_tmpdir):
```
gonzaponte:
```suggestion
def test_esmeralda_filters_events_with_too_many_hits(data_hdst, esmeralda_tracks, correction_map_filename, config_tmpdir):
```

invisible_cities/cities/esmeralda_test.py
```diff
@@ -317,3 +324,50 @@ def test_esmeralda_all_hits_after_drop_voxels(data_hdst, esmeralda_tracks, corre
...
+    PATH_IN   = data_hdst
+    PATH_OUT  = os.path.join(config_tmpdir, "exact_tracks_esmeralda_drop_voxels.h5")
```
gonzaponte:
```suggestion
    PATH_OUT  = os.path.join(config_tmpdir, "esmeralda_filters_long_events.h5")
```

invisible_cities/cities/esmeralda_test.py
```diff
@@ -317,3 +324,50 @@ def test_esmeralda_all_hits_after_drop_voxels(data_hdst, esmeralda_tracks, corre
...
+    assert np.all(summary[ summary.event.isin(paolina_events)].evt_ntrks >  0        )
+    assert np.all(summary[~summary.event.isin(paolina_events)].evt_ntrks == 0        )
+    assert np.all(summary[~summary.event.isin(paolina_events)].evt_nhits >  nhits_max)
+    #assert all hits and events are in hits table
+    assert len(hits) == 601
+    assert hits.event.nunique() == nevt_req
+    topology_filter = dio.load_dst(PATH_OUT, 'Filters', 'topology_select')
+    assert len(topology_filter) == nevt_req
+    assert np.all(topology_filter[ topology_filter.event.isin(paolina_events)].passed == 1)
+    assert np.all(topology_filter[~topology_filter.event.isin(paolina_events)].passed == 0)
```
gonzaponte: I would add some spaces between groups of assertions for readability

invisible_cities/config/esmeralda.conf
```diff
@@ -21,4 +21,5 @@ paolina_params      = dict(
    strict_vox_size  = True,
    energy_threshold = 10 * keV,
    min_voxels       = 2,
-   blob_radius      = 21 * mm)
+   blob_radius      = 21 * mm,
+   max_num_hits     = 10000)
```
gonzaponte, line 25: Is this number optimized in some manner? You should make sure it doesn't kill real event while killing outliers

invisible_cities/cities/esmeralda.py
```diff
@@ -170,6 +171,9 @@ def track_blob_info_creator_extractor(vox_size         : [float, float, float],
...
+            return df, hitc, True
```
gonzaponte, line 176: Is this `out_of_map = True` always correct? Does it have a meaning in this case at all?

gonzaponte (commented):
This PR adds a protection against events with an extreme number of hits avoiding the paolina step. The test looks good. Good job!

invisible_cities/cities/esmeralda.py
mmkekic: In practice it seems to be, in theory it doesnt have to.. Since the events will be saved in summary table the correct thing would be either to rename the column or add a new one (ie 'good_event'). However since most of the data has already been processed I think is best to leave it as is. I fixed it to True since most of the analysis using blobs and tracks properties will only be valid for events that are inside the active volume, so it will skip this weird large events.

invisible_cities/config/esmeralda.conf
mmkekic: We gave a look ad number of hits distribution, but in any case, this being a config file number, can easily be changed

gonzaponte: Yes, I mean for documentation kind of thing. It's good enough I guess

gonzaponte (approved):
New filter for Esmeralda skipping events with too many hits. Adds a good and readable test. Good job!

paolafer:
Do we know why these events appear? I guess it's only in data, right?

mmkekic:
@paolafer  Pau believes it is some energetic muon that provoked a cascade, but we dont really know. And yes, it is only in data

PR #685, opened 2019-12-12
Altered rwf_io to be able to write directly in root

andLaing:
EArray for the sensors can now be saved
in h5file.root as is done in the C++
detsim.

Tidied and added type hints, docstrings
and a test for the writer.

### DIFF
--- a/invisible_cities/io/rwf_io.py
+++ b/invisible_cities/io/rwf_io.py
@@ -22,7 +22,7 @@ def rwf_writer(h5out           : tb.file.File          ,
     group_name      : str
                       Name of the group in h5in.root
                       Known options: RD, BLR
-                      Setting to root will save directly in root
+                      Setting to None will save directly in root
     table_name      : str
                       Name of the table
                       Known options: pmtrwf, pmtcwf, sipmrwf
@@ -33,19 +33,18 @@ def rwf_writer(h5out           : tb.file.File          ,
     waveform_length : int
                       Number of samples per sensor
     """
-    if group_name is 'root':
+    if   group_name is None:
         rwf_group = h5out.root
+    elif group_name in h5out.root:
+        rwf_group = getattr           (h5out.root, group_name)
     else:
-        try:
-            rwf_group = getattr           (h5out.root, group_name)
-        except tb.NoSuchNodeError:
-            rwf_group = h5out.create_group(h5out.root, group_name)
+        rwf_group = h5out.create_group(h5out.root, group_name)
 
-    rwf_table = h5out.create_earray(rwf_group,
-                                   table_name,
-                                   atom    = tb.Int16Atom(),
-                                   shape   = (0, n_sensors, waveform_length),
-                                   filters = tbl.filters(compression))
+    rwf_table = h5out.create_earray(rwf_group                                ,
+                                    table_name                               ,
+                                    atom    =                  tb.Int16Atom(),
+                                    shape   = (0, n_sensors, waveform_length),
+                                    filters =        tbl.filters(compression))
     def write_rwf(waveform : np.ndarray) -> None:
         """
         Writes raw waveform arrays to file.
--- a/invisible_cities/io/rwf_io.py
+++ b/invisible_cities/io/rwf_io.py
@@ -1,23 +1,57 @@
+import numpy  as np
 import tables as tb
+
+from typing  import Callable
 
 from .. reco import tbl_functions as tbl
 
 
-def rwf_writer(file,
+def rwf_writer(h5out           : tb.file.File          ,
                *,
-               group_name      : 'options: RD, BLR',
-               table_name      : 'options: pmtrwf, pmtcwf, sipmrwf',
-               compression     = 'ZLIB4',
-               n_sensors       : 'number of pmts or sipms',
-               waveform_length : 'length of pmt or sipm waveform_length'):
-    try:                       rwf_group = getattr          (file.root, group_name)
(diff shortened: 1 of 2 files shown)


invisible_cities/io/rwf_io.py
```diff
@@ -1,23 +1,57 @@
...
-               compression     = 'ZLIB4',
-               n_sensors       : 'number of pmts or sipms',
-               waveform_length : 'length of pmt or sipm waveform_length'):
-    try:                       rwf_group = getattr          (file.root, group_name)
-    except tb.NoSuchNodeError: rwf_group = file.create_group(file.root, group_name)
+               group_name      :          str          ,
+               table_name      :          str          ,
+               compression     :          str = 'ZLIB4',
+               n_sensors       :          int          ,
+               waveform_length :          int          ) -> Callable:
```
gonzaponte, line 15: This will not work, arguments with default values (optional) cannot be followed by arguments without default values (mandatory)

invisible_cities/io/rwf_io.py
```diff
@@ -1,23 +1,57 @@
...
+                      Known options: pmtrwf, pmtcwf, sipmrwf
+    compression     : str
+                      file compression
+    n_sensors       : int
+                      number of sensors in the table (shape[0])
+    waveform_length : int
+                      Number of samples per sensor
+    """
+    if group_name is 'root':
+        rwf_group = h5out.root
```
gonzaponte: I think it is more natural to use an empty string than `"root"`, but it is less comprehensible. Maybe using `None` is a good compromise.

invisible_cities/io/rwf_io.py
```diff
@@ -1,23 +1,57 @@
...
+    else:
+        try:
+            rwf_group = getattr           (h5out.root, group_name)
+        except tb.NoSuchNodeError:
+            rwf_group = h5out.create_group(h5out.root, group_name)
```
gonzaponte: I think it is more correct to ask whether the node exists than trying to see if it works:
```
if group_name in h5out.root:
    h5out.create_group(h5out.root, group_name)

rwf_group = getattr(h5out.root, group_name)
```

invisible_cities/io/rwf_io.py
```diff
@@ -1,23 +1,57 @@
...
 
-    rwf_table = file.create_earray(rwf_group,
+    rwf_table = h5out.create_earray(rwf_group,
                                    table_name,
                                    atom    = tb.Int16Atom(),
                                    shape   = (0, n_sensors, waveform_length),
                                    filters = tbl.filters(compression))
```
gonzaponte: slight missalignment

invisible_cities/io/rwf_io_test.py
```diff
@@ -0,0 +1,45 @@
...
+                                 group_name      = group_name,
+                                 table_name      = table_name,
+                                 n_sensors       =    nsensor,
+                                 waveform_length =    nsample)
+
+        for sens in test_data:
+            rwf_writer_(sens)
+
+    with tb.open_file(ofile) as h5test:
+        if group_name is 'root':
```
gonzaponte: It is not correct to compare strings with `is`. `is` checks whether two objects have the same ID, i.e. if they refer to the same memory space. Curiously this works, interactively, but it is not guaranteed to work in general.

invisible_cities/io/rwf_io_test.py
```diff
@@ -0,0 +1,45 @@
...
+            group = h5test.root
+        else:
+            group = getattr(h5test.root, group_name)
+
+
+        assert table_name in group
+
+        table = getattr(group, table_name)
+        assert table.shape == (nevt, nsensor, nsample)
+        assert_allclose(test_data, table.read())
```
gonzaponte: since they are ints, you should be able to compare them exactly, right?

gonzaponte (commented):
I can't believe we didn't have `rwf_io_test.py`!

invisible_cities/io/rwf_io.py
andLaing: Curiously, that's not something that I changed but now that you mention it it shouldn't work. Must just be that we've always set everything explicitly or something. I'll change it to default to the NEW numbers

gonzaponte: You are right, it was already like that. I've only noticed it because you changed something else :laughing: 
If you want, leave it as is and we open a new PR fixing it everywhere. I don't think it should take any particular number.

invisible_cities/io/rwf_io_test.py
jacg: > Curiously this works, interactively

Yes, it depends on whether the string is interned or not ... and that depends on complex implementation-defined details (which are probably not guaranteed not to evolve with time).  So, to a first order approximation:

1. Do **not** use `is` to compare strings ...
2. ... unless 
    + you have benchmarked to prove that your code is too slow
    + you have profiled to prove that some string comparison is taking a significant portion of your CPU cycles

    If you ever find yourself in that situation (rather unlikely) then interning relevant strings and using `is` might speed it up.

TL;DR: do not use `is` to compare strigns.

invisible_cities/io/rwf_io.py
andLaing: OK, I'll leave it for now

gonzaponte (approved):
Nice extension to the rwf writer. Now it allows writing to the root group. It also adds a good test that covers both the new and the original cases, which were somehow untested.
Good job!

PR #686, opened 2019-12-17
Removing hard-coded parameters in irene

ausonandres:
The purpose of this PR is to remove some parameters (pmt_sample_frequency=25 ns, sipm_sample_frequency=1 mus, and ratio=40) that were hard-coded in some functions of `peak_functions.py`. Now, they are defined in the input of the city (since they should not change their values), and could be specified in config file.

### DIFF
--- a/invisible_cities/cities/irene.py
+++ b/invisible_cities/cities/irene.py
@@ -22,6 +22,7 @@ import tables as tb
 from .. types.ic_types import minmax
 from .. database       import load_db
 
+from .. core.system_of_units_c import units
 from .. reco                  import tbl_functions        as tbl
 from .. reco                  import  peak_functions      as pkf
 from .. core.random_sampling  import NoiseSampler         as SiPMsNoiseSampler
@@ -51,7 +52,8 @@ from .  components import wf_from_files
 def irene(files_in, file_out, compression, event_range, print_mod, detector_db, run_number,
           n_baseline, n_mau, thr_mau, thr_sipm, thr_sipm_type,
           s1_lmin, s1_lmax, s1_tmin, s1_tmax, s1_rebin_stride, s1_stride, thr_csum_s1,
-          s2_lmin, s2_lmax, s2_tmin, s2_tmax, s2_rebin_stride, s2_stride, thr_csum_s2, thr_sipm_s2):
+          s2_lmin, s2_lmax, s2_tmin, s2_tmax, s2_rebin_stride, s2_stride, thr_csum_s2, thr_sipm_s2,
+          pmt_sample_f=25*units.ns, sipm_sample_f=1*units.mus):
     if   thr_sipm_type.lower() == "common":
         # In this case, the threshold is a value in pes
         sipm_thr = thr_sipm
@@ -87,7 +89,7 @@ def irene(files_in, file_out, compression, event_range, print_mod, detector_db, 
                               item = "sipm")
 
     # Build the PMap
-    compute_pmap     = fl.map(build_pmap(detector_db, run_number,
+    compute_pmap     = fl.map(build_pmap(detector_db, run_number, pmt_sample_f, sipm_sample_f,
                                          s1_lmax, s1_lmin, s1_rebin_stride, s1_stride, s1_tmax, s1_tmin,
                                          s2_lmax, s2_lmin, s2_rebin_stride, s2_stride, s2_tmax, s2_tmin, thr_sipm_s2),
                               args = ("ccwfs", "s1_indices", "s2_indices", "sipm"),
@@ -152,7 +154,7 @@ def irene(files_in, file_out, compression, event_range, print_mod, detector_db, 
 
 
 
-def build_pmap(detector_db, run_number,
+def build_pmap(detector_db, run_number, pmt_sample_f, sipm_sample_f,
                s1_lmax, s1_lmin, s1_rebin_stride, s1_stride, s1_tmax, s1_tmin,
                s2_lmax, s2_lmin, s2_rebin_stride, s2_stride, s2_tmax, s2_tmin, thr_sipm_s2):
     s1_params = dict(time        = minmax(min = s1_tmin,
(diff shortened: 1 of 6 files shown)


invisible_cities/cities/irene.py
```diff
@@ -22,6 +22,7 @@
 from .. types.ic_types import minmax
 from .. database       import load_db
 
+from .. core.system_of_units_c import units
```
andLaing, line 28: Line up the imports

invisible_cities/cities/irene_test.py
```diff
@@ -4,6 +4,7 @@
 import tables as tb
 import numpy  as np
 
+import pytest
```
andLaing: I'd use the same format as for the other imports so:
instead of `import pytest`
use `from pytest import raises`

And the same for any other one that you need

invisible_cities/cities/irene.py
```diff
@@ -51,7 +52,8 @@
 def irene(files_in, file_out, compression, event_range, print_mod, detector_db, run_number,
           n_baseline, n_mau, thr_mau, thr_sipm, thr_sipm_type,
           s1_lmin, s1_lmax, s1_tmin, s1_tmax, s1_rebin_stride, s1_stride, thr_csum_s1,
-          s2_lmin, s2_lmax, s2_tmin, s2_tmax, s2_rebin_stride, s2_stride, thr_csum_s2, thr_sipm_s2):
+          s2_lmin, s2_lmax, s2_tmin, s2_tmax, s2_rebin_stride, s2_stride, thr_csum_s2, thr_sipm_s2,
+          pmt_sample_f=25*units.ns, sipm_sample_f=1*units.mus):
```
andLaing: I'm not sure that I'm happy with the names. The f means frequency, I assume but the value isn't really a frequency but a width. It's not that important maybe but something like 'pmt_samp_wid' might be a bit more correct without making the name too long

invisible_cities/cities/irene_test.py
```diff
@@ -386,3 +387,59 @@ def test_irene_empty_input_file(config_tmpdir, ICDATADIR):
...
+    conf.update(dict(run_number    = 2983,
+                 files_in      = PATH_IN,
+                 file_out      = PATH_OUT,
+                 pmt_sample_f  = pmt_sample_f,
+                 sipm_sample_f = sipm_sample_f))
+
+    msg  = "Shapes don't match!\n"
+    msg += "times has length 6\n"
+    msg += "pmts  has length 6 \n"
+    msg += "sipms has length 3\n"
```
andLaing, line 438: I'm not really a fan of this comparison where you're explicitly writing a copy of the error message. Unfortunately, I don't see an obvious fix for it without changing the PMap definition to have the error messages accessible.

invisible_cities/cities/irene_test.py
```diff
@@ -386,3 +387,59 @@ def test_irene_empty_input_file(config_tmpdir, ICDATADIR):
...
+    PATH_OUT = os.path.join(config_tmpdir, 'r2983_pmaps_diffparams.h5')
+    conf = configure('dummy invisible_cities/config/irene.conf'.split())
+
+    pmt_sample_f  = 25 * units.ns
+    sipm_sample_f =  2 * units.mus
+    conf.update(dict(run_number    = 2983,
+                 files_in      = PATH_IN,
+                 file_out      = PATH_OUT,
+                 pmt_sample_f  = pmt_sample_f,
+                 sipm_sample_f = sipm_sample_f))
```
andLaing: Line them up

invisible_cities/cities/irene_test.py
```diff
@@ -386,3 +387,59 @@ def test_irene_empty_input_file(config_tmpdir, ICDATADIR):
...
+
+    msg  = "Shapes don't match!\n"
+    msg += "times has length 6\n"
+    msg += "pmts  has length 6 \n"
+    msg += "sipms has length 3\n"
+
+    with pytest.raises(ValueError) as error:
+        assert irene(**conf)
```
andLaing: You don't need the assert here

invisible_cities/reco/peak_functions.py
```diff
@@ -109,32 +118,40 @@ def find_peaks(ccwfs, index,
                time, length,
                stride, rebin_stride,
                Pk, pmt_ids,
+               pmt_sample_f =25*units.ns,
```
andLaing: Watch spacing

invisible_cities/reco/peak_functions.py
```diff
@@ -109,32 +118,40 @@ def find_peaks(ccwfs, index,
...
+             s1_params, s2_params, thr_sipm_s2, pmt_ids,
+             pmt_sample_f, sipm_sample_f):
+    return PMap(find_peaks(ccwf, s1_indx, Pk=S1, pmt_ids=pmt_ids,
+                           pmt_sample_f=pmt_sample_f,
+                           **s1_params),
                 find_peaks(ccwf, s2_indx, Pk=S2, pmt_ids=pmt_ids,
                            sipm_wfs    = sipm_zs_wf,
                            thr_sipm_s2 = thr_sipm_s2,
+                           pmt_sample_f  = pmt_sample_f,
+                           sipm_sample_f = sipm_sample_f,
```
andLaing: alignment

andLaing (changes requested):
Generally good, just a few minor points.

I'd prefer there were a way to get the correct binning from the file or database to save errors but the PMap protections should be enough for now though I suspect there might have to be further changes to accommodate the version with PMT rebin pre-selection

You need to rebase too. Maybe do it pre changes to limit clashes.

ausonandres:
Thank you for the comments. Regarding the first one, the idea was to allow the possibility of specifying pmt and sipm sample widths in the config file, but if you don't want to, standard values will be set at the beginning of the city. The rest of suggestions have been applied.

andLaing (approved):
Some simple changes that remove hardwired numbers from the PMap construction steps. Makes things more flexible for future updates and is tested.

PR #683, opened 2019-12-05
Negative charge filtering

andLaing:
Adds to PR  #658 addressing some of the issues mentioned in issue #606 

Adds a filtering mask in PMap construction to avoid negative bins in accepted peaks.

PR #687, opened 2019-12-17
Protect rebin

andLaing:
This PR reopens #658 with an added patch to protect
against attempts to rebin using only negative charge bins.

Tests are added to demonstrate the problems and their
solution.

### DIFF
--- a/invisible_cities/reco/peak_functions.py
+++ b/invisible_cities/reco/peak_functions.py
@@ -142,7 +142,7 @@ def rebin_times_and_waveforms(times, widths, waveforms,
                               rebin_stride=2, slices=None):
     if rebin_stride < 2: return times, widths, waveforms
 
-    if not slices:
+    if slices is None:
         n_bins = int(np.ceil(len(times) / rebin_stride))
         reb    = rebin_stride
         slices = [slice(reb * i, reb * (i + 1)) for i in range(n_bins)]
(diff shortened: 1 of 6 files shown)


invisible_cities/reco/peak_functions.py
```diff
@@ -153,9 +153,10 @@ def rebin_times_and_waveforms(times, widths, waveforms,
     rebinned_wfs    = np.zeros((n_sensors, len(slices)))
 
     for i, s in enumerate(slices):
-        t  = times    [   s]
-        e  = waveforms[:, s]
-        w  = np.sum(e, axis=0) if np.any(e) else None
+        t = times    [   s]
+        e = waveforms[:, s]
+        w = np.sum(e, axis=0).clip(0) if np.any(e) else None
+        if not np.any(w): w = None
```
mmkekic: This is equivalent to a more readable version (at least to me):
```
w = np.sum(e, axis=0).clip(0) #note that we dont have to check if there are non-0s in e  since we will check the value of the sum later
if np.all(w == 0) : w = None
```
Maybe this can be made a bit more elegant so we dont rewrite `w`

mmkekic (commented):
The 2 tests show ill-behaved old version of rebinning algorithm and the need to fix the bugs. The fix ensures that the bin position after rebinning will be physical. This PR still doesn't fix the problem of negative waveforms after the rebinning (this was intended in PR #683) , as I understand this is left to the future PR of @ausonandres  and @msorel.

invisible_cities/reco/peak_functions_test.py
```diff
@@ -597,3 +600,12 @@ def test_rebin_times_and_waveforms_times_are_consistent(t_and_wf, stride):
                                                   np.ones_like(wfs), stride)
 
     assert np.sum(rb_times) * stride == approx(np.sum(times))
+
+
+def test_rebin_times_and_waveforms_negative_bins(ICDATADIR):
+    pmap_file = os.path.join(ICDATADIR, 'pmaps_negative_bins.h5')
```
mmkekic, line 608: maybe add a comment to explain what this test does? something like 'test that rebinning doesnt raise an error even when all the weights are negative". 
If in the future we introduce the masking of the negative signal we can reuse the test file to check if we indeed get rid of negative bins.

mmkekic (approved):
This PR demonstrates and  fixes the bug in weighted rebinning of the waveforms with negative weights. The change is minimal and clear, and the test are relevant for any kind of future refactoring of irene city. Good job!

PR #688, opened 2020-01-09
Upload database with gain values for different SiPM voltages

carmenromo:
In the last month, runs with different SiPMs voltages were taken, so database has been uploaded with the corresponding gains for each voltage.

This plot shows how the gains of the sensors change with the voltage.
gains_vs_voltages


jmbenlloch (approved):
Latest calibration properly added, all tests passed.

PR #690, opened 2020-02-03
Upload db

carmenromo:
This PR includes the latest sensor calibrations from run 7787. Database has been uploaded with the PMT and SiPMs gains and the pdf values.

pmtRunDifferencePlots

sipmRunDifferencePlots_test


andLaing:
Any reason why we seem to have less agreement on the 1pe sigmas than usual?

jmbenlloch (approved):
According to @carmenromo the issue that @andLaing mention has been already sorted out and the database is properly uploaded. Approved!

PR #693, opened 2020-02-25
Update mc readers

andLaing:
Updates the pandas DataFrame based readers
for the MC hits, particles and sensors taking into
account the new nexus output structure.

Adapts existing functions to return the same structure
for both old and new format files.


jmunozv:
With the last "load_mcsensor_response_df functionality extended" commit we get:
- The DB_name and run_number is not mandatory, so new geometries not stored in the DataBase can be handled without the need to pass arbitrary values to these variables.
- A new parameter called sns_name has been added for just the input files with the new hdf5 persistency format. This sns_name, if set, is used by the function to only return the sns_response of that type of sensors.

jmunozv:
Currently load_mcsensor_response_df returns a Tuple with 2 pandas DataFrames: sns_response and bin_widths. This is the only mc-reader with this behavior. The rest of the mc-readers just return what you ask (hits, particles).
Given that the bin_widths are easily got via get_sensor_binning() function provided by IC, my proposal, for consistency, is that this function should return only the sns_response.

andLaing:
With the commit 6ca6062 we remove the old readers which returned dictionaries of evm.event_model objects. @paolafer suggests a tag be made before merging these changes into the head.

mmkekic:
The MC structure table has been bothering me for a while and I am very happy to see it improving! However I think we should all agree on what we want to do and try and keep the code as clean as possible.
Since the MC table structure is changing we need duplicated readers to be able to read different data format during this transition period, is a bit ugly but I dont see a way around it.
Seems that the new readers are always returning pandas dataframe, so the question is do we still want to be able to read the MC information as our custom types (ie dictionaries)? @msorel , @ausonandres , @paolafer , @Aretno  might want to comment on this? If the answer is no, I suggest we make an IC tag and clean the unused types from the event model (and all the functions using those types). If the answer is yes we would need to provide a functions that transform dataframes to needed types.

paolafer:
I think that eliminating completely the event model requires a more thorough discussion, including @jjgomezcadenas. For the MC readers, we agreed on making a tag and then move on eliminating the readers that convert tables into dictionaries.

jacg:
> ... eliminating completely the event model ...

:-)

I wish I had the time to pay attention to this, or even come over and discuss it with you all in person. Unfortunately, I've just unleashed a deluge of activity in my life, so I'm completely out of the picture for a while.

paolafer:
> I wish I had the time to pay attention to this, or even come over and discuss it with you all in person. Unfortunately, I've just unleashed a deluge of activity in my life, so I'm completely out of the picture for a while.

I'm glad you commented :-), I didn't mention you because we know you're out of the picture, but your interventions whenever you find the time are greatly appreciated!

mmkekic:
> I think that eliminating completely the event model requires a more thorough discussion, including @jjgomezcadenas. 

No, I didnt mean to eliminate the whole event model, that doesnt seem feasible at the moment (though note that we are duplicating some functionalities to work with pandas directly when the performance is the issue... ). I meant to declutter it, ie remove types that are simply not used anywhere in IC (as MCParticle would be in this case we eliminate the reader) 

> For the MC readers, we agreed on making a tag and then move on eliminating the readers that convert tables into dictionaries.

I think that we should at least keep the possibility for a user to create MCHit type since it is needed to be fed in the paolina functions for example. My suggestion would be to have a function that transforms hits dataframe to a dictionary {event : List[MCHit]}. I am not familiar if the output of the  other readers (sns responses etc) is used anywhere.

mmkekic:
> > I wish I had the time to pay attention to this, or even come over and discuss it with you all in person. Unfortunately, I've just unleashed a deluge of activity in my life, so I'm completely out of the picture for a while.
> 
> I'm glad you commented :-), I didn't mention you because we know you're out of the picture, but your interventions whenever you find the time are greatly appreciated!

Indeed, your suggestion, even sporadic, are more than welcome! Thanks for still keeping an eye on us! :)

invisible_cities/reco/paolina_functions_test.py
```diff
@@ -239,20 +239,15 @@ def test_voxels_with_no_hits(ICDATADIR):
...
-                break
-
-        hits_dict = load_mchits(hit_file, (evt_line, evt_line+1))
-        voxels = voxelize_hits(hits_dict[evt_number], vox_size, strict_voxel_size=False)
-        for v in voxels:
-            assert sum(h.E for h in v.hits) == v.E
+    hits_df  = load_mchits_df(hit_file)
+    ## Patch to BHits, temporary?
+    evt_hits = hits_df.loc[evt_number][['x', 'y', 'z', 'energy']].values
+    bhit_seq = [BHit(*hit) for hit in evt_hits]
```
mmkekic: I think we should have an official function that casts mchits_df to a dictionary (the output of load_mchits) should be in mcinfo_io

invisible_cities/cities/penthesilea_test.py
```diff
@@ -195,15 +197,15 @@ def test_penthesilea_produces_mcinfo(KrMC_pmaps_filename, KrMC_hdst, config_tmpd
...
+    penthesilea_evts        = load_mchits_df(penthesilea_output_path)
     true_evts               = KrMC_true_hits.hdst
 
-    assert sorted(penthesilea_evts) == sorted(true_evts)
-    for evt_no, true_hits in true_evts.items():
-        penthesilea_hits = penthesilea_evts[evt_no]
-
-        assert len(penthesilea_hits) == len(true_hits)
-        (assert_MChit_equality(p_hit, t_hit) for p_hit, t_hit in zip(penthesilea_hits, true_hits))
+    assert np.all(penthesilea_evts.index.levels[0] == true_evts.index.levels[0])
```
mmkekic: Since you are reading all as dataframes now isnt `assert_dataframes_equal(penthesilea_evts, KrMC_true_hits.hdst)` enough?
Also maybe use the `assert_dataframes_equal` from our testing utils?

andLaing: Is there any reason to use our own assert_dataframes_equal and not `pandas.testing.assert_frame_equal`?

mmkekic: Maybe to be consistent with (mostly?) all other tests?  But also it might be easier to track in case something changes in future pandas/numpy version (ie, the testing used to be in pandas.utils.testing and now is pandas.testing...)
Also shouldn't assert close be more appropriate since those are floats?

andLaing:
While adjusting the cities to use the new paradigm for the copying of MC info I've come across a problem in a few tests that I'm not sure the best way to solve. The file used (electrons_40keV_z250_RWF.h5) seems to have an error in the MC tables as can be seen in the attached screenshot. The extents row for the last event gives the last hit and last particle one less than the total number of rows in the corresponding table. This causes an error when trying to read either of these tables into a coherent dataframe since the last value of the event_id column is a nan. I could probably catch the exception caused and replace the nan with the last valid event_id but I was worried that this would cover up other corrupt files. It's not relevant for new MC but I thought I'd better ask for opinions before I start rewriting functions or changing test files
Screenshot 2020-03-28 at 19 31 18

jmunozv:
Andrew, may be the error comes from the event number one (the one with id = 1) ... according to the plot you have sent it wouldn't have any particle

andLaing:
Nope it's definitely the last one, there are 50 hit rows and the last index is 48 not 49 as it needs to be

paolafer:
Hi! The `electrons_40keV_z250_RWF.h5` file has clearly an error in it, since, as Javi says, in the `extents` table you can see that event 0 and event 1 have the same `last_particle` value. I've checked that the MCRD correspondent file has the correct MC true information and that, if I run diomira on it, a correct table in the RWF file is produced, therefore, the IC code seems to be fine. For some reason, the `electrons_40keV_z250_RWF.h5` file was produced in a wrong way.

I would propose to use another RWF file in the tests that use it, we have some of them already in the `test_data` folder, if those tests don't assume anything in particular for the content of the file.

jmunozv:
That was exactly my point ... The event with id=1 has 4 hits and 0 particles what has no sense at all.

invisible_cities/io/mcinfo_io.py
```diff
@@ -171,13 +175,73 @@ def __call__(self, mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
         self.generator_table.flush()
 
 
-def copy_mc_info(h5in : tb.File, writer : Type[mc_info_writer], which_events : List[int]=None):
-    """Copy from the input file to the output file the MC info of certain events.
+def save_mc_table(file_in : str                         ,
```
mmkekic: This seems rather complex, aren't we supposed to save all the output in a unique (new) MC format? ie, in copy_mc_info we read all relevant data (using our universal for both old and new readers) and just write the relevant tables? 
I would prefer to have output files with a unique structure even if some of the data was not previously saved in old MC format and we have to fill it with NaNs

invisible_cities/io/mcinfo_io.py
```diff
@@ -186,26 +250,55 @@ def copy_mc_info(h5in : tb.File, writer : Type[mc_info_writer], which_events : L
...
+        def _try_unique_evt_itr(group, itr):
+            for elem in itr:
+                try:
+                    yield np.unique(getattr(group, elem).cols.event_id)
+                except tb.exceptions.NoSuchNodeError:
+                    pass
+
+        with tb.open_file(file_name, 'r') as h5in:
+            mc_tbls  = ['hits', 'particles', 'sns_response']
+            evt_list = list(_try_unique_evt_itr(h5in.root.MC, mc_tbls))
```
mmkekic: do we understand why are there events that are not in particles or hits table?

invisible_cities/io/mcinfo_io.py
```diff
@@ -330,32 +456,68 @@ def read_mcparticles_df(h5in    : tb.file.File,
...
-    sipm_bin = float(sipm_bin.split()[0]) * units_dict[sipm_bin.split()[1]]
+    if is_oldformat_file(file_name):
+        return sensor_binning_old(file_name)
+    else:
+        return sensor_binning_new(file_name)
+
+
+def sensor_binning_old(file_name : str) -> pd.DataFrame:
+    # Is a pre-Feb 2020 MC file
+    config = pd.read_hdf(file_name, 'MC/configuration').set_index('param_key')
```
mmkekic: I think is better to use `dst_io.load_dst` ( that opens file using pytables) since we already saw that in different pandas version this might not work, depending on how file was created (ie pytables or pandas writer)

invisible_cities/io/mcinfo_io.py
```diff
@@ -330,32 +456,68 @@ def read_mcparticles_df(h5in    : tb.file.File,
...
+    bins         = config[config.index.str.contains('binning')].copy()
```
mmkekic: since config was read in a previous line and not given as a mutable input, you don't have to copy it

invisible_cities/io/mcinfo_io.py
```diff
@@ -330,32 +456,68 @@ def read_mcparticles_df(h5in    : tb.file.File,
...
+    bins         = config.loc[sns_binnings].copy()
+    bins.columns = ['bin_width']
+    bins.index   = bins.index.rename('sns_name')
+    bins.index   = bins.index.str.strip('_binning')
+    bins.bin_width = bins.bin_width.str.split(expand=True).apply(
+        lambda x: float(x[0]) * getattr(units, x[1]), axis=1)
+    return bins
+
+
+def get_sensor_types(file_name : str) -> pd.DataFrame:
```
mmkekic, line 667: I am confused between get, load and read names, can't we make a more systematic naming?

invisible_cities/io/mcinfo_io.py
andLaing: It does output a unique format that's the same for all but you have to take into account that the input isn't always the same and output it in a readable format

invisible_cities/io/mcinfo_io.py
andLaing: This is a question for the NEXUS authors really.
There are certain types of simulations that have some and not other MC types saved, a simulation of optical photons for example would have an empty hits table as far as I understand.

invisible_cities/io/mcinfo_io.py
andLaing: Do you have an example of a fail? I've never seen it fail in this context due to pandas versions.

mmkekic: The beersheba_tests fail due to pandas not being able to read PSF file in version 0.25 (we use 0.23). This is an issue that needs to be solved but obviously not related to this PR

invisible_cities/io/mcinfo_io.py
paolafer: There are different kinds of simulations, where we save different kinds of output (some will save only sensor response information and not true particles, sometimes the opposite happen, for instance). We want to be able to read all kind of MC files.

paolafer: An option would be to resurrect the old `events` table, which contained two columns, namely the eventID and the total deposited true energy, but I don't know if it's worth it...

andLaing:
@gonzaponte , question related to PR #702 . With the changes I've made to the writer it's not so natural to have the `IndexError` raised in `mcinfo_io.copy_mc_info` as it was in the past. Would you prefer that I add a test there to raise in case of no overlap or should I add it to `components.copy_mc_info` where the overlap is already checked?

gonzaponte:
If I understand correctly the code, `read_mc_tables` takes care of reading the relevant event numbers which are then sent to the writer. So there is no possibility of raising an `IndexError` in the writing process. For me, `read_mc_tables` should raise the one raising the exception if the requested event number is not in the file. In that case, both `copy_mc_info` can forget about performing that check.

andLaing:
Would you remove the full overlap check in `components.copy_mc_info` then?
```
try:
            event_numbers_in_file = mcinfo_io.get_event_numbers_in_file(f)
            event_numbers_to_copy = list(evt for evt in event_numbers \
                                         if evt in event_numbers_in_file)
            mcinfo_io.copy_mc_info(f, writer, event_numbers_to_copy,
                                   db_file, run_number)
            continue
```
It would be a natural place to do the check and we'd have to call the same code again to do it elsehwere

gonzaponte:
Sorry, I was confused. `components.copy_mc_info` still has to do `event_numbers_to_copy = list(evt for evt in event_numbers if evt in event_numbers_in_file)` to limit the full set of selected events to those present in a given file. That is to be kept. Independently of that, `read_mc_tables` has to ensure that the event number is trying to read exists (right now it would just provide empty dataframes, is this the desired behavior?). `mcinfo_io.copy_mc_info` can forget about the overlap check.

invisible_cities/io/mcinfo_io.py
```diff
@@ -172,13 +181,97 @@ def __call__(self, mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
...
+
+
+def check_last_merge_index(h5out : tb.file.File) -> int:
+    """
+    Get the last file index used to index merged files
+    """
+    if 'MC' in h5out.root:
+        if 'event_mapping' in h5out.root.MC:
+            return h5out.root.MC.event_mapping.cols.file_index[-1]
+    return 0
```
mmkekic: I think here you should return -1 in case there is no table?

invisible_cities/io/mcinfo_io.py
```diff
@@ -172,13 +181,97 @@ def __call__(self, mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
...
+
+
+def read_mc_tables(file_in : str                 ,
+                   evt_arr : np.ndarray          ,
+                   db_file : Optional[str] = None,
+                   run_no  : Optional[int] = None) -> Dict:
+    """
+    Reads all the MC tables present in
+    file_in and sorts the Dataframes for
```
mmkekic: stores rather than sorts? also add arguments and return explanation in docstrings

invisible_cities/io/mcinfo_io.py
```diff
@@ -172,13 +181,97 @@ def __call__(self, mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
...
+    Get the last file index used to index merged files
+    """
+    if 'MC' in h5out.root:
+        if 'event_mapping' in h5out.root.MC:
+            return h5out.root.MC.event_mapping.cols.file_index[-1]
+    return 0
+
+
+def read_mc_tables(file_in : str                 ,
+                   evt_arr : np.ndarray          ,
```
mmkekic: the evt_arr should be optional, if not given the whole table should be read

invisible_cities/io/mcinfo_io.py
```diff
@@ -172,13 +181,97 @@ def __call__(self, mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
...
+                   db_file : Optional[str] = None,
+                   run_no  : Optional[int] = None) -> Dict:
+    """
+    Reads all the MC tables present in
+    file_in and sorts the Dataframes for
+    each into a dictionary.
+    """
+    mctbls = get_mc_tbl_list(file_in)
+    if (len(evt_arr) > 0 and
+        len(np.intersect1d(evt_arr, get_event_numbers_in_file(file_in))) == 0):
```
mmkekic: I dont understand this check, why do you want to raise the error only if none of the required events are contained in the file but allow empty selection list? I think more appropriate  would be to raise either Error or Warning when some of the events are not present.

invisible_cities/io/mcinfo_io.py
```diff
@@ -275,88 +443,134 @@ def load_mcparticles_df(file_name: str) -> pd.DataFrame:
...
+                              'final_y'           : p_tb.col('final_vertex')[:, 1],
+                              'final_z'           : p_tb.col('final_vertex')[:, 2],
+                              'final_t'           : p_tb.col('final_vertex')[:, 3],
+                              'initial_volume'    : p_tb.col('initial_volume').astype('U20'),
+                              'final_volume'      : p_tb.col('final_volume').astype('U20'),
+                              'initial_momentum_x': p_tb.col('momentum')[:, 0],
+                              'initial_momentum_y': p_tb.col('momentum')[:, 1],
+                              'initial_momentum_z': p_tb.col('momentum')[:, 2],
+                              'kin_energy'        : p_tb.col('kin_energy'),
+                              'creator_proc'      : p_tb.col('creator_proc').astype('U20')})
```
mmkekic, line 606: I realized that in new format there are new columns:
`'length', 'final_momentum_z', 'final_momentum_x', 'final_momentum_y', 'final_proc'` 
(please @paolafer , @jmunozv , @jmalbos  confirm if this is new standard)
which means that we would never be able to mix the new and old format, ie the writer will not work, are we happy with that?

invisible_cities/io/mcinfo_io.py
```diff
@@ -275,88 +443,134 @@ def load_mcparticles_df(file_name: str) -> pd.DataFrame:
...
+
+        # Adding event info
+        evt_part_df = extents[['last_particle', 'evt_number']]
+        evt_part_df.set_index('last_particle', inplace = True)
+        parts = parts.merge(evt_part_df         ,
+                            left_index  =   True,
+                            right_index =   True,
+                            how         = 'left')
+        parts.rename(columns={"evt_number": "event_id"}, inplace = True)
```
mmkekic: no spaces around '=' in oneliners

invisible_cities/io/mcinfo_io.py
```diff
@@ -187,26 +280,93 @@ def copy_mc_info(h5in : tb.File, writer : Type[mc_info_writer], which_events : L
...
+    def try_unique_evt_itr(group, itr):
+        for elem in itr:
+            try:
+                yield np.unique(getattr(group, elem).cols.event_id)
+            except tb.exceptions.NoSuchNodeError:
+                pass
+
+    evt_list = list(try_unique_evt_itr(h5in.root.MC, mc_tbls))
+    if len(evt_list) == 0:
+        raise tb.exceptions.NoSuchNodeError
```
mmkekic: this is a bit confusing, I am not sure that this function should raise an error if there are no tables with events or simply return an empty list?
In any case it is more clear to first get the name of the present tables and iterate over them extending the unique numpy array with `setdiff1d`or so

invisible_cities/io/mcinfo_io.py
```diff
@@ -172,13 +181,97 @@ def __call__(self, mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
...
+            config = load_mcconfiguration(file_in)
+            tbl_dict[tbl] = config
+        elif tbl is MCTableType.events               :
+            pass
+        elif tbl is MCTableType.event_mapping        :
+            evt_map = load_mcevent_mapping(file_in)
+            tbl_dict[tbl] = evt_map[evt_map.event_id.isin(evt_arr)]
+        else                                         :
+            raise TypeError("MC table has no reader")
+    return tbl_dict
```
mmkekic, line 179: maybe we should also return a list of events?

invisible_cities/io/mcinfo_io.py
```diff
@@ -172,13 +181,97 @@ def __call__(self, mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
...
+
+
+def copy_mc_info(file_in      : str                       ,
+                 writer       : Type[mc_writer]           ,
+                 which_events : Optional[List[int]] = None,
+                 db_file      : Optional[str]       = None,
+                 run_number   : Optional[int]       = None) -> None:
+    """
+    Copy from the input file to the output file the MC info of certain events.
```
mmkekic, line 188: description of `db_file` and `run_number`

invisible_cities/io/mcinfo_io.py
```diff
@@ -187,26 +280,93 @@ def copy_mc_info(h5in : tb.File, writer : Type[mc_info_writer], which_events : L
...
+
+
+def get_mc_tbl_list(file_name: str) -> List[MCTableType]:
+    """
+    Returns a list of the tables in
+    the MC group of a given file
+    """
+    with tb.open_file(file_name, 'r') as h5in:
+        mc_group = h5in.root.MC
+        return [MCTableType[tbl.name] for tbl in mc_group]
```
mmkekic, line 256: what happens if there is new table name in the file? Guess enum will complain, are we happy with the error?

invisible_cities/io/mcinfo_io.py
```diff
@@ -187,26 +280,93 @@ def copy_mc_info(h5in : tb.File, writer : Type[mc_info_writer], which_events : L
...
+    if len(evt_list) == 0:
+        raise tb.exceptions.NoSuchNodeError
+    return np.unique(np.concatenate(evt_list)).astype(int)
+
+
+def load_mcconfiguration(file_name : str) -> pd.DataFrame:
+    config           = load_dst(file_name, 'MC', 'configuration')
+    if is_oldformat_file(file_name):
+        binning_keys = config.param_key.str.contains('binning')
+        new_names    = config[binning_keys].param_key.str.split(
```
mmkekic: this line is a bit hard to read, maybe a comment explains why and what we do?

invisible_cities/io/mcinfo_io.py
```diff
@@ -187,26 +280,93 @@ def copy_mc_info(h5in : tb.File, writer : Type[mc_info_writer], which_events : L
...
+        config.loc[binning_keys, 'param_key'] = new_names
+    return config
+
+
+def load_mcsensor_positions(file_name : str,
+                            db_file   : Optional[str] = None,
+                            run_no    : Optional[int] = None) -> pd.DataFrame:
+    if is_oldformat_file(file_name):
+        sns_pos = load_dst(file_name, 'MC', 'sensor_positions')
+        if sns_pos.shape[0] > 0:
```
mmkekic, line 352: if I understand this, the 'sensor_position' is *always* present in the old format file, but it might be empty? And if it is present the `db_file`and `run_no` are not optional and the code will crash. Do we need a protection against that?

invisible_cities/io/mcinfo_io.py
```diff
@@ -187,26 +280,93 @@ def copy_mc_info(h5in : tb.File, writer : Type[mc_info_writer], which_events : L
...
+            sns_names = get_sensor_binning(file_name).index
+            pmt_ids   = DB.DataPMT(db_file, run_no).SensorID
+            pmt_name  = sns_names.str.contains('Pmt')
```
mmkekic, line 361: Again a comment of what and why is happening would be useful, ie the names are of pmts but no sipms names in sns_names ?

invisible_cities/io/mcinfo_io.py
```diff
@@ -275,88 +443,134 @@ def load_mcparticles_df(file_name: str) -> pd.DataFrame:
...
+        lambda x: float(x[0]) * getattr(units, x[1]), axis=1)
+    return bins
+
+
+def sensor_binning_new(file_name : str) -> pd.DataFrame:
+    config         = load_dst(file_name, 'MC',
+                              'configuration').set_index('param_key')
+    sns_names      = load_dst(file_name, 'MC', 'sns_positions').sensor_name
+    if len(sns_names) == 0:
+        return pd.DataFrame(index=['sns_name'], columns=['bin_width'])
```
mmkekic: There can be a file with empty 'sns_position' ? In any case you can read config after you check for the condition

invisible_cities/io/mcinfo_io.py
```diff
@@ -275,88 +443,134 @@ def load_mcparticles_df(file_name: str) -> pd.DataFrame:
...
+    sns_binnings   = [nm + '_binning' for nm in sns_names.unique()]
```
mmkekic: Either a comment or in-detail docstring would be useful

invisible_cities/io/mcinfo_io.py
```diff
@@ -275,88 +443,134 @@ def load_mcparticles_df(file_name: str) -> pd.DataFrame:
...
+    bins.drop('file_index', axis=1, inplace=True, errors='ignore')
+    bins.columns   = ['bin_width']
+    bins.index     = bins.index.rename('sns_name')
+    bins.index     = bins.index.str.strip('_binning')
+    bins.bin_width = bins.bin_width.str.split(expand=True).apply(
+        lambda x: float(x[0]) * getattr(units, x[1]), axis=1)
+    return bins[~bins.index.duplicated()]
+
+
+def get_sensor_types(file_name : str) -> pd.DataFrame:
```
mmkekic, line 667: is this even used any more?

invisible_cities/io/mcinfo_io.py
```diff
@@ -367,31 +581,43 @@ def load_mcsensor_response_df(file_name : str,
...
+                                       get_sensor_binning     (file_name),
+                                       load_mcsensor_positions(file_name,
+                                                               db_file  ,
+                                                               run_no   ))
+    else:
+        sns_resp = load_dst(file_name, 'MC', 'sns_response')
+        if return_raw:
+            return sns_resp
+        return convert_timebin_to_time(sns_resp                          ,
+                                       get_sensor_binning     (file_name),
```
mmkekic, line 738: I assume we can not have 'sns_response' without 'sns_position' table? But probably we do need a protection for that?

invisible_cities/cities/components.py
```diff
@@ -168,20 +174,21 @@ def copy_mc_info(files_in : List[str], h5out: tb.File, event_numbers: List[int])
...
-                mcinfo_io.copy_mc_info(h5in, writer, event_numbers_to_copy)
-            except tb.exceptions.NoSuchNodeError:
-                continue
+        try:
+            event_numbers_in_file = mcinfo_io.get_event_numbers_in_file(f)
+            event_numbers_to_copy = list(evt for evt in event_numbers \
+                                         if evt in event_numbers_in_file)
+            mcinfo_io.copy_mc_info(f, writer, event_numbers_to_copy,
+                                   db_file, run_number)
+        except tb.exceptions.NoSuchNodeError:
```
mmkekic: This is confusing, when do we expect NoSuchNodeError? Does this mean we skip a file that does not have MC information? This doesnt seem right... Also there is no protection that all `event_numbers` are eventually contained in the output file

mmkekic (commented):
All the functions should have more detailed docstrings, especially what type of information is contained in the loaders outputs.
There are bunch of tests still missing and it would be good to add some standardized test files with the same nexus event in both old and new format, for all combinations (full, up to hits, optical photons etc...) , @jmalbos  promised to help with that :)

invisible_cities/io/mcinfo_io.py
paolafer: Exactly, 'sns_response' is only filled with the positions of the sensors that have detected some photoelectron, at the moment. It's not a bad idea to add a protection, though, because we may want to save that table with all the sensor positions, in the future.

invisible_cities/io/mcinfo_io.py
paolafer: Yes, confirmed. I think that we are ok with not mixing new and old format.

invisible_cities/io/mcinfo_io.py
paolafer: In any case, all the tables are present in the file, always. They may be empty, though.

invisible_cities/io/mcinfo_io.py
mmkekic: If we want to make interface blind to the file format maybe would be a better idea to fill the missing columns with nans? Cause someone doing analysis with old and new files would still have to have 2 versions of the code to work with each type, and we wanted to avoid that?

jmunozv:
I still keep thinking that having that storing in this table the positions of all the sensors wouldn't make its size much bigger, and it would make our lives much easier. 

Javi. 

Javier Munoz Vidal
javier.munoz.vidal@gmail.com

invisible_cities/io/mcinfo_io.py
andLaing: This was to reproduce the behaviour of that was present from PR #702 and copy the tables without event number as a variable in case of 0 events passing the cuts in the city being launched

invisible_cities/io/mcinfo_io.py
andLaing: To what end?

We send an event list in most user cases and if not it will give you the information for all events and the complete event list can be got by calling the get_event_numbers_in_file function

invisible_cities/io/mcinfo_io.py
andLaing: I think so. All possible tables are there at the moment and have the same name as they do in file so if there's a change the Enum will give a readable error and the user will know that the code needs to be updated to allow for that table.

invisible_cities/io/mcinfo_io.py
andLaing: The exception was to maintain the expected try except structure approved for PR #702 but if you prefer I can look for ways to improve it?

We don't have access to the file name here to use the table list function and I'd prefer not to look at all tables in any case since many don't have an event_id column.

As to the extending, it seems like it'd need more gymnastics to do it that way since we'd have to initialise, right?

invisible_cities/io/mcinfo_io.py
andLaing: I think the exception should be raised if the database info isn't sent with an old format file. Maybe I should change the ordering so it's raised even if the sensor_positions table is empty?

invisible_cities/io/mcinfo_io.py
andLaing: It was a new function that @jmunozv requested. If you think it's better we can remove it from this PR and get him to make a separate one adding the function.

invisible_cities/io/mcinfo_io.py
andLaing: Whether we have all the positions saved or just the ones that are hit doesn't affect this function since we only use the names of the detectors, not the positions etc.

In case of an empty sns_positions table there might be a fail path but I'm not 100% sure what it would be. I'll have a look to see if I can understand it a bit better

invisible_cities/io/mcinfo_io.py
mmkekic: Its ok to be here, but in case this is only valid for new format raise Exception if the old format is passed

invisible_cities/io/mcinfo_io.py
mmkekic: The error is :
`TypeError: getattr(): attribute name must be string`
so I believe a more readable exception is in place.
Also I wonder if maybe we should throw a warning and return an empty DataFrame?

invisible_cities/io/mcinfo_io.py
mmkekic: Seems that `get_sensor_binning` gives name 'SiPMSensl' while in the new format is 'SiPM', should we change that?

invisible_cities/io/mcinfo_io.py
```diff
@@ -313,6 +313,47 @@ def load_mcconfiguration(file_name : str) -> pd.DataFrame:
...
+                DataFrame containing information about
+                the positions and types of sensors in
+                the MC simulation.
+    """
+    if is_oldformat_file(file_name):
+        sns_pos = load_dst(file_name, 'MC', 'sensor_positions')
+        if sns_pos.shape[0] > 0:
+            ## Add a column to the DataFrame so all info
+            ## is present like in the new format
+            sns_names = get_sensor_binning(file_name).index
```
mmkekic: This seems like you need a function that just extract sensor_names, for both old and new format from config table.This is also related to `sensor_binning_new` where you need to read 'sensor_positions' to get the sensor names.

invisible_cities/io/mcinfo_io.py
```diff
@@ -321,11 +321,22 @@ def load_mcconfiguration(file_name : str) -> pd.DataFrame:
     config           = load_dst(file_name, 'MC', 'configuration')
     if is_oldformat_file(file_name):
         binning_keys = config.param_key.str.contains('binning')
+        def binning_name(x):
```
mmkekic: how about something like this ?

```
config           = load_dst(file_name, 'MC', 'configuration')
config.param_key = config.param_key.str.replace('.*Pmt.*\_binning.*', 'Pmt_binning')
config.param_key = config.param_key.str.replace('.*SiPM.*\_binning.*', 'SiPM_binning')
config           = config.drop_duplicates('param_key').reset_index(drop=True)
```

mmkekic: Also, we should check if waveforms are present and non-empty, otherwise we should remove any `_binning` parameters from configuration

invisible_cities/io/mcinfo_io.py
```diff
@@ -655,72 +655,8 @@ def get_sensor_binning(file_name : str) -> pd.DataFrame:
...
-    if len(sns_names) == 0:
-        return pd.DataFrame(index=['sns_name'], columns=['bin_width'])
-    config         = load_dst(file_name, 'MC',
-                              'configuration').set_index('param_key')
-    ## In the configuration table the binning is named
-    ## <sensor type>_binning, adjust to search.
-    sns_binnings   = [nm + '_binning' for nm in sns_names.unique()]
-    bins           = config.loc[sns_binnings].copy()
+    config         = load_mcconfiguration(file_name).set_index('param_key')
+    bins           = config[config.index.str.contains('binning')].copy()
```
mmkekic: in case config does not contain binning information (ie bins is empty dataframe) I suggest we return empty df and a warning that the information is not present?

invisible_cities/io/mcinfo_io.py
```diff
@@ -321,29 +321,42 @@ def read_mc_tables(file_in : str                        ,
...
+        if MCTableType.event_mapping not in tbl_dict.keys():
+            new_key = MCTableType.event_mapping
+            tbl_dict[new_key] = pd.DataFrame(which_events, columns=['event_id'])
+        writer(tbl_dict)
     except tb.exceptions.NoSuchNodeError:
-        raise tb.exceptions.NoSuchNodeError(f"No MC info in file {h5in}.")
+        raise tb.exceptions.NoSuchNodeError(f"No MC info in file {file_in}.")
     except IndexError:
-        raise IndexError(f"No event {n} in file {h5in}.")
+        raise IndexError(f"No events {which_events} in file {file_in}.")
```
mmkekic: Remove `IndexError`, that was needed for the previous mc writer to work. Also remove the impossible `NoParticleInfoFile` exception
The `NoSuchNodeError` can happen only in `get_event_numbers_in_file` if none  of the tables are present. Id say remove that too and give string explanation in `get_event_numbers_in_file`as 'at least one of ... tables has to be present in the file'

invisible_cities/io/mcinfo_io.py
mmkekic: Remove this line, it was needed in PR #702 but note that previous writer worked on per-event basis. In this writer it really makes no sense to keep any check of this style

invisible_cities/cities/components.py
mmkekic: I believe that the error can be raised only if the file didnt contain some essential table in which case we do want user to react on that, and not to sneaky catch the exception and continue the work.

invisible_cities/io/mcinfo_io.py
mmkekic: In case there are tables in 'MC' group that are outside the enum class the `KeyError`will be raised. Maybe it needs a bit more explanation?

invisible_cities/cities/components.py
mmkekic: On the other hand we never check that all of the `event_numbers` are actually copied, guess we could append `event_numbers_to_copy` in a list and at the end compare that its the same as the `event_numbers`list?

invisible_cities/cities/components.py
```diff
@@ -47,6 +48,7 @@
 from .. io     .pmaps_io          import                load_pmaps
 from .. io     .hits_io           import              hits_from_df
 from .. io     .dst_io            import                  load_dst
+from .. io     .hits_io           import                 load_hits
```
mmkekic: why? XD

invisible_cities/cities/components.py
```diff
@@ -168,20 +175,23 @@ def copy_mc_info(files_in : List[str], h5out: tb.File, event_numbers: List[int])
...
+        if mcinfo_io.check_mc_present(f):
+            event_numbers_in_file = mcinfo_io.get_event_numbers_in_file(f)
+            event_numbers_to_copy = list(evt for evt in event_numbers \
+                                         if evt in event_numbers_in_file)
+            mcinfo_io.copy_mc_info(f, writer, event_numbers_to_copy,
+                                   db_file, run_number)
+        else:
+            warnings.warn(f' File does not contain MC tables.\
+             Use positve run numbers for data', UserWarning)
+            continue
```
mmkekic, line 203: I think it would be good at the end of file loop to check that all the events are actually found in the files; basically collect all `event_numbers_to_copy` and make sure that it is the same as `event_numbers` lits

invisible_cities/cities/esmeralda_test.py
```diff
@@ -90,7 +90,7 @@ def test_esmeralda_filters_events(KrMC_hdst_filename_toy, correction_map_MC_file
     with tb.open_file(PATH_OUT)  as h5out:
         event_info = get_event_info(h5out)
         assert length_of(event_info) == nevt_req
-        MC_num_evs = h5out.root.MC.extents[:]['evt_number']
+        MC_num_evs = load_mchits_df(PATH_OUT).index.levels[0]
```
mmkekic: shouldnt it use `get_event_numbers_in_file`?

invisible_cities/cities/penthesilea_test.py
```diff
@@ -1,9 +1,12 @@
 import os
 import numpy  as np
 import tables as tb
+import pandas as pd
 
 from pytest import mark
 
+from pandas.testing import assert_frame_equal
```
mmkekic: we should switch to the `assert_dataframes_close`

invisible_cities/cities/penthesilea_test.py
```diff
@@ -221,28 +219,41 @@ def test_penthesilea_read_multiple_files(ICDATADIR, output_tmpdir):
...
+    hits_in2      = load_mchits_df     (second_file)
+    particles_out = load_mcparticles_df(   file_out)
+    hits_out      = load_mchits_df     (   file_out)
+
+    # back to front beacause that's how the events are.
+    evt_in  = np.concatenate([hits_in2.index.levels[0],
+                              hits_in1.index.levels[0]])
+    first_event_out = 4
+    evt_out = hits_out.index.levels[0]
+    print(evt_in, evt_out)
```
mmkekic: forgotten print

invisible_cities/cities/penthesilea_test.py
```diff
@@ -221,28 +219,41 @@ def test_penthesilea_read_multiple_files(ICDATADIR, output_tmpdir):
...
+    assert all(evt_in[first_event_out:] == evt_out)
+
+    all_hit_in      = pd.concat([hits_in1     ,      hits_in2])
+    pd.testing.assert_frame_equal(all_hit_in.loc[evt_in[first_event_out:]],
```
mmkekic: again - switch to our assert function

invisible_cities/cities/penthesilea_test.py
```diff
@@ -221,28 +219,41 @@ def test_penthesilea_read_multiple_files(ICDATADIR, output_tmpdir):
...
+                                  hits_out                                )
+    all_particle_in = pd.concat([particles_in1, particles_in2])
+    pd.testing.assert_frame_equal(all_particle_in.loc[evt_in[first_event_out:]],
+                                  particles_out                                )
```
mmkekic: This test is very weird anyway, as I understand we want to check that all the events passed are in output file? @gonzaponte you were the original author of this test?

If I am correct isnt it better to do something like this:
```
input_files = glob(file_in)
events_passed = [14004, 14005, 14006, 14007, 14008, 14009]
all_inp_hits      = pd.concat([load_mchits_df(file_) for file_ in input_files], axis=0)
all_inp_particles = pd.concat([load_mcparticles_df(file_) for file_ in input_files], axis=0)
all_out_hits      = load_mchits_df(file_out)
all_out_particles = load_mcparticles_df(file_out)
assert_frame_equal(all_inp_hits.loc[events_passed, :], all_out_hits)
assert_frame_equal(all_inp_particles.loc[events_passed, :], all_out_particles)
```

ie, instead of hard coding the number of first events from first file that didnt pass filter just hardcode the event numbers that we expect in the output? and later check that the mcinfo is the same?

invisible_cities/io/mcinfo_io.py
```diff
@@ -1,397 +1,766 @@
...
+            if (key is MCTableType.configuration or
+                key is MCTableType.event_mapping   ):
+                try:
+                    orig_indx = tbl.file_index.unique()
+                    new_indx  = np.arange(first_indx                 ,
+                                          first_indx + len(orig_indx))
+                    tbl.file_index.replace(orig_indx, new_indx, inplace=True)
+                except AttributeError:
+                    tbl['file_index'] = first_indx
+            mcwriter_(df=tbl, table_name=key.name)
```
mmkekic: you should specify :
`columns_to_index` - it should be event for most of the relevant tables (hits, particles etc)
`str_col_length` - seems some of the strings are longer than 32 characters, can you identify for which table and then increase the number?

invisible_cities/cities/irene_test.py
```diff
@@ -171,6 +176,7 @@ def test_irene_output_file_structure(config_tmpdir):
         assert "runInfo"      in h5out.root.Run
 
 
+@mark.xfail
```
mmkekic: Instead of skipping can we just set `run_number`to something else and fake 'data' so we dont try to copy MC?

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -206,54 +89,187 @@ def test_copy_mc_info_which_events_out_of_range(ICDATADIR, config_tmpdir):
...
-def test_load_mchits_correct_number_of_hits(mc_all_hits_data):
-    efile, number_of_hits, evt_number = mc_all_hits_data
-    mchits_dict = load_mchits(efile)
+        writer = mc_writer(h5out)
+        copy_mc_info(file_in, writer, which_events      ,
+                     db_file = 'new', run_number = -6400)
+
+    ## Check that we can read the output but it's empty
+    hits = load_mchits_df(file_out)
+    assert hits.shape == (0, 6)
```
mmkekic: maybe hits.empty?

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -206,54 +89,187 @@ def test_copy_mc_info_which_events_out_of_range(ICDATADIR, config_tmpdir):
...
+    with tb.open_file(file_out_old, 'w') as h5out:
+        writer = mc_writer(h5out)
+        copy_mc_info(file_in_old, writer, db_file='new', run_number=-6400)
+    with tb.open_file(file_out_new, 'w') as h5out:
+        writer = mc_writer(h5out)
+        copy_mc_info(file_in_new, writer)
+
+    old_tbls = read_mc_tables(file_out_old)
+    new_tbls = read_mc_tables(file_out_new)
+    assert len(old_tbls) == len(new_tbls)
```
mmkekic: why just length and not by element comparison?

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -206,54 +89,187 @@ def test_copy_mc_info_which_events_out_of_range(ICDATADIR, config_tmpdir):
...
+
+
+@mark.parametrize("fn_oldformat fn_newformat".split(),
+                  (("nexus_scint.oldformat.sim.h5"         ,
+                    "nexus_scint.newformat.sim.h5"         ),
+                   ("nexus_new_kr83m_fast.oldformat.sim.h5",
+                    "nexus_new_kr83m_fast.newformat.sim.h5"),
+                   ("nexus_new_kr83m_full.oldformat.sim.h5",
+                    "nexus_new_kr83m_full.newformat.sim.h5")))
+def test_copy_mc_info_same_results(ICDATADIR   , config_tmpdir,
```
mmkekic: Would be good to split this test in 2, one is just `read_mc_tables`and other copy+read. Also give more descriptive name, ie 'test_copy_mc_info_same_results_old_and_new_format' or so, the idea is to understand what the test does from the test name solely

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -206,54 +89,187 @@ def test_copy_mc_info_which_events_out_of_range(ICDATADIR, config_tmpdir):
...
+        else:
+            ## Check dummy columns in old style (mcparticles)
+            old_notnan = old_tbls[key].notna().all()
+            assert new_tbls[key].notna().all().sum() == new_tbls[key].shape[1]
+            assert_dataframes_equal(new_tbls[key].loc[:, old_notnan],
+                                    old_tbls[key].loc[:, old_notnan],
+                                    check_types=False)
+
+
+def test_read_mc_tables(mc_particle_and_hits_nexus_data_new):
```
mmkekic: again change name, this test checks that the tables shape are the same, so more of 'test_read_mc_tables_correct_shape'

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -206,54 +89,187 @@ def test_copy_mc_info_which_events_out_of_range(ICDATADIR, config_tmpdir):
...
+        for tbl in h5in.root.MC:
+            shape_dict[MCTableType[tbl.name]] = (tbl.shape[0], len(tbl.cols))
+
+    tbl_dict    = read_mc_tables(file_in, all_evt)
+    for tbl, shp in shape_dict.items():
+        assert tbl in tbl_dict.keys()
+        assert shp == tbl_dict[tbl].shape
+
+
+def test_mc_writer(mc_particle_and_hits_nexus_data_new, config_tmpdir):
```
mmkekic: 'test_mc_writer_contains_all_tables' (or any more correct version).
Also a good test would be to read, save and read back checking that the read tables are the same. As well as copy_mc_tables comparing that read+write = copy

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -206,54 +89,187 @@ def test_copy_mc_info_which_events_out_of_range(ICDATADIR, config_tmpdir):
...
+        mcwrite(tbl_dict)
 
-    assert len(mchits_dict[evt_number]) == number_of_hits
+    saved_tbls = get_mc_tbl_list(file_out)
+    assert np.all(input_tbls == saved_tbls)
 
 
-def test_load_mchits(mc_particle_and_hits_nexus_data):
-    efile, _, _, _, _, _, _, X, Y, Z, E, t = mc_particle_and_hits_nexus_data
+def test_mc_writer_oldformat(mc_sensors_nexus_data, config_tmpdir):
```
mmkekic: again rename to be more explicit

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -206,54 +89,187 @@ def test_copy_mc_info_which_events_out_of_range(ICDATADIR, config_tmpdir):
...
+    k_eng = 1.0
+    vi    = [-396.4495 , -351.88654,  318.9442 , 0.0      ]
+    vf    = [-389.12915, -368.3956 ,  304.85535, 0.1059136]
+    p     = [ 0.7167202, 0.3602375 , -1.174112]
+
+    return efile, name, vi, vf, p, k_eng, X, Y, Z, E, t
 
 
-def test_read_mchits_df(mc_particle_and_hits_nexus_data):
+def test_load_mchits_df(mc_particle_and_hits_nexus_data):
```
mmkekic: Either rename the test (`test_load_mchits_df_oldfile`) or do the trick with indirect fixtures.

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -264,62 +280,45 @@ def test_load_mchits_df(mc_particle_and_hits_nexus_data):
...
-    evt  = 0
-    p_id = 1
-    particle = mcparticle_df.loc[evt].loc[p_id]
-    assert particle.particle_name == name
-    assert np.isclose(particle.kin_energy, k_eng)
 
-    init_vtx = particle[['initial_x', 'initial_y',
-                         'initial_z', 'initial_t']]
-    assert_allclose(init_vtx.tolist(), vi)
+def test_load_mcparticles_df(mc_particle_and_hits_nexus_data):
```
mmkekic: the same as with `test_load_mchits_df`

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -341,137 +340,99 @@ def test_read_mcparticles_df(mc_particle_and_hits_nexus_data):
...
 
-    sns_number = 0
-    wvf        = waveforms[sns_number]
-    bins       = [t / wvf.bin_width for t in wvf.times]
-    samples    = list(zip(bins, wvf.charges))
 
-    assert samples[0]   == pmt0_first
-    assert samples[-1]  == pmt0_last
-    assert len(samples) == pmt0_tot_samples
+@fixture(scope = 'module')
```
mmkekic, line 479: Maybe you can manually construct the exact binning dataframe you expect for both new and old format (since it is 2-3 rows) so the check is more obvious?

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -206,54 +89,187 @@ def test_copy_mc_info_which_events_out_of_range(ICDATADIR, config_tmpdir):
...
+    ## The extents and events tables are not saved.
+    assert len(input_tbls) == len(saved_tbls) + 2
+    assert MCTableType.configuration in saved_tbls
+    assert MCTableType.hits          in saved_tbls
+    assert MCTableType.particles     in saved_tbls
+    assert MCTableType.sns_positions in saved_tbls
+    assert MCTableType.sns_response  in saved_tbls
+
+
+@fixture(scope = 'module')
```
mmkekic, line 311: if  `mc_particle_and_hits_nexus_data` is only used in this module you can move it here from conftest.py

invisible_cities/io/mcinfo_io.py
mmkekic: No need to `copy`

invisible_cities/io/mcinfo_io.py
```diff
@@ -1,397 +1,766 @@
...
-                              run_no    : int) -> Tuple:
+    returns
+    -------
+    sns_pos : pd.DataFrame
+              Sensor position info for the MC sensors
+              which saw light in this simulation.
+    """
+    if is_oldformat_file(file_name):
+        raise TypeError('Old format files not valid for get_sensor_types')
+    sns_pos = load_dst(file_name, 'MC', 'sns_positions').copy()
```
mmkekic: no need to `copy`

mmkekic (commented):
Maybe you can add tests:
- Error when input file contains table outside of McTableType
- Explicit test for `get_event_numbers_in_file` and the Error in case no table is present
- Trivial test for `is_old_format`
- Tests for `load_mcsensor_positions`: for old format that the error is raised; for both that the sensors in sensor_responses are presend in mcsensor_position df
- Tests for `load_mcsensor_response_df` with `return_raw` both `True` and `False` for both formats

mmkekic (commented):
in all city tests remove comments and replace the tables with new MC tables we expect for 2 types of test:
- check if table present
- exact comparison. for this one you have to replace the exact_comparison file

After this and some squashing the PR is ready to be approved (finally)

mmkekic (approved):
This PR modifies MC io module to adapt for the new nexus output format, ensuring that the main readers/writer can be applied to the old format as well. The new functions use extensively dataframes that are faster to manipulate with than the previous dictionary based readers/writer. All the functionalities are well tested, and new MC test dataset is introduced to guarantee the equality in information read from the old and new format. 
Some functionalities are repeated due to different old/new table structures, and are planned to be removed in the future once there is no need for the 'old' files readers.
The road to user-transparent compatibility was bumpy but it sparks joy to finally approve this PR! Well done!

PR #692, opened 2020-02-17
Add detsim city source

gondiaz:
Following issue #691:
1. a source function for data flow city detsim is upload in module reco/detsim_functions.py.

This function reads MC files and creates an event dictionary with the information:
1. event number
2. splited hits in x, y, z and energy

Module imports are set in order to account for subsequent PRs.

PR #694, opened 2020-02-26
Detsim source function

gondiaz:
This PR is the same as PR #692 but with another branch name.

detsim_PR branch is now detsim_source

Note that a commit was added that remove unnecessary imports for this function, rename and relocate module in invisible_cities/cities, since this function will probably be relocated in invisible_cities/cities/components.py in the future.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -62,6 +62,7 @@ from .. io     .dst_io            import                  load_dst
 from .. io     .event_filter_io   import       event_filter_writer
 from .. io     .pmaps_io          import               pmap_writer
 from .. io     .rwf_io            import             buffer_writer
+from .. io     .mcinfo_io         import            load_mchits_df
 from .. types  .ic_types          import                        xy
 from .. types  .ic_types          import                        NN
 from .. types  .ic_types          import                       NNN
@@ -504,6 +505,23 @@ def hits_and_kdst_from_files(paths: List[str]) -> Iterator[Dict[str,Union[HitCol
                            run_number = run_number,
                            event_number = event_number,
                            timestamp = timestamp)
+
+
+def MC_hits_from_files(files_in : List[str]) -> Generator:
+    for filename in files_in:
+        try:
+            hits_df = load_mchits_df(filename)
+        except tb.exceptions.NoSuchNodeError:
+            continue
+        for evt, hits in hits_df.groupby(level=0):
+            yield dict(event_number = evt,
+                       x            = hits.x     .values,
+                       y            = hits.y     .values,
+                       z            = hits.z     .values,
+                       energy       = hits.energy.values,
+                       time         = hits.time  .values,
+                       label        = hits.label .values,
+                       timestamp    = 0)
 
 
 def sensor_data(path, wf_type):


invisible_cities/cities/detsim_source.py
```diff
@@ -0,0 +1,24 @@
...
+            extents = pd.read_hdf(filename, 'MC/extents')
+            event_ids  = extents.evt_number
+            hits_df    = read_mchits_df(h5in, extents)
+            for evt in event_ids:
+                hits = hits_df.loc[evt, :, :]
+                yield dict(event_number = evt,
+                           x      = hits["x"]     .values,
+                           y      = hits["y"]     .values,
+                           z      = hits["z"]     .values,
+                           energy = hits["energy"].values)
```
gonzaponte: align

invisible_cities/cities/detsim_source.py
```diff
@@ -0,0 +1,24 @@
...
+############### SOURCE ################
+#######################################
+def load_MC(files_in : List[str]) -> Generator:
+    for filename in files_in:
+        with tb.open_file(filename) as h5in:
+            extents = pd.read_hdf(filename, 'MC/extents')
+            event_ids  = extents.evt_number
+            hits_df    = read_mchits_df(h5in, extents)
+            for evt in event_ids:
+                hits = hits_df.loc[evt, :, :]
```
gonzaponte: This can also be done like
```
for evt, hits in hits_df.groupby(level=0): # iteration over first level of multiindex
    ...
```
which would eliminate the need to define `event_ids`

invisible_cities/cities/detsim_source.py
```diff
@@ -0,0 +1,24 @@
...
+
+from invisible_cities.io.mcinfo_io import read_mchits_df
+
+#######################################
+############### SOURCE ################
+#######################################
+def load_MC(files_in : List[str]) -> Generator:
+    for filename in files_in:
+        with tb.open_file(filename) as h5in:
+            extents = pd.read_hdf(filename, 'MC/extents')
```
gonzaponte: This line looks misplaced. It should either go outside the `with` statement or load the data using the already-opened file rather than the filename.

gonzaponte: I think the same function accepts the file object, so it can be kept there changing `filename` by `h5in`.

gonzaponte: In that case, align with the rest

invisible_cities/cities/detsim_source.py
```diff
@@ -0,0 +1,24 @@
+import numpy  as np
```
gonzaponte: `numpy` is not used

invisible_cities/cities/detsim_source.py
```diff
@@ -0,0 +1,24 @@
...
+import tables as tb
+import pandas as pd
+
+from typing import List, Generator
```
gonzaponte: Split into two lines

invisible_cities/cities/detsim_source.py
gonzaponte: Also, `hits.x.values` etc. looks a bit better

gonzaponte (commented):
Some minor comments. Looks good.

andLaing:
A lot of the things that need to be changed here depend highly on PR #693 so some of @gonzaponte 's suggested changes should maybe wait til it's approved (or you can test with the branch for that PR). Also, we'll need the hit time returned too, remember.

invisible_cities/cities/detsim_source.py
gondiaz: Do you mean to align the =?

invisible_cities/cities/detsim_source.py
gondiaz: `extents` must be a `pd.DataFrame`, personally I prefer it to be inside the with statement. But a `to_df=pd.DataFrame.from_records` should be added in the module.
What is your preference?

gonzaponte: Well, since you are loading everything at once you can simply use `load_mchits_df` which gives you `hits_df` directly. I had not seen this function before, but it is doing exactly what you are doing, so better use i). Otherwise, you would need to move it outside or do `extents = pd.DataFrame.from_records(h5in.root.MC.extents.read())`

invisible_cities/cities/detsim_source.py
```diff
@@ -0,0 +1,22 @@
+import tables as tb
+import pandas as pd
+
```
gonzaponte: Not needed

gonzaponte (changes requested):
I think the function should go in components.py with the rest of the sources.

invisible_cities/cities/detsim_source.py
```diff
@@ -0,0 +1,22 @@
...
+from typing import List
+from typing import Generator
+
+from invisible_cities.io.mcinfo_io import load_mchits_df
+
+#######################################
+############### SOURCE ################
+#######################################
+def load_MC(files_in : List[str]) -> Generator:
+    for filename in files_in:
```
mmkekic: here shouldn't we have `try except` in case some of the input files are corrupted/empty?

mmkekic (commented):
Do we want a test to check that the dictionary yielded contains the information we expect? Though seems most of the sources do not have a test...

invisible_cities/cities/detsim_source.py
gonzaponte: Good point

gonzaponte:
We should eventually have tests for the sources, but we can open a PR for all of them at once.

invisible_cities/cities/detsim_source.py
gondiaz: In case the file is empty, you will loop through nothing, so you will get an empty output file.

gondiaz:
Added try-except statement, and moved source function to components.

gondiaz:
Commit history cleaned, that PR should be ready for approval. Recall that this PR do not contain tests as other source functions.

gonzaponte (approved):
Source for the detsim city. All good.

invisible_cities/cities/components.py
```diff
@@ -493,6 +494,24 @@ def hits_and_kdst_from_files(paths: List[str]) -> Iterator[Dict[str,Union[HitCol
                            timestamp = timestamp)
 
 
+def load_MC(files_in : List[str]) -> Generator:
```
mmkekic: As @andLaing  pointed out the name is too generic, and should probably be something of `MC_hits_from_files` or similar

gondiaz: Totally agree. Also, we must add the hits `label` in the source. I will push this change right now.

PR #696, opened 2020-02-26
Upload database with gains from latest calibrations

carmenromo:
This PR includes the latest sensor calibrations from run 7852. Database has been uploaded with the PMT and SiPMs gains and the pdf values.

pmtRunDifferencePlots

sipmRunDifferencePlots


jmbenlloch (approved):
Latest calibration added with all test passing!

Issue #691, opened 2020-02-17
Detector simulation


andLaing:
The functionality to sort nexus full simulation into buffers or fully simulate the detector response in python (generally referred to as detsim) needs to be integrated into IC.

We propose the following grouping of pull requests to gradually integrate the functionality:
1) Additional source components to read relevant MC info into dataflow
2) New writer to save all the relevant information and add the possibility to split long events into multiple 'triggers'
3) Functions to generate scintillation light and its detection by the sensors
4) Functions to simulate electron drift (diffusion/attachment etc)
5) Functions to simulate EL light and its detection by the sensors
6) Functions to produce buffers and position s1 and s2
7) Utility functions for light level scaling etc
8) Dataflow (city) reading mc hits, simulating detector response, and saving relevant information
9) Dataflow (city) reading full simulation sensor response and production buffers and other relevant info.

All PRs will reference this issue for book keeping.

gondiaz:
10. Functions to manipulate the psf functions.

gonzaponte:
In order to coordinate the PRs, please agree on a common folder for the code that doesn't belong in the cities folder and move your code there (just before the PR is merged, don't do it in between other commits).

andLaing:
In PR #737 I put the new functions in a folder `detector_simulation`. If everyone's happy with that we could use it. If other PRs are accepted before it it'll make things a bit more awkward but it shouldn't be a big problem.

If that name's a bit verbose we could change it in all.

gonzaponte:
Both `detector_simulation` and `detsim` look good to me. I think the detsim name is pretty well known among the software people and it is shorter, which may be nicer to see, but I don't have a strong opinion. Whatever you decide. But I would like all PRs to share this folder before merging them.

mmkekic:
Closed by approving final parts: #742 and #769

PR #695, opened 2020-02-26
Detsim electron simulation functions

gondiaz:
Following issue #691:
4. upload functions to simulate electron drift (diffusion/attachment etc) in invisible_cities/cities/detsim_simulate_electrons.py

This module contains three functions, that generate, drift and diffuse the secondary electrons produced by the energy hits.

PR #697, opened 2020-02-26
Detsim photon simulation functions

gondiaz:
Following issue #691:
3. & 5. upload functions to generate scintillation (S1) and EL (S2) light and function to compute the pes at sensors based on the psf function in invisible_cities/cities/detsim_simulate_photons.py

This module contains three functions: S1 photons generation, S2 photon generation and computation of pes at sensors based on the psf.


invisible_cities/cities/detsim_simulate_photons.py
```diff
@@ -0,0 +1,44 @@
+import numpy  as np
+
+from typing import Callable
+
+
+#######################################
+########## PHOTON SIMULATION ##########
+#######################################
+def generate_s1_photons(energies : np.array,
```
andLaing: The type hint for a numpy array is np.ndarray

invisible_cities/cities/detsim_simulate_photons.py
```diff
@@ -0,0 +1,44 @@
...
+
+def generate_s2_photons(x              : np.array,
+                        el_gain        : float,
+                        el_gain_sigma  : float) -> np.array:
+    """generate number of EL-photons produced by secondary electrons that reach
+    the EL (after drift and diffusion)
+    """
+    n = len(x)
+    nphs = np.random.normal(el_gain, el_gain_sigma, size = n)
+    return nphs
```
andLaing: Should the nphs not take into account the x? What is x, actually? an array of electrons?

If you just wanted to calculate a yield value then the name should be changed and I'd just send the size rather than the full array.

invisible_cities/cities/detsim_simulate_photons_test.py
```diff
@@ -0,0 +1,63 @@
...
+import invisible_cities.database.load_db as db
+
+from invisible_cities.io.mcinfo_io import read_mchits_df
+
+from invisible_cities.cities.detsim_simulate_photons import *
+
+## same as in detsim_simulate_electrons_test
+@fixture(params = ["Kr83_nexus_v5_03_00_ACTIVE_7bar_3evts.MCRD.h5"])
+def nexusMC_filename(request, ICDATADIR):
+    return os.path.join(ICDATADIR, request.param)
```
andLaing: I think this file is already in a session scope fixture in `invisible_cities/conftest.py` , I'm not sure you gain much by redefining it

invisible_cities/cities/detsim_simulate_photons_test.py
```diff
@@ -0,0 +1,63 @@
...
+
+## same as in detsim_simulate_electrons_test
+@fixture
+def random_event_hits(nexusMC_filename):
+    filename = nexusMC_filename
+
+    with tb.open_file(filename) as h5in:
+        extents = pd.read_hdf(filename, 'MC/extents')
+        event_ids  = extents.evt_number
+        hits_df    = read_mchits_df(h5in, extents)
```
andLaing: We'll have to change this with the new readers implementation so it works for the new format too. It's in PR #693 if you want to have a look, hopefully it'll be ready soon

invisible_cities/cities/detsim_simulate_photons_test.py
```diff
@@ -0,0 +1,63 @@
...
+
+        evt = np.random.choice(event_ids)
+        hits = hits_df.loc[evt, :, :]
```
andLaing: `hits_df.loc[evt, :, :]` can just be hits_df.loc[evt]

invisible_cities/cities/detsim_simulate_photons_test.py
```diff
@@ -0,0 +1,63 @@
...
+    photons = generate_s1_photons(energies, 10)
+
+    pes = pes_at_sensors(xs, ys, zs, photons,
+                         x_sensors, y_sensors, z_sensors,
+                         psf)
+
+    nsensors = len(sensor_data)
+    nhits    = len(xs)
+
+    assert pes.shape == (nsensors, nhits)
```
andLaing: Again, add individual tests for each function

andLaing (changes requested):
Again these functions don't belong in invisible_cities/cities but in another folder. Personally my preference would be a new folder like invisible_cities/detsim or invisible_cities/optical_simulation but maybe a @mmkekic or @gonzaponte have more informed opinions.

Other than that a few clarifications on the functions and more tests needed.

gonzaponte:
> my preference would be a new folder like invisible_cities/detsim or invisible_cities/optical_simulation

I agree. I think this is a good time to create a simulation-focused folder where we can gather all non-generic, non-reco functions (diomira + detsim, essentially).

As for the name, I don't have a strong preference, but it would like to be generic enough to include everything simulation-related. "simulation" or "detector_simulation" sound good to me, but you can come up with a better choice, probably.

invisible_cities/cities/detsim_simulate_photons_test.py
jerenner: Do we need a test even for functions that are as simple as `generate_s2_photons`, which would essentially be testing `np.random.normal`?

invisible_cities/cities/detsim_simulate_photons.py
gondiaz: x is an array with the size of the number of secondary electrons that arrive to the EL, so any of the arrays from the output of `diffuse_electrons` in PR #695. The reason why this is implemented in that way is because in the dataflow dictionary (PR #700) just the arrays appear. In my opinion this must be modified, so suggestions are welcome

jahernando: I think 'x' is in this case a bad name, maybe in this case you can pass _n_ as argument, the len of _x_ and _number_of_electrons_

invisible_cities/cities/detsim_simulate_photons_test.py
jahernando: The function is quite simple, and I guess the test can be quite simple too. Maybe it is fast to do the test anyhow.

jahernando:
Respect the location of the code, I will go for a folder called 'detector_simulation' and include the functions there.

gonzaponte (commented):
Not sure these need to be separate functions. They are one-liners that are used in exactly one place (as far as I know). Unless you feel like they add something essential I would just replace the function call by the line itself.

gondiaz:
This functions will go in the city dataflow module.

PR #698, opened 2020-02-26
Detsim get psf functions

gondiaz:
Following issue #691:
10. upload functions to read an manipulate kr maps and sipm psfs for their transformation into a callable that can be used in `pes_at_sensors` function from PR #697 in module invisible_cities/cities/detsim_get_psf.py

This module contains two functions to get the referred psfs (pmt and sipm), an auxiliary function to construct a required callable to be used at `pes_at_sensors` and a psf temporal analytical function.

The psf analytical function must be substituted in the future by a light table equivalent.


invisible_cities/cities/detsim_get_psf.py
```diff
@@ -0,0 +1,66 @@
+import numpy  as np
+import tables as tb
+import pandas as pd
+
+from typing import Tuple, Callable
```
andLaing: Line per import

invisible_cities/cities/detsim_get_psf.py
```diff
@@ -0,0 +1,66 @@
...
+                       xaxis : Tuple[float],
+                       yaxis : Tuple[float])->Callable:
+    xmin, xmax, dx = xaxis
+    ymin, ymax, dy = yaxis
+    def function(x, y, z=0):
+        x = np.clip(x, xmin, xmax-0.1*dx)
+        y = np.clip(y, ymin, ymax-0.1*dy)
+        ix = ((x-xmin)//dx).astype(int)
+        iy = ((y-ymin)//dy).astype(int)
+        return array[ix, iy]
```
andLaing: I'd prefer it had a more explicative name, function seems a bit generic as a name.

Also, you don't use z. Did you forget to push part of the code or is z not used? It shouldn't be an argument of the function if it has no use.

Also the type hint for a numpy array is np.ndarray. I'd give it a descriptive name other than just array which is confusing.

invisible_cities/cities/detsim_get_psf_test.py
```diff
@@ -0,0 +1,88 @@
...
+####################################
+@fixture
+def extent_and_function():
+    xmin, xmax, dx = 12, 100, 3.
+    ymin, ymax, dy =-50, 50, 13.
+
+    xaxis = (xmin, xmax, dx)
+    yaxis = (ymin, ymax, dy)
+
+    func = lambda x, y: - x**2
```
andLaing: why does it have y as an argument if it only uses x?

invisible_cities/cities/detsim_get_psf_test.py
```diff
@@ -0,0 +1,88 @@
...
+    xbins = np.arange(xmin, xmax+dx, dx)
+    ybins = np.arange(ymin, ymax+dy, dy)
+
+    x, y = np.meshgrid(xbins, ybins)
+    x, y = x.flatten(), y.flatten()
+    f = func(x, y)
+
+    #test create_function
+    H, _ = np.histogramdd((x, y), weights=f, bins=(xbins, ybins))
+    callable = create_xy_function(H, xaxis, yaxis)
```
andLaing: give it a name, callable is a type

andLaing (changes requested):
Needs to be moved too.

Few clarifications needed on the individual functions

invisible_cities/cities/detsim_get_psf.py
gondiaz: Now, I changed this `create_xy_function` by `create_xyx_function`.  The fact of adding the `z` variable is because when we compute `pes = photons * psf(dxs, dys, dzs)` (PR #697 in `pes_at_sensors ` function), this `psf` is in general a x, y, z function. 

For krmap and sipmpsf, z dimension is not used, but it is added to do not include something duplicate like:
```
if S2:
    pes = photons * psf(dxs, dys, dzs)
elif S1:
    pes = photons * psf(dxs, dys)
```
Besides, we can add light tables, which are x, y, z functions.

invisible_cities/cities/detsim_get_psf_test.py
gondiaz: No reason for that, the function `func` is just a x, y, z (new code adds the z) for testing `create_xyz_function`. In the case it is there, just a x dependence is assumed.

andLaing: It's a lambda function in a fixture that takes 3 arguments and only uses one?! Why? If you want it as a place holder for a more complex function which will use the other arguments write the full function and make it return `-x**2` if y and z are 0 or something like that. If it's always just going to give `-x**2` than just define it with one parameter `x`

gondiaz: This is just a matter of taste. This function can be modified to whatever xyz function you want. Note that its aim is in `test_create_xyz_function`. I changed it with explicit x, y, z dependence inside, maybe now it looks better for you.

andLaing: Well no, but thanks for changing it so that y and z are relevant

gondiaz:
This PR is overriden by new ones #756 #757.

PR #699, opened 2020-02-26
Detsim waveform creation functions

gondiaz:
Following issue #691:
6. upload the functions that create the waveforms in invisible_cities/cities/detsim_waveforms.py

This module contains three functions: one is auxiliary (probably already existing in IC); from the other two, one of them basically calls recursively (for each sensor) to the third. 

The last is a function that takes a vector of pes and buffer times and creates the waveform. EL width and S1 emission time are simulated with the nsamples argument.


invisible_cities/cities/detsim_waveforms.py
```diff
@@ -0,0 +1,53 @@
...
+    if nsamples==0:
+        sel = pes>0
+        t = np.repeat(times[sel], pes[sel])
+        t = np.clip  (t, 0, bins[-1])
+        indexes, counts = bincounter(t, wf_bin_time)
+        wf[indexes] = counts
+        return wf
+
+    sel = pes>0
+    t = np.repeat(times[sel], pes[sel])
```
jerenner: The lines:
```
    sel = pes>0
    t = np.repeat(times[sel], pes[sel])
```
could be placed above:
```
    if nsamples==0:
```
to avoid repeating them after the `if` statement.

jerenner (commented):
Some short docstrings describing the inputs and return values would be helpful

invisible_cities/cities/detsim_waveforms.py
```diff
@@ -0,0 +1,53 @@
+import numpy as np
+
+##################################
+######### WAVEFORMS ##############
+##################################
+def bincounter(xs : np.array,
+               dx : float =1.,
+               x0 : float =0.):
+    ixs = ((xs - x0) // dx).astype(int)
+    return np.unique(ixs, return_counts=True)
```
gonzaponte: Have you checked [np.digitize](https://docs.scipy.org/doc/numpy/reference/generated/numpy.digitize.html)? It does precisely that if you provide him with the bin edges.

gonzaponte: Add return type hint.
It would also be helpful to expand the output of `np.unique` (in case this implementation remains as is) to some descriptive variable names for readability.

invisible_cities/cities/detsim_waveforms.py
```diff
@@ -0,0 +1,53 @@
...
+                    bins  : np.array,
+                    wf_bin_time : float,
+                    nsamples    : int) -> np.array:
+    wf = np.zeros(len(bins))
+
+    if np.sum(pes)==0:
+        return wf
+    if nsamples==0:
+        sel = pes>0
+        t = np.repeat(times[sel], pes[sel])
```
gonzaponte: no need to do this selection, wherever `pes` is zero the entry in `times` will not be propagated to `t`.
(how can you have a recorded time with zero pes anyway?)

invisible_cities/cities/detsim_waveforms.py
```diff
@@ -0,0 +1,53 @@
...
+               dx : float =1.,
+               x0 : float =0.):
+    ixs = ((xs - x0) // dx).astype(int)
+    return np.unique(ixs, return_counts=True)
+
+
+def create_waveform(times : np.array,
+                    pes   : np.array,
+                    bins  : np.array,
+                    wf_bin_time : float,
```
gonzaponte: rename to `bin_width` or `wf_bin_width`

gonzaponte: Actually the is redundancy between `bins` and `wf_bin_time`, as they are dependent on each other. I think this function would be much better working with bin_edges and using `np.digitize`

invisible_cities/cities/detsim_waveforms.py
```diff
@@ -0,0 +1,53 @@
...
+                    nsamples    : int) -> np.array:
+    wf = np.zeros(len(bins))
+
+    if np.sum(pes)==0:
+        return wf
+    if nsamples==0:
+        sel = pes>0
+        t = np.repeat(times[sel], pes[sel])
+        t = np.clip  (t, 0, bins[-1])
```
gonzaponte: Are you sure you want to clip like that? stuff that wouldn't fall in the buffer are going to fall in the edges.

invisible_cities/cities/detsim_waveforms.py
```diff
@@ -0,0 +1,53 @@
...
+        return wf
+
+    sel = pes>0
+    t = np.repeat(times[sel], pes[sel])
+    t = np.clip  (t, 0, bins[-nsamples])
+    indexes, counts = bincounter(t, wf_bin_time)
+
+    spread_pes = np.repeat(counts[:, np.newaxis]/nsamples, nsamples, axis=1)
+    for index, counts in zip(indexes, spread_pes):
+        wf[index:index+nsamples] = wf[index:index+nsamples] + counts
```
gonzaponte: for consistency, rename `spread_pes` to `spread_counts`

gonzaponte: you can also use `+=` for incrementing the counts.

gonzaponte: What happens if `counts = 5` and `nsamples = 2`? You add 2.5 pes to each bin?
There is a risk of adding (or removing) counts unless the Poisson line in `create_sensor_waveforms` is managed properly, which might be the case, but maybe is good idea to clarify or leave some inline comment in the code if that is the case.

invisible_cities/cities/detsim_waveforms.py
```diff
@@ -0,0 +1,53 @@
...
+                            wf_bin_time    : float,
+                            nsamples : int,
+                            poisson  : bool =False) -> np.array:
+    bins = np.arange(0, wf_buffer_time, wf_bin_time)
+    wfs = np.array([create_waveform(times, pes, bins, wf_bin_time, nsamples) for pes in pes_at_sensors])
+
+    if poisson:
+        wfs = np.random.poisson(wfs)
+
+    return wfs
```
gonzaponte: `times` and `pes_at_sensors` are not very descriptive. If I understand correctly, they are sequences containing a sequence of times/how many pes at that time per sensor. Is that right? Maybe rename the variables and/or add an inline comment clarifying

invisible_cities/cities/detsim_waveforms.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def create_sensor_waveforms(times   : np.array,
+                            pes_at_sensors : np.array,
+                            wf_buffer_time : float,
+                            wf_bin_time    : float,
+                            nsamples : int,
+                            poisson  : bool =False) -> np.array:
+    bins = np.arange(0, wf_buffer_time, wf_bin_time)
+    wfs = np.array([create_waveform(times, pes, bins, wf_bin_time, nsamples) for pes in pes_at_sensors])
```
gonzaponte: better use `np.stack`

invisible_cities/cities/detsim_waveforms_test.py
```diff
@@ -0,0 +1,20 @@
...
+    nsamples = 2
+    times = np.array([1, 2, 5, 6, 9, 10])
+    pes   = np.array([2, 2, 2, 4, 1,  3])
+
+    expected = np.array([0, 1, 2, 1, 0, 1, 3, 2, 0, 2, 2], dtype=float)
+
+    # test
+    waveform = create_waveform(times, pes, bins, wf_bin_time, nsamples)
+
+    assert (waveform == expected).all()
```
gonzaponte:
```suggestion
    assert np.all(waveform == expected)
```

invisible_cities/cities/detsim_waveforms.py
```diff
@@ -0,0 +1,53 @@
...
+def create_waveform(times : np.array,
+                    pes   : np.array,
+                    bins  : np.array,
+                    wf_bin_time : float,
+                    nsamples    : int) -> np.array:
+    wf = np.zeros(len(bins))
+
+    if np.sum(pes)==0:
+        return wf
+    if nsamples==0:
```
gonzaponte: The code outside the `if` is generic enough to handle `nsamples = 1`, which is what you are dealing with in this `if` statement, right? Maybe you should forbid `nsamples=0` and if the input is `1` you know it works (a test to demonstrate it would be nice)

invisible_cities/cities/detsim_waveforms_test.py
```diff
@@ -0,0 +1,20 @@
+import numpy as np
+
+from invisible_cities.cities.detsim_waveforms import *
```
gonzaponte: `*` imports are forbidden! Import specific elements

invisible_cities/cities/detsim_waveforms_test.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+
+def test_create_waveform():
```
gonzaponte: this is good, but you need to add maybe a more generic test with hypothesis and also extreme cases like those that you check within the function: sum is zero or nsamples is zero.

Also, the other two functions that are not tested at all!

gonzaponte (commented):
First look to the code from my side. I think that changing the approach in the main function can help.

invisible_cities/cities/detsim_waveforms.py
gondiaz: Right

invisible_cities/cities/detsim_waveforms.py
gondiaz: If I understand your last comment, there should be no problem with the poisson since for X\~Poiss(a) and Y\~Poiss(b), then X+Y~Poiss(a+b). This is one of the reasons to do not apply poisson until the end in `create_sensor_waveforms`. The other reason is because the function could be accurately tested if no poisson is applied.

gonzaponte: No, my concern was mostly about the meaning of things and the connection with other parts of the code.

Since the Poisson randomization seems to be optional (it is an argument to `create_sensor_waveforms`), I would say that the output of `create_waveform` means "exact number of photons in these time bins". Essentially because in a simulation I don't think it makes sense to say you received 2.5 photons.

Now, assuming this is _not_ necessarily correct, then there must be a connection between `create_waveform` and applying the Poisson sampling in `create_sensor_waveforms`, otherwise, I don't understand it. In this case, I assume the output of `create_waveform` can mean "_average_ number of photons in these time bins" and then we can take that average and use it as the parameter of a Poisson distribution to get the actual number of photons detected. This makes sense to me, but it also introduces a subtle connection between one function and the other.

It can also be that you decided to use it both ways:
- If `nsamples = 1` (or 0 or whatever) then the output of `create_waveform` is directly the number of photons detected.
- Otherwise, if `nsamples > 1`, the output of `create_waveform` is not the number of photons detected, but the average, and to get the actual number of photons detected we need to sample a Poisson distribution.

If this is the case, that is OK, but it needs to be documented in detail.
Also, if this code is meant to work as I described it, then `poisson` is not an independent argument, as it is determined from the value of `nsamples`. But this is an extrapolation from what I think is going on, which may not be correct.

gondiaz: Your point is correct, I have two proposals for `create_waveform`:

Note that `pes` values must be a vector of integers as it is the function now. In detsim, pes from an emission point will come from the product of `pes = #photons x LT`, where `#photons` is the number of photons emited in an emission point and `LT` the Light Table probability for that emission point. Since the `LT`probability is a float, `pes` could be a non integer.

1) We could convert `pes` values to integer inside the function and explicitly comment this as well as the fact that for `nsamples>1` the counts could be no integer. 

2) We could accept no integer values in `pes` and change the histogram procedure using `np.histogram` with `pes` as weights. 

I haven't probe it, but I think the 1) is better in performance since a `np.histagram` could be time consuming.

invisible_cities/cities/detsim_waveforms.py
```diff
@@ -0,0 +1,53 @@
...
+                            pes_at_sensors : np.array,
+                            wf_buffer_time : float,
+                            wf_bin_time    : float,
+                            nsamples : int,
+                            poisson  : bool =False) -> np.array:
+    bins = np.arange(0, wf_buffer_time, wf_bin_time)
+    wfs = np.array([create_waveform(times, pes, bins, wf_bin_time, nsamples) for pes in pes_at_sensors])
+
+    if poisson:
+        wfs = np.random.poisson(wfs)
```
andLaing: Don't quite understand what this bit does?

gondiaz: First the waveforms are created for each sensor obtaining an array with size (#sensor, len(bins)), here call `wfs`. This contains the counts in each time bin. We want each time bin to be a poisson variable, which is allowed by this bit.

invisible_cities/cities/detsim_waveforms.py
gonzaponte: If I understand correctly, 1 is not correct. If by convert you mean `pes = int(pes)`, then that is not correct, because the original `pes` means "_average_ number of photons" and therefore it needs to be "poissonized". 2. Is more correct and I seriously doubt that `np.histogramp` has performance issues (because essentially does the same as you, but in C).

gondiaz: You are right, `pes` is "poissonized" in `pes_at_sensor` function in #697 . So I would keep the function as it is, but explicitly adding a comment for `nsamples>1`. Would this be enought?

gonzaponte: Mmm, I think poissonization should be the last step no? in the end we are producing waveforms with integer number of photoelectrons

gondiaz: Maybe I am confusing something. As it is now, we poissonize in `pes_at_sensors` function first (in module detsim_simulate_photons in #697). So the argument variable `pes_at_sensors` in `create_sensor_waveforms` is an array of integers.
If we remove this poissonization we are in 1 above, where photoelectrons are  no necessarily integers.
Probably this poissonization is not necessary and we could just simply have `pes=int(#photons x LT)`.

Then, another poissonization is done at the end of `create_sensor_waveforms` if poisson argument is set to True. In this case the output waveforms have an integer number of photoelectrons as you say. I do not understand your last comment since this second poissonization is done at the end.

invisible_cities/cities/detsim_waveforms.py
andLaing: But zeros subject to a poisson will be Poisson distributed too and the electronics noise is not Poissonian not do we want to add noise in this step. Would you not need to select only the bins with signal?

invisible_cities/cities/detsim_waveforms.py
gonzaponte: > Probably this poissonization is not necessary and we could just simply have pes=int(#photons x LT).

No, this is physically incorrect, particularly when the average number of photons detected is very low. The poissonization **must** be done.

We have for sure to poissonize the number of photons arriving to a sensor, but after that I don't think it makes sense to distribute those photons in multiple bins and then poissonize again.

> Then, another poissonization is done at the end of create_sensor_waveforms

This is maybe wrong, according to what I said before.

>  if poisson argument is set to True

Why is this optional?

msorel: I have not followed any details, but doing Poisson smearing twice on same simulated data sounds very wrong to me (If this is actually what the code does).

gondiaz: This was in @jahernando original code, which comes from @gonzaponte too. Strictly speaking, the poissonization was included in `create_waveform`.

In my opinion, before thinking better, the last poisson makes no sense since when you redistribute in `nsamples` you are summing up poisson variables (if the first poissonization was already done) which is a poisson itself, so "poissonizing" is not correct.

This was set as optional argument for two reasons: one because you can make a test for the no-poissonized function; second because I wasn't totally sure about this was correct.

invisible_cities/cities/detsim_waveforms.py
gondiaz: The `np.random.poisson(0) = 0`. Nevertheless, "poissonization" is deeper discussed here https://github.com/nextic/IC/pull/699#discussion_r391819474

invisible_cities/cities/detsim_waveforms.py
gonzaponte: > the last poisson makes no sense

I also think that it is not correct, but there are other considerations.

When you think about the physics, there are a number of processes that have fluctuations (considering S2 only):
- When you have an energy deposition in the detector, you compute the number of ionization electrons (ie), this number is poissonized, because what you have as an estimator is the average number of ie produced.
- For each ie, you have to estimate the number of photons emitted (nphot), which is gaussianly distributed according to the EL gain and the Fano Factor.
- Given a value for nphot, you can compute the number of pes in a given sensor multiplying nphot by the probability from a light table or the parametrization. This number is also an average: the _average_ number of photons detected by that sensor. Therefore, in order to get the actual number of photons detected in a given sensor (npes) you need to poissonize, which you do in `pes_at_sensors`.
- Now, given that quantity you may need to redistribute them over a number of time bins (nsample), which you do in `create_waveform`. That's a simulation trick that you may want to implement but, if you do, you must keep the same number of photons, i.e., you shouldn't say you have npes/nsample photoelectrons in a time bin and poissonize again, even if statistically it is equivalent.

> one because you can make a test for the no-poissonized function

I understand the logic, but I think if you fix the random seed for your test you should be able to test the function without tricks of this kind. We can discuss it if you find troubles.

invisible_cities/cities/detsim_waveforms.py
gondiaz: As it is now, `nsamples=1` redistribute the counts in the same bin, which is equivalent to do not redistribute at all. In this case the function runs the count distribution loop. 

In order to afford this and improve performance (to do not enter in the loop), I suggest to keep the conditional for `nsamples=0` and `nsamples=1` such as: `if nsamples in [0, 1]` or equivalent.

gonzaponte: Unless there is a performance impact as you claim, I don't see a good reason to do that. It adds a bit of noise and, more importantly, it duplicates the code, which is not good. In any case, we can leave the time profiling to the end when the code is more mature and stable.

In any case, having who values in which `nsamples` implies a different behaviour is a bit fishy. Stick to one, which I would say is `nsamples = 1`. And unless there is a relevant performance difference, don't make it a special case at all!

invisible_cities/cities/detsim_waveforms.py
```diff
@@ -0,0 +1,76 @@
...
+
+    Input:
+    pes_at_sensors: an array with size (#sensors, len(times)). It is the same
+    as pes argument in create_waveform but for each sensor in axis 0.
+    wf_buffer_time: a float with the waveform extent (in default IC units)
+    bin_width: a float with the time distance between bins in the waveform buffer.
+    poisson: a bool. This must be set to True to poisson distribute each bin count.
+    """
+
+    bins = np.arange(0, wf_buffer_time, bin_width)
```
gonzaponte: actually, because these are bin edges, you need to have one more entry. Otherwise your waveform goes from 0 to `wf_buffer_time - bin_width`
```suggestion
    bins = np.arange(0, wf_buffer_time + bin_width, bin_width)
```
which means that in `create_waveforms` you need to do `wf = np.zeros(len(bins) - 1)`

invisible_cities/cities/detsim_waveforms.py
```diff
@@ -0,0 +1,74 @@
...
+    each of the buffer times in times argument.
+    bins: a vector with the output waveform bin times (for example [0, 25, 50, ...] if
+    the detector has a sampling time of 25).
+    nsamples: an integer that controlls the distribution of the photoelectrons in each of
+    the waveform bins. The counts (N) in a given time bin (T) are distributed in the waveform
+    such as the nsamples posterior to T would have N/nsamples counts (included T).
+    nsamples must be >=1 an <len(bins).
+    """
+    if nsamples<1:
+        raise Exception("nsamples<1. nsamples must be >= 1")
```
gonzaponte: raise a more specific exception. `ValueError` is appropriate here.

invisible_cities/cities/detsim_waveforms_test.py
gondiaz: I added a test for Exception when `nsamples<1` and Warning for `nsamples>len(bins)` and for the conservation of total pes, that is sum of initial pes is equal to the sum of counts in waveform. Note this sum conservation is only true if times values lie within bin edges, if not, pes in the exterior are removed (recall conversation https://github.com/nextic/IC/pull/699#discussion_r390937048 which was solved in commit 5962c8d)

In my opinion, it is correct and adecuate to have the test `test_create_waveform` since is an human made check for the case `nsamples` different to 1.

invisible_cities/cities/detsim_waveforms.py
```diff
@@ -0,0 +1,74 @@
...
+    if nsamples>len(bins):
+        warn("nsamples > number of bins. Shrinking it to number of bins")
+        nsamples = len(bins)
```
gonzaponte: Exceeding the number of bins doesn't sound like a mild thing. I would also raise an exception here.

invisible_cities/cities/detsim_waveforms.py
```diff
@@ -0,0 +1,74 @@
...
+
+    wf = np.zeros(len(bins))
+    if np.sum(pes)==0:
+        return wf
+
+    t = np.repeat(times, pes)
+    sel = (bins[0]<=t) & (t<=bins[-1])
```
gonzaponte:
```suggestion
    sel = in_range(t, bins[0], bins[-1])
```
`in_range` from core_functions

invisible_cities/cities/detsim_waveforms_test.py
```diff
@@ -0,0 +1,48 @@
...
+    assert np.allclose(waveform, expected)
+
+
+@composite
+def times_and_pes_from_bins(draw, bins):
+    min, max = bins[0], bins[-1]
+
+    times = draw(lists(integers(min_value=min, max_value=max), min_size=1))
+    pes   = draw(lists(integers(min_value=0), min_size=len(times), max_size=len(times)))
+    return times, pes, bins
```
gonzaponte: You may want to fix the bins in the composite to some sensible value and not take arguments at all since you are fixing them anyway elsewhere.

invisible_cities/cities/detsim_waveforms_test.py
```diff
@@ -0,0 +1,48 @@
...
+def test_create_waveform_Exception_Warning_Sum(times_and_pes, nsamples):
+    times, pes, bins = times_and_pes
+
+    if nsamples<1:
+        pytest.raises(Exception,  create_waveform, times, pes, bins, nsamples)
+    elif nsamples>len(bins):
+        pytest.warns(UserWarning, create_waveform, times, pes, bins, nsamples)
+    else:
+        waveform = create_waveform(times, pes, bins, nsamples)
+        assert np.allclose(np.sum(pes), np.sum(waveform))
```
gonzaponte: Split the test to consider the three cases separately. If you implement my suggestion of raising an error also for `nsamples > nbins` you can parametrize one test to consider both wrong values of `nsamples` and leave the other one for the normal case in which all inputs are correct. You will also need to add a test that checks that if `sum(pes) == 0` you get all zeros.

invisible_cities/cities/detsim_waveforms.py
gonzaponte: You can combine both checks into a single one

invisible_cities/cities/detsim_waveforms_test.py
gonzaponte: I agree that this test should stay, I meant that we need more :)

invisible_cities/cities/detsim_waveforms.py
gondiaz: `bins` are bin centers instead of bin edges

invisible_cities/cities/detsim_waveforms.py
gondiaz: It would be `sel = in_range(t, bins[0], bins[-1], right_closed=True)`. Nevertheless, I do not see a strong reason for this line to be changed, since in my opinion it is clearer for a code reader/reviewer to read as it is than looking to `in_range` to figure out what this function does.

invisible_cities/cities/detsim_waveforms.py
gonzaponte: No, they are bin edges.
If they were bin centers, you would have a bin center that is zero (unphysical) and
```
sel = (bins[0]<=t) & (t<=bins[-1])
```
and
```
np.digitize(t, bins)
```
would be wrong.
Besides, everything is easier if you work with bin edges whenever possible. It takes a bit of a mental switch for us but it is worth it.

invisible_cities/cities/detsim_waveforms.py
gonzaponte: > It would be sel = in_range(t, bins[0], bins[-1], right_closed=True).

Not if you use bin edges.

> Nevertheless, I do not see a strong reason for this line to be changed, since in my opinion it is clearer for a code reader/reviewer to read as it is than looking to in_range to figure out what this function does.

This function has been part of the IC utilities for very long. It was included because that functionality is required in many cases.Therefore we have a somewhat tricky piece of code that
- is not repeated everywhere (typos less likely)
- is tested
- has simpler syntax

It is ok if you didn't know about this function, but it has some clear benefits even if the first time you have to invest some time in learning it. Maybe the name should be changed to `in_interval`, which would be more appropriate.

invisible_cities/cities/detsim_waveforms.py
gondiaz: I think it is better to think in bin edges, so I changed it as you say. Nevertheless, `np.digitize` can be think to be working between bin centers as I was thinking in it before, which I know it is less natural.

I changed the distribution method. Now, the waveform is expanded to distribute the counts out of the real waveform. This is more natural since in this way we avoid clipping, which accumulate counts in a single bin. In this way, count distribution in nsamples bins is now continuos, in the sense that if we attach another waveform to this one, the distribution of bins would be maintaining its nsamples distribution correct. At the end, we shrink the waveform to the real size.

Note that commit 5b41aef is wrong, and it is corrected after.

invisible_cities/cities/detsim_waveforms_test.py
gondiaz: Do you suggest any other necessary tests?

gondiaz:
This PR is overriden by new ones #756 #757.

PR #700, opened 2020-02-26
Detsim city with IC dataflow structure

gondiaz:
Following issue #691:
8. upload detsim city with dataflow structure in invisible_cities/cities/detsim.py

This detsim city read MC energy hits and performs the physics probabilistic simulation, which is not the full simulation. The output is in waveform form.

Comments: 
-A modification in the city imports would be required once the PRs are approved: #694 #695 #697 #698 #699.
-For the full simulation city consult PR #


gondiaz:
Close deprecated PR. A new open will be open.

PR #701, opened 2020-03-05
Add documentation on how to prepare commits

gonzaponte:
This is a first attempt in documenting how to make a good commit, building on what's already described in `contributing.rst` and making it a bit more specific.

I intend to update this PR with some common practices that don't follow these guidelines to make a clearer picture.

### DIFF
--- a/doc/commits.rst
+++ b/doc/commits.rst
@@ -1,0 +1,48 @@
+How to create a good commit history
+===================================
+
+Commit structure
+-----------------
+
+- Focus on a specific change. Don't try to change everything at once.
+  If the plan is to implement a series of things, do them one by one.
+
+- Do not mix changes concerning different features in one commit.
+  For instance, do not mix cosmetical changes with other, more relevant ones.
+
+- Commit frequently and in an incremental manner.
+  Do not be afraid of making too many commits, they can be combined afterwards
+  (this is known as squashing in git).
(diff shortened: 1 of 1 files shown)


doc/commits.rst
```diff
@@ -0,0 +1,46 @@
...
+-----------------
+
+- Focus on a specific change. Don't try to change everything at once.
+  If the plan is to implement a series of things, do them one by one.
+
+- Do not mix changes concerning different features in one commit.
+  For instance, do not mix cosmetical changes with other, more relevant ones.
+
+- Commit frequently and in an incremental manner.
+  Do not be afraid of making too many commits, they can be merged afterwards.
```
jacg: Careful, "merge" has a very specific meaning in git, and not the one you are trying to convey here: "squash".

gonzaponte: Yup, you are right. I was actually thinking that maybe it would be more descriptive for people that are not familiar with git, but it cannot be used like this.

jacg: I understand completely where you're coming from. How about: "they can be combined afterwards. (This is known as *squashing* in git.)"?

gonzaponte: Yes, that's nice.

mmkekic (approved):
This PR adds a best practice guideline on how and when to commit. Hopefully the developers will follow it, which would not only make the history cleaner but also immensely ease the review process. Thanks for taking time to make this document!

Issue #641, opened 2019-03-11
EventMixer code in IC


msorel:
For MC physics studies, we mix several input streams into a single output stream through the EventMixer python code. The same code also tells us the effective exposure (in years) of a certain number of MC events above some visible energy threshold.

This mixing is necessary because:
- we mix close to 100 background sources, where each source is a given (isotope, Geant4 volume) combination, so the process is cumbersome and relies on detailed activity and detection efficiency assumptions for each (isotope, Geant4 volume) combination
- we want analysers to use a consistent event mix assumption in each analysis cycle, so that results can be easily compared

Background info:
- The code currently lives here https://next.ific.uv.es:8888/nextsw/PyToNE/blob/master/PyToNE/EventMixer.py
- Explanation of what this simple code does at NEXT-doc-287, particularly slide 5: https://next.ific.uv.es/cgi-bin/DocDB/private/ShowDocument?docid=287
- Example configuration file for Nov 2018 background MC production can be found here: https://next.ific.uv.es:8888/nextsw/PyToNE/blob/master/jobOptions/EventMixer_NEW_DepletedXe.job

What we want to change, after discussions with @paolafer and Pau:
- Code currently mixes nexus output files in GATE format. After mixing, all subsequent processing steps starting from the CPU-consuming detsim one are made only for this mixed output stream
- There are scenarios where we want to change normalisation assumptions for the ~100 background sources relatively frequently, yet keeping control of what analysers are using. Reprocessing everything from the detsim stage  can be a pain, and the detsim, diomira, irene steps are fairly stable
- The plan would therefore to allow the possibility to mix irene output files in hdf5 format, rather than nexus output files in GATE format. Apart from that, the code would look similar to existing one, specifying input streams, activities and detection efficiencies for each in a long configuration file. However, it would not make sense for this code to be stand-alone anymore, as it would use IC output. So, it should most likely live in IC. In this new scheme, all ~100 background sources would be separately processed all the way through irene, then mixed in correct proportions after irene and before penthesilea.

This issue is about receiving guidance from the experts on how to structure this piece of code in IC:

- Should this be a stand-alone IC city? Something else?
- Is irene the right stage where to mix input streams?
[...]

PR #702, opened 2020-03-08
New function to copy MCInfo to output file outside the main loop for selected events

jmalbos:
This PR introduces a new function (`copy_mc_info`) in mcinfo_io to copy the MCInfo from the input files to the output file for a list of events.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -152,6 +152,14 @@ def print_every_alternative_implementation(N):
     return print_every_loop
 
 
+def collect():
+    """Return a future/sink pair for collecting streams into a list."""
+    def append(l,e):
+        l.append(e)
+        return l
+    return fl.reduce(append, initial=[])()
+
+
 # TODO: consider caching database
 def deconv_pmt(dbfile, run_number, n_baseline, selection=None):
     DataPMT    = load_db.DataPMT(dbfile, run_number = run_number)
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -272,7 +272,6 @@ def wf_from_files(paths, wf_type):
                 run_number  = get_run_number  (h5in)
                 pmt_wfs     = get_pmt_wfs     (h5in, wf_type)
                 sipm_wfs    = get_sipm_wfs    (h5in, wf_type)
-                mc_info     = get_mc_info_safe(h5in, run_number)
                 (trg_type ,
                  trg_chann) = get_trigger_info(h5in)
             except tb.exceptions.NoSuchNodeError:
@@ -284,12 +283,9 @@ def wf_from_files(paths, wf_type):
                 event_number, timestamp         = evtinfo.fetch_all_fields()
                 if trtype  is not None: trtype  = trtype .fetch_all_fields()[0]
 
-                yield dict(pmt=pmt, sipm=sipm, mc=mc_info,
-                           run_number=run_number, event_number=event_number, timestamp=timestamp,
+                yield dict(pmt=pmt, sipm=sipm, run_number=run_number,
+                           event_number=event_number, timestamp=timestamp,
                            trigger_type=trtype, trigger_channels=trchann)
-            # NB, the monte_carlo writer is different from the others:
-            # it needs to be given the WHOLE TABLE (rather than a
-            # single event) at a time.
 
 
 def pmap_from_files(paths):
@@ -303,7 +299,6 @@ def pmap_from_files(paths):
             try:
                 run_number  = get_run_number(h5in)
                 event_info  = get_event_info(h5in)
-                mc_info     = get_mc_info_safe(h5in, run_number)
             except tb.exceptions.NoSuchNodeError:
                 continue
             except IndexError:
@@ -313,11 +308,9 @@ def pmap_from_files(paths):
 
             for evtinfo in event_info:
                 event_number, timestamp = evtinfo.fetch_all_fields()
-                yield dict(pmap=pmaps[event_number], mc=mc_info,
-                           run_number=run_number, event_number=event_number, timestamp=timestamp)
-            # NB, the monte_carlo writer is different from the others:
-            # it needs to be given the WHOLE TABLE (rather than a
-            # single event) at a time.
+                yield dict(pmap=pmaps[event_number], run_number=run_number,
+                           event_number=event_number, timestamp=timestamp)
+
 
 def cdst_from_files(paths: List[str]) -> Iterator[Dict[str,Union[pd.DataFrame, MCInfo, int, float]]]:
     """Reader of the files, yields collected hits,
@@ -351,7 +344,8 @@ def cdst_from_files(paths: List[str]) -> Iterator[Dict[str,Union[pd.DataFrame, M
             # single event) at a time.
 
 def hits_and_kdst_from_files(paths: List[str]) -> Iterator[Dict[str,Union[HitCollection, pd.DataFrame, MCInfo, int, float]]]:
-    """Reader of the files, yields HitsCollection, pandas DataFrame with kdst info, mc_info, run_number, event_number and timestamp"""
+    """Reader of the files, yields HitsCollection, pandas DataFrame with
+    kdst info, run_number, event_number and timestamp."""
     for path in paths:
         try:
             hits_df = load_dst (path, 'RECO', 'Events')
@@ -363,7 +357,6 @@ def hits_and_kdst_from_files(paths: List[str]) -> Iterator[Dict[str,Union[HitCol
             try:
                 run_number  = get_run_number(h5in)
                 event_info  = get_event_info(h5in)
-                mc_info     = get_mc_info_safe(h5in, run_number)
             except (tb.exceptions.NoSuchNodeError, IndexError):
                 continue
 
@@ -372,11 +365,12 @@ def hits_and_kdst_from_files(paths: List[str]) -> Iterator[Dict[str,Union[HitCol
             for evtinfo in event_info:
                 event_number, timestamp = evtinfo.fetch_all_fields()
                 hits = hits_from_df(hits_df.loc[hits_df.event == event_number])
-                yield dict(hits = hits[event_number], kdst = kdst_df.loc[kdst_df.event==event_number], mc=mc_info, run_number=run_number,
-                           event_number=event_number, timestamp=timestamp)
-            # NB, the monte_carlo writer is different from the others:
-            # it needs to be given the WHOLE TABLE (rather than a
-            # single event) at a time.
+                yield dict(hits = hits[event_number],
+                           kdst = kdst_df.loc[kdst_df.event==event_number],
+                           run_number = run_number,
+                           event_number = event_number,
+                           timestamp = timestamp)
+
 
 def sensor_data(path, wf_type):
     with tb.open_file(path, "r") as h5in:
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -44,6 +44,8 @@ from .. filters.s1s2_filter       import               S12Selector
 from .. filters.s1s2_filter       import               pmap_filter
 from .. database                  import                   load_db
 from .. sierpe                    import                       blr
+from .. reco                      import             tbl_functions as tbl
+from .. io                        import                 mcinfo_io
 from .. io     .pmaps_io          import                load_pmaps
 from .. io     .hits_io           import              hits_from_df
 from .. io     .dst_io            import                  load_dst
@@ -158,6 +160,32 @@ def collect():
         l.append(e)
         return l
     return fl.reduce(append, initial=[])()
+
+
+def copy_mc_info(files_in : List[str], h5out: tb.File, event_numbers: List[int]) -> None:
+    """Copy to an output file the MC info of a list of selected events.
+
+    Parameters
+    ----------
+    files_in : List of strings
+        Name of the input files.
+    file_out : tables.File
+        The output h5 file.
+    event_numbers : List[int]
+        List of event numbers for which the MC info is copied to the output file.
+    """
+
+    writer = mcinfo_io.mc_info_writer(h5out)
+
+    for f in files_in:
+        with tb.open_file(f, "r") as h5in:
+            try:
+                event_numbers_in_file = h5in.root.MC.extents.cols.evt_number[:]
+                event_numbers_to_copy = list(evt for evt in event_numbers \
+                                             if evt in event_numbers_in_file)
+                mcinfo_io.copy_mc_info(h5in, writer, event_numbers_to_copy)
+            except tb.exceptions.NoSuchNodeError:
+                continue
 
 
 # TODO: consider caching database
--- a/invisible_cities/cities/components_test.py
+++ b/invisible_cities/cities/components_test.py
@@ -17,6 +17,8 @@ from .. core.exceptions import ClusterEmptyList
 from .. core.system_of_units_c import units
 
 from .  components import event_range
+from .  components import collect
+from .  components import copy_mc_info
 from .  components import WfType
 from .  components import wf_from_files
 from .  components import pmap_from_files
@@ -24,7 +26,9 @@ from .  components import compute_xy_position
 from .  components import city
 from .  components import hits_and_kdst_from_files
 
-from .. database import load_db
+from .. dataflow   import dataflow as fl
+
+from .. database   import load_db
 
 
 def _create_dummy_conf_with_event_range(value):
@@ -151,7 +155,7 @@ def test_hits_and_kdst_from_files(ICDATADIR):
     event_number = 1
     timestamp    = 0.
     num_hits     = 13
-    keys = ['hits', 'mc', 'kdst', 'run_number', 'event_number', 'timestamp']
+    keys = ['hits', 'kdst', 'run_number', 'event_number', 'timestamp']
     file_in     = os.path.join(ICDATADIR    ,  'Kr83_nexus_v5_03_00_ACTIVE_7bar_3evts.HDST.h5')
     generator = hits_and_kdst_from_files([file_in])
     output = next(generator)
@@ -160,3 +164,29 @@ def test_hits_and_kdst_from_files(ICDATADIR):
     assert output['timestamp']      == timestamp
     assert len(output['hits'].hits) == num_hits
     assert type(output['kdst'])     == pd.DataFrame
+
+
+def test_collect():
+    the_source    = list(range(0,10))
+    the_collector = collect()
+    the_result    = fl.push(source = the_source,
+                            pipe   = fl.pipe(the_collector.sink),
+                            result = the_collector.future)
+    assert the_source == the_result
+
+
+def test_copy_mc_info_noMC(ICDATADIR, config_tmpdir):
+    file_in  = os.path.join(ICDATADIR, 'run_2983.h5')
+    file_out = os.path.join(config_tmpdir, 'dummy_out.h5')
+    with tb.open_file(file_out, "w") as h5out:
+        copy_mc_info([file_in], h5out, [])
+
+@mark.xfail
+def test_copy_mc_info_repeated_event_numbers(ICDATADIR, config_tmpdir):
+    file_in  = os.path.join(ICDATADIR, "Kr83_nexus_v5_03_00_ACTIVE_7bar_10evts.sim.h5")
+    file_out = os.path.join(config_tmpdir, "dummy_out.h5")
+
+    with tb.open_file(file_out, 'w') as h5out:
+        copy_mc_info([file_in, file_in], h5out, [0,1,0,9])
+        events_in_h5out = h5out.root.MC.extents.cols.evt_number[:]
+        assert events_in_h5out.tolist() == [0,1,0,9]
--- a/invisible_cities/cities/diomira.py
+++ b/invisible_cities/cities/diomira.py
@@ -1,6 +1,6 @@
 """
 -----------------------------------------------------------------------
-                                Diomira                                
+                                Diomira
 -----------------------------------------------------------------------
 
 From Germanic, Teodomiro/Teodemaro: famous among its people.
@@ -37,7 +37,6 @@ from .. reco                    import    wfm_functions as wfm
 from .. sierpe                  import fee              as FE
 from .. core.random_sampling    import NoiseSampler     as SiPMsNoiseSampler
 from .. io.rwf_io               import           rwf_writer
-from .. io.       mcinfo_io     import       mc_info_writer
 from .. io.run_and_event_io     import run_and_event_writer
 from .. io. event_filter_io     import  event_filter_writer
 from .. filters.trigger_filters import TriggerFilter
@@ -47,12 +46,11 @@ from .. evm.pmaps               import S2
 from .. types.ic_types          import minmax
 
 from .. dataflow          import dataflow as fl
-from .. dataflow.dataflow import push
-from .. dataflow.dataflow import pipe
-from .. dataflow.dataflow import fork
 
 from .  components import city
 from .  components import print_every
+from .  components import collect
+from .  components import copy_mc_info
 from .  components import sensor_data
 from .  components import deconv_pmt
 from .  components import WfType
@@ -85,33 +83,38 @@ def diomira(files_in, file_out, compression, event_range, print_mod, detector_db
         write_sipm = fl.sink(RWF(table_name='sipmrwf', n_sensors=sd.NSIPM, waveform_length=sd.SIPMWL                    ), args="sipm_sim")
 
         write_event_info_ = run_and_event_writer(h5out)
-        write_mc_         = mc_info_writer      (h5out) if run_number <= 0 else (lambda *_: None)
         write_evt_filter_ = event_filter_writer (h5out, "trigger", compression=compression)
 
         write_event_info = fl.sink(write_event_info_, args=("run_number", "event_number", "timestamp"   ))
-        write_mc         = fl.sink(write_mc_        , args=(        "mc", "event_number"                ))
         write_evt_filter = fl.sink(write_evt_filter_, args=(              "event_number", "trigger_pass"))
 
         event_count_in = fl.spy_count()
 
-        return push(
-            source = wf_from_files(files_in, WfType.mcrd),
-            pipe   = pipe(fl.slice(*event_range, close_all=True),
-                          event_count_in.spy                    ,
-                          print_every(print_mod)                ,
-                          simulate_pmt_response_                ,
-                          emulate_trigger_                      ,
-                          trigger_pass                          ,
-                          fl.branch(write_evt_filter)           ,
-                          trigger_filter.filter                 ,
-                          simulate_sipm_response_               ,
-                          fork(write_pmt                        ,
-                               write_blr                        ,
-                               write_sipm                       ,
-                               write_mc                         ,
-                               write_event_info)                ),
-            result = dict(events_in     = event_count_in.future,
-                          events_filter = trigger_filter.future))
+        evtnum_collect = collect()
+
+        result = fl.push(source = wf_from_files(files_in, WfType.mcrd),
+                         pipe   = fl.pipe(fl.slice(*event_range, close_all=True)        ,
+                                          event_count_in.spy                            ,
+                                          print_every(print_mod)                        ,
+                                          simulate_pmt_response_                        ,
+                                          emulate_trigger_                              ,
+                                          trigger_pass                                  ,
+                                          fl.branch(write_evt_filter)                   ,
+                                          trigger_filter.filter                         ,
+                                          simulate_sipm_response_                       ,
+                                          fl.branch("event_number", evtnum_collect.sink),
+                                          fl.fork(write_pmt                             ,
+                                                  write_blr                             ,
+                                                  write_sipm                            ,
+                                                  write_event_info)                     ),
+                         result = dict(events_in     = event_count_in.future,
+                                       evtnum_list   = evtnum_collect.future,
+                                       events_filter = trigger_filter.future))
+
+        if run_number <= 0:
+            copy_mc_info(files_in, h5out, result.evtnum_list)
+
+        return result
 
 
 def compute_pe_resolution(rms, adc_to_pes):
--- a/invisible_cities/cities/penthesilea.py
+++ b/invisible_cities/cities/penthesilea.py
@@ -32,7 +32,6 @@ from .. core.system_of_units_c import                units
 from .. reco                   import        tbl_functions as tbl
 from .. reco.  pmaps_functions import          RebinMethod
 from .. io  .          hits_io import          hits_writer
-from .. io  .        mcinfo_io import       mc_info_writer
 from .. io  . run_and_event_io import run_and_event_writer
 from .. io  .          kdst_io import            kr_writer
 from .. io  .  event_filter_io import  event_filter_writer
@@ -43,11 +42,13 @@ from .. dataflow.dataflow import     push
 from .. dataflow.dataflow import     pipe
 
 from .  components import                  city
+from .  components import          copy_mc_info
 from .  components import           print_every
 from .  components import       peak_classifier
 from .  components import   compute_xy_position
 from .  components import       pmap_from_files
 from .  components import           hit_builder
+from .  components import               collect
 from .  components import build_pointlike_event as build_pointlike_event_
 
 @city
@@ -82,31 +83,35 @@ def penthesilea(files_in, file_out, compression, event_range, print_mod, detecto
     event_count_in  = df.spy_count()
     event_count_out = df.spy_count()
 
+    evtnum_collect = collect()
+
     with tb.open_file(file_out, "w", filters = tbl.filters(compression)) as h5out:
 
         # Define writers...
-        write_mc_             = mc_info_writer(h5out) if run_number <= 0 else (lambda *_: None)
-
-        write_mc              = df.sink(write_mc_                  , args=(        "mc", "event_number"             ))
         write_event_info      = df.sink(run_and_event_writer(h5out), args=("run_number", "event_number", "timestamp"))
         write_hits            = df.sink(         hits_writer(h5out), args="hits")
         write_pointlike_event = df.sink(           kr_writer(h5out), args="pointlike_event")
         write_pmap_filter     = df.sink( event_filter_writer(h5out, "s12_selector"), args=("event_number", "pmap_passed"))
 
-        return push(source = pmap_from_files(files_in),
-                    pipe   = pipe(
-                        df.slice(*event_range, close_all=True),
-                        print_every(print_mod)                ,
-                        event_count_in       .spy             ,
-                        classify_peaks                        ,
-                        pmap_passed                           ,
-                        df.branch(write_pmap_filter)          ,
-                        pmap_select          .filter          ,
-                        event_count_out      .spy             ,
-                        df.fork((build_hits           , write_hits           ),
-                                (build_pointlike_event, write_pointlike_event),
-                                                        write_mc             ,
-                                                        write_event_info     )),
-                    result = dict(events_in  = event_count_in .future,
-                                  events_out = event_count_out.future,
-                                  selection  = pmap_select    .future))
+        result = push(source = pmap_from_files(files_in),
+                      pipe   = pipe(df.slice(*event_range, close_all=True)                ,
+                                    print_every(print_mod)                                ,
+                                    event_count_in.spy                                    ,
+                                    classify_peaks                                        ,
+                                    pmap_passed                                           ,
+                                    df.branch(write_pmap_filter)                          ,
+                                    pmap_select          .filter                          ,
+                                    event_count_out      .spy                             ,
+                                    df.branch("event_number", evtnum_collect.sink)        ,
+                                    df.fork((build_hits           , write_hits           ),
+                                            (build_pointlike_event, write_pointlike_event),
+                                                                    write_event_info    )),
+                      result = dict(events_in   = event_count_in .future,
+                                    events_out  = event_count_out.future,
+                                    evtnum_list = evtnum_collect .future,
+                                    selection   = pmap_select    .future))
+
+        if run_number <= 0:
+            copy_mc_info(files_in, h5out, result.evtnum_list)
+
+        return result
--- a/invisible_cities/dataflow/dataflow.py
+++ b/invisible_cities/dataflow/dataflow.py
@@ -239,19 +239,6 @@ def spy_count():
     return FutureSpy(future = pair.future, spy = branch(pair.sink))
 
 
-def collect_into_list(arg=None):
-    def _append(l,e):
-        l.append(e)
-        return l
-    pair = reduce(_append, initial=[])()
-    if arg is None:
-        return FutureSpy(future = pair.future, spy = branch(pair.sink))
-    elif _exactly_one(arg):
-        return FutureSpy(future = pair.future, spy = branch(arg, pair.sink))
-    else:
-        raise ValueError("dataflow.collect_into_list only accepts one argument")
-
-
 def stop_when(predicate):
     @sink
     def stop_when_loop(item):
--- a/invisible_cities/dataflow/dataflow.py
+++ b/invisible_cities/dataflow/dataflow.py
@@ -239,6 +239,19 @@ def spy_count():
     return FutureSpy(future = pair.future, spy = branch(pair.sink))
 
 
+def collect_into_list(arg=None):
+    def _append(l,e):
+        l.append(e)
+        return l
+    pair = reduce(_append, initial=[])()
+    if arg is None:
+        return FutureSpy(future = pair.future, spy = branch(pair.sink))
+    elif _exactly_one(arg):
+        return FutureSpy(future = pair.future, spy = branch(arg, pair.sink))
+    else:
+        raise ValueError("dataflow.collect_into_list only accepts one argument")
+
+
 def stop_when(predicate):
     @sink
     def stop_when_loop(item):
--- a/invisible_cities/io/mcinfo_io.py
+++ b/invisible_cities/io/mcinfo_io.py
@@ -6,6 +6,7 @@ import pandas as pd
 from .. reco            import tbl_functions as tbl
 from .. core            import system_of_units as units
 from .. core.exceptions import SensorBinningNotFound
+from .. core.exceptions import NoParticleInfoInFile
 
 from .. evm.event_model import MCParticle
 from .. evm.event_model import MCHit
@@ -21,6 +22,8 @@ from .. database import load_db as DB
 from typing import Mapping
 from typing import Sequence
 from typing import Tuple
+from typing import List
+from typing import Type
 
 # use Mapping (duck type) rather than dict
 
@@ -166,6 +169,34 @@ class mc_info_writer:
             new_row['region']        = g[3]
             new_row.append()
         self.generator_table.flush()
+
+
+def copy_mc_info(h5in : tb.File, writer : Type[mc_info_writer], which_events : List[int]=None):
+    """Copy from the input file to the output file the MC info of certain events.
+
+    Parameters
+    ----------
+    h5in  : tb.File
+        Input h5 file.
+    writer : instance of class mcinfo_io.mc_info_writer
+        MC info writer to h5 file.
+    which_events : None or list of ints
+        List of IDs (i.e. event numbers) that identify the events to be copied
+        to the output file. If None, all events in the input file are copied.
+    """
+    try:
+        if which_events is None:
+            which_events = h5in.root.MC.extents.cols.evt_number[:]
+        mcinfo = tbl.get_mc_info(h5in)
+        for n in which_events:
+            writer(mctables=mcinfo, evt_number=n)
+    except tb.exceptions.NoSuchNodeError:
+        raise tb.exceptions.NoSuchNodeError(f"No MC info in file {h5in}.")
+    except IndexError:
+        raise IndexError(f"No event {n} in file {h5in}.")
+    except NoParticleInfoInFile as err:
+        print(f"Warning: {h5in}", err)
+        pass
 
 
 def load_mchits(file_name: str,
--- a/invisible_cities/io/mcinfo_io_test.py
+++ b/invisible_cities/io/mcinfo_io_test.py
@@ -17,6 +17,7 @@ from .  mcinfo_io import get_sensor_binning
 from .  mcinfo_io import load_mcsensor_response
 from .  mcinfo_io import load_mcsensor_response_df
 from .  mcinfo_io import mc_info_writer
+from .  mcinfo_io import copy_mc_info
 from .  mcinfo_io import read_mcinfo_evt
 
 from .. core            import system_of_units as units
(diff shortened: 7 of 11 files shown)


jmalbos:
I've modified for now Penthesilea. What other cities should make use of the new function?

gonzaponte:
All of them. `Dorothea` is a special case because back in the day it was decided not to copy the information. I think we can skip that one for now unless there is a good reason to include it.

invisible_cities/io/mcinfo_io.py
```diff
@@ -3,6 +3,9 @@
 import numpy  as np
 import pandas as pd
 
+from .. cities.components import get_run_number
+from .. cities.components import get_mc_info_safe
+
```
gonzaponte: by design, `components` should only be used in the cities. Those functions belong somewhere else.

invisible_cities/cities/components.py
```diff
@@ -277,7 +277,7 @@ def pmap_from_files(paths):
 
             for evtinfo in event_info:
                 event_number, timestamp = evtinfo.fetch_all_fields()
-                yield dict(pmap=pmaps[event_number], mc=mc_info,
+                yield dict(pmap=pmaps[event_number], #mc=mc_info,
```
gonzaponte: Remove instead of commenting out.

invisible_cities/cities/components.py
```diff
@@ -267,7 +267,7 @@ def pmap_from_files(paths):
             try:
                 run_number  = get_run_number(h5in)
                 event_info  = get_event_info(h5in)
-                mc_info     = get_mc_info_safe(h5in, run_number)
+                #mc_info     = get_mc_info_safe(h5in, run_number)
```
gonzaponte: Remove instead of commenting out.

invisible_cities/cities/penthesilea.py
```diff
@@ -85,28 +86,32 @@ def penthesilea(files_in, file_out, compression, event_range, print_mod, detecto
...
+                                    event_count_in       .spy             ,
+                                    classify_peaks                        ,
+                                    pmap_passed                           ,
+                                    df.branch(write_pmap_filter)          ,
+                                    pmap_select          .filter          ,
+                                    event_count_out      .spy             ,
+                                    df.fork((build_hits           , write_hits           ),
+                                            (build_pointlike_event, write_pointlike_event),
+                                                                    events_selected       ,
+                                                                    write_event_info     )),
```
gonzaponte: maybe change the order so it doesn't break the symmetry between the writes?

invisible_cities/cities/penthesilea.py
```diff
@@ -85,28 +86,32 @@ def penthesilea(files_in, file_out, compression, event_range, print_mod, detecto
...
-                        event_count_out      .spy             ,
-                        df.fork((build_hits           , write_hits           ),
-                                (build_pointlike_event, write_pointlike_event),
-                                                        write_mc             ,
-                                                        write_event_info     )),
-                    result = dict(events_in  = event_count_in .future,
-                                  events_out = event_count_out.future,
-                                  selection  = pmap_select    .future))
+        event_ids = []
+        events_selected = df.sink(event_ids.append, args=("event_number"))
```
gonzaponte: Mmm. This is **not** wrong, but we tend to use futures for this kind of things. My approach would be to use the `reduce` component, but I can't think of a good reason not to take this approach. The only (very minor thing) is that using something named `result.events_selected` (or similar) would be slightly more indicative of what is happening, but it's not a big deal at all. I kinda like the fact that you can specify the argument to take :+1:

gonzaponte: The parenthesis around `"event_number" can be removed.

invisible_cities/cities/penthesilea.py
```diff
@@ -85,28 +86,32 @@ def penthesilea(files_in, file_out, compression, event_range, print_mod, detecto
...
+                                    event_count_out      .spy             ,
+                                    df.fork((build_hits           , write_hits           ),
+                                            (build_pointlike_event, write_pointlike_event),
+                                                                    events_selected       ,
+                                                                    write_event_info     )),
+                      result = dict(events_in  = event_count_in .future,
+                                    events_out = event_count_out.future,
+                                    selection  = pmap_select    .future))
+
+        copy_mc_info(files_in, h5out, event_ids)
```
gonzaponte: I think it is safer and more verbose for someone trying to understand the code to make `copy_mc_info` run only when the run number is `<= 0` and not let it figure it out from the run number in the file

invisible_cities/io/mcinfo_io.py
```diff
@@ -168,6 +172,37 @@ def __call__(self, mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
...
+                 List of events (event_id) to be copied.
+    """
+
+    writer = mc_info_writer(h5out)
+
+    for f in files_in:
+        try:
+            with tb.open_file(f, "r") as h5in:
+                run_number = get_run_number(h5in)
+                mc_info    = get_mc_info_safe(h5in, run_number)
```
gonzaponte: Related to what I said in another comment. Do not let it figure it out. This is a base function and I think it is safer and better to make it work only for the case it is meant to be used. In my opinion, it should complain if there is not a MC table in the input file.

invisible_cities/io/mcinfo_io.py
```diff
@@ -168,6 +172,37 @@ def __call__(self, mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
...
+                for n in event_list[:]:
+                    try:
+                        writer(mctables=mc_info, evt_number=n)
+                        # We succeeded copying the event. Remove its id
+                        # from the event list
+                        event_list.remove(n)
+                    except IndexError:
+                        # The event id was not found in the current file
+                        continue
+        except:
```
gonzaponte: catch specific exceptions and, in case they are too generic, put some comment about what you are catching.

invisible_cities/io/mcinfo_io.py
```diff
@@ -168,6 +172,37 @@ def __call__(self, mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
...
+                mc_info    = get_mc_info_safe(h5in, run_number)
+                for n in event_list[:]:
+                    try:
+                        writer(mctables=mc_info, evt_number=n)
+                        # We succeeded copying the event. Remove its id
+                        # from the event list
+                        event_list.remove(n)
+                    except IndexError:
+                        # The event id was not found in the current file
+                        continue
```
gonzaponte: A few things:
- There is nothing to ensure that `event_list` is empty at the end of all loops... I wonder if we can make sure of that.
- Maybe we can get rid of this `try/except` clause by using `set`s. If you take the event numbers of the input file and make the intersection with `event_list` you would only run over the event numbers in that file without having to check them all.
- I think we don't have any protection against repeated event numbers in the writer, but it's something to think about (not for this PR, though)

gonzaponte (commented):
We need some tests for the new function :)

invisible_cities/io/mcinfo_io.py
```diff
@@ -168,6 +172,37 @@ def __call__(self, mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
         self.generator_table.flush()
 
 
+def copy_mc_info(files_in, h5out, event_list):
+    """
```
mmkekic: I think this function can go to components, it seems not general enough for io module to me. Also add type hints

invisible_cities/io/mcinfo_io.py
gonzaponte: Apparently, this comment was confusing. I meant that `copy_mc_info` should not assume anything about whether the sign of run number means something or not. I would like `copy_mc_info` to assume whatever files come in to be MC files and have the relevant MC tables. If something goes wrong it should crash (there should be an explicit check to make the crash more obvious). Let the user of this function (aka the cities or the components) handle that situation (that is, calling `copy_mc_info` only for MC files).

invisible_cities/cities/penthesilea.py
gonzaponte: Another side effect of calling `copy_mc_info` always is that the MC tables are created even for non-MC files!

invisible_cities/io/mcinfo_io.py
jmalbos: Not sure what am I suppose to do then; notice that those are pre-existing functions. Should I move them to` mcinfo_io`?

gonzaponte: Well, as discussed below, you will not need them, you will need `get_mc_info`, which is in reco.tbl_function. I think that function should also be moved from there to mcinfo_io, but maybe that's for another PR.

invisible_cities/cities/penthesilea.py
jmalbos: The main reason why I didn't use `reduce` in the end is that it forced me to write something like this:
`add_to_list = df.reduce(lambda l,x: l+[x], initial=[])`
It works, but it looks a bit ugly. I can change it though; it isn't particularly important.

invisible_cities/cities/penthesilea.py
jmalbos: OK!

invisible_cities/cities/penthesilea.py
gonzaponte: No, I like this approach very much!

jacg: TL:DR the `sink` approach is fine, but just for fun and education ...

> `add_to_list = df.reduce(lambda l,x: l+[x], initial=[])`

... is `O(N)` per insertion (where `N` is the length of the stream) and hence `O(N^2)` in total. You can *almost* do it in `amortized O(1)` per insertion, `O(N)` total with `add_to_list = df.reduce(list.append, initial=[])` except that `list.append` returns `None` rather than the list itself, which is a Python wart, as far as I am concerned, and throws a spanner in the works here.

But you could hack around it with `df.reduce(lambda l,x: l.append(x) or l, initial=[])`, which is a bit ugly, so, you could provide

````python
def append(list_, element):
    list_.append(element)
    return list_
````
which would allow you to say
````python
df.reduce(append, initial=[])
````
cleanly, and keep using the futures pattern that we are used to seeing, and thus address @gonzaponte's comment that

> using something named `result.events_selected` (or similar) would be slightly more indicative of what is happening

Comment (gonzaponte) [reply invisible_cities/cities/penthesilea.py]: >  except that list.append returns None rather than the list itself

For that, we had thought of using `np.append` which does return the object, at the expense of making a copy of the array every time.

Comment (jacg) [reply invisible_cities/cities/penthesilea.py]: Which makes the time-cost of the whole process quadratic rather than linear in the length of the source.

Comment (gonzaponte) [reply invisible_cities/cities/penthesilea.py]: Yes, of course. I wasn't suggesting that it was a good choice, but rather an attempt to leave some trace that this was considered :)

Comment (jacg) [reply invisible_cities/cities/penthesilea.py]: Yup, good.

Just one more reminder: it's great to theorize, but it's also good to back the theory up with measurement. For which `timeit` (and the way it is embedded in IPython) can be very useful:

````
In [1]: import numpy as np                                                                                                                                                                         

In [2]: from functools import reduce                                                                                                                                                               

In [3]: def a(N): 
   ...:     return reduce(np.append, range(N), np.array([])) 
   ...:

In [4]: def b(N): 
   ...:     return reduce(lambda l,e: l.append(e) or l, range(N), []) 
   ...:

In [5]: assert (a(10**3) == b(10**3)).all()

In [6]: for fn in (a,b): 
   ...:     for p in range(3,6): 
   ...:         %timeit fn(10**p) 
   ...:     print('='*60) 
   ...:

11.9 ms ± 74.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
143 ms ± 1.67 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
4.97 s ± 41.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
============================================================
211 µs ± 12.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
2.12 ms ± 32.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
21.4 ms ± 155 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
============================================================

In [7]: np.array([143, 4970]) / 11.9
Out[7]: array([ 12.01680672, 417.64705882])

In [8]: np.array([2120, 21400]) / 211
Out[8]: array([  10.04739336, 101.42180095])

````

In summary, the `lambda l,e: l.append(e) or l` approach is significantly faster for nontrivial input size: (211, 2120, 21400) vs (11900, 143000, 4970000); and grows more slowly with increasing size: (1, 10, 111) vs (1, 12, 418).

File: invisible_cities/cities/penthesilea.py [hunk 11]
```diff
@@ -83,6 +82,14 @@ def penthesilea(files_in, file_out, compression, event_range, print_mod, detecto
     event_count_in  = df.spy_count()
     event_count_out = df.spy_count()
 
+    def append(l, e):
+        l.append(e)
+        return l
+
+    add_to_list = df.reduce(append, initial=[])
```
jacg: I'd be tempted to

1. rename `add_to_list` to `collect_into_list`
2. contribute it to `dataflow` (If you do that, then rename `append` to `_append` when you put it into `dataflow`, or just use the `lambda` version; and, of course, add a documenting test.)

jmalbos:
- I've added a new function `collect_into_list` to `dataflow`, following @jacg's suggestion and with the invaluable help of @mmkekic. 
- I've added as well three tests documenting the use of the new function.
- Penthesilea has been amended to make use of the new function to create a list with the selected events.

invisible_cities/io/mcinfo_io.py
jmalbos: Done.

invisible_cities/cities/penthesilea.py
jmalbos: Done.

mmkekic:
> Is it really worthwhile? Is it better to have to come up with and remember a name for branch(collect_into_list) (and have users remember that it exists) rather than just writing branch(collect_into_list)?

The main annoyance is that `reduce` doesnt accept arguments, and the only way (to my knowledge) to constrain the data flowing through `reduce` is to use (string, pipe...) syntax, that, as you said, is not well documented... Other possibilities is to define a sink (as @jmalbos  did in  6c6b397ade3545f7ca49e653b7415437f2c9137e , initializing empty list and defining a sink that appends to it) in every city that needs it, or modify `reduce` to accept arguments. In that case I believe it would be clear enough to use `branch(collect_into_list)`, where now `collect_into_list` could have optional arguments. 

> If it is worthwhile doing it for this particular sink, why not for any of the others?

Notice that `spy_count` is exactly that (sink +  branch) (Also isn't `count` just a specific `reduce(lambda count, _ : count+1)`, but instead it has its own implementation?)
On the other hand, I agree that it is not clear if this particular spy is general enough to be in dataflow or should go to components and be linked to IC.

mmkekic:
> However, the parallel-flows-in-dicts idea was added to support important patterns required by IC, and `dataflow` contains some features (`args`, `out` and `item` in `map`; `args` in the `filter`s; the optional strings at the beginning of pipes) which are completely undocumented!
> 
> Let's see if I can find the time to fix this.

Guess the third option is to fix unclear parts of dataflow, add documentation and maybe polish it by raising appropriate readable errors if the input types or syntax is wrong etc, but I believe that can not be done without your active involvement...

jacg:
> The main annoyance is that `reduce` doesnt accept arguments, and the only way (to my knowledge) to constrain the data flowing through reduce is to use (string, pipe...) syntax

Is using (selector-string, rest-of-pipe) really that bad? Do you really gain enough with the abstraction for it to be worth while?

> that, as you said, is not well documented...

That definitely needs fixing in the dataflow documentation, but the selector-strings are used frequently in IC, so it's documented there by example, and I get the impression that most users get their dataflow documentation by looking at how it's used in the cities, rather than the actual documentation tests. Which is not to say that the dataflow documentation doesn't need improving: it does.

BTW, unless I've missed some detail, it should be trivial to extend the selector-string-at-beginning-of-pipe feature to admit a selector tuple:

1. Would this be useful? 
2. This would make tuples serve as both implicit pipes in fork and as selectors at the beginning of pipes: would this overloading make it confusing to read?

> Other possibilities is to define a sink (as @jmalbos did in 6c6b397 , initializing empty list and defining a sink that appends to it) in every city that needs it,

Which is fine as a one-off, but if it's going to be used frequently, I think it should be provided as a ready-to-use component, in which case it should be symmetric with others which play a similar role, at which point the `reduce` version makes sense. This was @gonzaponte's original point.

> or modify `reduce` to accept arguments.

That might be a useful thing to do. At the moment I'm not convined that it's worth the effort and increase in complexity ...

> In that case I believe it would be clear enough to use `branch(collect_into_list)`, where now `collect_into_list` could have optional arguments.

... which would allow us to write

    branch(collect_into_list(arg='banana'))

as opposed to

    branch('banana', collect_into_list)

which is possible with the most simple implementation of `collect_into_list`.

I'm not at all convinced that the former is better than the latter, and even less convinced that it's worth the extra effort and complexity of implementation.

Fundamentally, if you provide a bunch of very simple, orthogonal and composable components, you can stick them together to make the more complex ones trivially. If you create components which combine a bunch of features, they are less reusable, more complicated to test, more difficult to remember.

Rather than making such components, we should aim to optimize the composition features, so that creating something by composition on the fly is just as cheap or cheaper (syntactically) than using a pre-made more complex component. I think that `dataflow` does this quite well already (which is not to say that it can't be improved) and in the case of `branch('banana', collect_into_list)` which is a composition of 3 features

1. `branch`
2. selector string at the beginning of pipes
3. a sink

the direct composition is (IMO) cleaner than the special-case pre-made component.

Incidentally, lest I appear to be contradicting myself: earlier I said that @jmalbos' 6c6b397 approach (if used) should be packed up as a utility, and here I'm saying that making a utility is not worthwhile. Here is my reasoning: the former needs to be defined out of line, the latter is trivally implementable inline. That's a big qualitative difference, which suggest to me that providing the former as a pre-made solution is worthwhile, while doing so for the latter might not be worth it, or even more trouble than it's worth.

> (Also isn't `count` just a specific `reduce(lambda count, _ : count+1)`, but instead it has its own implementation?)

Yes. `count` could be implemented in terms of `reduce`. Its hand-coded implementation is an historical artefact. (Feel free to re-implement it in terms of reduce. But if you do, please keep the old implementation under an alternative (probably private) name, and run both through the same tests. I think that having the direct comparison of the two available is useful.)

And I think that `collect_into_list` should be almost identical to `count` (in terms of functionality, not implemetation): it should be a simple specific use of reduce, without the other bells and whistles (args) which are adequately catered for by the orthogonal args selection features of pipes.

So why do `map` and `filter` have arg-handling features built into them. There's also a good reason for that, but this message is already far too long, so I'll leave that for now.

> Notice that `spy_count` is exactly that (sink + branch)

Yes, but notice how simple it is. It is minimal and composable. The current implementation of `collect_into_list` is complex and less composable.

By all means implement both `collect_into_list` and `spy_collect_into_list`. Both can be done in far fewer lines in total, that the current implementation of `collect_into_list`.

jacg:
> Guess the third option is to fix unclear parts of dataflow,

Which parts would you say fall in this category?

> add documentation

Definitiley a good idea.

> and maybe polish it by raising appropriate readable errors if the input types or syntax is wrong etc, 

Would be useful. Do you have examples of situations?

mmkekic:
> > Guess the third option is to fix unclear parts of dataflow,
> 
> Which parts would you say fall in this category?
> 
My original comment should have been ' fix unclear parts of dataflow : add documentation and maybe... '

> > add documentation
> 
> Definitiley a good idea.
> 
> > and maybe polish it by raising appropriate readable errors if the input types or syntax is wrong etc,
> 
> Would be useful. Do you have examples of situations?

For example if you try to pass a tuple as a selector the error is :
`TypeError: 'tuple' object is not callable`
and I believed that something of 
`TypeError: selectors can only be of type str`
would be more clear

>  BTW, unless I've missed some detail, it should be trivial to extend the selector-string-at-beginning-of-pipe feature to admit a selector tuple:

Well, trivial  for you :)

> 
> Would this be useful?
> This would make tuples serve as both implicit pipes in fork and as selectors at the beginning of pipes: would this overloading make it confusing to read?

I agree, the only case would be `reduce` with selected elements, since maps, filters and sinks already accept arguments.

Is your suggestion then to have 

```
def collect_into_list_sink():
    def append(l,e):
        l.append(e)
        return l
   return reduce(append, initial=[])()
   ```
and call it inside the city flow as 
```
collect_evt_number_into_list = collect_into_list_sink()
df.branch('eventinfo', collect_evt_number_into_list.sink)
```

or you prefer to also implement a spy based on the collect_into_list_sink that would accept arguments (basically divide already existing collect_into_list in sink and spy)?

 Also do you think collect_into_list should stay in dataflow or go into components?

**Comment by jacg** [comment]
> Is your suggestion then to have
>
>````python
>def collect_into_list_sink():
>    def append(l,e):
>        l.append(e)
>        return l
>   return reduce(append, initial=[])()
>````
> and call it inside the city flow as `df.branch('eventinfo', collect_into_list_sink())`

Well, more like `events = collect_into_list_sink()` followed by `df.branch('eventinfo', events.sink)` and `result = dict(events = events.future)`.

> or you prefer to also implement a spy based on the collect_into_list_sink that would accept arguments (basically divide already existing collect_into_list in sink and spy)?

I think it should parallel the situation we have with `count` and `spy_count`:

````python
def collect():
    return reduce(lambda l,e: l.append(e) or l, initial=[])

def spy_collect():
    future, sink = collect()
    return FutureSpy(future, branch(sink))
````
though written out more verbosely than I have done here.

> Also do you think `collect_into_list` should stay in dataflow or go into components?

It's so obvious and fundamental that I think it would be perfectly reasonable to stick it all into dataflow.

And, as hinted in my example above, I'd call them simply `collect` and `spy_collect`.

Actually, I think there's an argument for adding something along the lines of:

````python
def sink_spy(sink):
    future, sink = sink()
    return FutureSpy(future=future, spy=branch(sink))
````

Then you could provide `count` and `collect` and not bother with `spy_count` and `spy_collect` because they could be spelt: `sink_spy(count)` and `sink_spy(collect)`. This reflects my earlier point about making combination of basic components so cheap that it obviates the need for packaging up such combinations.

**Comment by mmkekic** [comment]
> > Is your suggestion then to have
> > ```python
> > def collect_into_list_sink():
> >    def append(l,e):
> >        l.append(e)
> >        return l
> >   return reduce(append, initial=[])()
> > ```
> > 
> > 
> > and call it inside the city flow as `df.branch('eventinfo', collect_into_list_sink())`
> 
> Well, more like `events = collect_into_list_sink()` followed by `df.branch('eventinfo', events.sink)` and `result = dict(events = events.future)`.
> 

I corrected myself but obviously not fast enough! (:

> Actually, I think there's an argument for adding something along the lines of:
> 
> ```python
> def sink_spy(sink):
>     future, sink = sink()
>     return FutureSpy(future=future, spy=branch(sink))
> ```
> 
> Then you could provide `count` and `collect` and not bother with `spy_count` and `spy_collect` because they could be spelt: `sink_spy(count)` and `sink_spy(collect)`. This reflects my earlier point about making combination of basic components so cheap that it obviates the need for packaging up such combinations.

But then we wouldnt be able to pass arguments to the collect sink right? The sink_spy would have to accept optional arguments then or am I missing something?

**Comment by jacg** [comment]
> The sink_spy would have to accept optional arguments then [...]?

Yes. I think that there's a strong argument for `sink_spy` to accept `args` à la `map`. Is that what you meant?

This conversation is going way off-topic for this PR. **I'm going to extract dataflow into a separate repo/package**[*] outside of IC, and we can continue this conversation there.

We'll need a name. A seach for 'dataflow' on PyPI comes up with a bazillion hits. Suggestions welcome. Maybe @jjgomezcadenas can come up with something poetic. (Key concepts to use for inspiration: data flows/networks/graphs/streams/pipelines; **composability** of simple and orthogonal functions; discouraging physicists' 2000-line functions; improvement of signal to noise ratio in code; code clarity; liquid, for historical IC reasons; ...)

As for this PR ... I propose you write
````python
collect = lambda : reduce(lambda l,e: l.append(e) or l, initial=[])
````
(or some more verbose equivalent) inside the city's definition, and use it. Leave a TODO about replacing it once dataflow has been extracted, revamped, cleaned up, and published.

**Comment by gonzaponte** [comment]
> As for this PR ... I propose you write
>
> `collect = lambda : reduce(lambda l,e: l.append(e) or l, initial=[])`
>
> or some more verbose equivalent) inside the city's definition

Just to clarify, this is going to be used in many cities so we need it as a standard component. Then, each city should take care of importing and introducing it in the pipeline.

**Comment by jmalbos** [comment]
> > As for this PR ... I propose you write
> > `collect = lambda : reduce(lambda l,e: l.append(e) or l, initial=[])`
> > or some more verbose equivalent) inside the city's definition
> 
> Just to clarify, this is going to be used in many cities so we need it as a standard component. Then, each city should take care of importing and introducing it in the pipeline.

Should I add it then to components?

**Comment by gonzaponte** [comment]
> Should I add it then to components?

I would say so

**Comment by jmalbos** [comment]
> > Should I add it then to components?
> 
> I would say so

OK… Can you please clarify where should I add it? Your comment appeared to suggest that @jacg's proposal of adding the function inside each city's definition was inconvenient.

**Comment by gonzaponte** [comment]
> Can you please clarify where should I add it? 

Since that line is somewhat complex and that it needs to be repeated in every city, I suggest to add it to components.py with a proper `def` declaration and then import it in each city. 

> Your comment appeared to suggest that @jacg's proposal of adding the function inside each city's definition was inconvenient.

We can wait for @jacg to comment on that to see if there is something wrong about it, maybe I missed something during the discussion.

**Comment by jmalbos** [comment]
> > Can you please clarify where should I add it?
> 
> Since that line is somewhat complex and that it needs to be repeated in every city, I suggest to add it to components.py with a proper `def` declaration and then import it in each city.
> 
> > Your comment appeared to suggest that @jacg's proposal of adding the function inside each city's definition was inconvenient.
> 
> We can wait for @jacg to comment on that to see if there is something wrong about it, maybe I missed something during the discussion.

It's alright; let's add it in `components`. I had misread your comment ("I would say no", I thought you wrote), and I was puzzled about what you meant.

**Comment by jacg** [comment]
If it is going to be used in multiple cities right now, then write it once, and import it in each city that needs it. Don't overengineer it: do the simlpest thing for now.

Are you going to be using the spy version immediately? If not, forget it; if you will be using in *now* then go with something like:

````python

def spy_sink(sink, args=None):
    p = sink()
    if args is None: spy = branch(      p.sink)
    else           : spy = branch(args, p.sink)
    return FutureSpy(future = p.future, spy = spy)
````

This should immediately work for `arg` being a single string selector, and should automatically start working for a tuple of selectors, if and when that feature is added to pipes (which, in spite of @mmkekic's doubts, should be trivial to implement for any of you :-) [Hint: add two lines to `string_to_pick` (which should then probably be renamed to `items_to_pick`.)]

If you're going to do any of this, then it might as well go straight into dataflow, rather than components.

We should spend more time and care thinking about this interface, but for the purposes of *this PR* this should be fine. That care should be taken in the new extracted dataflow.

**Comment by jacg** [comment]
There is a copy of dataflow at https://github.com/jacg/dataflow.

I suggest that any further discussion of dataflow be continued over there.

The idea is that this gradually become a fully-fledged separate package with decent documentation (and a catchy name), on which IC can depend.

**Comment by jmalbos** [comment]
Let's try again:

- Added a new function `components.collect`. It's written in the more verbose style discussed above because I find that the expression `l.append(e) or l` is somewhat cryptic for a casual reader.

- Penthesilea has been modified to make use of the new function.

I will add tomorrow the corresponding tests and will modify the other cities if you are happy with the approach described above.

File: invisible_cities/cities/components.py [hunk 12]
```diff
@@ -152,6 +152,17 @@ def print_every_loop(target):
     return print_every_loop
 
 
+def collect():
+    """
+    Collect into a list an argument (selector string) passed through a pipe.
+    The list can be retrieved from a future.
```
jacg: I wouldn't mention the selector string, as it's completely orthogonal: `collect` collects whatever comes down the pipe, which might or might not have been selected with a selector string.

Saying that it can be retrieved from a future also seems a bit odd, because this is absolutely standard behaviour for any sink that produces a value. So, either don't mention the future at all, or describe fully how it returns a a future and a sink pair.

Also, https://www.python.org/dev/peps/pep-0257/#id17

OTOH, I wouldn't worry too much about all this, as it will be documented properly in the stand-alone dataflow package. So you might as well leave it as it is, for the purposes of this PR.

invisible_cities/io/mcinfo_io.py
mmkekic: > Maybe we can get rid of this try/except clause by using sets. If you take the event numbers of the input file and make the intersection with event_list you would only run over the event numbers in that file without having to check them all.

This function will be modified in #693, so I dont think we need to fine tune it too much. However this implementation has 2 issues:

1. it sneakily modifies event_list
2. in case event numbers are the same across different input files just the first occurrence (file, event number) will be written.  I am not sure how big of an issue it is cause we do not have any protection anywhere against repeated event numbers in our readers and writers ...

jmalbos: I can fix somewhat the _sneaky_ behavior of the function. However, there were several things about `event_list` I wasn't sure I could assume; namely:
- Do event ids come in order?
- Is that order correlated with the `files_in` order?
- Can event ids repeat?
I understand that the answer to the first two is 'no', and that we really don't know the answer of the third one, correct?

gonzaponte: > Do event ids come in order?

Not necessarily

> Is that order correlated with the `files_in` order?

They will keep the order in which they are written within the same file, and the order of the input files is kept. That is if you have `file_1` with events `[2, 6, 4]` and `file_2` with events `[1, 3, 5]`
you will receive `[2, 6, 4, 1, 3, 5]`

> Can event ids repeat?

They can, in theory. It shouldn't happen, but there is nothing preventing it. For instance if you run over two different MC productions at the same time, the event numbers will repeat unless you take care of it explicitly. It can also happen if someone doesn't take care of organizing the event number in a `nexus` run that is split into many jobs. So far we haven't had issues, but I think it can happen in #641.

paolafer: So far, we have assumed that event IDs don't repeat, because we were using dictionaries in the event data model, correct me if I'm wrong. I'm not sure this is the case anymore, so we may want to discuss this point. Not to repeat event IDs is particularly cumbersome when we do event mixing in the background model simulations, so, if we can drop this requirement, it may be a simplification.

mmkekic: Uhh, this is getting complicated... There are couple of options:

- we assume events ids and files_in order is correlated (in that case removing the event_id from the list should work well even in the case of duplicates)
- we write all combinations of files_in, events_ids that are possible (probably not very useful)
- we move this function to components and not worry about general usage 

What is your opinion @gonzaponte ?

mmkekic: Now that i think of it the safest would be to have a dictionary `{file_in : event_list} `, that way there can not be confusion. However we will not be able to read this file since as @paolafer  said we do expect to have unique IDs at the reader level...

gonzaponte: How about making this function for a single file and expand it for many in components? This would make the utility available in general (which is probably useful) while keeping these intricacies away from the base functions. I don't think we need to deal with repeated events or the ordering here. My feeling is that we should assume that order doesn't matter and that there shouldn't be repeated events for practical purposes. In case, we can add a tool that ensures that somewhere.

jmalbos: > How about making this function for a single file and expand it for many in components? This would make the utility available in general (which is probably useful) while keeping these intricacies away from the base functions. I don't think we need to deal with repeated events or the ordering here. My feeling is that we should assume that order doesn't matter and that there shouldn't be repeated events for practical purposes. In case, we can add a tool that ensures that somewhere.

But how do I correlate files with selected events then? The main loop in push is reading all input files together.

gonzaponte: I propose to do the following:

In `mcinfo_io.copy_mc_info`
- Make this function take `h5in`,`h5out` and `which_events` (unofficial names)
- If `which_events` is `None` (probably default), read the event numbers from the input file and assign it to `which_events`. This is a detail that doesn't affect the current usage, but that it would be good to have.
- Assume that each event in `which_events` is stored in the file.
- For each event in `which_events`, copy it.

In `components.copy_mc_info` (unofficial name):
- take `filenames_in`, `h5out` and `all_filtered_events`
- for each filename
  - open the file and call it `h5in` (unofficial name)
  - read the event numbers in that file and make the intersection with `all_filtered_events`. Let's call that `filtered_events_in_file`
  - call `mc_info_io.copy_mc_info(h5in, h5out, filtered_events_in_file)`

With this:
- Duplicated event numbers in different files are written
- The order of the files is irrelevant
- The order of the events within a file matters (as it has mattered so far), but hopefully this will be addressed in the future

jmalbos: Sounds good, @gonzaponte.

mmkekic: Also dont forget to change get_mc_info_safe to get_mc_info directly!

invisible_cities/cities/components.py
```diff
@@ -152,6 +154,39 @@ def print_every_loop(target):
...
 
+def collect():
+    """Return a future/sink pair for collecting streams into a list."""
+    def append(l,e):
+        l.append(e)
+        return l
+    return fl.reduce(append, initial=[])()
+
+
+def copy_mc_info(files_in : List[str], h5out: tb.File, event_numbers: List[int]):
```
gonzaponte: return type hint `-> None`

invisible_cities/cities/components.py
```diff
@@ -152,6 +154,39 @@ def print_every_loop(target):
...
+
+    writer = mcinfo_io.mc_info_writer(h5out)
+
+    for f in files_in:
+        with tb.open_file(f, "r") as h5in:
+            try:
+                event_numbers_in_file = h5in.root.MC.extents.cols.evt_number[:]
+                event_numbers_to_copy = set(event_numbers_in_file).intersection(event_numbers)
+                mcinfo_io.copy_mc_info(h5in, writer, list(event_numbers_to_copy))
+            except tb.exceptions.NoSuchNodeError:
```
gonzaponte, line 187: I don't think this exception should pass silently. If there is no data, it should complain.

invisible_cities/cities/components.py
```diff
@@ -152,6 +154,39 @@ def print_every_loop(target):
...
+    """
+
+    writer = mcinfo_io.mc_info_writer(h5out)
+
+    for f in files_in:
+        with tb.open_file(f, "r") as h5in:
+            try:
+                event_numbers_in_file = h5in.root.MC.extents.cols.evt_number[:]
+                event_numbers_to_copy = set(event_numbers_in_file).intersection(event_numbers)
+                mcinfo_io.copy_mc_info(h5in, writer, list(event_numbers_to_copy))
```
```suggestion
                mcinfo_io.copy_mc_info(h5in, writer, sorted(event_numbers_to_copy))
```
How aout this? I am afraid the set returns them in a random order

invisible_cities/cities/components.py
```diff
@@ -305,7 +332,7 @@ def hits_and_kdst_from_files(paths: List[str]) -> Iterator[Dict[str,Union[HitCol
             for evtinfo in event_info:
                 event_number, timestamp = evtinfo.fetch_all_fields()
                 hits = hits_from_df(hits_df.loc[hits_df.event == event_number])
-                yield dict(hits = hits[event_number], kdst = kdst_df.loc[kdst_df.event==event_number], mc=mc_info, run_number=run_number,
+                yield dict(hits = hits[event_number], kdst = kdst_df.loc[kdst_df.event==event_number], run_number=run_number,
```
gonzaponte: can we take the chance to split this line?

invisible_cities/cities/components.py
```diff
@@ -305,7 +332,7 @@ def hits_and_kdst_from_files(paths: List[str]) -> Iterator[Dict[str,Union[HitCol
...
                            event_number=event_number, timestamp=timestamp)
             # NB, the monte_carlo writer is different from the others:
```
gonzaponte: You forgot to remove this comment that you removed from other readers.

invisible_cities/cities/components_test.py
```diff
@@ -160,3 +164,30 @@ def test_hits_and_kdst_from_files(ICDATADIR):
...
     assert type(output['kdst'])     == pd.DataFrame
+
+
+def test_collect():
+    the_source    = list(range(0,10))
+    the_collector = collect()
+    fl.push(source = the_source,
+            pipe   = fl.pipe(the_collector.sink))
+
+    assert the_source == the_collector.future.result()
```
gonzaponte: You are not meant to call the `result` method yourself. Rather you should use the `result` argument of `push` to define your futures. Something like
```
result = fl.push(source = ..., pipe =..., result = the_collector.future)

assert the_source == result
```
This is (still) undocumented, but `result` (the variable) in this snippet can be different things depending on what you write in the `result` argument:
- If you write a single future, it `result` will be the value of that future
- If you write a tuple of futures, it will be a tuple with the value of each future
- If you write a mapping of futures, it will be a namespace with the value of each future

invisible_cities/cities/components_test.py
```diff
@@ -160,3 +164,30 @@ def test_hits_and_kdst_from_files(ICDATADIR):
...
+
+
+def test_copy_mc_info_noMC(ICDATADIR, config_tmpdir):
+    file_in  = os.path.join(ICDATADIR, 'run_2983.h5')
+    file_out = os.path.join(config_tmpdir, 'dummy_out.h5')
+    with tb.open_file(file_out, "w") as h5out:
+        copy_mc_info([file_in], h5out, [])
```
gonzaponte, line 182: Is this meant to check that it doesn't fail without MC tables?
As I said before I think it should and the test should reflect that.

invisible_cities/cities/components_test.py
```diff
@@ -160,3 +164,30 @@ def test_hits_and_kdst_from_files(ICDATADIR):
...
+
+@mark.xfail
+def test_copy_mc_info_repeated_event_numbers(ICDATADIR, config_tmpdir):
+    file_in  = os.path.join(ICDATADIR, "Kr83_nexus_v5_03_00_ACTIVE_7bar_10evts.sim.h5")
+    file_out = os.path.join(config_tmpdir, "dummy_out.h5")
+
+    with tb.open_file(file_out, 'w') as h5out:
+        copy_mc_info([file_in, file_in], h5out, [0,1,0,9])
+        events_in_h5out = h5out.root.MC.extents.cols.evt_number[:]
+        print(events_in_h5out)
```
gonzaponte: Remove print

invisible_cities/cities/diomira.py
```diff
@@ -85,33 +83,37 @@ def diomira(files_in, file_out, compression, event_range, print_mod, detector_db
...
+                                          event_count_in.spy                            ,
+                                          print_every(print_mod)                        ,
+                                          simulate_pmt_response_                        ,
+                                          emulate_trigger_                              ,
+                                          trigger_pass                                  ,
+                                          fl.branch(write_evt_filter)                   ,
+                                          trigger_filter.filter                         ,
+                                          simulate_sipm_response_                       ,
+                                          fl.branch("event_number", evtnum_collect.sink),
+                                          fl.fork(write_pmt                             ,
```
gonzaponte, line 106: Why not have it as one more pipe in the fork?

invisible_cities/cities/diomira.py
```diff
@@ -85,33 +83,37 @@ def diomira(files_in, file_out, compression, event_range, print_mod, detector_db
...
+                                                  write_blr                             ,
+                                                  write_sipm                            ,
+                                                  write_event_info)                     ),
+                         result = dict(events_in     = event_count_in.future,
+                                       evtnum_list   = evtnum_collect.future,
+                                       events_filter = trigger_filter.future))
+
+        copy_mc_info(files_in, h5out, result.evtnum_list) if run_number <= 0 else None
```
gonzaponte: Mmm this syntax is a bit fishy... Don't use inline `if/else` for something that is not being collected anywhere.
Better
```
if run_number <= 0:
    copy_mc_info(...)
```

invisible_cities/dataflow/dataflow.py
```diff
@@ -239,6 +239,19 @@ def spy_count():
...
+        return l
+    pair = reduce(_append, initial=[])()
+    if arg is None:
+        return FutureSpy(future = pair.future, spy = branch(pair.sink))
+    elif _exactly_one(arg):
+        return FutureSpy(future = pair.future, spy = branch(arg, pair.sink))
+    else:
+        raise ValueError("dataflow.collect_into_list only accepts one argument")
+
+
```
gonzaponte: Is this a leftover from the discussion or are we keeping it for some reason?

gonzaponte: Same for the tests affecting this

invisible_cities/io/mcinfo_io.py
```diff
@@ -3,9 +3,12 @@
 import numpy  as np
 import pandas as pd
 
+from copy import deepcopy
+
```
gonzaponte: why this import? I guess is some leftover from different attempts

invisible_cities/io/mcinfo_io.py
```diff
@@ -168,6 +173,30 @@ def __call__(self, mctables: (tb.Table, tb.Table, tb.Table, tb.Table),
...
+    """
+
+    try:
+        if which_events is None:
+            which_events = h5in.root.MC.extents.cols.evt_number[:]
+        mcinfo = tbl.get_mc_info(h5in)
+        for n in which_events:
+            writer(mctables=mcinfo, evt_number=n)
+    except NoParticleInfoInFile:
+        pass
```
gonzaponte: Again, I don't think this should pass silently. I think we want it to crash whenever there is a problem in the input data. That's why we have `if run_number <= 0 ` to deal with that case in the cities.

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -176,6 +177,18 @@ def test_mc_info_writer_filter_first_event_of_first_file(output_tmpdir, ICDATADI
...
+    file_out = os.path.join(config_tmpdir, "dummy_out.h5")
+
+    with tb.open_file(file_out, 'w') as h5out:
+        writer = mc_info_writer(h5out)
+        with tb.open_file(file_in, 'r') as h5in:
+            copy_mc_info(h5in, writer)
+            events_in_h5in  = h5in .root.MC.extents.cols.evt_number[:]
+            events_in_h5out = h5out.root.MC.extents.cols.evt_number[:]
+            assert all(events_in_h5in == events_in_h5out)
+
```
gonzaponte, line 191: This is good. Can we have another one with a subset of events?

invisible_cities/cities/components.py
mmkekic: But wait, this is wrong, we want to have them in the same order as in all the other tables, which means they cant be sorted but have the exact order of event_numbers list

gonzaponte: You are right... I suggest
```
events_numbers_to_copy = tuple(evt for evt in event_numbers if evt in event_numbers_in_file)
```
or some other implementation using `filter`

invisible_cities/cities/components.py
mmkekic: Maybe throwing a warning? Notice that we want to be able to run the cities on empty files

gonzaponte: Ah, if that is the reason I guess it makes sense. But maybe we can introduce a function to detect empty files and use that. With this implementation we cannot distinguish between that and a file that has something wrong.

invisible_cities/dataflow/dataflow.py
jmalbos: Yes, I have yet to push the commits removing the code from `dataflow`.

mmkekic: On a side note, since this PR will need some history cleaning, @gonzaponte  do you want to keep these changes in the history if they turn out to be useful in the future dataflow refinements or you prefer to remove them?

invisible_cities/cities/components.py
jmalbos: > Ah, if that is the reason I guess it makes sense. But maybe we can introduce a function to detect empty files and use that. With this implementation we cannot distinguish between that and a file that has something wrong.

That is the reason, yes.

invisible_cities/dataflow/dataflow.py
gonzaponte: Good question... We can maybe leave them there, if that doesn't add too much noise.

invisible_cities/cities/components.py
jmalbos: In fact, independently of the implementation, there is still a problem that I documented with a test. If the list `event_numbers` contains event numbers that are present in more than one file, then the output file will contain more events than `len(event_numbers)`. I'm not sure there is a way to fix that easily.

invisible_cities/cities/components.py
gonzaponte: @mmkekic., do you know whether an empty file contains the `runInfo` table with zero entries? This could be the key to detect empty files vs "corrupt" files.

mmkekic: They do! So maybe thats the first thing we can check per file_in, if its empty do nothing, if its not and the reading doesnt work raise an error

invisible_cities/cities/components.py
mmkekic: I dont think you can solve the duplicated events in one file problem in this PR, so Id say forget about that possibility for now...And the efficiency of lists vs sets is really not important in this function, since the slowest part (by couple of orders of magnitude) is the writer itself

gonzaponte: Yeah, we cannot deal with that right now

invisible_cities/cities/components.py
jmalbos: The test `cities_test.test_city_empty_input_file` uses an empty input file that contains no tables.

gonzaponte: I can't confirm, but I would say that this is not realistic. It depends on the decoder, which is not in this repo, but I think these tables are always there, even if empty. In any case, a function that checks whether a file is empty or not could check both whether they exist and, if so, check whether they contain any entries.

jmalbos: Sure, but shouldn't that be a task for a different PR?

gonzaponte: I would prefer that, but that would delay this one, right? I am open to introducing it here, if necessary. @mmkekic ?

mmkekic: I just want to accept this PR so even a warning would be fine for me :p
But if you dont want to do sloppy job add a function that checks if the file is empty in this PR

jmalbos: > I just want to accept this PR so even a warning would be fine for me :p

I'd honestly follow this approach. Print a warning for now and leave as a to-do the "empty-file" checks, which could even require modifying the existing tests. I would even volunteer to do that…

gonzaponte: Fine, go ahead.

jmalbos:
I just pushed a number of commits that, I believe, address all the concerns raised by the reviewers. We leave as a to-do for a future PR improving the handling of empty or corrupted input files.

invisible_cities/cities/components.py
```diff
@@ -44,6 +44,8 @@
 from .. filters.s1s2_filter       import               pmap_filter
 from .. database                  import                   load_db
 from .. sierpe                    import                       blr
+from .. reco                      import             tbl_functions as tbl
```
gonzaponte, line 47: Rmove unused import

gonzaponte (approved):
This PR applies a new paradigm to the propagation of the MC information through the cities. Instead of introducing it in the dataflow pipe, it is copied at the end, taking into account which events were filtered out. This will be useful in conjunction with upcoming PRs where the format of the MC data is modified.

The new functions introduced here are very modular and well tested. Depending on the evolution of the dataflow package, the implementation of the evt_id collector may change in the future, as indicated in the commit history.

Great first PR. Good job!

Issue #501, opened 2018-07-05
Reorganization of folders


paolafer:
This is a minor issue, but I think that the readability of the code would improve with some reorganization of folders. 
For instance, I find confusing that the code in `sensor_functions.py`, which simulates the sensor response, lives in the `reco` folder. The same happens for `olivia.py`, which has the name of a city, but doesn't live in `cities`. In `reco` there is also `calib_sensors_functions.py`, which, just looking at the name, seems to belong better to the `calib` folder.
Another confusing issue, which has been mentioned in the past, is the existence of an `icaro` folder in IC, which contains functions for histogramming, but does not contain functions for fitting, which, in turn, live in `core`, whereas in the ICARO code there is an `invisible_cities` folder (but this is another story...).

PR #707, opened 2020-03-10
Clean up IC

gonzaponte:
This PRs goes for an exhaustive cleanup of the repository (at least that's what I tried). Among others, the changes are:
- Remove histogram functions for monitoring
- Remove monitor functions
- Remove unnecessary Cython stuff
- Remove unused and deprecated parts of the code
- Remove old corrections and rename corrections_new to corrections
- Fix imports across the repo
- Fix repeated test names
- Remove unused variables across the repo
- Remove some duplicities
- Move get_mc_info to mcinfo_io
- Rename hist_io to histogram_io

Most of the stuff deleted has been moved to another repo so it is not completely lost, but at least we keep IC clean.


andLaing:
Everything looks reasonable to me.

mmkekic:
I think I am ok with all the changes but I think it would be easier if you reorder your commits. Maybe have first removed unused full modules (with corresponding fix in imports in other files if necessary), then remove unused functions, (and corresponding tests), fixtures, and lastly remove unused variables?

gonzaponte:
I have tried to do that. I see that there are a couple of things that got done in the wrong order, mostly because I realized there were things like that after I had finished. Anyway, I am afraid that some of those will become a super messy rebase... Will do my best.

invisible_cities/cities/components_test.py
```diff
@@ -12,7 +12,6 @@
```
paolafer, line 12: This test was supposed to fail if the implementation of `compute_xy_position` doesn't take into account the correct run number. With your modification what fails is the code itself, since an exception is raised. My understanding is that in the test we want a test to fail, not the code itself, to make it clear what is failing; in fact, the code could be failing with good reason, tell me if I'm wrong.

gonzaponte: > This test was supposed to fail if the implementation of compute_xy_position doesn't take into account the correct run number

And the result of not taking into account the correct run number is that it raises a `ClusterEmptyList` exception, which is a valid failure for a test. 

> With your modification what fails is the code itself, since an exception is raised. My understanding is that in the test we want a test to fail, not the code itself.

What do you mean that what fails is the code? A test passes if and only if no exception is raised, otherwise it fails.  Most tests contain `assert`, which raises an `AssertionError` exception if a condition is not met, but `assert` is just a convenience for something like
```
if not <condition>:
    raise AssertionError
```

> in fact, the code could be failing with good reason

What do you mean? If the code fails for a different reason the test will still fail (in both cases)

paolafer: Yes, I understand that logically it is exactly the same, I just thought it was clearer to have an assertion in the test, but it doesn't matter.

invisible_cities/types/ic_types.py
```diff
@@ -74,7 +74,7 @@ def interval(self): return (self.min, self.max)
     def center(self): return (self.max + self.min) / 2
```
paolafer, line 61: I'm afraid I don't understand the comment of the commit... This change affects wherever in the code this method of the `minmax` class is used, even if the cython implementation disappears, right?

invisible_cities/core/core_functions_performance.py
```diff
@@ -1,32 +0,0 @@
-
```
paolafer, line 1: What's the rationale behind removing this file? Where was it used?

paolafer:
Some of the points of issue #501 have been addressed in this PR. There is one more that I would like to bring up, namely the content of the `reco` folder. What is this folder supposed to contain, from a logical point of view? I see functions related to calibration, to waveform, sensors... Have we thought about that? Do we have a consistency?

paolafer:
Some of the changes of this PR are likely to break the analysis code of some users (I think about `system_of_units`, for instance). Is there a way to highlight this so that users can be aware and know where to look at to fix their code?

invisible_cities/types/ic_types.py
gonzaponte: The thing is that this class was not being used in favor of the cython implementation. There was this subtle difference with respect to the cython implementation that I had to fix in order to migrate all the code to use the python implementation.

invisible_cities/core/core_functions_performance.py
gonzaponte: Nowhere, that's why it is removed. I think it was some kind of script to compare performances, that didn't belong here, in my opinion.

gonzaponte:
> What is this folder supposed to contain, from a logical point of view?

Functions specific to NEXT reconstruction.

> I see functions related to calibration, to waveform, sensors.

Yes, it has become a bit of our default folder and everything new was added here instead of having a proper folder. I think I also mistakenly suggested that calibration stuff was added there (back when @andLaing was working on writing all those utilities).

> Have we thought about that?

With detsim, the same issue was raised. We have to discuss and propose a sensible scheme. But I propose to deal with reorganization in another PR, this has become a bit of a monster already.

> Is there a way to highlight this so that users can be aware and know where to look at to fix their code?

Hmm, I don't think so... But the fix is simple:
Replace `from invisible_cities.core.system_of_units_c import units`
by `import invisible_cities.core.system_of_units as units`

I don't think the other utilities are used by anyone outside IC.

paolafer:
Ok, with leaving the `reco` folder reorganization for another PR (but let's not forget about it...).

> I don't think the other utilities are used by anyone outside IC

Some functions for histogramming in the `icaro` folder were used by at least one user (me 😅). I don't know if it's possible to write some comment on this when merging the PR.

gonzaponte:
Sure! I will be in touch with @carmenromo or @bpalmeiro (whoever merges it) to make sure we leave some note regarding that.

If you need specific functions, they should be incorporated into ICAROS, but we have to make a clean addition of selected functions, not a dump of everything that was here. We can coordinate with @mmkekic for that.

paolafer (approved):
This PR simplifies the IC structure, eliminating obsolete code and doing cosmetics fixes. Overall it is a long-due, very nice clean-up! The only caveat, which will be addressed separately, is that some of the eliminated functions could be used in the analysis. They will be moved to a more appropriate place.

paolafer:
What about the message in the merging... 😢?

carmenromo:
It is in the message of the merge commit

gonzaponte:
Isn't that what you meant?

paolafer:
Ah, now I see it! Yes, it is perfect!

PR #708, opened 2020-03-12
Fix tests

mmkekic:
This PR fixes improper calls / typos  in several tests.


gonzaponte (approved):
While working on #707 we have realized about mistakes in some tests. @mmkekic has collected them in this PR and added some more. Luckily the tests pass after fixing the writing.

Good job.

PR #709, opened 2020-03-12
Improve df to tables writer

mmkekic:
This PR removes underscore in front of store_pandas_as_tables writer since the method is being used outside the module. It also modifies the method to improve speed when writing large dataframes, no tests needed because it is just internal change.

### DIFF
--- a/invisible_cities/io/dst_io.py
+++ b/invisible_cities/io/dst_io.py
@@ -39,9 +39,14 @@ def _make_tabledef(column_types : pd.Series, str_col_length : int=32) -> dict:
             tabledef[colname] = tb.Col.from_type(coltype, pos=indx)
     return tabledef
 
-def _store_pandas_as_tables(h5out : tb.file.File, df : pd.DataFrame, group_name : str, table_name : str, compression : str='ZLIB4', descriptive_string : [str]="", str_col_length : int=32) -> None:
-    if len(df) == 0:
-        warnings.warn(f'dataframe is empty', UserWarning)
+def store_pandas_as_tables(h5out              : tb.file.File ,
+                           df                 : pd.DataFrame ,
+                           group_name         : str          ,
+                           table_name         : str          ,
+                           compression        : str = 'ZLIB4',
+                           descriptive_string : str = ""     ,
+                           str_col_length     : int = 32
+                           ) -> None:
     if group_name not in h5out.root:
         group = h5out.create_group(h5out.root, group_name)
 
--- a/invisible_cities/io/dst_io.py
+++ b/invisible_cities/io/dst_io.py
@@ -29,15 +29,31 @@ def load_dsts(dst_list, group, node):
     dsts = [load_dst(filename, group, node) for filename in dst_list]
     return pd.concat(dsts, ignore_index=True)
 
-def _make_tabledef(column_types : pd.Series, str_col_length : int=32) -> dict:
+def _make_tabledef(column_types : np.dtype, str_col_length : int) -> dict:
     tabledef = {}
-    for indx, colname in enumerate(column_types.index):
+    for indx, colname in enumerate(column_types.names):
         coltype = column_types[colname].name
         if coltype == 'object':
             tabledef[colname] = tb.StringCol(str_col_length, pos=indx)
         else:
             tabledef[colname] = tb.Col.from_type(coltype, pos=indx)
     return tabledef
+
+def _check_castability(arr : np.ndarray, table_types : np.dtype):
+    arr_types = arr.dtype
+
+    if set(arr_types.names) != set(table_types.names):
+        raise TableMismatch(f'dataframe differs from already existing table structure')
+
+    for name in arr_types.names:
+        if arr_types[name].name == 'object':
+            max_str_length = max(map(len, arr[name]))
+            if max_str_length > table_types[name].itemsize:
+                warnings.warn(f'dataframe contains strings longer than allowed', UserWarning)
+
+        elif not np.can_cast(arr_types[name], table_types[name], casting='same_kind'):
+            raise TableMismatch(f'dataframe numeric types not consistent with the table existing ones')
+
 
 def store_pandas_as_tables(h5out              : tb.file.File ,
                            df                 : pd.DataFrame ,
@@ -49,23 +65,26 @@ def store_pandas_as_tables(h5out              : tb.file.File ,
                            ) -> None:
     if group_name not in h5out.root:
         group = h5out.create_group(h5out.root, group_name)
-
     group = getattr(h5out.root, group_name)
+
+    arr = df.to_records(index=False)
+
     if table_name not in group:
-        tabledef = _make_tabledef(df.dtypes)
+        tabledef = _make_tabledef(arr.dtype, str_col_length=str_col_length)
         table    =  make_table(h5out,
                                group       = group_name,
                                name        = table_name,
                                fformat     = tabledef,
                                description = descriptive_string,
                                compression = compression)
+    else:
+        table = getattr(group, table_name)
 
-    table = getattr(group, table_name)
-    if not np.array_equal(df.columns, table.colnames):
-        raise TableMismatch(f'dataframe differs from already existing table structure')
-    for indx in df.index:
-        tablerow = table.row
-        for colname in table.colnames:
-            tablerow[colname] = df.at[indx, colname]
-        tablerow.append()
-    table.flush()
+    data_types = table.dtype
+    if len(arr) == 0:
+        warnings.warn(f'dataframe is empty', UserWarning)
+    else:
+        _check_castability(arr, data_types)
+        arr = arr.astype(data_types)
+        table.append(arr)
+        table.flush()
--- a/invisible_cities/io/dst_io.py
+++ b/invisible_cities/io/dst_io.py
@@ -85,6 +85,7 @@ def store_pandas_as_tables(h5out              : tb.file.File ,
         warnings.warn(f'dataframe is empty', UserWarning)
     else:
         _check_castability(arr, data_types)
-        arr = arr.astype(data_types)
+        columns = list(data_types.names)
+        arr = arr[columns].astype(data_types)
         table.append(arr)
         table.flush()
--- a/invisible_cities/io/dst_io_test.py
+++ b/invisible_cities/io/dst_io_test.py
@@ -8,7 +8,7 @@ from ..core.testing_utils import assert_dataframes_equal
 from ..core.exceptions    import TableMismatch
 from . dst_io             import load_dst
 from . dst_io             import load_dsts
-from . dst_io             import _store_pandas_as_tables
+from . dst_io             import store_pandas_as_tables
 from . dst_io             import _make_tabledef
 
 import warnings
@@ -104,7 +104,7 @@ def test_store_pandas_as_tables_exact(config_tmpdir, df):
     group_name = 'test_group'
     table_name = 'table_name_1'
     with tb.open_file(filename, 'w') as h5out:
-        _store_pandas_as_tables(h5out, df, group_name, table_name)
+        store_pandas_as_tables(h5out, df, group_name, table_name)
     df_read = load_dst(filename, group_name, table_name)
     assert_dataframes_close(df_read, df, False, rtol=1e-5)
 
@@ -114,8 +114,8 @@ def test_store_pandas_as_tables_2df(config_tmpdir, df1, df2):
     group_name = 'test_group'
     table_name = 'table_name_2'
     with tb.open_file(filename, 'w') as h5out:
-        _store_pandas_as_tables(h5out, df1, group_name, table_name)
-        _store_pandas_as_tables(h5out, df2, group_name, table_name)
+        store_pandas_as_tables(h5out, df1, group_name, table_name)
+        store_pandas_as_tables(h5out, df2, group_name, table_name)
     df_read = load_dst(filename, group_name, table_name)
     assert_dataframes_close(df_read, pd.concat([df1, df2]).reset_index(drop=True), False, rtol=1e-5)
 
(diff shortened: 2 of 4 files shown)


invisible_cities/io/dst_io.py
```diff
@@ -58,9 +58,12 @@ def _store_pandas_as_tables(h5out : tb.file.File, df : pd.DataFrame, group_name
...
-    for indx in df.index:
-        tablerow = table.row
-        for colname in table.colnames:
-            tablerow[colname] = df.at[indx, colname]
-        tablerow.append()
+
+    arr = df.to_records(index=False)
+    #hack to transform object numpy dtype to string
+    data_types = [(col, arr[col].dtype if arr[col].dtype != 'O' else \
+                   f'S{str_col_length}') for col in arr.dtype.names]
```
gonzaponte: It's a bit hard to read. Can we define a function (even inline) to do that?

gonzaponte (commented):
I think the tests are also checking that the types remain unchanged after writing & reading, right? I would like to have a specific one, but it's ok as is.

invisible_cities/io/dst_io.py
mmkekic: How about now?

mmkekic:
> I think the tests are also checking that the types remain unchanged after writing & reading, right? I would like to have a specific one, but it's ok as is.

I just noticed that the check type was to False in the tests, I changed that

invisible_cities/io/dst_io.py
```diff
@@ -29,25 +29,28 @@ def load_dsts(dst_list, group, node):
...
     tabledef = {}
     for indx, colname in enumerate(column_types.index):
         coltype = column_types[colname].name
         if coltype == 'object':
+            if df[colname].str.len().max() > str_col_length:
+                warnings.warn(f'dataframe contains strings longer than allowed', UserWarning)
             tabledef[colname] = tb.StringCol(str_col_length, pos=indx)
         else:
             tabledef[colname] = tb.Col.from_type(coltype, pos=indx)
     return tabledef
```
gonzaponte: Actually I don't think `_make_tabledef` cares about the length of the object. I would revert this back to the original version and perform the check in `store_pandas_as_tables`. What do you think?

invisible_cities/io/dst_io.py
```diff
@@ -58,9 +61,16 @@ def _store_pandas_as_tables(h5out : tb.file.File, df : pd.DataFrame, group_name
...
         raise TableMismatch(f'dataframe differs from already existing table structure')
-    for indx in df.index:
-        tablerow = table.row
-        for colname in table.colnames:
-            tablerow[colname] = df.at[indx, colname]
-        tablerow.append()
+
+    arr = df.to_records(index=False)
+    #hack to transform object numpy dtype to string
+    def _cast_type(dtype : np.dtype):
```
gonzaponte: I understand where the underscore comes from, but since it is inside a function it doesn't need to have special syntax :)

invisible_cities/io/dst_io_test.py
```diff
@@ -155,4 +156,14 @@ def test_store_pandas_as_tables_raises_warning_empty_dataframe(config_tmpdir, em
...
+            store_pandas_as_tables(h5out, empty_dataframe, group_name, table_name)
+
+@given(df=lstrings_dataframe)
+def test_store_pandas_as_tables_raises_warning_long_string(config_tmpdir, df):
+    filename   = config_tmpdir + 'dataframe_to_table_exact.h5'
+    group_name = 'test_group'
+    table_name = 'table_name_3'
+    with tb.open_file(filename, 'w') as h5out:
+        with pytest.warns(UserWarning, match='dataframe contains strings longer than allowed'):
+            store_pandas_as_tables(h5out, df, group_name, table_name)
```
gonzaponte: I suggest that here you set explicitly the string column length in the call to `store_pandas_as_tables` to make the test more obvious.

invisible_cities/io/dst_io_test.py
```diff
@@ -118,7 +118,7 @@ def test_store_pandas_as_tables_2df(config_tmpdir, df1, df2):
         store_pandas_as_tables(h5out, df1, group_name, table_name)
         store_pandas_as_tables(h5out, df2, group_name, table_name)
     df_read = load_dst(filename, group_name, table_name)
-    assert_dataframes_close(df_read, pd.concat([df1, df2]).reset_index(drop=True), False, rtol=1e-5)
+    assert_dataframes_equal(df_read, pd.concat([df1, df2]).reset_index(drop=True))
```
gonzaponte: these dataframes contain floats, `assert_dataframes_close` is more appropriate.

invisible_cities/io/dst_io.py
mmkekic: Agree. I added a check in `store_pandas_as_tables ` directly. I also modified to cast numpy records array to the table type if it was an already existing one. However I feel like this 3x search for strings and type castings can be somehow better optimized...

invisible_cities/io/dst_io.py
```diff
@@ -39,28 +39,43 @@ def _make_tabledef(column_types : pd.Series, str_col_length : int=32) -> dict:
...
         table    =  make_table(h5out,
                                group       = group_name,
                                name        = table_name,
                                fformat     = tabledef,
                                description = descriptive_string,
                                compression = compression)
+        #hack to transform object numpy dtype to string
+        def _cast_type(dtype : np.dtype):
+            if dtype != 'O': return dtype
+            else:            return f'S{str_col_length}'
```
gonzaponte: Remove underscore.

Don't you find it weird to test for `!=` instead of `==`? There is nothing wrong with it, but it feels counterintuitive.

invisible_cities/io/dst_io.py
```diff
@@ -39,28 +39,43 @@ def _make_tabledef(column_types : pd.Series, str_col_length : int=32) -> dict:
...
-        tablerow.append()
+
+    #check which columns are of type object and throw warning if length exceeds
+    for colname in df.columns:
+        if (df[colname].dtype.name == 'object'):
+            if df[colname].str.len().max() > data_types[colname].itemsize:
+                warnings.warn(f'dataframe contains strings longer than allowed', UserWarning)
+        else:
+            if arr.dtype[colname] != data_types[colname]:
+                    warnings.warn(f'dataframe numeric types not consistent with the table existing ones', UserWarning)
```
gonzaponte: I think this part belongs to the case in which there is already a table object stored. It doesn't make sense to check it for both cases

invisible_cities/io/dst_io.py
```diff
@@ -39,28 +39,43 @@ def _make_tabledef(column_types : pd.Series, str_col_length : int=32) -> dict:
...
         raise TableMismatch(f'dataframe differs from already existing table structure')
-    for indx in df.index:
-        tablerow = table.row
-        for colname in table.colnames:
-            tablerow[colname] = df.at[indx, colname]
-        tablerow.append()
+
+    #check which columns are of type object and throw warning if length exceeds
+    for colname in df.columns:
+        if (df[colname].dtype.name == 'object'):
```
gonzaponte: Maybe this could be written as
```
for colname, col in filter(lambda (_, c): c.dtype.name == "object", df.items()):
    if col.str.len().max() > data_types[colname].itemsize:
        warnings.warn...
```
or
```
for colname, col in df.items():
    if col.dtype.name == "object" and col.str.len().max() > data_types[colname].itemsize:
        warnings.warn...
```

mmkekic: Right... Let me try to rewrite the whole function in a more sensible way...

invisible_cities/io/dst_io.py
```diff
@@ -29,53 +29,52 @@ def load_dsts(dst_list, group, node):
...
-                    warnings.warn(f'dataframe numeric types not consistent with the table existing ones', UserWarning)
-
-    arr = arr.astype(data_types)
-    table.append(arr)
-    table.flush()
+    data_types = table.dtype
+    if len(arr) == 0:
+        warnings.warn(f'dataframe is empty', UserWarning)
+    else:
+        _can_cast(arr, data_types)
```
gonzaponte: I suggest a better name like `_check_castability` or something like that. The use of "can" suggests that it returns a boolean value.

invisible_cities/io/dst_io.py
```diff
@@ -75,6 +75,7 @@ def store_pandas_as_tables(h5out : tb.file.File, df : pd.DataFrame, group_name :
         warnings.warn(f'dataframe is empty', UserWarning)
     else:
         _can_cast(arr, data_types)
-        arr = arr.astype(data_types)
+        columns = list(data_types.names)
+        arr = arr[columns].astype(data_types)
```
gonzaponte, line 89: where is the sorting?

invisible_cities/io/dst_io.py
```diff
@@ -29,38 +29,53 @@ def load_dsts(dst_list, group, node):
...
+    for name in arr_types.names:
+        if arr_types[name].name == 'object':
+            max_str_length = max(map(len, arr[name]))
+            if max_str_length > table_types[name].itemsize:
+                warnings.warn(f'dataframe contains strings longer than allowed', UserWarning)
+        elif arr_types[name] != table_types[name]:
+            warnings.warn(f'dataframe numeric types not consistent with the table existing ones', UserWarning)
+
+
+def store_pandas_as_tables(h5out : tb.file.File, df : pd.DataFrame, group_name : str, table_name : str, compression : str='ZLIB4', descriptive_string : [str]="", str_col_length : int=32) -> None:
```
gonzaponte: I know it was already like that, but can we take the opportunity to split this into many for readability?

invisible_cities/io/dst_io.py
```diff
@@ -29,38 +29,53 @@ def load_dsts(dst_list, group, node):
...
+    arr_types = arr.dtype
+    if set(arr_types.names) != set(table_types.names):
+        raise TableMismatch(f'dataframe differs from already existing table structure')
+    for name in arr_types.names:
+        if arr_types[name].name == 'object':
+            max_str_length = max(map(len, arr[name]))
+            if max_str_length > table_types[name].itemsize:
+                warnings.warn(f'dataframe contains strings longer than allowed', UserWarning)
+        elif arr_types[name] != table_types[name]:
+            warnings.warn(f'dataframe numeric types not consistent with the table existing ones', UserWarning)
```
gonzaponte: Shouldn't these be actual errors rather than just warnings?

invisible_cities/io/dst_io.py
mmkekic: It is not sorted, is just assured that the columns of record array have the same order as the ones in table

invisible_cities/io/dst_io.py
mmkekic: I am not sure, probably we do want error in case you try to store string in numeric column, but we do not want to stop the process if someone tries to put an Int32 value into Int64 columns... And writing all possible combinations that make sense seems like a bit tedious job, so I opted for 'leave it to user responsibility to check the warning'

invisible_cities/io/dst_io.py
gonzaponte: OK. Got slightly confused by the commit message.

invisible_cities/io/dst_io.py
gonzaponte: Mmm, if you think about the cities, that shouldn't happen as the types should be consistent through the iterations. If you think about other usages, I don't see that it is a big deal for the user to deal with slightly different types.

For the first warning the code will probably just chop part of the string, and the second one would either crash or store useless data. And, being realistic, warnings are not usually checked..

mmkekic:
I think I addressed all previous comments. @gonzaponte  if you are happy with the code I can do the history cleaning?

invisible_cities/io/dst_io.py
```diff
@@ -29,38 +29,60 @@ def load_dsts(dst_list, group, node):
...
+    arr_types = arr.dtype
+    if set(arr_types.names) != set(table_types.names):
+        raise TableMismatch(f'dataframe differs from already existing table structure')
+    for name in arr_types.names:
+        if arr_types[name].name == 'object':
+            max_str_length = max(map(len, arr[name]))
+            if max_str_length > table_types[name].itemsize:
+                warnings.warn(f'dataframe contains strings longer than allowed', UserWarning)
+        elif not np.can_cast(arr_types[name], table_types[name], casting='same_kind'):
+            raise TableMismatch(f'dataframe numeric types not consistent with the table existing ones')
```
```suggestion
def _check_castability(arr : np.ndarray, table_types : np.dtype):
    arr_types = arr.dtype
    
    if set(arr_types.names) != set(table_types.names):
        raise TableMismatch(f'dataframe differs from already existing table structure')
    
    for name in arr_types.names:
        if arr_types[name].name == 'object':
            max_str_length = max(map(len, arr[name]))
            if max_str_length > table_types[name].itemsize:
                warnings.warn(f'dataframe contains strings longer than allowed', UserWarning)
        
        elif not np.can_cast(arr_types[name], table_types[name], casting='same_kind'):
            raise TableMismatch(f'dataframe numeric types not consistent with the table existing ones')
```
Helps readability.

invisible_cities/io/dst_io_test.py
```diff
@@ -155,4 +155,36 @@ def test_store_pandas_as_tables_raises_warning_empty_dataframe(config_tmpdir, em
...
+    filename   = config_tmpdir + 'dataframe_to_table_long_string.h5'
+    group_name = 'test_group'
+    table_name = 'table_name_lstr'
+    with tb.open_file(filename, 'w') as h5out:
+        with pytest.warns(UserWarning, match='dataframe contains strings longer than allowed'):
+            store_pandas_as_tables(h5out, df, group_name, table_name, str_col_length=1)
+
+
+@given(df=dataframe)
+def test_store_pandas_as_tables_raises_warning_inconsistent_types(config_tmpdir, df):
```
gonzaponte:
```suggestion
def test_store_pandas_as_tables_raises_TableMismatch_inconsistent_types(config_tmpdir, df):
```

gonzaponte (commented):
After these cosmetical changes we are good to go

gonzaponte (approved):
This PR makes the function `store_pandas_as_table` public, as it should be. It also improves the performance of the writer and adds type checks for robustness. A bunch of tests have been added, which is always good.

Good job!

PR #711, opened 2020-03-20
Tidy up .gitignore file

jmalbos:
Rewrite of the `.gitignore` file removing redundancies and adding a few new entries.
Please let me know if you find anything is missing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,13 +1,25 @@
+# Byte-compiled / optimized
+__pycache__/
+*.py[cod]
 
-*.pyc
-*.o
-*.so
+# C extensions
 *.c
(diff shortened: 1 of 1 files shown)


jacg:
> `*.*~`

Why not `*~`?

IOW, why shouldn't we ignore backups of files which don't have a filename extension?

jmalbos:
BTW, the file I've uploaded is essentially a selective copy-paste from this other one:
https://github.com/github/gitignore/blob/master/Python.gitignore

gonzaponte (approved):
A better `.gitignore` file for those still using the command line ;)

Good job.

PR #712, opened 2020-03-25
Change assert

MiryamMV:
After replacing the _compare_dataframes IC function with the pandas.testing function assert_frame_equal (with @mmkekic help) there are 3 test that fail. These are beersheba_test.py (not clear), esmeralda_test.py (it fails because the index of the compared dataframes is different) and corrections_test.py (not clear to me either, but apparently not related to the changes I made). 
@mmkekic suggested to open this PR  so we can discuss what approach to take @andLaing @gonzaponte

### DIFF
--- a/invisible_cities/cities/beersheba_test.py
+++ b/invisible_cities/cities/beersheba_test.py
@@ -27,7 +27,7 @@ def test_create_deconvolution_df(ICDATADIR):
                                                   CutType.abs, ecut, 3) for _, t in true_dst.groupby('event')])
     true_dst = true_dst.loc[true_dst.E > ecut, :].reset_index(drop=True)
 
-    assert_dataframes_close(new_dst, true_dst)
+    assert_dataframes_close(new_dst .reset_index(drop=True), true_dst.reset_index(drop=True))
 
 
 @mark.parametrize("cut_type", CutType.__members__)
--- a/invisible_cities/cities/esmeralda_test.py
+++ b/invisible_cities/cities/esmeralda_test.py
@@ -163,7 +163,9 @@ def test_esmeralda_tracks_exact(data_hdst, esmeralda_tracks, correction_map_file
     #some events are not in df_tracks_exact
     events = df_tracks_exact.event.unique()
     df_tracks_cut  = df_tracks[df_tracks.event.isin(events)]
-    assert_dataframes_close (df_tracks_cut[columns2], df_tracks_exact[columns2])
+
+    assert_dataframes_close (df_tracks_cut[columns2]  .reset_index(drop=True),
+                             df_tracks_exact[columns2].reset_index(drop=True))
     #make sure out_of_map is true for events not in df_tracks_exact
     diff_events = list(set(df_tracks.event.unique()).difference(events))
     df_summary  = dio.load_dst(PATH_OUT, 'Summary', 'Events')
--- a/invisible_cities/core/testing_utils.py
+++ b/invisible_cities/core/testing_utils.py
@@ -1,4 +1,5 @@
 import numpy  as np
+import pandas as pd
 
 from pytest                 import approx
 from numpy.testing          import assert_array_equal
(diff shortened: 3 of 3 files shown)


mmkekic:
I believe that all the `beersheba_test.py::test_create_deconvolution_df`  and `esmeralda_test.py::test_esmeralda_tracks_exact` should be fixed by resetting index when comparing, what do you think @gonzaponte ? And based on the Travis I believe those are the only two tests that fail, @MiryamMV  the fails were local right?

gonzaponte:
If that fixes it, go ahead. I don't think there is a way to ignore the index column in that function.

As for the one in corrections_test, don't worry about it. That is a flaky test that will be removed soon.

MiryamMV:
> I believe that all the `beersheba_test.py::test_create_deconvolution_df` and `esmeralda_test.py::test_esmeralda_tracks_exact` should be fixed by resetting index when comparing, what do you think @gonzaponte ? And based on the Travis I believe those are the only two tests that fail, @MiryamMV the fails were local right?

Yes!! The fails were local

andLaing:
It appears for me that travis still hasn't finished, is there something wrong with travis?

invisible_cities/cities/beersheba_test.py
```diff
@@ -27,7 +27,9 @@ def test_create_deconvolution_df(ICDATADIR):
                                                   CutType.abs, ecut, 3) for _, t in true_dst.groupby('event')])
     true_dst = true_dst.loc[true_dst.E > ecut, :].reset_index(drop=True)
 
-    assert_dataframes_close(new_dst, true_dst)
+    #new_dst .reset_index(drop=True)#
+    #true_dst.reset_index(drop=True)#
```
mmkekic: remove instead of comment

invisible_cities/cities/esmeralda_test.py
```diff
@@ -163,7 +163,8 @@ def test_esmeralda_tracks_exact(data_hdst, esmeralda_tracks, correction_map_file
     #some events are not in df_tracks_exact
     events = df_tracks_exact.event.unique()
     df_tracks_cut  = df_tracks[df_tracks.event.isin(events)]
-    assert_dataframes_close (df_tracks_cut[columns2], df_tracks_exact[columns2])
+
+    assert_dataframes_close (df_tracks_cut[columns2].reset_index(drop=True), df_tracks_exact[columns2].reset_index(drop=True))
```
mmkekic: maybe split the line, is a bit long

invisible_cities/core/testing_utils.py
```diff
@@ -106,11 +107,18 @@ def _compare_dataframes(assertion, df1, df2, check_types=True, **kwargs):
 
 
 def assert_dataframes_equal(df1, df2, check_types=True, **kwargs):
-    _compare_dataframes(assert_array_equal, df1, df2, check_types, **kwargs)
+    pd.testing.assert_frame_equal(df1.sort_index(axis=1), df2.sort_index(axis=1), check_names=True, check_dtype=check_types, check_exact=True, **kwargs)
 
+def assert_dataframes_close(df1, df2, check_types=True, rtol=None, atol=None, **kwargs):
+    if rtol:
```
mmkekic, line 113: Can someone else also check this makes sense, @gonzaponte , @andLaing ?

mmkekic (commented):
I believe this PR should have 2 commits, one where the function in testing_utils are replaced with the pandas one, and other where esmeralda and beersheba test are corrected for different indices.  (Basically squash first three commits and give it a meaningful name)

mmkekic (commented):
This PR replaces dataframe comparison from the custom to pandas one, that is more suitable when comparing mixed (floats and strings) dataframes. Good job @MiryamMV  and welcome to IC developers!

mmkekic (approved):
This PR replaces dataframe comparison from the custom to pandas one, that is more suitable when comparing mixed (floats and strings) dataframes. Good job @MiryamMV and welcome to IC developers!

PR #713, opened 2020-03-30
Change rwf file isidora

andLaing:
Removes the use of the file electrons_40keV_z250_RWF.h5 in
the isidora tests and in the commandline tests for isidora
and irene.

This file was found to have badly written MC info in PR #693

### DIFF
--- a/invisible_cities/cities/irene_test.py
+++ b/invisible_cities/cities/irene_test.py
@@ -68,8 +68,8 @@ def job_info_missing_pmts(config_tmpdir, ICDATADIR):
     pmt_active  = list(filter(lambda x: x not in pmt_missing, range(12)))
 
 
-    ifilename = os.path.join(ICDATADIR    , 'electrons_40keV_z250_RWF.h5')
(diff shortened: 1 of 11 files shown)


paolafer:
There are other places in the code where this file is used, maybe it would be safer to replace it everywhere, checking that the tests that use it don't rely on the specific content of the file. What do you think about it?

andLaing:
Seems reasonable. The initial commits just replaced it where the test accessed the MC info. I'll do another set of commits removing it completely. Should I remove the file from the repository too?

paolafer:
Yeah, I think we can remove it, once it is not used anymore.

paolafer:
For instance, to me `electrons_40keV_z25_RWF.h5` seems the most similar file we have: it has the same number of events (10), it's the same kind of particle (40-keV electrons), the only difference being the z initial position of the electrons.

paolafer (approved):
This PR eliminates a test file that had wrong MC true information. The tests that used it have been modified to use a different file with similar content and the file has been removed. All the tests succeed now, so I approve this PR.

PR #714, opened 2020-04-02
Use IC io `load_dst` instead of pandas `read_hdf` to load PSFs

Aretno:
The use of `read_hdf` function as it is fails in higher pandas versions. This uses the standard IC function to load dataframes to discouple the version problem from `beersheba` city.


mmkekic:
I think this PR can be approved, however, annoyingly, Travis failed... @jmbenlloch  can you maybe restart Travis build so hopefully it will not fail and we could merge it?

mmkekic (approved):
This PR fixes the reader of psf_file that was doomed to fail in more recent pandas version. Good job, we are one step closer to updating our environment!

PR #716, opened 2020-04-08
Index dataframes tables

mmkekic:
The PR adds option to flag tables columns to be indexed.

### DIFF
--- a/invisible_cities/io/dst_io.py
+++ b/invisible_cities/io/dst_io.py
@@ -63,15 +63,15 @@ def _check_castability(arr : np.ndarray, table_types : np.dtype):
             raise TableMismatch(f'dataframe numeric types not consistent with the table existing ones')
 
 
-def store_pandas_as_tables(h5out              : tb.file.File ,
-                           df                 : pd.DataFrame ,
-                           group_name         : str          ,
-                           table_name         : str          ,
-                           compression        : str = 'ZLIB4',
-                           descriptive_string : str = ""     ,
-                           str_col_length     : int = 32     ,
-                           columns_to_index   : Optional[Sequence[str]] = None
-                           ) -> None:
+def df_writer(h5out              : tb.file.File ,
+              df                 : pd.DataFrame ,
+              group_name         : str          ,
+              table_name         : str          ,
+              compression        : str = 'ZLIB4',
+              descriptive_string : str = ""     ,
+              str_col_length     : int = 32     ,
+              columns_to_index   : Optional[Sequence[str]] = None
+              ) -> None:
     """ The function writes a dataframe to open pytables file.
     Parameters:
     h5out              : open pytable file for writing
--- a/invisible_cities/io/dst_io.py
+++ b/invisible_cities/io/dst_io.py
@@ -9,6 +9,9 @@ from    tables          import HDF5ExtError
 from .. core.exceptions import TableMismatch
 from .  table_io        import make_table
 
+from    typing          import Optional
+from    typing          import Sequence
+
 
 def load_dst(filename, group, node):
     """load a kdst if filename, group and node correctly found"""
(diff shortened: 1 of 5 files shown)


invisible_cities/io/dst_io.py
```diff
@@ -89,3 +104,9 @@ def store_pandas_as_tables(h5out              : tb.file.File ,
         arr = arr[columns].astype(data_types)
         table.append(arr)
         table.flush()
+
+    if columns_to_index:
```
```suggestion
    if columns_to_index is not None:
```

invisible_cities/io/dst_io_test.py
```diff
@@ -188,3 +191,35 @@ def test_store_pandas_as_tables_unordered_df(config_tmpdir, df1, df2):
         store_pandas_as_tables(h5out, df2, group_name, table_name)
     df_read = load_dst(filename, group_name, table_name)
     assert_dataframes_equal(df_read, pd.concat([df1, df2], sort=True).reset_index(drop=True))
+
+
```
gonzaponte, line 212: Another good test could be opening the file and checking that the table is actually indexed:

`a_table.indexed` is a property that indicates whether a table is indexed
`a_table.cols.<col_name>.indexed` is a property that indicates whether a specific column is indexed.

invisible_cities/io/dst_io.py
```diff
@@ -89,3 +104,9 @@ def store_pandas_as_tables(h5out              : tb.file.File ,
...
+        if set(columns_to_index).issubset(set(df.columns)):
+            table.set_attr('columns_to_index', columns_to_index)
+        else:
+            raise KeyError('column to index not present in the dataframe')
```
gonzaponte: Clarifying which column failed would help the user.

gonzaponte (changes requested):
Cool. I would like just one more test to be added.

invisible_cities/io/dst_io.py
```diff
@@ -9,14 +9,20 @@
 from .. core.exceptions    import TableMismatch
 from .  table_io           import make_table
 from    typing             import Optional
+from    typing             import Sequence
+
+def _decode_df_to_str(df):
```
gonzaponte: how about `decode_str_columns`?

invisible_cities/io/dst_io_test.py
```diff
@@ -98,6 +101,12 @@ def test_load_dsts_warns_if_not_existing_file(ICDATADIR):
     with pytest.warns(UserWarning, match='does not exist'):
         load_dsts([good_file, wrong_file], group, node)
 
+def test_load_dst_converts_from_bytest(ICDATADIR, fixed_dataframe):
```
gonzaponte:
```suggestion
def test_load_dst_converts_from_bytes(ICDATADIR, fixed_dataframe):
```
small typo

invisible_cities/io/dst_io_test.py
mmkekic: But this is just labeling columns to be indexed, is not indexing them (thats whats done at the end of the city for example using  `index_tables`function of components)
We could index table  in store_pandas_as_tables but in case it is called repeatedly inside cities it doesn't seem efficient.

gonzaponte: We have io/indexation_test.py that creates a dummy city and ensures every writer indexes the tables it writes. This should be added there. However this writer has a slightly different structure than the other, so it may require some adaptation.

We also need another PR to index tables in the cities that use this writer.

gonzaponte (approved):
An improved version of the dataframe writer with table indexation. The writer is now renamed to `df_writer`, which is much more comfortable. This PR also includes the indexation of the `event` column in the output of Beersheva and Esmeralda.

This change suggests that we should ensure each city indexes its output properly, but that is left for a different PR. 

The new feature is properly and thoughtfully tested. Nice job!

PR #719, opened 2020-05-21
Update NEXT100 DB and related tests

paolafer:
In this PR, the NEXT100 database has been updated according to the current design of the detector. Some tables are empty, since we don't have information about electronics yet, therefore, two tests have been marked to be skipped in the case of the NEXT100 detector.

Also, the previous NEXT100DB database has been renamed to NEXT100DB_old.


andLaing (approved):
NEXT100 database updated to have the channel positions correct for the current design.

Skip added to tests for that database which test database tables which are not yet filled.

Approved.

PR #720, opened 2020-05-22
Add low threshold info

bpalmeiro:
As shown by @msorel, the number of hits per blob for the low threshold case is a useful variable to keep. This PR adds it to the track info table, both for the low and high threshold case.

### DIFF
--- a/invisible_cities/cities/esmeralda.py
+++ b/invisible_cities/cities/esmeralda.py
@@ -48,6 +48,8 @@ from .. io.run_and_event_io import run_and_event_writer
 from .. io. event_filter_io import event_filter_writer
 from .. io.          dst_io import df_writer
 
+from .. reco.hits_functions import number_of_hits_per_blob
+
 
 types_dict_summary = OrderedDict({'event'     : np.int32  , 'evt_energy' : np.float64, 'evt_charge'    : np.float64,
                                   'evt_ntrks' : np.int    , 'evt_nhits'  : np.int    , 'evt_x_avg'     : np.float64,
@@ -56,19 +58,22 @@ types_dict_summary = OrderedDict({'event'     : np.int32  , 'evt_energy' : np.fl
                                   'evt_r_min' : np.float64, 'evt_x_max'  : np.float64, 'evt_y_max'     : np.float64,
                                   'evt_z_max' : np.float64, 'evt_r_max'  : np.float64, 'evt_out_of_map': bool      })
 
-types_dict_tracks = OrderedDict({'event'           : np.int32  , 'trackID'       : np.int    , 'energy'      : np.float64,
-                                 'length'          : np.float64, 'numb_of_voxels': np.int    , 'numb_of_hits': np.int    ,
-                                 'numb_of_tracks'  : np.int    , 'x_min'         : np.float64, 'y_min'       : np.float64,
-                                 'z_min'           : np.float64, 'r_min'         : np.float64, 'x_max'       : np.float64,
-                                 'y_max'           : np.float64, 'z_max'         : np.float64, 'r_max'       : np.float64,
-                                 'x_ave'           : np.float64, 'y_ave'         : np.float64, 'z_ave'       : np.float64,
-                                 'r_ave'           : np.float64, 'extreme1_x'    : np.float64, 'extreme1_y'  : np.float64,
-                                 'extreme1_z'      : np.float64, 'extreme2_x'    : np.float64, 'extreme2_y'  : np.float64,
-                                 'extreme2_z'      : np.float64, 'blob1_x'       : np.float64, 'blob1_y'     : np.float64,
-                                 'blob1_z'         : np.float64, 'blob2_x'       : np.float64, 'blob2_y'     : np.float64,
-                                 'blob2_z'         : np.float64, 'eblob1'        : np.float64, 'eblob2'      : np.float64,
+types_dict_tracks = OrderedDict({'event'           : np.int32  , 'trackID'        : np.int    , 'energy'      : np.float64,
+                                 'length'          : np.float64, 'numb_of_voxels' : np.int    , 'numb_of_hits': np.int    ,
+                                 'numb_of_tracks'  : np.int    , 'x_min'          : np.float64, 'y_min'       : np.float64,
+                                 'z_min'           : np.float64, 'r_min'          : np.float64, 'x_max'       : np.float64,
+                                 'y_max'           : np.float64, 'z_max'          : np.float64, 'r_max'       : np.float64,
+                                 'x_ave'           : np.float64, 'y_ave'          : np.float64, 'z_ave'       : np.float64,
+                                 'r_ave'           : np.float64, 'extreme1_x'     : np.float64, 'extreme1_y'  : np.float64,
+                                 'extreme1_z'      : np.float64, 'extreme2_x'     : np.float64, 'extreme2_y'  : np.float64,
+                                 'extreme2_z'      : np.float64, 'blob1_x'        : np.float64, 'blob1_y'     : np.float64,
+                                 'blob1_z'         : np.float64, 'blob2_x'        : np.float64, 'blob2_y'     : np.float64,
+                                 'blob2_z'         : np.float64, 'eblob1'         : np.float64, 'eblob2'      : np.float64,
                                  'ovlp_blob_energy': np.float64,
-                                 'vox_size_x'      : np.float64, 'vox_size_y'    : np.float64, 'vox_size_z'  : np.float64})
+                                 'nhits_blob1_hth' : np.int    , 'nhits_blob2_hth': np.int    ,
+                                 'nhits_blob1_lth' : np.int    , 'nhits_blob2_lth': np.int    ,
+                                 'eblob1_lth'      : np.float64, 'eblob2_lth'     : np.float64,
+                                 'vox_size_x'      : np.float64, 'vox_size_y'     : np.float64, 'vox_size_z'  : np.float64})
 
 def hits_threshold_and_corrector(map_fname        : str  ,
                                  threshold_charge : float,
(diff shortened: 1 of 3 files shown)


msorel (commented):
Function number_of_hits_per_blob function looks good to me, @bpalmeiro ! I think it makes sense to use a single function to compute nhits and Eblob (if asked), to speed up things.

As minor comment, in the function comments looks like you have a redundant threshold_charge parameter, to be removed since it is not there?

msorel (commented):
@bpalmeiro , changes in esmeralda.py look good to me except one typo in list_of_vars, line 254: change one nh1_high_th to nh2_high_th.

bpalmeiro:
@msorel typos solved, you don't miss any! Thanks for the review!

invisible_cities/cities/esmeralda.py
```diff
@@ -227,12 +232,27 @@ def create_extract_track_blob_info(hitc):
...
                 e_blob1, e_blob2, hits_blob1, hits_blob2 = plf.blob_energies_and_hits(t, blob_radius)
                 overlap = float(sum(h.Ep for h in set(hits_blob1).intersection(set(hits_blob2))))
+
+                low_th_hits = number_of_hits_per_blob(hitc_low,
+                                                      blob_pos1,
+                                                      blob_pos2,
+                                                      blob_radius)
+                nh1_low_th, nh2_low_th, eb1_low_th, eb2_low_th = low_th_hits
+
+                nh1_high_th, nh2_high_th = number_of_hits_per_blob(hitc,
```
mmkekic, line 242: This might be confusing, the `hitc`object is not the same as the input one, it has dropped `nan` energy hits, and the end-point-voxels hits, not sure this is what you want. 
If the performance is not too affected Id suggest to add another function that takes  `("cor_low_th_hits", "paolina_hits" , "topology_info")` and return a new tracking dataframe with the added info. In that case we leave already hard-to-follow `create_extract_track_blob_info` as is.

invisible_cities/reco/hits_functions.py
```diff
@@ -63,3 +64,52 @@ def threshold_hits(hits : List[evm.Hit], th : float, on_corrected : bool=False)
...
+    A tuple with the number of hits for both blobs and, if asked,
+    the energy of the blobs.
+    """
+    nhits1     = 0
+    eb1        = 0.
+    nhits2     = 0
+    eb2        = 0.
+
+    for hit in hitc.hits:
+        e_h   = hit.Ec if not np.isnan(hit.Ec) and hit.Ec>0 else 0.
```
mmkekic, line 100: `np.clip` and `np.nan_to_num`? 
However if the hit has `nan`or negative energy (the `Ec` before correction is -1, but it should not happen after the correction, if this is the case please report it, might be a bug!), do you still want to count +1 for the number of hits? @msorel ?

invisible_cities/reco/hits_functions.py
```diff
@@ -63,3 +64,52 @@ def threshold_hits(hits : List[evm.Hit], th : float, on_corrected : bool=False)
                 x.E = es[i]
                 new_hits.append(x)
         return new_hits
+
+def number_of_hits_per_blob(hitc       : evm.HitCollection,
```
mmkekic, line 68: Is `evm.HitCollection`indeed the type you need? Seems it only needs a sequence of hits, in which case the type should be `Sequence[evm.Hit]`

invisible_cities/reco/hits_functions_test.py
```diff
@@ -151,3 +154,52 @@ def test_threshold_hits_all_larger_than_th(hits, th):
...
+@dataclass
+class hits_test:
+    Ec : float
+    X  : float
+    Y  : float
+    Z  : float
+
+@dataclass
+class hitsc_test:
+    hits : list
```
mmkekic, line 167: It is confusing that you ask for `Hit`type but then pass the custom `dataclass`that has some of the `Hit` attributes - ie just construct the needed `evm` type. Also maybe think of more general test: ie - large enough blob radius to encompass all hits should return energy of all hits, 0-radius or far enough center should return0s, if position of both blobs are the same the information should be repeated etc..

mmkekic (commented):
I think we should try to keep some consistency in naming. Till now 'blob' was always related to paolina functions and tracks where the  distances are calculated along the tracks (ie eblob in esmeralda tracking table). It is confusing to have two 'blobs' with different meaning, I am not sure what a good naming convention should be here @msorel , @paolafer ?

paolafer:
> I think we should try to keep some consistency in naming. Till now 'blob' was always related to paolina functions and tracks where the distances are calculated along the tracks (ie eblob in esmeralda tracking table). It is confusing to have two 'blobs' with different meaning, I am not sure what a good naming convention should be here @msorel , @paolafer ?

I agree with @mmkekic, calling `blob` this new definition of energy-around-the-extreme looks confusing to me, because it's different from the current one.

Also, I'm wondering if it makes sense to find the blob centres with high threshold hits and then build these new energy-around-the-extreme objects using low threshold hits. Have you tried to use the same low threshold for everything (doing tracks, find extremes, build these new variables...)? The reason of the high threshold hits was the performance of the 2-blob cut.

msorel:
Thanks @mmkekic and @paolafer ! Several points:

1. Variable naming and use of paolina functions
These are in fact blob energies and blob hits as in Paolina's functions as far as I can tell, although computed also for the low-threshold hit collection and not just the high-thr collection. By looking closer at Paolina's code, you are right that I see a difference, though: the euclidean difference between blob center and hits is made not only in end voxels and contiguous voxels as in Paolina functions, but for all voxels. I agree it would be clearer and more consistent to use directly Paolina's functions hits_in_blob and blob_energies_hits_and_centres. It  should not make any appreciable difference and it will be more symmetric.

2. Blob center computation
@paolafer you are right that there is a slight inconsistency, and it would be better to compute the low-threshold blob nhits and low-thr blob energies using  the hit distances from the low-thr barycenter and not from the high-thr barycenter as now. Not sure we can use Paolina's function blob_centre for this, though. voxel.hits returns the high-thr hits, right?

3. Use same hit threshold for everything in Track table
@paolafer , you are again right this would be the ideal case, and we thought about this. It is quite possible that we can find a unique threshold (lower than 30 PEs) that works well for everything (fiducial cuts, track connectivity, topological performance,...). In this case one would only need to add two variables to the Track table: the number of hits at the two track extremes, all Track quantities computed with same threshold, no need to specify "high" and "low". But this requires running quite a few test productions and data/MC comparisons to find if such a unique optimal threshold exists, re-optimize all cuts (incl. fiducial), re-compute data/MC disagreements, etc.. No time since we are trying to get a first bb2nu paper out in few weeks, if we can. We propose this "hybrid" Track table with quantities from two hit collections as a patch for now, and we revisit later, if that's OK.

Thanks!

invisible_cities/reco/hits_functions.py
msorel: Hi @mmkekic , forgot to comment on this one. If a hit has nan or negative energy, I would be inclined NOT to count it as +1. I guess we exclude those hits when computing blob energies, and I would use the same set of hits to compute blob energy and blob number of hits. Otherwise things like (average hit energy) = (blob energy) / (blob nhits) are ill-defined.

In any case, I'd like to understand better if such hits can be real or are pure artifacts. What kind of physics situation creates  a hit, but one with nan or negative energy?

paolafer:
> 2. Blob center computation
@paolafer you are right that there is a slight inconsistency, and it would be better to compute the low-threshold blob nhits and low-thr blob energies using the hit distances from the low-thr barycenter and not from the high-thr barycenter as now. Not sure we can use Paolina's function blob_centre for this, though. voxel.hits returns the high-thr hits, right?

The code as it is computes tracks, extremes, blob centres... using high threshold hits. One should compute all these topological objects using the low threshold hits, meaning doing a parallel calculation. I understand that you're in a hurry, and that the analysis can be refined later, however, I'm wondering if the correct way to proceed is to approve and merge in IC a temporary code. Is it anti-kosher to use branches for this kind of things?

invisible_cities/reco/hits_functions.py
bpalmeiro: I agree, I was simply mimicking @msorel procedure, where this protection was added... Did you have to do it because you found misbehaviour or just it was general protection just in case?

bpalmeiro:
> The code as it is computes tracks, extremes, blob centres... using high threshold hits. One should compute all these topological objects using the low threshold hits, meaning doing a parallel calculation. I understand that you're in a hurry, and that the analysis can be refined later, however, I'm wondering if the correct way to proceed is to approve and merge in IC a temporary code. Is it anti-kosher to use branches for this kind of things?

I agree, but it is not mandatory at all to be merged so we can use the branch, this can lives as a pr while addressing those issues.

msorel:
> > The code as it is computes tracks, extremes, blob centres... using high threshold hits. One should compute all these topological objects using the low threshold hits, meaning doing a parallel calculation. I understand that you're in a hurry, and that the analysis can be refined later, however, I'm wondering if the correct way to proceed is to approve and merge in IC a temporary code. Is it anti-kosher to use branches for this kind of things?
> 
> I agree, but it is not mandatory at all to be merged so we can use the branch, this can lives as a pr while addressing those issues.

If I understood you well, @paolafer and @bpalmeiro , you suggest we keep this PR open, without approving nor merging to master, and we process data from this branch for now until we settle on a longer-term code? That would be OK with me. Still, let us know if you have further suggestions also for the temporary code.

invisible_cities/reco/hits_functions.py
msorel: The protection against nan energies is necessary, I think. The code crashed otherwise. We are doing things like
eb1    += e_h
here, which make no sense with nan e_h, I guess.

paolafer:
Yes, in that case (if we don't merge this code to IC), I understand that the correct procedure is to keep doing the official production in Canfranc with the tagged Esmeralda code and you can re-run Esmeralda with your branch. Would it be fine?

jacg:
Another PR with a red cross. This one has been stalled for over 10 months.

In this case, the test report seems to be unretrievable! Please rebase to re-run tests in GHA. (Re-run if they fail spuriously, otherwise fix.)

msorel:
Others should comment, too, but I suspect this is now obsolete and not needed by anybody in standard NEXT data reconstruction. We are now computing blob energies after running Beersheba + Isaura, using Esmeralda low-threshold hits as input to those cities. 

Is this right?

bpalmeiro:
This temporal fix has been incorporated into the normal processing flow. Just kept for archiving proposes, not needed any longer now.

PR #721, opened 2020-05-25
Source for nexus full simulation

andLaing:
Adds a source to compoments which reads the nexus sensor response output and returns a dictionary.

Addresses point 1 of issue #691 for the case of nexus full optical simulation.

Valid for NEW, NEXT100 and DEMO++

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -10,6 +10,7 @@ from enum        import Enum
 from typing      import Callable
 from typing      import Iterator
 from typing      import Mapping
+from typing      import Generator
 from typing      import List
 from typing      import Dict
 from typing      import Tuple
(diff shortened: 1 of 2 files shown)


jmalbos:
The new function (`components.mcsensors_from_file`) is quite straightforward, given that it relies on existing functions to do most of the legwork. The two new tests only evaluate the failure conditions for non-valid input files. Do we need to test as well that the data yielded is correct (i.e. what we expect)?

andLaing:
I could add another simple test to look that the dictionary structure is what we expect. Most of this is already tested in the tests of the actual reader though. Maybe it's worthwhile adding something to make sure that we get the right number of sensors out of the splitting into pmts and sipms etc. I'll look into it.

andLaing:
could I get someone to have a look at this most recent version. It's the only ingredient missing to write the city.

jmalbos:
> could I get someone to have a look at this most recent version. It's the only ingredient missing to write the city.

I'll do it.

invisible_cities/cities/components.py
```diff
@@ -337,6 +338,61 @@ def check_lengths(*iterables):
...
+                        run_number:      int ) -> Generator:
+    """
+    Loads the nexus MC sensor information into
+    a pandas DataFrame using the IC function
+    load_mcsensor_response_df.
+    Returns info event by event as a
+    generator in the structure expected by
+    the dataflow.
+
+    file_names : List of strings
```
jmalbos: `file_names` should be `paths`

jmalbos (changes requested):
This _pull request_ adds a new data source that reads the _nexus_ sensor response output and yields a dictionary as expected by `dataflow`. The code is straightforward, given that it relies on existing functions to do most of the legwork. 

Three new tests are added to IC in this PR: the first two check the behavior of the new function when the input datafile does not conform to the standard; the third test makes sure that the returned information is correct.

Once a small typo is fixed in the docstring of the function, I'll approve the PR.

jmalbos (approved):

Issue #715, opened 2020-04-06
Merging MC files 


mmkekic:
Dealing with  #693  we run into a problem of merging several nexus files information of the configuration table. Seems that it makes sense to have a configuration information (such as Geometry/Physics used to generate files) unique for all concatenated files, however, random_seed is a per-file information that needs to be saved and we are not sure what is the best way to save it. Maybe the best option is to have another table that will match event numbers and random seed?

This issue is somewhat related to

1.  future version of detsim https://github.com/nextic/detsim.git is able to split long nexus events, making nexus event_id repeating, meaning we need a mapping between detsim_event_id and nexus_event_id (maybe random_seed can go into this table?). There is also important issue of assigning detsim_event_id in a way to ensure that it is unique per run.

2.  in general how to deal with merging files with repeated event_id. This issue is certainly present when mixing several MC productions and we still dont have a good solution for it.

gonzaponte:
> in general how to deal with merging files with repeated event_id. This issue is certainly present when mixing several MC productions and we still dont have a good solution for it.

How about adding a `sub_event_id` column? We would need to add some stuff to deal with MC and data separately and transparently, but it sounds feasible.

jmalbos:
> Dealing with #693 we run into a problem of merging several nexus files information of the configuration table. Seems that it makes sense to have a configuration information (such as Geometry/Physics used to generate files) unique for all concatenated files, however, random_seed is a per-file information that needs to be saved and we are not sure what is the best way to save it. Maybe the best option is to have another table that will match event numbers and random seed?

There may be cases (e.g. mixing of background events from different sources) in which not only the random seed but also other configuration parameters could be different. It may be simpler to copy all the configuration tables, labelling them with a  _subrun_ tag (or something similar).

andLaing:
I think that this is something that needs to be solved in the medium term but I'd propose a staged approach since we really need to get PR #693 completed so that the integration of the detector simulation code isn't delayed too much.

I think we need to come up with a more general solution (please keep suggesting here) but I'd propose a minimal protection in PR #693 so that the configuration info isn't confusing/clashing (still not that simple, really) and that we deal with it in a more complete way in the PRs related with event splitting etc. I think that the favoured production paradigm needs to remain what we've used up til now in the short term -- processing single files per job.

A patch could be adding file number/name in the param_keys somewhere in the configuration information and adding a check for overlap in the merging of other MC tables.

Thoughts?

jmalbos:
[...]

PR #722, opened 2020-05-28
Integrated buffer writer

andLaing:
Adds a new io option to rwf_io which uses the basic rwf_writer and other table writers to write all event info in one step. In this way long MC events can be split into multiple trigger-like events in the output file and the event numbers can be logged and mapped accordingly.

Some issues remain for the logging portion that are under debate.

Addresses point 2 of issue #691

### DIFF
--- a/invisible_cities/evm/nh5.py
+++ b/invisible_cities/evm/nh5.py
@@ -43,6 +43,12 @@ class MCExtentInfo(tb.IsDescription):
     last_particle = tb.UInt64Col(pos=2)
 
 
+class MCEventMap(tb.IsDescription):
+    """Map between event index and original event."""
+    evt_number = tb.Int32Col(shape=(), pos=0)
+    nexus_evt  = tb.Int32Col(shape=(), pos=1)
+
+
 class MCHitInfo(tb.IsDescription):
     """Stores the simulated hits as metadata using Pytables.
     """
--- a/invisible_cities/evm/nh5.py
+++ b/invisible_cities/evm/nh5.py
@@ -46,7 +46,7 @@ class MCExtentInfo(tb.IsDescription):
 class MCEventMap(tb.IsDescription):
     """Map between event index and original event."""
     evt_number = tb.Int32Col(shape=(), pos=0)
-    nexus_evt  = tb.Int32Col(shape=(), pos=1)
+    sub_evt    = tb.Int32Col(shape=(), pos=1)
 
 
 class MCHitInfo(tb.IsDescription):
--- a/invisible_cities/io/rwf_io.py
+++ b/invisible_cities/io/rwf_io.py
@@ -102,6 +102,13 @@ def buffer_writer(h5out, *,
                   Default directly in root
     compression : Optional[str] default 'ZLIB4'
                   Compression level for output file.
+
+    returns
+    -------
+    write_buffers : Callable
+                    A function which takes event information
+                    for the tracking and energy planes and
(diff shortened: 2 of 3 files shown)


invisible_cities/io/rwf_io.py
```diff
@@ -54,3 +59,102 @@ def write_rwf(waveform : np.ndarray) -> None:
         """
         rwf_table.append(waveform.reshape(1, n_sensors, waveform_length))
     return write_rwf
+
+
+class EventMap(tb.IsDescription):
+    """
+    Map between event index and original
```
jmalbos: Any reason why this shouldn't be a one-line docstring?

invisible_cities/io/rwf_io.py
```diff
@@ -54,3 +59,102 @@ def write_rwf(waveform : np.ndarray) -> None:
...
+    length_eng  : int
+                  Number of samples per waveform for energy plane.
+    length_trk  : int
+                  Number of samples per waveform for tracking plane.
+    group_name  : Optional[str] default None
+                  Group name within root where waveforms to be saved.
+                  Default directly in root
+    compression : Optional[str] default 'ZLIB4'
+                  Compression level for output file.
+    """
```
jmalbos, line 105: You could add here a description of what the function returns.

invisible_cities/io/rwf_io.py
```diff
@@ -54,3 +59,102 @@ def write_rwf(waveform : np.ndarray) -> None:
         """
         rwf_table.append(waveform.reshape(1, n_sensors, waveform_length))
     return write_rwf
+
+
+class EventMap(tb.IsDescription):
```
jmalbos: This definition should be moved to `evm/nh5.py`.

invisible_cities/io/rwf_io.py
```diff
@@ -54,3 +59,102 @@ def write_rwf(waveform : np.ndarray) -> None:
...
+                            table_name      =    'sipmrd',
+                            n_sensors       =  n_sens_trk,
+                            waveform_length =  length_trk)
+
+    run_and_event = partial(run_and_event_writer(h5out                  ,
+                                                 compression=compression),
+                            run_number = run_number                      )
+
+    evt_group = getattr(h5out.root, 'Run')
+    nexus_map = h5out.create_table(evt_group, "eventMap", EventMap,
```
jmalbos: You could use for this the function `make_table` in `io/table_io.py`.

invisible_cities/evm/nh5.py
```diff
@@ -43,6 +43,12 @@ class MCExtentInfo(tb.IsDescription):
     last_particle = tb.UInt64Col(pos=2)
 
 
+class MCEventMap(tb.IsDescription):
+    """Map between event index and original event."""
```
jmalbos, line 47: I found this description and the row names below a bit misleading. After reading the code and discussing it with @andLaing, I understand that `evt_number` is always going to be the event number read from the NEXUS file, and that `nexus_evt` is actually a subevent-like index unrelated to NEXUS. I'd change the docstring and row names to reflect that.

jmalbos:
This PR adds a new writer (and the corresponding test) to `rwf_io` than can handle the splitting of long (MC) events into several _subevents_. A new table (`MCEventMap`) is used to associate the new subevents to the original event.

Nevertheless, this new writer has limited use until a decision is taken regarding #715.

jmalbos (approved):

PR #723, opened 2020-06-03
Update env

mmkekic:
This PR updates the environment packages.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -7,8 +7,6 @@ from hypothesis import Verbosity
 settings.register_profile("travis-ci" , settings(max_examples = 1000, deadline=300))
 settings.register_profile("hard"      , settings(max_examples = 1000))
 settings.register_profile("dev"       , settings(max_examples =   10))
-settings.register_profile("hard_nocov", settings(max_examples = 1000, use_coverage=False))
-settings.register_profile("dev_nocov" , settings(max_examples =   10, use_coverage=False))
 settings.register_profile("debug"     , settings(max_examples =   10,
                                                 verbosity=Verbosity.verbose))
 settings.load_profile(os.getenv(u'HYPOTHESIS_PROFILE', 'dev'))
--- a/manage.sh
+++ b/manage.sh
@@ -61,9 +61,9 @@ function install_conda {
     else
         echo Installing conda for $CONDA_OS
         if which wget; then
-            wget https://repo.continuum.io/miniconda/Miniconda3-4.5.4-${CONDA_OS}-x86_64.sh -O miniconda.sh
+            wget https://repo.continuum.io/miniconda/Miniconda3-py${PYTHON_VERSION//.}_4.8.2-${CONDA_OS}-x86_64.sh -O miniconda.sh
         else
-            curl https://repo.continuum.io/miniconda/Miniconda3-4.5.4-${CONDA_OS}-x86_64.sh -o miniconda.sh
+            curl https://repo.continuum.io/miniconda/Miniconda3-py${PYTHON_VERSION//.}_4.8.2-${CONDA_OS}-x86_64.sh -o miniconda.sh
         fi
         bash miniconda.sh -b -p $HOME/miniconda
         CONDA_SH=$HOME/miniconda/etc/profile.d/conda.sh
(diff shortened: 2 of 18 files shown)


mmkekic:
6 tests related to `deconv_functions` and `beersheba` are failing. I think the main problem is in `interpolate_signal` in the call  of `scipy.interpolate.griddata` but @gonzaponte  and me couldn't figure out why would this change between versions. @Aretno  could you check that your algorithm works with updated numpy/scipy? Also, as I understand the interpolation is always done on the regular grid, why aren't you using scipy interpolation methods that work on gridded data?

Aretno:
Will look into it. Regarding the gridded data, no strong reason but that won't be the case when working on 3D as the Z position of the hits is not gridded.

conftest.py
```diff
@@ -7,8 +7,8 @@
 settings.register_profile("travis-ci" , settings(max_examples = 1000, deadline=300))
 settings.register_profile("hard"      , settings(max_examples = 1000))
 settings.register_profile("dev"       , settings(max_examples =   10))
-settings.register_profile("hard_nocov", settings(max_examples = 1000, use_coverage=False))
-settings.register_profile("dev_nocov" , settings(max_examples =   10, use_coverage=False))
+settings.register_profile("hard_nocov", settings(max_examples = 1000))
+settings.register_profile("dev_nocov" , settings(max_examples =   10))
```
gonzaponte: How can we deactivate coverage now? Or we don't care anymore? In any case, we have duplicated profiles now.

manage.sh
```diff
@@ -57,22 +57,27 @@ function install_conda {
     esac
 
     if conda --version ; then
-        echo Conda already installed. Skipping conda installation.
+	if  [ `(conda --version | cut -d' ' -f2)` == "4.8.2" ]; then
+            echo Conda already installed.
+	else
+	    conda activate base
+	    conda install conda=4.8.2
```
gonzaponte: `conda deactivate` after update?

manage.sh
```diff
@@ -57,22 +57,27 @@ function install_conda {
...
-            curl https://repo.continuum.io/miniconda/Miniconda3-4.5.4-${CONDA_OS}-x86_64.sh -o miniconda.sh
+            curl https://repo.continuum.io/miniconda/Miniconda3-py${PYTHON_VERSION//.}_4.8.2-${CONDA_OS}-x86_64.sh -o miniconda.sh
         fi
         bash miniconda.sh -b -p $HOME/miniconda
-        CONDA_SH=$HOME/miniconda/etc/profile.d/conda.sh
-        source $CONDA_SH
-        echo Activated conda by sourcing $CONDA_SH
+	CONDA_SH=$HOME/miniconda/etc/profile.d/conda.sh
+	source $CONDA_SH
+	echo Activated conda by sourcing $CONDA_SH
```
gonzaponte: why did you dedent? It shouldn't be dedented as it refers to the miniconda autoinstalled.

manage.sh
gonzaponte: Another thing that is bothering me is that, as I understand it, the update affects all environments and therefore it has side effects for other environments we may have (I have a few, at least). Is this the case? If so maybe we shouldn't force the conda version. If not then this should be done just for the IC environment.

gonzaponte (changes requested):
We should have an automated test for this thing. Since we don't we should have a non-automated one at least. Try to run the script both installing and not installing conda. _neutrinos_ would be a good test bench for this as many people will use it there.

manage.sh
mmkekic: I dont think there should be any problems with keeping old environments with updated conda version, but maybe is safer to just print a message saying that the conda is outdated and one can update it with a given command

conftest.py
mmkekic:
```
settings._define_setting(
    "use_coverage",
    default=not_set,
    deprecation_message="""
use_coverage no longer does anything and can be removed from your settings.
""",
    deprecated_since="2017-09-14",
    description="""
A flag to enable a feature that no longer exists. This setting is present
only for backwards compatibility purposes.
""",
)
```
Seems that we couldn't deactivate for some time already!

gonzaponte (approved):
This PR updates our packages to a recent version, much needed after freezing them for 1.5 years.
There are some caveats:
- Some tests have been skipped due to a change in behavior of scipy.interpolate that doesn't affect the final result. These are to be fixed soon.
- Some tests have moved from a single-input to a hypothesis-generated input, which may increase test-execution time (we haven't evaluated it).

Otherwise, this is ready to be merged. Good job.

gonzaponte:
Please rebase and merge.

PR #725, opened 2020-06-11
Update db

carmenromo:
This PR includes the latest sensor calibrations from run 8041. Database has been updated with the PMT and SiPMs gains and the pdf values.
DICE 12 shows an small offset in the baseline of the spectra (maybe due to the last spark?), but it does not affect the gains computation. We will keep an eye on it...

pmtRunDifferencePlots

sipmRunDifferencePlots

chann12


jmbenlloch (approved):
Latest calibration added to the DB with all test passing. Approved!

PR #726, opened 2020-06-18
Change rounding in PMT simulation

andLaing:
Changes the way the output of the PMT simulation used in `diomira` is converted to integer
to better simulate the ADC.

This change should remove or significantly improve the ~15% loss of signal observed
in passing MC through the `diomira` city.

### DIFF
--- a/invisible_cities/cities/diomira.py
+++ b/invisible_cities/cities/diomira.py
@@ -159,7 +159,7 @@ def simulate_pmt_response(detector, run_number):
         rwf, blr = sf.simulate_pmt_response(0, pmtrd[np.newaxis],
                                             adc_to_pes, pe_resolution,
                                             detector, run_number)
-        return rwf.astype(np.int16), blr.astype(np.int16)
+        return np.round(rwf).astype(np.int16), np.round(blr).astype(np.int16)
     return simulate_pmt_response
 
 
--- a/invisible_cities/cities/diomira_test.py
+++ b/invisible_cities/cities/diomira_test.py
@@ -10,10 +10,14 @@ from .. core.configure     import                       all as all_events
 from .. core.configure     import                 configure
 from .. core.testing_utils import   assert_dataframes_close
 from .. core.testing_utils import    assert_tables_equality
+from .. database           import                   load_db
 from .. reco               import             tbl_functions as tbl
 from .. io  .mcinfo_io     import get_event_numbers_in_file
 from .. io  .mcinfo_io     import            load_mchits_df
 from .. io  .mcinfo_io     import       load_mcparticles_df
+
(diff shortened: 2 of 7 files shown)


andLaing:
These changes were implemented in discussion with Vicente Herrero where we agreed that the `round` functionality better represented the actual functionality of the PMT ADC. See attached plot.
kibcoanaboodncme

andLaing:
We found some other problems in the electronics simulation that caused loss of charge.

Specifically, the daq_decimate function was using an order 30 Chebyshev filter which affected the DC part of the signal, effectively reducing the magnitude. We have changed the function to use the scipy default which, while a bit slower, does not have a noticeable effect on the integrated signal. We plan to study `diomira` for bottlenecks in any case.

In order to set the ADC central values correctly we ran tests using files with 1, 2, 10, 20. 50 & 100 pe in various forms (1 bin, 2 bin, 5 bin,...) and checked that, without gain dispersion nor charge fluctuation, what value we got for the signal and that the other charge levels were the appropriate multiple of that value.

There is one test which I have marked skip, `fee_test.test_spe_to_adc`, this is an old test with a lot of hardwired numbers which tests a function which doesn't seem to be used in any other part of the code and simulates the electronics with the functions in an order which differs with respect to the sequence used in diomira. How should I deal with it? Change it to be more similar to the current paradigm or are we happy to remove the test?

andLaing:
[comp_PR726.pdf](https://github.com/nextic/IC/files/4955597/comp_PR726.pdf)
Here's a pdf with some plots comparing the input/output for the simplified electronics simulation (`pmtblr`) for the current master and this PR.

Similar reduction in systematics are seen comparing to the irene selection with more complicated distribution due to electronics noise and the full simulation and deconvolution steps.

mmkekic (commented):
Since the relevant changes in algorithm were approved by VH I can only review the code. 
Can you squash:
-  Update single_pe tests for new decimate, Mark test_diomira_reproduces_singlepe veryslow,  Test to show loss of single pe signal in pmt simulation;
-  Update description of marker, Add functions to run very slow tests with commandline option
-  Use new file in test_diomira_exact_result, New comparison file for test_diomira_exact_result

and I am happy to approve

mmkekic (approved):
This PR fixes ~15% signal loss after passing through diomira. The main changes are:
1. using round instead of floor for adc converter simulation (a test is added to show that for each PMT the mean ADC is equivalent to the gain)
2. change daq_decimate to use scipy default  (a test is added to show that the signal is maintained after the down-sampling) 
3. change  ADC central values 
All the algorithm changes are discussed and approved by V. Herrero. Good job Andrew in recovering our MC signal!

PR #728, opened 2020-06-23
Install and initialize git lfs on Travis OSX

### LINKED ISSUE #552: Travis failing on OS X, post LFS
It seems that git lfs is not working on OS X on Travis, making all our OS X builds fail. As these are formally allowed to fail (in order to reduce the delays before getting the green light, because the Travis OS X jobs sometimes take significantly longer to get going) we haven't really noticed or cared.

Git LFS needs to be properly configured on OS X Travis.

We should also review our current policy of allowing OS X builds to fail.

jacg:
Travis has been failing on OSX since we enabled Git LFS.

This is becasue Travis enables LFS on its Linux images but not on the OSX ones.

This PR fixes the problem by using `brew` to install Git LFS on the OSX image.

Closes #552.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,3 +22,11 @@ install:
 
 script:
   - HYPOTHESIS_PROFILE=travis-ci bash manage.sh run_tests_par
+
+
+before_install:
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then brew install git-lfs; fi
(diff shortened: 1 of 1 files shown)


jacg:
@jjgomezcadenas Is there any good reason to delay the review and merging of this almost trivial PR?

jmbenlloch (approved):
Simple fix to get OSX working again on Travis.

PR #733, opened 2020-07-17
Update beersheba related tests after package updates

Aretno:
When updating some packages in #723 there were some test that were marked as skipped after checking final result did not change performance-wise. This PR aims to update the values of the test so they can be enabled again.


gonzaponte (approved):
Scipy interpolate changed behavior in the recent package upgrade. The exact methodology change hasn't been found in the changelog, which is annoying. The differences in the values obtained in the tests vary around the percent level, but the overall performance of the algorithm hasn't been affected, according to @Aretno. Thus, the update simply relies on freezing these new values, hoping that they will not change again in a future release of scipy.

PR #734, opened 2020-07-24
Detsim signal simulation module.

gondiaz:
Following issue #691:

This module contains the functions that compute the photoelectrons at each sensor using the LightTables and PSF. Also, the functions that generate the S1 times distribution.


gondiaz:
Added new signal simulation module for the loop-based implementation of detsim.

gondiaz:
This PR is overriden by #756 #757.

PR #735, opened 2020-07-27
Fix non-integer sensor position in beersheba.

Aretno:
To accommodate for NEXT100 geommetry (15.55 mm pitch) we have had to modify the input to the interpolation function as well as the interpolation points themselves. The main issue was that the float precission was all over the place, and the ranges, min and max functions were returning values with tiny deviations. This, when acting slice by slice, caused a missmatch between slices. After discussing with @gonzaponte we have failed to see any other solution that add some rounding. In a real-life situation, the result is a fixed grid at n + 0.5 mm (n= 1, 2, etc.) when running at 1 mm deconvolution.

The code has already been used by @paolafer in NEXT100 reconstruction, I tested it with NEW. The only test that required some change, as the performance changed, was the `interpolate_signal` function as the grid that the test used is not the same as the grid coming from the modified function.

UPDATE:

After discussion with @mmkekic, we thought that the best solution to the problem was to generate a detector grid at the beginning of the process and then apply a range selection on a slice by slice basis before interpolation. This way a common reference point is guaranteed no matter the chosen deconvolution bin size. As a result, the general grid is now an argument needed in most of the deconvolution functions and is loaded in beersheba.

### DIFF
--- a/invisible_cities/cities/beersheba.py
+++ b/invisible_cities/cities/beersheba.py
@@ -39,6 +39,8 @@ from .. dataflow               import dataflow                as fl
 from .. dataflow.dataflow      import push
 from .. dataflow.dataflow      import pipe
 
+from .. database.load_db       import DataSiPM
+
 from .. reco.deconv_functions  import find_nearest
 from .. reco.deconv_functions  import cut_and_redistribute_df
 from .. reco.deconv_functions  import drop_isolated_sensors
@@ -66,7 +68,8 @@ class DeconvolutionMode(AutoNameEnumBase):
     separate = auto()
 
 
-def deconvolve_signal(psf_fname       : str,
+def deconvolve_signal(det_db          : pd.DataFrame,
+                      psf_fname       : str,
                       e_cut           : float,
                       n_iterations    : int,
                       iteration_tol   : float,
@@ -86,6 +89,7 @@ def deconvolve_signal(psf_fname       : str,
 
     Parameters
     ----------
+    det_db          : Detector database.
     psf_fname       : Point-spread function.
     e_cut           : Cut in relative value to the max voxel over the deconvolution output.
     n_iterations    : Number of Lucy-Richardson iterations
@@ -113,7 +117,10 @@ def deconvolve_signal(psf_fname       : str,
     diffusion     = np.asarray(diffusion              )
 
     psfs          = load_dst(psf_fname, 'PSF', 'PSFs')
-    deconvolution = deconvolve(n_iterations, iteration_tol, sample_width, bin_size, inter_method)
+    det_grid      = [np.arange(det_db[var].min() + bs/2, det_db[var].max() - bs/2 + np.finfo(np.float32).eps, bs)
+                     for var, bs in zip(dimensions, bin_size)]
+    deconvolution = deconvolve(n_iterations, iteration_tol,
+                               sample_width, det_grid       , inter_method)
 
     if not isinstance(energy_type , HitEnergy          ):
         raise ValueError(f'energy_type {energy_type} is not a valid energy type.')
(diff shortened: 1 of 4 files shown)


invisible_cities/cities/beersheba.py
```diff
@@ -113,7 +117,10 @@ def deconvolve_signal(psf_fname       : str,
     diffusion     = np.asarray(diffusion              )
 
     psfs          = load_dst(psf_fname, 'PSF', 'PSFs')
-    deconvolution = deconvolve(n_iterations, iteration_tol, sample_width, bin_size, inter_method)
+    xygrid         = [np.arange(det_db[var].min() + bs/2, det_db[var].max() - bs/2 + np.finfo(np.float32).eps, bs)
```
mmkekic: maybe the name is not the best, in principle this function should work even with ndim=3 and then it would be xyzgrid? how about `det_grid` (detector grid)?

invisible_cities/cities/beersheba.py
```diff
@@ -409,9 +416,9 @@ def beersheba(files_in, file_out, compression, event_range, print_mod, detector_
 
     with tb.open_file(file_out, "w", filters = tbl.filters(compression)) as h5out:
         # Define writers
-        write_event_info = fl.sink(run_and_event_writer(h5out), args=("run_number", "event_number", "timestamp"))
-        write_deconv     = fl.sink(  deconv_writer(h5out=h5out), args =  "deconv_dst"         )
-        write_summary    = fl.sink( summary_writer(h5out=h5out), args =  "summary"            )
+        write_event_info = fl.sink(run_and_event_writer (h5out), args=("run_number", "event_number", "timestamp"))
```
mmkekic: since you already touched the white spaces, maybe you can align `=` in which case you should put white space before and after `=`

invisible_cities/reco/deconv_functions.py
```diff
@@ -7,12 +7,12 @@
...
-from scipy        import interpolate
-from scipy.signal import fftconvolve
-from scipy.signal import convolve
+from scipy                  import interpolate
+from scipy.signal           import fftconvolve
+from scipy.signal           import convolve
 from scipy.spatial.distance import cdist
 
-from ..core .core_functions import shift_to_bin_centers
+from ..core .core_functions import shift_to_bin_centers, in_range
```
mmkekic: add separate line for `in_range` import

invisible_cities/reco/deconv_functions.py
```diff
@@ -129,21 +129,20 @@ def deconvolution_input(data        : Tuple[np.ndarray, ...],
                            ) -> Tuple[np.ndarray, Tuple[np.ndarray, ...]]:
 
         ranges = [[coord.min() - 1.5 * sw, coord.max() + 1.5 * sw]     for coord,   sw in zip(data      , sample_width)]
-        nbin   = [np.ceil(np.diff(rang)/bs).astype('int')[0]           for bs   , rang in zip(bin_size  ,       ranges)]
```
mmkekic, line 132: remove white spaces in previous line since you dont need to align it any more

mmkekic (commented):
Can we add another test that would demonstrate the problem? If you cant have next100 file even new file with bin size of say 9mm should show the problem? a test should check that for 2 z slices (obviously slices with different sipms touched), choosing 2 hits in both slices, that the offset of X and Y coordinates among those 2 hits are multiple of bin size

mmkekic (approved):
This PR ensures that the grid of deconvoluted image is fixed according to detector geometry and desired bin size, hence not varying between z-slices. This makes the code more flexible to accept arbitrary pitch/bin ratio which is needed for NEXT-100 geometry. Tests are added to demonstrate previous inconsistent behavior in case of non-integer pitch/bin ratio that pass after the fix. Good job!

PR #736, opened 2020-07-27
Modify 'load_dst' and 'load_dsts' to accept an event list.

Aretno:
Added argument to the usual load dst functions to just load events in
a given list. Usually loading several hdsts (and the alike) is a
memory heavy process as the full dst needs to be loaded and user does
not always need the full dst. This option allows for much faster and
lighter loading in such cases. If the new argument is not passed, the
function will behave as usual.

### DIFF
--- a/invisible_cities/io/dst_io.py
+++ b/invisible_cities/io/dst_io.py
@@ -18,13 +18,18 @@ def _decode_str_columns(df):
     df        = df.apply(to_string)
     return df
 
-def load_dst(filename, group, node):
+def load_dst(filename, group, node, evt_list=None):
     """load a kdst if filename, group and node correctly found"""
     try:
         with tb.open_file(filename) as h5in:
             try:
-                table = getattr(getattr(h5in.root, group), node).read()
-                return _decode_str_columns(pd.DataFrame.from_records(table))
+                table  = getattr(getattr(h5in.root, group), node)
+                if evt_list is None:
+                    values = table.read()
+                else:
+                    events = table.read(field='event')
+                    values = table.read_coordinates(np.where(np.isin(events, evt_list)))
+                return _decode_str_columns(pd.DataFrame.from_records(values, columns=table.colnames))
             except NoSuchNodeError:
                 warnings.warn(f' not of kdst type: file= {filename} ', UserWarning)
     except HDF5ExtError:
@@ -33,8 +38,8 @@ def load_dst(filename, group, node):
         warnings.warn(f' does not exist: file = {filename} ', UserWarning)
 
 
-def load_dsts(dst_list, group, node):
-    dsts = [load_dst(filename, group, node) for filename in dst_list]
+def load_dsts(dst_list, group, node, evt_list=None):
+    dsts = [load_dst(filename, group, node, evt_list) for filename in dst_list]
     return pd.concat(dsts, ignore_index=True)
 
 def _make_tabledef(column_types : np.dtype, str_col_length : int) -> dict:
--- a/invisible_cities/io/dst_io_test.py
+++ b/invisible_cities/io/dst_io_test.py
@@ -54,11 +54,31 @@ def test_load_dst(KrMC_kdst):
                             False  , rtol=1e-5)
 
 
+def test_load_dst_event_list(KrMC_kdst):
+    event_list = [0, 4, 7]
+    df_read    = load_dst(*KrMC_kdst[0].file_info, evt_list=event_list)
+    df_check   = KrMC_kdst[0].true
+    df_check   = df_check[df_check.event.isin(event_list)].reset_index(drop=True)
+    assert_dataframes_close(df_read, df_check,
+                            False  , rtol=1e-5)
+
+
 def test_load_dsts_single_file(KrMC_kdst):
     tbl     = KrMC_kdst[0].file_info
     df_read = load_dsts([tbl.filename], tbl.group, tbl.node)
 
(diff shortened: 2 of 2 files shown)


invisible_cities/io/dst_io.py
```diff
@@ -18,13 +18,15 @@ def _decode_str_columns(df):
...
+def load_dst(filename, group, node, evt_list=None):
     """load a kdst if filename, group and node correctly found"""
     try:
         with tb.open_file(filename) as h5in:
             try:
-                table = getattr(getattr(h5in.root, group), node).read()
-                return _decode_str_columns(pd.DataFrame.from_records(table))
+                table  = getattr(getattr(h5in.root, group), node)
+                values = (table.read() if evt_list is None
+                          else [ x for evt in evt_list for x in table.read_where(f"""(event == {evt})""") ])
```
mmkekic: Im not sure this is the best approach, guess this depends heavily on the user case. If the event list contains almost all the events from the file, this will be much slower than simply reading the table and selecting the desired events for two reasons:
1. lists are less efficient than numpy array
2. loop over `evt_list`. This is especially relevant in `load_dsts`where you will loop over `evt_list` that contains set of events from all the files you intend to read.

Im not sure whats the best option, maybe to read the event column, select indices and use `read_coordinates` table method? Since this PR is optimizing a piece of code guess the best would be to actually profile both time and memory of different options for your user case?

Aretno: The problem for me is not so much speed as memory, once you get to hdst and above loading several files becomes a nightmare. Specially in the case of deconvolution, where things become dramatically huge and slower, in my computer, is mostly impossible to load all the deconvolved events of even a few files and I must load event by event (getting the event list from the Run table). I have also found it to be pretty useful when looking into high energy peaks starting from penthesilea and in cases when you just have an event list and dont' know in which files are the events.  In any case, with the current implementation, the user may still use the old approach, no need to go through the event list (which, of course, for kdsts will be much slower, yeah)

The `read_coordinates` function could do the trick, yeah, I'll test it and report on the results.

Aretno: Did some quick test, just checking the time and memory peaks in my computer, and the improvement on velocity with the read_coordinates approach is amazing, starting from a single file with 10 million rows dst and keeping only 250k rows (15 events) after event filtering:
- load all events and selecting afterwards ~ 3 seconds, memory comsumption up to 2 gb.
- load iterating through lists ~ 41 seconds, memory comsumption up to 250 mb
- load iterating through read_coordinates ~ 3.5 seconds, memory comsumption up to 250 mb

so I would be very much inclined to use the `read_coordinates` approach and go ahead as we still keep the 'classical' functionality of the function.

mmkekic: Ok, that sounds good! You are still reading the full event column to select the indices right? Does that fit in the memory?

Aretno: I also tried to see what would happen with several files, using `load_dsts`, and keeping now 250 events. Thing just explodes when trying to load several files (around 200 million rows in total) before being able to filter out the events, following the current dst loading in IC. Using the coordinate thing, things are kept at around 500 mb and loads in around 1 minute, is not so fast but it manages to keep things really light.

Indeed I need to load the full event list but that does not seem to be problematic, the code would be as follows:

```
def load_dst(filename, group, node, evt_list=None):
    """load a kdst if filename, group and node correctly found"""
    try:
        with tb.open_file(filename) as h5in:
            try:
                table  = getattr(getattr(h5in.root, group), node)
                if evt_list is None:
                    values = table.read()
                else:
                    events = table.read(field='event')
                    values = table.read_coordinates(np.where(np.isin(events, evt_list)))
                return _decode_str_columns(pd.DataFrame.from_records(values, columns=table.colnames))
            except NoSuchNodeError:
                warnings.warn(f' not of kdst type: file= {filename} ', UserWarning)
    except HDF5ExtError:
        warnings.warn(f' corrupted: file = {filename} ', UserWarning)
    except IOError:
        warnings.warn(f' does not exist: file = {filename} ', UserWarning)
```

invisible_cities/io/dst_io_test.py
```diff
@@ -54,6 +54,15 @@ def test_load_dst(KrMC_kdst):
                             False  , rtol=1e-5)
 
 
+def test_load_dst_event_list(KrMC_kdst):
+    evt_list = [0, 4, 7]
+    df_read  = load_dst(*KrMC_kdst[0].file_info, evt_list)
```
mmkekic: maybe put explicit keyword for readability

invisible_cities/io/dst_io_test.py
```diff
@@ -62,6 +71,17 @@ def test_load_dsts_single_file(KrMC_kdst):
                             False  , rtol=1e-5)
 
 
+def test_load_dsts_single_file_event_list(KrMC_kdst):
```
mmkekic, line 74: if is not too much of a hassle (and test files are already in the test dataset) maybe do `load_dsts`for 2 files?

mmkekic (commented):
I think this PR is ready, maybe squash the commits (or leave if you think the previous version is good to be documented).

mmkekic (approved):
This simple change allows to load dataframes of only subset of events using pytables functions, hence being less  time and memory consuming than selecting events once the dataframe is read. A code is tested and profiled, showing that the current version is the most efficient one. Good job!

PR #737, opened 2020-08-08
Bufferization functions

andLaing:
Adds functions and utilities used to sort MC sensor data (full optical simulation) into buffers for IC processing.

Addresses point 6 of issue #691 

Functions can be seen in context in https://github.com/andLaing/detsim/tree/IC-integration-tests where the `position_signal.py ` script will form the basis for a future IC city.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -7,10 +7,12 @@ from os.path     import expandvars
 from itertools   import count
 from itertools   import repeat
 from enum        import Enum
+from typing      import Callable
 from typing      import Iterator
 from typing      import Mapping
 from typing      import List
 from typing      import Dict
+from typing      import Tuple
 from typing      import Union
 import tables as tb
 import numpy  as np
@@ -37,6 +39,7 @@ from .. core   .exceptions        import InvalidInputFileStructure
 from .. core   .configure         import                EventRange
 from .. core   .configure         import          event_range_help
 from .. core   .random_sampling   import              NoiseSampler
+from .. detsim                    import          buffer_functions as  bf
 from .. reco                      import           calib_functions as  cf
 from .. reco                      import          sensor_functions as  sf
 from .. reco                      import   calib_sensors_functions as csf
@@ -203,6 +206,53 @@ def copy_mc_info(files_in     : List[str],
             continue
     if len(np.setdiff1d(event_numbers, copied_events)) != 0:
         raise MCEventNotFound(f' Some events not found in MC tables')
+
+
+def wf_binner(max_buffer: int) -> Callable:
+    """
+    Returns a function to be used to convert the raw
+    input MC sensor info into data binned according to
+    a set bin width, effectively
+    padding with zeros inbetween the separate signals.
+
+    Parameters
+    ----------
+    max_buffer : float
+                 Maximum event time to be considered in nanoseconds
+    """
+    def bin_sensors(sensors  : pd.DataFrame,
+                    bin_width: float       ,
+                    t_min    : float       ,
+                    t_max    : float       ) -> Tuple[np.ndarray, pd.Series]:
+        return bf.bin_sensors(sensors, bin_width, t_min, t_max, max_buffer)
+    return bin_sensors
+
+
+def signal_finder(buffer_len   : float,
+                  bin_width    : float,
+                  bin_threshold:   int) -> Callable:
+    """
+    Decides where there is signal-like
+    charge according to the configuration
+    and the PMT sum in order to give
+    a useful position for buffer selection.
+    Currently simple threshold on binned charge.
+
+    Parameters
+    ----------
+    buffer_len    : float
+                    Configured buffer length in mus
+    bin_width     : float
+                    Sampling width for sensors
+    bin_threshold : int
+                    PE threshold for selection
+    """
+    # The stand_off is the minumum number of samples
+    # necessary between candidate triggers.
+    stand_off = int(buffer_len / bin_width)
+    def find_signal(wfs: pd.Series) -> List[int]:
+        return bf.find_signal_start(wfs, bin_threshold, stand_off)
+    return find_signal
 
 
 # TODO: consider caching database
--- a/invisible_cities/detsim/buffer_functions.py
+++ b/invisible_cities/detsim/buffer_functions.py
@@ -1,0 +1,141 @@
+import numpy  as np
+import pandas as pd
+
+from typing import  Callable
+from typing import      List
+from typing import     Tuple
+
+from .. reco.peak_functions import indices_and_wf_above_threshold
+from .. reco.peak_functions import                 split_in_peaks
+
+
+def weighted_histogram(data: pd.DataFrame, bins: np.ndarray) -> np.ndarray:
+    return np.histogram(data.time, weights=data.charge, bins=bins)[0]
+
+
+def bin_sensors(sensors   : pd.DataFrame,
+                bin_width : float       ,
+                t_min     : float       ,
+                t_max     : float       ,
+                max_buffer: int         ) -> Tuple[np.ndarray, pd.Series]:
+    """
+    Raw data binning function.
+
+    Parameters
+    ----------
+    sensors : List of Waveforms for one sensor type
+    Should be sorted into one type/binning
+    t_min   : float
+              Minimum time to be used to define bins.
+    t_max   : float
+              As t_min but the maximum to be used
+    """
+    max_time    = min(t_max, t_min + max_buffer)
+    min_bin     = np.floor(t_min    / bin_width) * bin_width
+    max_bin     = np.ceil (max_time / bin_width) * bin_width
+
+    bins        = np.arange(min_bin, max_bin + bin_width, bin_width)
+    bin_sensors = sensors.groupby('sensor_id').apply(weighted_histogram, bins)
+    return bins[:-1], bin_sensors
+
+
+## !! to-do: clarify for non-pmt versions of next
+## !! to-do: Check on integral instead of only threshold?
+def find_signal_start(wfs          : pd.Series,
+                      bin_threshold: float    ,
+                      stand_off    : int      ) -> List[int]:
+    """
+    Finds signal in the binned waveforms and
+    identifies candidate triggers.
+    """
+    eng_sum = wfs.sum()
+    indices = indices_and_wf_above_threshold(eng_sum,
+                                             bin_threshold).indices
+    ## Just using this and the stand_off for now
+    ## taking first above sum threshold.
+    ## !! To-do: make more robust with min int? or similar
+    all_indx = split_in_peaks(indices, stand_off)
+    return [pulse[0] for pulse in all_indx]
+
+
+def pad_safe(sensors: np.ndarray, padding: Tuple) -> np.ndarray:
+    """Pads zeros around each sensor in a 2D array"""
+    if not sensors.shape[0]:
+        return np.empty((0, padding[0] + padding[1] + 1))
+    return np.apply_along_axis(np.pad, 1, sensors, padding, "constant")
+
+
+def buffer_calculator(buffer_len: float, pre_trigger: float,
+                      pmt_binwid: float, sipm_binwid: float) -> Callable:
+    """
+    Calculates the output buffers for all sensors
+    based on a configured buffer length and pretrigger.
+    Synchronising the clock between the two sensor types.
+
+    Parameters
+    ----------
+    buffer_len  : float
+                  Length of buffer expected in mus
+    pre_trigger : float
+                  Time in buffer before identified signal in mus
+    pmt_binwid  : float
+                  Width in mus of PMT sample integration
+    sipm_binwid : float
+                  Width in mus of SiPM sample integration
+    """
+    pmt_buffer_samples  = int(buffer_len  //  pmt_binwid)
+    sipm_buffer_samples = int(buffer_len  // sipm_binwid)
+    sipm_pretrg         = int(pre_trigger // sipm_binwid)
+    sipm_postrg         = sipm_buffer_samples - sipm_pretrg
+    pmt_pretrg_base     = int(pre_trigger //  pmt_binwid)
+    pmt_postrg_base     = pmt_buffer_samples - pmt_pretrg_base
+
+    def generate_slice(trigger    : int       ,
+                       pmt_bins   : np.ndarray,
+                       pmt_charge : np.ndarray,
+                       sipm_bins  : np.ndarray,
+                       sipm_charge: np.ndarray) -> Tuple:
+        """
+        Synchronises the clocks between the SiPMs and PMTs
+        and slices the histograms where this synchronisation
+        indicates.
+        """
+        npmt_bin  = len(pmt_bins)
+        nsipm_bin = len(sipm_bins)
+
+        trg_bin    = np.where(sipm_bins <= pmt_bins[trigger])[0][-1]
+        bin_corr   = (pmt_bins[trigger] - sipm_bins[trg_bin]) // pmt_binwid
+        pmt_pretrg = pmt_pretrg_base + int(bin_corr)
+        pmt_postrg = pmt_postrg_base - int(bin_corr)
+
+        pmt_pre    = 0       , trigger - pmt_pretrg
+        pmt_pos    = npmt_bin, trigger + pmt_postrg
+        pmt_slice  = slice(max(pmt_pre), min(pmt_pos))
+        pmt_pad    = -min(pmt_pre), max(0, pmt_pos[1] - npmt_bin)
+
+        sipm_pre   = 0        , trg_bin - sipm_pretrg
+        sipm_pos   = nsipm_bin, trg_bin + sipm_postrg
+        sipm_slice = slice(max(sipm_pre), min(sipm_pos))
+        sipm_pad   = -min(sipm_pre), max(0, sipm_pos[1] - nsipm_bin)
+
+        return (pad_safe( pmt_charge[:,  pmt_slice],  pmt_pad),
+                pad_safe(sipm_charge[:, sipm_slice], sipm_pad))
+
+
+    def position_signal(triggers   :       List,
+                        pmt_bins   : np.ndarray,
+                        pmt_charge :  pd.Series,
+                        sipm_bins  : np.ndarray,
+                        sipm_charge:  pd.Series
+                        ) -> List[Tuple[np.ndarray, np.ndarray]]:
+        """
+        Synchronises the SiPMs and PMTs for each identified
+        trigger and calls the padding function to fill with
+        zeros where necessary.
+        """
+        pmt_q  = np.asarray(pmt_charge.tolist())
+        sipm_q = np.empty((0,0))\
+          if sipm_charge.empty else np.asarray(sipm_charge.tolist())
+        return [generate_slice(trigger, pmt_bins, pmt_q, sipm_bins, sipm_q)
+                for trigger in triggers                                    ]
+    return position_signal
--- a/invisible_cities/conftest.py
+++ b/invisible_cities/conftest.py
@@ -74,6 +74,11 @@ def nohits_sim_file(request, ICDATADIR):
 def sns_only_sim_file(request, ICDATADIR):
     return os.path.join(ICDATADIR, request.param)
 
+@pytest.fixture(scope  = 'session',
+                params = ['Kr83_full_nexus_v5_03_01_ACTIVE_7bar_1evt.sim.h5'])
+def full_sim_file(request, ICDATADIR):
+    return os.path.join(ICDATADIR, request.param)
+
 
 @pytest.fixture(scope='session')
 def mc_all_hits_data(krypton_MCRD_file):
--- a/invisible_cities/database/test_data/binned_simwfs.h5
+++ b/invisible_cities/database/test_data/binned_simwfs.h5
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:7254828b6e8cbec760007e4a23ab22507c214d9c3de00a0f597f9b572d130b67
+size 2331616
--- a/invisible_cities/detsim/buffer_functions_test.py
+++ b/invisible_cities/detsim/buffer_functions_test.py
@@ -1,0 +1,111 @@
+import numpy  as np
+import pandas as pd
+
+from pytest import mark
+
+from .. core            import system_of_units as units
+
+from . buffer_functions import       bin_sensors
+from . buffer_functions import buffer_calculator
+from . buffer_functions import find_signal_start
+
+
+def test_bin_sensors(mc_waveforms, pmt_ids, sipm_ids):
+    max_buffer = 10 * units.minute
+
+    evts, pmt_binwid, sipm_binwid, all_wfs = mc_waveforms
+
+    evt   = evts[0]
+    wfs   = all_wfs.loc[evt]
+
+    pmts  = wfs[wfs.index.isin( pmt_ids)]
+    sipms = wfs[wfs.index.isin(sipm_ids)]
+
+    ## Assumes pmts the triggering sensors as in new/next-100
+    bins_min = pmts.time.min()
+    bins_max = pmts.time.max() + pmt_binwid
+    pmt_bins ,  pmt_wf = bin_sensors(pmts ,  pmt_binwid,
+                                     bins_min, bins_max, max_buffer)
+    sipm_bins, sipm_wf = bin_sensors(sipms, sipm_binwid,
+                                     bins_min, bins_max, max_buffer)
+
+    assert  pmt_bins[ 0] >= bins_min
+    assert  pmt_bins[-1] <= min(bins_max, max_buffer)
+    assert np.all(np.diff( pmt_bins) ==  pmt_binwid)
+
+    assert sipm_bins[ 0] >= bins_min
+    assert sipm_bins[-1] <= min(bins_max, max_buffer)
+    assert np.all(np.diff(sipm_bins) == sipm_binwid)
+
+    ## In current DAQ, pmts have higher sample frequency than SiPMs
+    assert pmt_bins[ 0] >= sipm_bins[ 0]
+    assert pmt_bins[-1] >= sipm_bins[-1]
+
+    pmt_sum  = pmts .charge.sum()
+    sipm_sum = sipms.charge.sum()
+    assert pmt_wf .sum().sum() ==  pmt_sum
+    assert sipm_wf.sum().sum() == sipm_sum
+
+
+@mark.parametrize("signal_thresh", (2, 10))
+def test_find_signal_start(binned_waveforms, signal_thresh):
+
+    pmt_bins, pmt_wfs, *_ = binned_waveforms
+
+    buffer_length = 800 * units.mus
+    bin_width     = np.diff(pmt_bins)[0]
+    stand_off     = int(buffer_length / bin_width)
+
+    pmt_sum       = pmt_wfs.sum()
+    pulses        = find_signal_start(pmt_wfs, signal_thresh, stand_off)
+
+    assert np.all(pmt_sum[pulses] > signal_thresh)
+
+
+def test_find_signal_start_correct_index():
+
+    thresh     = 5
+    thr_bin    = 3
+    simple_evt = pd.Series([np.zeros(5)] * 12)
+    simple_evt[1][thr_bin] = thresh
+
+    pulses     = find_signal_start(simple_evt, thresh, 5)
+    assert len(pulses) == 1
+    assert pulses[0]   == thr_bin
+
+
+@mark.parametrize("pre_trigger signal_thresh".split(),
+                  ((100 * units.mus,  2),
+                   (400 * units.mus, 10)))
+def test_buffer_calculator(mc_waveforms, binned_waveforms,
+                           pre_trigger ,    signal_thresh):
+
+    _, pmt_binwid, sipm_binwid, _ = mc_waveforms
+
+    pmt_bins, pmt_wfs, sipm_bins, sipm_wfs = binned_waveforms
+
+    buffer_length     = 800 * units.mus
+    bin_width         = np.diff(pmt_bins)[0]
+    stand_off         = int(buffer_length / bin_width)
+
+    pulses            = find_signal_start(pmt_wfs, signal_thresh, stand_off)
+
+    calculate_buffers = buffer_calculator(buffer_length,
+                                          pre_trigger  ,
+                                          pmt_binwid   ,
+                                          sipm_binwid  )
+
+    buffers           = calculate_buffers(pulses, *binned_waveforms)
+    pmt_sum           = pmt_wfs.sum()
+
+    assert len(buffers) == len(pulses)
+    for i, (evt_pmt, evt_sipm) in enumerate(buffers):
+        sipm_trg_bin = np.where(sipm_bins <= pmt_bins[pulses[i]])[0][-1]
+        diff_binedge = pmt_bins[pulses[i]] - sipm_bins[sipm_trg_bin]
+        pre_trg_samp = int(pre_trigger / pmt_binwid + diff_binedge)
+
+        assert pmt_wfs .shape[0] == evt_pmt .shape[0]
+        assert sipm_wfs.shape[0] == evt_sipm.shape[0]
+        assert evt_pmt .shape[1] == int(buffer_length /  pmt_binwid)
+        assert evt_sipm.shape[1] == int(buffer_length / sipm_binwid)
+        assert np.sum(evt_pmt, axis=0)[pre_trg_samp] == pmt_sum[pulses[i]]
--- a/invisible_cities/detsim/conftest.py
+++ b/invisible_cities/detsim/conftest.py
@@ -1,0 +1,49 @@
+import os
+
+import numpy  as np
+import pandas as pd
+import tables as tb
+
+from pytest import fixture
+from pytest import    mark
+
+from .. core               import           system_of_units as units
+from .. database           import                   load_db
+from .. io      .mcinfo_io import        get_sensor_binning
+from .. io      .mcinfo_io import load_mcsensor_response_df
+
+from .  buffer_functions   import               bin_sensors
+
+
+@fixture(scope="module")
+def mc_waveforms(full_sim_file):
+    file_in = full_sim_file
+    wfs     = load_mcsensor_response_df(file_in, db_file='new', run_no=-6400)
+
+    sns_bins    = get_sensor_binning(file_in)
+    pmt_binwid  = sns_bins.bin_width[sns_bins.index.str.contains( 'Pmt')]
+    sipm_binwid = sns_bins.bin_width[sns_bins.index.str.contains('SiPM')]
+    return wfs.index.levels[0], pmt_binwid.iloc[0], sipm_binwid.iloc[0], wfs
+
+
+## !! to-do: generalise for all detector configurations
+@fixture(scope="module")
+def pmt_ids():
+    return load_db.DataPMT('new', 6400).SensorID.values
+
+
+@fixture(scope="module")
+def sipm_ids():
+    return load_db.DataSiPM('new', 6400).SensorID.values
+
+
+@fixture(scope="module")
+def binned_waveforms(ICDATADIR):
+    binned_file = os.path.join(ICDATADIR, 'binned_simwfs.h5')
+    with tb.open_file(binned_file) as h5in:
+        pmt_bins  = h5in.root.BINWIDTHS.pmt_binwid .read()
+        sipm_bins = h5in.root.BINWIDTHS.sipm_binwid.read()
+
+    pmt_wf  = pd.read_hdf(binned_file,  'pmtwfs')
+    sipm_wf = pd.read_hdf(binned_file, 'sipmwfs')
+    return pmt_bins, pmt_wf, sipm_bins, sipm_wf
--- a/invisible_cities/detsim/sensor_utils.py
+++ b/invisible_cities/detsim/sensor_utils.py
@@ -1,0 +1,96 @@
+import numpy  as np
+import pandas as pd
+
+from typing import Callable
+from typing import     List
+from typing import    Tuple
+
+from .. database.load_db   import            DataPMT
+from .. database.load_db   import           DataSiPM
+from .. io      .mcinfo_io import get_sensor_binning
+
+
+def trigger_times(trigger_indx: List[int] ,
+                  event_time  :      float,
+                  time_bins   : np.ndarray) -> List[float]:
+    """
+    Calculates trigger time for all identified candidates
+    according to the nexus event time and the time within
+    the event where the trigger was identified.
+
+    Parameters
+    ----------
+    trigger_indx : List[int]
+                   The indices in the histogrammed nexus sensor
+                   data where triggers were identified.
+    event_time   : float
+                   Time of nexus event
+    time_bins    : np.ndarray
+                   The binning used to histogram the nexus sensor
+                   data.
+
+    Returns
+    -------
+    triggered event times : List[float]
+    """
+    return event_time + time_bins[trigger_indx]
+
+
+def first_and_last_times(pmt_wfs    : pd.DataFrame,
+                         sipm_wfs   : pd.DataFrame,
+                         pmt_binwid : float       ,
+                         sipm_binwid: float       ) -> Tuple[float, float]:
+    """
+    Returns the maximum and minimum time of an
+    event given the two types of detector.
+    """
+    min_time  = min(pmt_wfs.time.min(), sipm_wfs.time.min())
+    max_time  = max(pmt_wfs.time.max(), sipm_wfs.time.max())
+    max_time += min(pmt_binwid        ,         sipm_binwid)
+    return min_time, max_time
+
+
+def sensor_order(detector_db: str, run_number : int,
+                 length_pmt : int, length_sipm: int) -> Callable:
+    """
+    Casts the event sensor info into the correct order
+    adding zeros for sensors which didn't see any signal.
+    """
+    pmt_ids       = DataPMT (detector_db, run_number).SensorID
+    sipm_ids      = DataSiPM(detector_db, run_number).SensorID
+    n_pmt, n_sipm = get_n_sensors(detector_db, run_number)
+    pmt_shape     = (n_pmt , length_pmt )
+    sipm_shape    = (n_sipm, length_sipm)
+    def ordering(sensor_order : pd.Int64Index  ,
+                 sensor_resp  : np.ndarray     ,
+                 sensor_shape : Tuple[int, int]) -> np.ndarray:
+        sensors = np.zeros(sensor_shape, np.int)
+        sensors[sensor_order] = sensor_resp
+        return sensors
+        
+    def order_and_pad(pmt_resp    : pd.Series                          ,
+                      sipm_resp   : pd.Series                          ,
+                      evt_buffers : List[Tuple[np.ndarray, np.ndarray]]
+                      ) -> List[Tuple]:
+        pmt_ord  = pmt_ids [ pmt_ids.isin( pmt_resp.index)].index
+        sipm_ord = sipm_ids[sipm_ids.isin(sipm_resp.index)].index
+
+        return [(ordering(pmt_ord , pmts , pmt_shape ),
+                 ordering(sipm_ord, sipms, sipm_shape))
+                    for pmts, sipms in evt_buffers]
+    return order_and_pad
+
+
+def get_n_sensors(detector_db: str, run_number: int) -> Tuple[int, int]:
+    """Get the number of sensors for this run"""
+    npmt  = DataPMT (detector_db, run_number).shape[0]
+    nsipm = DataSiPM(detector_db, run_number).shape[0]
+    return npmt, nsipm
+
+
+def pmt_and_sipm_bin_width(file_name: str) -> Tuple[float, float]:
+    """returns pmt and sipm bin widths as set in nexus"""
+    sns_bins = get_sensor_binning(file_name)
+    pmt_wid  = sns_bins.bin_width[sns_bins.index.str.contains( 'Pmt')].iloc[0]
+    sipm_wid = sns_bins.bin_width[sns_bins.index.str.contains('SiPM')].iloc[0]
+    return pmt_wid, sipm_wid
--- a/invisible_cities/detsim/sensor_utils_test.py
+++ b/invisible_cities/detsim/sensor_utils_test.py
@@ -1,0 +1,99 @@
+import numpy  as np
+import pandas as pd
+
+from pytest import fixture
+from pytest import    mark
+
+from .. core import system_of_units as units
+
+from . sensor_utils import   first_and_last_times
+from . sensor_utils import          get_n_sensors
+from . sensor_utils import           sensor_order
+from . sensor_utils import pmt_and_sipm_bin_width
+from . sensor_utils import          trigger_times
+
+
+def test_trigger_times():
+    event_time = 22 * units.ns
+    bin_width  = 0.1
+    time_bins  = np.arange(0, 10, bin_width)
+    triggers   = [15, 75]
+
+    trgr_times = trigger_times(triggers, event_time, time_bins)
+    assert len(trgr_times) == len(triggers)
+    assert trgr_times[0]   == event_time + triggers[0] * bin_width
+    assert trgr_times[1]   == event_time + triggers[1] * bin_width
+
+
+@mark.parametrize("bin_min bin_max bin_wid".split(),
+                  ((22, 40, 1), (1, 5, 0.1)))
+def test_first_and_last_times(bin_min, bin_max, bin_wid):
+    test_bins1 = np.arange(bin_min, bin_max, bin_wid)
+    test_bins2 = np.arange(bin_min, bin_max - 1)
+
+    bmin, bmax = first_and_last_times(pd.DataFrame({'time': test_bins1}),
+                                      pd.DataFrame({'time': test_bins2}),
+                                      bin_wid, 1)
+
+    assert np.isclose(bmin, bin_min)
+    assert np.isclose(bmax, bin_max)
+
+
+@fixture(scope = 'function')
+def sensor_info():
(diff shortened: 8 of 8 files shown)


andLaing:
There's probably some scope for change and certainly we should work out the way that all or part of these functions could be used as the final part of the detsim flow too.

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,201 @@
...
+from typing    import Generator
+from typing    import      List
+from typing    import     Tuple
+
+from .. core                import                system_of_units as units
+from .. reco.peak_functions import indices_and_wf_above_threshold
+from .. reco.peak_functions import                 split_in_peaks
+
+
+@wraps(np.histogram)
```
gonzaponte: I think your intention is to indicate that this is just a special `np.histogram` call, but `wraps` would be misleading, as you would expect exactly the same call signature, which is not the case. If this function is only used in `wf_binner.bin_data`, you can define it in place, if you want.

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,201 @@
...
+def weighted_histogram(data: pd.DataFrame, bins: np.ndarray) -> np.ndarray:
+    return np.histogram(data.time, weights=data.charge, bins=bins)[0]
+
+
+def wf_binner(max_buffer: int) -> Callable:
```
gonzaponte: This looks like a city component, rather than a generic function. Maybe the scheme should rather be to define `bin_data` as a generic function with an extra argument `max_buffer` and `wf_binner` a city component that calls it fixing that argument.

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,201 @@
...
+                                                         np.append(bins, b_pad))
+        return bins, bin_sensors
+    return bin_data
+
+
+## !! to-do: clarify for non-pmt versions of next
+## !! to-do: Check on integral instead of only threshold?
+def signal_finder(buffer_len   : float,
+                  bin_width    : float,
+                  bin_threshold:   int) -> Callable:
```
gonzaponte: Same. This looks like a city component.

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,201 @@
...
+    buffer_len    : float
+                    Configured buffer length in mus
+    bin_width     : float
+                    Sampling width for sensors
+    bin_threshold : int
+                    PE threshold for selection
+    """
+
+    stand_off = int(buffer_len * units.mus / bin_width)
+    def find_signal(wfs: pd.Series) -> List[int]:
```
gonzaponte: In this case, `find_signal` could work taking `stand_off` and `bin_threshold` as parameters and `signal_finder` should calculate `stand_off`.

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,201 @@
...
+    ----------
+    buffer_len    : float
+                    Configured buffer length in mus
+    bin_width     : float
+                    Sampling width for sensors
+    bin_threshold : int
+                    PE threshold for selection
+    """
+
+    stand_off = int(buffer_len * units.mus / bin_width)
```
gonzaponte: Maybe an inline comment explaining what `stand_off` means in this context? I think I get it, seeing how it is used, but it would help to have a short clarification.

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,201 @@
...
+    def find_signal(wfs: pd.Series) -> List[int]:
+
+        eng_sum = wfs.sum(0)
```
gonzaponte: I'm a bit puzzled by this. If `wfs` was a `np.ndarray` this would mean "sum over the outermost axis". But this being a series, it is 1D, isn't it? But then you use it as an array. Maybe the type annotation is wrong.

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,201 @@
...
+        all_indx = split_in_peaks(indices, stand_off)
+        return [pulse[0] for pulse in all_indx]
+    return find_signal
+
+
+def padder(sensors: np.ndarray, padding: Tuple) -> np.ndarray:
+    """Pads zeros around each sensor in a 2D array"""
+    if not sensors.shape[0]:
+        return np.empty((0, padding[0] + padding[1] + 1))
+    return np.apply_along_axis(np.pad, 1, sensors, padding, "constant")
```
gonzaponte: If `padding` is a tuple with presumably different values (otherwise you would have used an integer) and `sensors` is a (presumably) homogeneous `np.ndarray` (all columns or rows have the same length), padding a different number of elements would result in a non-homogeneous `np.ndarray`. Is that desired?

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,201 @@
...
+                  Width in mus of PMT sample integration
+    sipm_binwid : float
+                  Width in mus of SiPM sample integration
+    """
+    pmt_buffer_samples  = int(buffer_len * units.mus /  pmt_binwid)
+    sipm_buffer_samples = int(buffer_len * units.mus / sipm_binwid)
+    sipm_pretrg         = int(pre_trigger * units.mus / sipm_binwid)
+    sipm_postrg         = sipm_buffer_samples - sipm_pretrg
+    pmt_pretrg_         = int(pre_trigger * units.mus / pmt_binwid)
+    pmt_postrg_         = pmt_buffer_samples - pmt_pretrg_
```
gonzaponte: why the trailing underscore?

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,201 @@
...
+    def sipm_trg_bin(sipm_bins: np.ndarray,
+                     pmt_bins : np.ndarray) -> Callable[[int], int]:
+        """
+        Identifies the SiPM bin in which the trigger
+        was identified in the PMTs.
+        Assumes finer or equal bin size for PMTs compared to SiPMs
+        """
+        def get_sipm_bin(trigger: int) -> int:
+            return np.where(sipm_bins <= pmt_bins[trigger])[0][-1]
+        return get_sipm_bin
```
gonzaponte: Any reason not to define it as a single function with three arguments?
Since it is a one-liner used only once you can also just plug it in the corresponding place.

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,201 @@
...
+        and slices the histograms where this synchronisation
+        indicates.
+        """
+        npmt_bin  = len(pmt_bins)
+        nsipm_bin = len(sipm_bins)
+        sipm_trg  = sipm_trg_bin(sipm_bins, pmt_bins)
+
+        def generate_slices(triggers: List) -> Tuple:
+
+            for trg in triggers:
```
gonzaponte: Seen how you use it, I think it would be more useful to define it for a single trigger and move the loop outside.

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,201 @@
...
+                        sipm_bins  : np.ndarray,
+                        sipm_charge:  pd.Series) -> List:
+        """
+        Synchronises the SiPMs and PMTs for each identified
+        trigger and calls the padding function to fill with
+        zeros where necessary.
+        """
+
+        sipm_q = np.empty((0,0))\
+          if sipm_charge.empty else np.array(sipm_charge.tolist())
```
gonzaponte: Isn't `np.array(sipm_charge.tolist())` the same as `sipm_charge.values` or do you need to make a copy?
Why is the shape `(0,0)` if the series is empty? shouldn't it be `(0,)`?
If `(0,)` is the correct shape, then `sipm_q = sipm_charge.values` would cover both cases.

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,201 @@
...
+        slice_and_pad = slice_generator(pmt_bins                     ,
+                                        np.array(pmt_charge.tolist()),
```
gonzaponte: `pmt_charge.values`?

invisible_cities/detector_simulation/sensor_utils.py
```diff
@@ -0,0 +1,95 @@
...
+                 sensor_shape : Tuple        ) -> np.ndarray:
+        sensors = np.zeros(sensor_shape, np.int)
+        sensors[sensor_order] = sensor_resp
+        return sensors
+        
+    def order_and_pad(pmt_resp    : pd.Series  ,
+                      sipm_resp   : pd.Series  ,
+                      evt_buffers : List[Tuple]) -> List[Tuple]:
+        pmt_ord  = pmt_ids [ pmt_ids.isin( pmt_resp.index.tolist())].index
+        sipm_ord = sipm_ids[sipm_ids.isin(sipm_resp.index.tolist())].index
```
gonzaponte: is `tolist()` necessary?

invisible_cities/detector_simulation/sensor_utils.py
```diff
@@ -0,0 +1,95 @@
...
+
+        return [(ordering(pmt_ord , pmts , pmt_shape ),
+                 ordering(sipm_ord, sipms, sipm_shape))
+                    for pmts, sipms in evt_buffers]
+    return order_and_pad
+
+
+def get_no_sensors(detector_db: str, run_number: int) -> Tuple:
```
gonzaponte: May I suggest changing `no` to `n` or `number_of`?

gonzaponte (commented):
I haven't taken a look to the tests just yet, but they will probably change a bit with this.

invisible_cities/detector_simulation/buffer_functions.py
andLaing: That was my first thought but `.values` on the Series with lists as values (they're histograms for each sensor) gives and array of arrays and not a 2D np.ndarray which is what we want so we can use slicing to get everything into buffers.

invisible_cities/detector_simulation/buffer_functions.py
andLaing: Using `(0,0)` protects agains the use of 2D slicing in the other functions in the rare cases where we don't have SiPM info.

invisible_cities/detector_simulation/buffer_functions.py
andLaing: Padding is a Tuple because what you need to pad before the charge is rarely the same as what you need after.

If you have identified the first bin with charge as the trigger, for example, you need to pad by the number of samples corresponding to pretrigger but at the end it could be almost any number

invisible_cities/detector_simulation/sensor_utils.py
andLaing: No, I must have forgotten to check after changing something. I'll sort it.

invisible_cities/detector_simulation/buffer_functions.py
andLaing: Because I correct it later to be the actual used post trigger samples but it's not clear, I'll sort it

andLaing:
I think I addressed the initial comments in the last set of pushes. Whenever the reviewers have some time, the PR can be checked again.

invisible_cities/cities/components.py
```diff
@@ -205,6 +208,53 @@ def copy_mc_info(files_in     : List[str],
...
+    buffer_len    : float
+                    Configured buffer length in mus
+    bin_width     : float
+                    Sampling width for sensors
+    bin_threshold : int
+                    PE threshold for selection
+    """
+    # The stand_off is the minumum number of samples
+    # necessary between candidate triggers.
+    stand_off = int(buffer_len * units.mus / bin_width)
```
gonzaponte: We are lacking the city here, but I suppose `buffer_len` will be read from config file and if so, will it not come already with units? In general, units should only be applied when inputting data to the program (either by hardcoding parameters or by reading them from somewhere where they are assumed to be in a specific unit). (Most) functions should assume everything comes in "default" units.

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,149 @@
...
+    """
+    max_time    = min(t_max, t_min + max_buffer)
+    min_bin     = np.floor(t_min    / bin_width) * bin_width
+    max_bin     = np.ceil (max_time / bin_width) * bin_width
+
+    bins        = np.arange(min_bin, max_bin, bin_width)
+    b_pad       = bins[-1] + bin_width
+    bin_sensors = sensors.groupby('sensor_id').apply(weighted_histogram    ,
+                                                     np.append(bins, b_pad))
+    return bins, bin_sensors
```
gonzaponte: Maybe there is an edge case where this doesn't work (?) but, isn't it simpler:
```suggestion
    bins        = np.arange(min_bin, max_bin + bin_width, bin_width)
    bin_sensors = sensors.groupby('sensor_id').apply(weighted_histogram, bins)
    return bins[:-1], bin_sensors
```

Also, is it intended to not return the last upper bin edge?

invisible_cities/detector_simulation/buffer_functions_test.py
```diff
@@ -0,0 +1,90 @@
...
+    bins_max = pmts.time.max() + pmt_binwid
+    pmt_bins ,  pmt_wf = bin_sensors(pmts ,  pmt_binwid,
+                                     bins_min, bins_max, max_buffer)
+    sipm_bins, sipm_wf = bin_sensors(sipms, sipm_binwid,
+                                     bins_min, bins_max, max_buffer)
+
+    assert np.all(np.diff( pmt_bins) ==  pmt_binwid)
+    assert np.all(np.diff(sipm_bins) == sipm_binwid)
+    assert pmt_bins[ 0] >= sipm_bins[ 0]
+    assert pmt_bins[-1] >= sipm_bins[-1]
```
gonzaponte: Maybe comment that this is because of the different sensor sampling freqs?

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,149 @@
...
+
+def weighted_histogram(data: pd.DataFrame, bins: np.ndarray) -> np.ndarray:
+    return np.histogram(data.time, weights=data.charge, bins=bins)[0]
+
+
+def bin_sensors(sensors   : pd.Series,
+                bin_width : float    ,
+                t_min     : float    ,
+                t_max     : float    ,
+                max_buffer: int      ) -> Tuple:
```
gonzaponte: Probably the output annotation should be `Tuple[np.ndarray, pd.Series]`. Also in the corresponding city component.

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,149 @@
...
+from .. core                import                system_of_units as units
+from .. reco.peak_functions import indices_and_wf_above_threshold
+from .. reco.peak_functions import                 split_in_peaks
+
+
+def weighted_histogram(data: pd.DataFrame, bins: np.ndarray) -> np.ndarray:
+    return np.histogram(data.time, weights=data.charge, bins=bins)[0]
+
+
+def bin_sensors(sensors   : pd.Series,
```
gonzaponte: `sensors`' type hint should be `pd.DataFrame`, right? If so, check also the city component.

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,149 @@
...
+    bins        = np.arange(min_bin, max_bin, bin_width)
+    b_pad       = bins[-1] + bin_width
+    bin_sensors = sensors.groupby('sensor_id').apply(weighted_histogram    ,
+                                                     np.append(bins, b_pad))
+    return bins, bin_sensors
+
+
+## !! to-do: clarify for non-pmt versions of next
+## !! to-do: Check on integral instead of only threshold?
+def find_signal(wfs          : pd.Series,
```
gonzaponte: I find this confusing. Cause I think of `pd.Series` as 1D objects, but these ones contain arrays as items, therefore are effectively 2D. I am trying to see if this is the best container for these objects, but for now maybe we should annotate it, at least. Try `pd.Series[np.ndarray]` for now.

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,149 @@
...
+    identifies candidate triggers.
+    """
+    eng_sum = wfs.sum()
+    indices = indices_and_wf_above_threshold(eng_sum,
+                                             bin_threshold).indices
+    ## Just using this and the stand_off for now
+    ## taking first above sum threshold.
+    ## !! To-do: make more robust with min int? or similar
+    all_indx = split_in_peaks(indices, stand_off)
+    return [pulse[0] for pulse in all_indx]
```
gonzaponte: Without the overall picture, it's difficult to see why you only care about the first index of each, but I assume this will be clear once the corresponding city is implemented, right? Maybe I would rename the function to `find_signal_raise/start`/`find_triggers` or whatever you find more appropriate.

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def padder(sensors: np.ndarray, padding: Tuple) -> np.ndarray:
```
gonzaponte: `padder` sounds like a city component. `pad` or `pad_safe`, maybe?

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,149 @@
...
+                  Width in mus of PMT sample integration
+    sipm_binwid : float
+                  Width in mus of SiPM sample integration
+    """
+    pmt_buffer_samples  = int(buffer_len * units.mus /  pmt_binwid)
+    sipm_buffer_samples = int(buffer_len * units.mus / sipm_binwid)
+    sipm_pretrg         = int(pre_trigger * units.mus / sipm_binwid)
+    sipm_postrg         = sipm_buffer_samples - sipm_pretrg
+    pmt_pretrg_base     = int(pre_trigger * units.mus / pmt_binwid)
+    pmt_postrg_base     = pmt_buffer_samples - pmt_pretrg_base
```
gonzaponte: Same comment as before regarding the units, will they not come from config file or similar and come with units already?

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def generate_slice(trigger    : List      ,
```
gonzaponte:
```suggestion
    def generate_slice(trigger    : int       ,
```
right?

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,149 @@
...
+        """
+        Synchronises the clocks between the SiPMs and PMTs
+        and slices the histograms where this synchronisation
+        indicates.
+        """
+        npmt_bin  = len(pmt_bins)
+        nsipm_bin = len(sipm_bins)
+
+        trg_bin    = np.where(sipm_bins <= pmt_bins[trigger])[0][-1]
+        bin_corr   = (pmt_bins[trigger] - sipm_bins[trg_bin]) / pmt_binwid
```
gonzaponte:
```suggestion
        bin_corr   = (pmt_bins[trigger] - sipm_bins[trg_bin]) // pmt_binwid
```
It doesn't change anything, cause there is an `int` truncation just after this line, but it is more correct algorithmically.

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,149 @@
...
+        pmt_pretrg = pmt_pretrg_base + int(bin_corr)
+        pmt_postrg = pmt_postrg_base - int(bin_corr)
+
+        pmt_pre    = 0       , trigger - pmt_pretrg
+        pmt_pos    = npmt_bin, trigger + pmt_postrg
+        pmt_sl     = slice(max(pmt_pre), min(pmt_pos))
+        pmt_pad    = (int(-min(pmt_pre)),
+                      int( max(0, pmt_pos[1] - npmt_bin)))
```
gonzaponte: these `int` conversions are not necessary, right? same for sipms

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,149 @@
...
+
+        sipm_pre   = 0        , trg_bin - sipm_pretrg
+        sipm_pos   = nsipm_bin, trg_bin + sipm_postrg
+        sipm_sl    = slice(max(sipm_pre), min(sipm_pos))
+        sipm_pad   = (int(-min(sipm_pre)),
+                      int( max(0, sipm_pos[1] - nsipm_bin)))
+
+        return (padder(pmt_charge [:,  pmt_sl],  pmt_pad),
+                padder(sipm_charge[:, sipm_sl], sipm_pad))
```
gonzaponte:
```suggestion
        return (padder( pmt_charge[:,  pmt_sl],  pmt_pad),
                padder(sipm_charge[:, sipm_sl], sipm_pad))
```

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,149 @@
...
+        """
+        Synchronises the SiPMs and PMTs for each identified
+        trigger and calls the padding function to fill with
+        zeros where necessary.
+        """
+        sipm_q = np.empty((0,0))\
+          if sipm_charge.empty else np.array(sipm_charge.tolist())
+        return [generate_slice(trigger                      ,
+                               pmt_bins                     ,
+                               np.array(pmt_charge.tolist()),
```
gonzaponte: Aesthetically maybe better to define `pmt_q = np.array(pmt_charge.tolist())` alongside `sipm_q`

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,149 @@
...
+
+        return (padder(pmt_charge [:,  pmt_sl],  pmt_pad),
+                padder(sipm_charge[:, sipm_sl], sipm_pad))
+
+
+    def position_signal(triggers   :       List,
+                        pmt_bins   : np.ndarray,
+                        pmt_charge :  pd.Series,
+                        sipm_bins  : np.ndarray,
+                        sipm_charge:  pd.Series) -> List:
```
gonzaponte:
```suggestion
                        sipm_charge:  pd.Series) -> List[Tuple[np.ndarray, np.ndarray]]:
```
maybe ?

invisible_cities/detector_simulation/buffer_functions_test.py
```diff
@@ -0,0 +1,90 @@
...
+    ## Assumes pmts the triggering sensors as in new/next-100
+    bins_min = pmts.time.min()
+    bins_max = pmts.time.max() + pmt_binwid
+    pmt_bins ,  pmt_wf = bin_sensors(pmts ,  pmt_binwid,
+                                     bins_min, bins_max, max_buffer)
+    sipm_bins, sipm_wf = bin_sensors(sipms, sipm_binwid,
+                                     bins_min, bins_max, max_buffer)
+
+    assert np.all(np.diff( pmt_bins) ==  pmt_binwid)
+    assert np.all(np.diff(sipm_bins) == sipm_binwid)
```
gonzaponte: Perhaps check that the first and last bins are within the time limits for completeness.

invisible_cities/detector_simulation/buffer_functions_test.py
```diff
@@ -0,0 +1,90 @@
...
+    pmt_bins, pmt_wf, *_ = binned_waveforms
+
+    buffer_length = 800
+    bin_width     = np.diff(pmt_bins)[0]
+    stand_off     = int(buffer_length * units.mus / bin_width)
+
+    pmt_sum       = pmt_wf.sum()
+    pulses        = find_signal(pmt_wf, signal_thresh, stand_off)
+
+    assert np.all(pmt_sum[pulses] > signal_thresh)
```
gonzaponte: Another test checking that it finds the right indices for a simple event.

invisible_cities/detector_simulation/buffer_functions_test.py
```diff
@@ -0,0 +1,90 @@
...
+
+    pmt_bins, pmt_wf, sipm_bins, sipm_wf = binned_waveforms
+
+    buffer_length     = 800
+    bin_width         = np.diff(pmt_bins)[0]
+    stand_off         = int(buffer_length * units.mus / bin_width)
+
+    pulses            = find_signal(pmt_wf, signal_thresh, stand_off)
+
+    buffer_calculator = calculate_buffers(buffer_length,
```
gonzaponte: They should be named the other way around:
`calculate_buffers = buffer_calculator(...)`

invisible_cities/detector_simulation/buffer_functions_test.py
```diff
@@ -0,0 +1,90 @@
...
+                                          pre_trigger  ,
+                                          pmt_binwid   ,
+                                          sipm_binwid  )
+
+    buffers           = buffer_calculator(pulses, *binned_waveforms)
+    pmt_sum           = pmt_wf.sum()
+
+    assert len(buffers) == len(pulses)
+    for i, evt in enumerate(buffers):
```
gonzaponte:
```suggestion
    for i, (evt_pmt, evt_sipm) in enumerate(buffers):
```
to avoid `evt[0 / 1]` later

invisible_cities/detector_simulation/buffer_functions_test.py
```diff
@@ -0,0 +1,90 @@
...
+        sipm_trg_bin = np.where(sipm_bins <= pmt_bins[pulses[i]])[0][-1]
+        diff_binedge = pmt_bins[pulses[i]] - sipm_bins[sipm_trg_bin]
+        pre_trg_samp = int(pre_trigger * units.mus / pmt_binwid + diff_binedge)
+
+        assert pmt_wf .shape[0] == evt[0].shape[0]
+        assert sipm_wf.shape[0] == evt[1].shape[0]
+        assert evt[0] .shape[1] == int(buffer_length * units.mus / pmt_binwid)
```
gonzaponte: why not check also `evt[1].shape[1]`?

invisible_cities/detector_simulation/conftest.py
```diff
@@ -0,0 +1,58 @@
...
+@fixture(scope="module")
+def mc_waveforms(ICDATADIR):
+    file_in = os.path.join(ICDATADIR                                         ,
+                           'Kr83_full_nexus_v5_03_01_ACTIVE_7bar_1evt.sim.h5')
+    wfs     = load_mcsensor_response_df(file_in, db_file='new', run_no=-6400)
+    
+    sns_bins    = get_sensor_binning(file_in)
+    pmt_binwid  = sns_bins.bin_width[sns_bins.index.str.contains( 'Pmt')]
+    sipm_binwid = sns_bins.bin_width[sns_bins.index.str.contains('SiPM')]
+    return wfs.index.levels[0], pmt_binwid.iloc[0], sipm_binwid.iloc[0], wfs
```
gonzaponte: You can probably benefit from defining a namedtuple here for this output. Also in binned_waveforms.

invisible_cities/detector_simulation/conftest.py
```diff
@@ -0,0 +1,58 @@
...
+    pmts  = wfs[wfs.index.isin( pmt_ids)]
+    sipms = wfs[wfs.index.isin(sipm_ids)]
+
+    ## Assumes pmts the triggering sensors as in new/next-100
+    bins_min = pmts.time.min()
+    bins_max = pmts.time.max() + pmt_binwid
+    pmt_bins ,  pmt_wf = bin_sensors(pmts ,  pmt_binwid,
+                                     bins_min, bins_max, max_buffer)
+    sipm_bins, sipm_wf = bin_sensors(sipms, sipm_binwid,
+                                     bins_min, bins_max, max_buffer)
```
gonzaponte: We should avoid using our own functions in fixtures. Would it be possible to just create some fake waveforms that would work for this?

invisible_cities/detector_simulation/sensor_utils.py
```diff
@@ -0,0 +1,95 @@
...
+    -------
+    triggered event times : List[float]
+    """
+    return [event_time + time_bins[trg] for trg in trigger_indx]
+
+
+def first_and_last_times(pmt_wfs    : pd.Series,
+                         sipm_wfs   : pd.Series,
+                         pmt_binwid : float    ,
+                         sipm_binwid: float    ) -> Tuple:
```
gonzaponte:
```suggestion
def first_and_last_times(pmt_wfs    : pd.DataFrame,
                         sipm_wfs   : pd.DataFrame,
                         pmt_binwid : float       ,
                         sipm_binwid: float       ) -> Tuple[float, float]:
```

invisible_cities/detector_simulation/sensor_utils.py
```diff
@@ -0,0 +1,95 @@
...
+    Casts the event sensor info into the correct order
+    adding zeros for sensors which didn't see any signal.
+    """
+    pmt_ids    = DataPMT (detector_db, run_number).SensorID
+    sipm_ids   = DataSiPM(detector_db, run_number).SensorID
+    pmt_shape  = (n_pmt , length_pmt )
+    sipm_shape = (n_sipm, length_sipm)
+    def ordering(sensor_order : pd.Int64Index,
+                 sensor_resp  : np.ndarray   ,
+                 sensor_shape : Tuple        ) -> np.ndarray:
```
gonzaponte:
```suggestion
                 sensor_shape : Tuple[float, float]) -> np.ndarray:
```

invisible_cities/detector_simulation/sensor_utils.py
```diff
@@ -0,0 +1,95 @@
...
+        pmt_ord  = pmt_ids [ pmt_ids.isin( pmt_resp.index)].index
+        sipm_ord = sipm_ids[sipm_ids.isin(sipm_resp.index)].index
+
+        return [(ordering(pmt_ord , pmts , pmt_shape ),
+                 ordering(sipm_ord, sipms, sipm_shape))
+                    for pmts, sipms in evt_buffers]
+    return order_and_pad
+
+
+def get_n_sensors(detector_db: str, run_number: int) -> Tuple:
```
gonzaponte:
```suggestion
def get_n_sensors(detector_db: str, run_number: int) -> Tuple[int, int]:
```

invisible_cities/detector_simulation/sensor_utils.py
```diff
@@ -0,0 +1,95 @@
...
+    """Get the number of sensors for this run"""
+    npmt  = DataPMT (detector_db, run_number).shape[0]
+    nsipm = DataSiPM(detector_db, run_number).shape[0]
+    return npmt, nsipm
+
+
+def pmt_and_sipm_bin_width(file_name: str) -> Tuple:
```
gonzaponte:
```suggestion
def pmt_and_sipm_bin_width(file_name: str) -> Tuple[float, float]:
```

invisible_cities/detector_simulation/sensor_utils_test.py
```diff
@@ -0,0 +1,94 @@
...
+    test_bins1 = np.arange(bin_min, bin_max)
+    test_bins2 = np.arange(bin_min, bin_max - 1)
+
+    bmin, bmax = first_and_last_times(pd.Series({'time': test_bins1}),
+                                      pd.Series({'time': test_bins2}),
+                                      1, 1)
+
+    assert np.round(bmin) == bin_min
+    assert np.round(bmax) == bin_max
+
```
gonzaponte:
```suggestion


```

invisible_cities/detector_simulation/sensor_utils_test.py
```diff
@@ -0,0 +1,94 @@
...
+@fixture(scope = 'function')
+def id_dict():
+    id_dict = {'pmt_ids' :    (2, 5, 7), 'pmt_ord' :  (2, 5, 7),
+               'sipm_ids': (1010, 5023), 'sipm_ord':  (10, 279)}
+    return id_dict
+def test_order_sensors(id_dict, pmt_ids, sipm_ids):
```
gonzaponte: two-line spacing

invisible_cities/detector_simulation/sensor_utils_test.py
```diff
@@ -0,0 +1,94 @@
...
+    detector_db = 'new'
+    run_number  = 6400
+    n_pmt       = len( pmt_ids)
+    n_sipm      = len(sipm_ids)
+    nsamp_pmt   = 5000
+    nsamp_sipm  =    5
+
+    sensor_order = order_sensors(detector_db, run_number,
```
gonzaponte: swap names. `order_sensors = sensor_order(er?)(...)`

invisible_cities/detector_simulation/sensor_utils_test.py
```diff
@@ -0,0 +1,94 @@
...
+                                 n_pmt      ,  nsamp_pmt,
+                                 n_sipm     , nsamp_sipm)
+
+    pmt_resp   = pd.Series([[1]*nsamp_pmt ]*len(id_dict[ 'pmt_ids']),
+                          index = id_dict[ 'pmt_ids'])
+    sipm_resp  = pd.Series([[1]*nsamp_sipm]*len(id_dict['sipm_ids']),
+                          index = id_dict['sipm_ids'])
+    pmt_q      = np.array( pmt_resp.tolist())
+    sipm_q     = np.array(sipm_resp.tolist())
```
gonzaponte: Can this be defined in the same fixture `id_dict` (changing the name accordingly)?

invisible_cities/detector_simulation/sensor_utils_test.py
```diff
@@ -0,0 +1,94 @@
...
+    padded_evt = sensor_order(pmt_resp, sipm_resp, [(pmt_q, sipm_q)])
+
+    pmt_out  = padded_evt[0][0]
+    sipm_out = padded_evt[0][1]
+    assert pmt_out.shape == (n_pmt, nsamp_pmt)
+    pmt_nonzero = np.argwhere(pmt_out.sum(axis=1) != 0)
+    assert np.all(pmt_nonzero.flatten() == id_dict['pmt_ord'])
+    assert sipm_out.shape == (n_sipm, nsamp_sipm)
+    sipm_nonzero = np.argwhere(sipm_out.sum(axis=1) != 0)
+    assert np.all(sipm_nonzero.flatten() == id_dict['sipm_ord'])
```
gonzaponte:
```suggestion
    assert  pmt_out.shape == (n_pmt , nsamp_pmt)
    assert sipm_out.shape == (n_sipm, nsamp_sipm)

    pmt_nonzero  = np.argwhere( pmt_out.sum(axis=1) != 0)
    sipm_nonzero = np.argwhere(sipm_out.sum(axis=1) != 0)

    assert np.all( pmt_nonzero.flatten() == id_dict[ 'pmt_ord'])
    assert np.all(sipm_nonzero.flatten() == id_dict['sipm_ord'])
```

A bit os cosmetics for readability.

invisible_cities/detector_simulation/sensor_utils_test.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def test_get_n_sensors(pmt_ids, sipm_ids):
+    npmt, nsipm = get_n_sensors('new', 6400)
+
+    assert npmt  == len( pmt_ids)
+    assert nsipm == len(sipm_ids)
```
gonzaponte: parametrize to test positive and negative run numbers

invisible_cities/detector_simulation/sensor_utils_test.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def test_pmt_and_sipm_bin_width(ICDATADIR):
+    file_in = os.path.join(ICDATADIR                                         ,
+                           'Kr83_full_nexus_v5_03_01_ACTIVE_7bar_1evt.sim.h5')
```
gonzaponte: This filename is defined in the fixture `mc_waveforms`. It should probably be defined in an independent fixture to be reused here. Actually I've just seen that the same file is also used in other modules, so it should probably go in the global conftest.py file and be fixed later in the other modules.

gonzaponte (changes requested):
Overall I get the feeling that we should have named outputs in several places. Specifically, I am very tempted to just declare type synonyms like
```
ChargeArray = np.ndarray # 1D
WFset       = np.ndarray # 2D
Waveforms   = pd.Series  # 2D
```
(names not optimal). I think this would probably increase the readability and understanding of what data types are going through each function because it's not always obvious.

This PR looks very good. Although there are lots of comments, most of them are minor.

invisible_cities/detector_simulation/buffer_functions.py
andLaing: I agree that the annotation is probably a good idea but I quite like the container as it gives an object per sensor.

andLaing: pytest (python) doesn't seem to like this:
`ImportError while loading conftest '/Users/laingandrew/Documents/NEXT/IC_stuff/IC/invisible_cities/detector_simulation/conftest.py'.
invisible_cities/detector_simulation/conftest.py:13: in <module>
    from .  buffer_functions   import               bin_sensors
invisible_cities/detector_simulation/buffer_functions.py:46: in <module>
    def find_signal(wfs          : pd.Series[np.ndarray],
E   TypeError: 'type' object is not subscriptable`

gonzaponte: Mmm ok. I think this works in favor of declaring type synonymns.

invisible_cities/detector_simulation/buffer_functions_test.py
andLaing: The function is called `calculate_buffers`, you want me to change it?

gonzaponte: Yes, the convention is for functions that return callables to be named with nouns and the output to be named as verbs.

invisible_cities/detector_simulation/sensor_utils_test.py
andLaing: I'm not sure what we gain by the parametrizing the test. Positive and negative run numbers are essentially identical as load_db uses the absolute value. The +/- difference is only used to invoke the copy of mc tables in the cities

gonzaponte: Right. The idea was to make sure it worked in both cases. Since this code is going to be used in MC, you should use a negative run number, right?

andLaing: It seems irrelevant but I can make it negative. As I mentioned above the load_db uses absolute value so the difference should be tested there.

invisible_cities/detector_simulation/conftest.py
andLaing: I'll see what I can do but it's quite awkward and verbose to get something that's actually challenging the later functions.

invisible_cities/detector_simulation/buffer_functions_test.py
```diff
@@ -0,0 +1,109 @@
...
+
+    assert  pmt_bins[ 0] >= bins_min
+    assert  pmt_bins[-1] <= min(bins_max, max_buffer)
+    assert np.all(np.diff( pmt_bins) ==  pmt_binwid)
+    assert sipm_bins[ 0] >= bins_min
+    assert sipm_bins[-1] <= min(bins_max, max_buffer)
+    assert np.all(np.diff(sipm_bins) == sipm_binwid)
+    ## In current DAQ, pmts have higher sample frequency than SiPMs
+    assert pmt_bins[ 0] >= sipm_bins[ 0]
+    assert pmt_bins[-1] >= sipm_bins[-1]
```
gonzaponte:
```suggestion
    assert  pmt_bins[ 0] >= bins_min
    assert  pmt_bins[-1] <= min(bins_max, max_buffer)
    assert np.all(np.diff( pmt_bins) ==  pmt_binwid)ç
    
    assert sipm_bins[ 0] >= bins_min
    assert sipm_bins[-1] <= min(bins_max, max_buffer)
    assert np.all(np.diff(sipm_bins) == sipm_binwid)
    
    ## In current DAQ, pmts have higher sample frequency than SiPMs
    assert pmt_bins[ 0] >= sipm_bins[ 0]
    assert pmt_bins[-1] >= sipm_bins[-1]
```
just some spaces for readability

invisible_cities/detector_simulation/sensor_utils_test.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def test_first_and_last_times():
+    bin_min    = 22
+    bin_max    = 40
+    test_bins1 = np.arange(bin_min, bin_max)
+    test_bins2 = np.arange(bin_min, bin_max - 1)
+
+    bmin, bmax = first_and_last_times(pd.Series({'time': test_bins1}),
+                                      pd.Series({'time': test_bins2}),
```
gonzaponte: The input should be DataFrames. Are you using Series because is the only column used?

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,143 @@
...
+                  Width in mus of PMT sample integration
+    sipm_binwid : float
+                  Width in mus of SiPM sample integration
+    """
+    pmt_buffer_samples  = int(buffer_len /  pmt_binwid)
+    sipm_buffer_samples = int(buffer_len / sipm_binwid)
+    sipm_pretrg         = int(pre_trigger / sipm_binwid)
+    sipm_postrg         = sipm_buffer_samples - sipm_pretrg
+    pmt_pretrg_base     = int(pre_trigger / pmt_binwid)
+    pmt_postrg_base     = pmt_buffer_samples - pmt_pretrg_base
```
gonzaponte:
```suggestion
    pmt_buffer_samples  = int(buffer_len  //  pmt_binwid)
    sipm_buffer_samples = int(buffer_len  // sipm_binwid)
    sipm_pretrg         = int(pre_trigger // sipm_binwid)
    sipm_postrg         = sipm_buffer_samples - sipm_pretrg
    pmt_pretrg_base     = int(pre_trigger  // pmt_binwid)
    pmt_postrg_base     = pmt_buffer_samples - pmt_pretrg_base
```
technically integer division is more appropriate, although there is no effective change due to int conversion

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,143 @@
...
+                        pmt_charge :  pd.Series,
+                        sipm_bins  : np.ndarray,
+                        sipm_charge:  pd.Series
+                        ) -> List[Tuple[np.ndarray, np.ndarray]]:
+        """
+        Synchronises the SiPMs and PMTs for each identified
+        trigger and calls the padding function to fill with
+        zeros where necessary.
+        """
+        pmt_q  = np.array(pmt_charge.tolist())
```
gonzaponte: Minor detail: use `np.asarray`. The elements of `pmt_charge.tolist()` are arrays, this way we avoid copying them.

invisible_cities/detector_simulation/buffer_functions.py
```diff
@@ -0,0 +1,143 @@
...
+        sipm_q = np.empty((0,0))\
+          if sipm_charge.empty else np.array(sipm_charge.tolist())
```
gonzaponte: I think this would do what you want, if you want to give it a try:
```suggestion
        sipm_q = np.asarray(sipm_charge.tolist(), ndmin=2)
```

invisible_cities/detector_simulation/buffer_functions_test.py
```diff
@@ -0,0 +1,109 @@
...
+
+@mark.parametrize("pre_trigger signal_thresh".split(),
+                  ((100 * units.mus,  2),
+                   (400 * units.mus, 10)))
+def test_buffer_calculator(mc_waveforms, binned_waveforms,
+                           pre_trigger ,    signal_thresh):
+
+    _, pmt_binwid, sipm_binwid, _ = mc_waveforms
+
+    pmt_bins, pmt_wf, sipm_bins, sipm_wf = binned_waveforms
```
gonzaponte:
```suggestion
    pmt_bins, pmt_wfs, sipm_bins, sipm_wfs = binned_waveforms
```

invisible_cities/detector_simulation/buffer_functions_test.py
```diff
@@ -0,0 +1,109 @@
...
+    pmt_sum  = pmts .charge.sum()
+    sipm_sum = sipms.charge.sum()
+    assert pmt_wf .sum().sum() ==  pmt_sum
+    assert sipm_wf.sum().sum() == sipm_sum
+
+
+@mark.parametrize("signal_thresh", (2, 10))
+def test_find_signal_start(binned_waveforms, signal_thresh):
+
+    pmt_bins, pmt_wf, *_ = binned_waveforms
```
gonzaponte:
```suggestion
    pmt_bins, pmt_wfs, *_ = binned_waveforms
```

invisible_cities/detector_simulation/sensor_utils.py
```diff
@@ -0,0 +1,95 @@
...
+                   Time of nexus event
+    time_bins    : np.ndarray
+                   The binning used to histogram the nexus sensor
+                   data.
+
+    Returns
+    -------
+    triggered event times : List[float]
+    """
+    return [event_time + time_bins[trg] for trg in trigger_indx]
```
gonzaponte: I just noticed that this can be written as `event_time + time_bins[trigger_indx]`so you may want to get rid of this function, as it is pretty straightforward. Up to you.

invisible_cities/detector_simulation/sensor_utils.py
```diff
@@ -0,0 +1,95 @@
...
+    def ordering(sensor_order : pd.Int64Index  ,
+                 sensor_resp  : np.ndarray     ,
+                 sensor_shape : Tuple[int, int]) -> np.ndarray:
+        sensors = np.zeros(sensor_shape, np.int)
+        sensors[sensor_order] = sensor_resp
+        return sensors
+        
+    def order_and_pad(pmt_resp    : pd.Series  ,
+                      sipm_resp   : pd.Series  ,
+                      evt_buffers : List[Tuple]) -> List[Tuple]:
```
gonzaponte:
```suggestion
                      evt_buffers : List[Tuple[np.ndarray, np.ndarray]]) -> List[Tuple]:
```
This is getting a bit out of hand.

invisible_cities/detector_simulation/sensor_utils.py
```diff
@@ -0,0 +1,95 @@
...
+                 n_pmt      : int, length_pmt : int,
+                 n_sipm     : int, length_sipm: int) -> Callable:
+    """
+    Casts the event sensor info into the correct order
+    adding zeros for sensors which didn't see any signal.
+    """
+    pmt_ids    = DataPMT (detector_db, run_number).SensorID
+    sipm_ids   = DataSiPM(detector_db, run_number).SensorID
+    pmt_shape  = (n_pmt , length_pmt )
+    sipm_shape = (n_sipm, length_sipm)
```
gonzaponte: Maybe you prefer to use `get_n_sensors` here instead of taking `n_pmt` and `n_sipm` as parameters?

invisible_cities/detector_simulation/sensor_utils_test.py
```diff
@@ -0,0 +1,96 @@
...
+                                      1, 1)
+
+    assert np.round(bmin) == bin_min
+    assert np.round(bmax) == bin_max
```
gonzaponte: Are these `np.round` calls to account for float precision? If so use `np.isclose`.

You may want to add another test with differently-grained test bins as well.

invisible_cities/detector_simulation/sensor_utils_test.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+@fixture(scope = 'function')
+def id_dict():
+    id_dict = {'pmt_ids' :    (2, 5, 7), 'pmt_ord' :  (2, 5, 7),
+               'sipm_ids': (1010, 5023), 'sipm_ord':  (10, 279)}
+    return id_dict
+def test_sensor_order(id_dict, pmt_ids, sipm_ids):
```
gonzaponte: two blank lines between the fixture and the test

invisible_cities/detector_simulation/sensor_utils_test.py
```diff
@@ -0,0 +1,96 @@
...
+                                      1, 1)
+
+    assert np.round(bmin) == bin_min
+    assert np.round(bmax) == bin_max
+
+
+@fixture(scope = 'function')
+def id_dict():
+    id_dict = {'pmt_ids' :    (2, 5, 7), 'pmt_ord' :  (2, 5, 7),
+               'sipm_ids': (1010, 5023), 'sipm_ord':  (10, 279)}
```
gonzaponte: From the name of the function, I guess you also want to account for unordered numbers. Maybe you want to try to swap a couple of numbers to test that case as well.

gonzaponte (commented):
Just some very minor changes left. Also, please run `pyflakes` on the modified files to look for unused variables or imports.

gonzaponte:
Neat.

gonzaponte (approved):
A complete set of functions and components to split events into buffers and create waveforms in each of them. It reuses existing code as much as needed and tests every function implemented. All functions have some documentation. Very nice job.

invisible_cities/detsim/buffer_functions.py
```diff
@@ -0,0 +1,141 @@
...
+    min_bin     = np.floor(t_min    / bin_width) * bin_width
+    max_bin     = np.ceil (max_time / bin_width) * bin_width
+
+    bins        = np.arange(min_bin, max_bin + bin_width, bin_width)
+    bin_sensors = sensors.groupby('sensor_id').apply(weighted_histogram, bins)
+    return bins[:-1], bin_sensors
+
+
+## !! to-do: clarify for non-pmt versions of next
+## !! to-do: Check on integral instead of only threshold?
```
jacg, line 43: Many development tools recognize `TODO` (exactly four characters, all uppercase, no spaces, no dashes)  and offer assistance with finding, managing, and warning about stuff that has been marked to be done,  but this relies on it being written exactly like this: `TODO`.

So let us all, please, develop the habit of marking things to be done exactly thus: `TODO`.

PR #738, opened 2020-09-01
Update url of miniconda repository

jmalbos:
This PR fixes the installation process in macOS and, most likely, in any other system using `curl` to download the miniconda installer.

The url used until now (continuum.io) had been redirected to a new one (anaconda.com). While `wget` appears to accept the redirection by default, `curl` does not. This can be fixed with the `-L` flag, but I decided that it was probably simpler to avoid the redirect altogether updating the url.

### DIFF
--- a/manage.sh
+++ b/manage.sh
@@ -60,10 +60,11 @@ function install_conda {
         echo Conda already installed. Skipping conda installation.
     else
         echo Installing conda for $CONDA_OS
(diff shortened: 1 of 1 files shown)


jmalbos:
Incidentally: we're still installing conda 4.8.2, but there's a newer version (4.8.3).

jerenner (approved):
Successfully downloads and installs miniconda on macOS Catalina (10.15.5)

PR #739, opened 2020-09-03
Update Flex100DB to fix SiPM numbering

jmunozv:
This MR updates the IC tables related with Flex100 geometry to fix a little bug in TP SiPMs affecting their numbering.

The corresponding test has been updated too.


jmbenlloch (approved):
This PR keeps the Flex100 database synchronized with the nexus geometry. Approved!

PR #742, opened 2020-09-23
Bufferization city

andLaing:
Adds a new dataflow city which takes in sensor information from the nexus output and sorts them into buffers for each sensor type.

According to configurable pretrigger and trigger threshold the sensor signal can be positioned as an S1-like triggered event or an S2-like triggered event.

### DIFF
--- a/invisible_cities/cities/buffy.py
+++ b/invisible_cities/cities/buffy.py
@@ -1,0 +1,126 @@
+"""
+-----------------------------------------------------------------------
+                                Buffy
+-----------------------------------------------------------------------
+
+ -- Sorts MC sensor info into buffers, slays vampires etc
+
+This city reads nexus Monte Carlo Sensor information and sorts the
+information into data-like buffers without the addition of
+electronics noise.
+
+Uses a configured buffer length, pretrigger and threshold for the
+positioning of trigger like signal. If more than one trigger is found
+separated from each other by more than a buffer width the nexus event
+can be split into multiple data-like triggers.
+"""
+
+import numpy  as np
+import tables as tb
+
+from .. detsim.buffer_functions import      buffer_calculator
+from .. detsim.sensor_utils     import   first_and_last_times
+from .. detsim.sensor_utils     import          get_n_sensors
+from .. detsim.sensor_utils     import           sensor_order
+from .. detsim.sensor_utils     import pmt_and_sipm_bin_width
+from .. detsim.sensor_utils     import          trigger_times
+from .. io    .rwf_io           import          buffer_writer
+from .. reco                    import          tbl_functions as tbl
+
+from .. dataflow   import            dataflow as fl
+
+from .  components import                city
+from .  components import             collect
+from .  components import        copy_mc_info
+from .  components import mcsensors_from_file
+from .  components import         print_every
+from .  components import       signal_finder
+from .  components import           wf_binner
+
+
+@city
+def buffy(files_in     , file_out   , compression      , event_range,
+          print_mod    , detector_db, run_number       , max_time   ,
+          buffer_length, pre_trigger, trigger_threshold):
+
+    npmt, nsipm       = get_n_sensors(detector_db, run_number)
+    pmt_wid, sipm_wid = pmt_and_sipm_bin_width(files_in[0])
+    nsamp_pmt         = int(buffer_length /  pmt_wid)
+    nsamp_sipm        = int(buffer_length / sipm_wid)
+
+    extract_tminmax   = fl.map(first_and_last_times                ,
+                               args = ("pmt_resp"   ,  "sipm_resp",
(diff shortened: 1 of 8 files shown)


paolafer (changes requested):
I think that a test of an event that is split in two buffers would be useful, too. The same Kr test file can be used, with a shorter buffer, if you don't want to look for a specific event.

paolafer (approved):
The code works as expected, tests have been added to cover the two possible cases of simple and split events. I approve this PR.

invisible_cities/cities/buffy.py
```diff
@@ -0,0 +1,126 @@
...
+from .  components import           wf_binner
+
+
+@city
+def buffy(files_in     , file_out   , compression      , event_range,
+          print_mod    , detector_db, run_number       , max_time   ,
+          buffer_length, pre_trigger, trigger_threshold):
+
+    npmt, nsipm       = get_n_sensors(detector_db, run_number)
+    pmt_wid, sipm_wid = pmt_and_sipm_bin_width(files_in[0])
```
mmkekic: If I understand correctly, reading of bin widths from nexus configuration table is repeated here and in the buffy source (though in the source you dont use `pmt_and_sipm_bin_width` explicitly)? So the `pmt_wid` and coroutine `pmt_binwid` arguments are the same? I assume it does not change per event basis so ideally we should always use `pmt_wid`?
On the other hand this line is not safe in case `files_in` is an empty list (it will fail with `IndexError` ) or in case `filesin[0]` is an empty file. The first one is easy to fix but not sure what is the best protection against the second one? Though unless the file is corrupt (and we dont have protections in IC for that) even empty nexus files will contain the config table?

andLaing:
I rebased and implemented the changes suggested by @mmkekic. New commits from "Change pmt_and_sipm_bin_width to protect against empty files_in"

invisible_cities/cities/components.py
```diff
@@ -357,40 +357,33 @@ def mcsensors_from_file(paths     : List[str],
                  Run number for database
     """
 
-    pmt_ids  = load_db.DataPMT (db_file, run_number).SensorID
+    pmt_ids  = load_db.DataPMT(db_file, run_number).SensorID
 
     for file_name in paths:
         sns_resp = mcinfo_io.load_mcsensor_response_df(file_name              ,
```
mmkekic, line 363: I think we should skip empty or corrupted files with a `try except`, producing a warning that the file couldn't be processed?

andLaing: The errors that you suggest are already caught in `dst_io.load_dst` which is used in `load_mcsensor_response_df`. I'd need to make the warnings catchable.

jmalbos: > I think we should skip empty or corrupted files with a try except, producing a warning that the file couldn't be processed?

Skipping the files with a warning may not be a good idea. As far as we can tell from these last 6 months of productions, corrupted files do not appear to be a common issue (in fact, we've seen 0 corrupted files), while completely empty files are usually a symptom of errors in the production (errors in the job configuration of nexus, for example) and a good reason to stop the processing. Empty events may be legit, and should be filtered out (i.e. skipped but counted).

paolafer: I think that making a job crash because one file is empty or corrupt is annoying. 
I agree that we should not simply forget about those files, but that could be dealt with at the nexus stage, looking at nexus outputs and making sure that we understand why some files are not ok. What I mean is that trying to process them doesn't seem necessarily the right place for this sanity check to me. 
However, both approaches can be done, I have no strong opinion on any of them.

mmkekic: So, to conclude:
- No protection against corrupted files, can be added afterwards if really needed 
- A filter that will filter events that left no signal in the detector (the filter in the flow, not just skipping the event in the source)

andLaing: If there's no signal left in the sensors for an event the event doesn't appear in the table in the input file and, as such, can't be skipped or processed because it's not there.

I also can't remove the protection for corrupt files without catchable warnings since the underlying reader `dst_io.load_dst` already converts the exception to a warning. If this is to be changed it should be a separate PR with all the checks required, I think.

What should I do? Revert to having the `MCEventNotFound` error for an empty?

This is getting a bit confused since most of the limitations related to this error vs. warning discussion come from functions that aren't directly related to the PR.

andLaing:
Think that implements the suggestions for filtering. I added a test to make sure we get an empty event on a file which we know doesn't have any sensor responses.

mmkekic (approved):
This PR adds a city that reads nexus file and outputs detector-like RWFs. After @paolafer review a filter to filter-out events with empty sensor responses and appropriate tests are added.  Good job!

PR #744, opened 2020-09-29
Upload DB with the lastest calibrations in NEW

carmenromo:
This PR includes the latest sensor calibrations from run 8151. Database has been uploaded with the PMT and SiPMs gains and the pdf values. Apparently sensor 15060 is dead, so it has been masked.

pmt_comparison_with_previous_runs

sipmRunDifferencePlots_test


jmbenlloch (approved):
Latest calibration added. Approved!

PR #745, opened 2020-10-07
Update NEXT100 local db after bug fix changing SiPM positions

jmalbos:


jmbenlloch (approved):
Fixes the NEXT100 database with the correct SiPM positions.

PR #746, opened 2020-10-22
Include dice board ratios in DEMOPP database

carmenromo:
This PR adds  the gain values of the sensors modified with the ratios between every dice board and the dice board 1 to the DEMOPP database. The min run is 1008744 and the max run 1999999.


neuslopez (approved):
Correction using qmax added. All tests passed successfully.

PR #747, opened 2020-10-26
Update demo DB

carmenromo:
In the last DEMO++ db related PR, correction using qmax was added only for the SiPMs. This PR, updates the PMTs max run to avoid errors when executing the cities for high run numbers.


neuslopez (approved):
All tests passed successfully.

PR #751, opened 2020-10-28
Unique mc evtnumber

andLaing:
Adds functions to generate unique event numbers for MC to allow for safe processing of split nexus events and simplify MC event mixing.

In need of more tests and a file number reader but ready for discussion.

Discussion continues from issue #715

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -183,32 +183,37 @@ def copy_mc_info(files_in     : List[str],
 
     Parameters
     ----------
-    files_in : List of strings
-        Name of the input files.
-    file_out : tables.File
-        The output h5 file.
-    event_numbers : List[int]
-        List of event numbers for which the MC info is copied
-        to the output file.
+    files_in     : List of strings
+                   Name of the input files.
+    h5out        : tables.File
+                   The output h5 file.
+    event_numbers: List[int]
+                   List of event numbers for which the MC info is copied
+                   to the output file.
+    db_file      : str
+                   Name of database to be used where necessary to
+                   read the MC info (used for pre-2020 format files)
+    run_number   : int
+                   Run number for database (used for pre-2020 format files)
     """
 
     writer = mcinfo_io.mc_writer(h5out)
 
-    copied_events = []
+    copied_events   = []
     for f in files_in:
         if mcinfo_io.check_mc_present(f):
-            event_numbers_in_file = mcinfo_io.get_event_numbers_in_file(f)
-            event_numbers_to_copy = np.intersect1d(event_numbers        ,
-                                                   event_numbers_in_file)
-            mcinfo_io.copy_mc_info(f, writer, event_numbers_to_copy,
+            evt_copied  = mcinfo_io.safe_copy_nexus_eventmap(h5out        ,
+                                                             event_numbers,
+                                                             f            )
+            mcinfo_io.copy_mc_info(f, writer, evt_copied['nexus_evt'],
                                    db_file, run_number)
-            copied_events.extend(event_numbers_to_copy)
+            copied_events.extend(evt_copied['evt_number'])
         else:
-            warnings.warn(f' File does not contain MC tables.\
+            warnings.warn('File does not contain MC tables.\
              Use positve run numbers for data', UserWarning)
             continue
     if len(np.setdiff1d(event_numbers, copied_events)) != 0:
-        raise MCEventNotFound(f' Some events not found in MC tables')
+        raise MCEventNotFound('Some events not found in MC tables')
 
 
 def wf_binner(max_buffer: int) -> Callable:
@@ -368,7 +373,7 @@ def mcsensors_from_file(paths     : List[str],
 
         ## MC uses dummy timestamp for now
         ## Only in case of evt splitting will be non zero
-        timestamp = 0
+        timestamp     = 0
 
         for evt in mcinfo_io.get_event_numbers_in_file(file_name):
             try:
@@ -831,6 +836,7 @@ def compute_and_write_pmaps(detector_db, run_number, pmt_samp_wid, sipm_samp_wid
 
 
 def calculate_and_save_buffers(buffer_length    : float        ,
+                               max_time         : int          ,
                                pre_trigger      : float        ,
                                pmt_wid          : float        ,
                                sipm_wid         : float        ,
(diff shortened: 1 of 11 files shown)


mmkekic:
This seems closely related to the Issue #145  and corresponding PRs:
 #168 where a solution was to introduce first_event parameter in the configuration file and leave it up to a user to make sure the parameter is sensible;
#245 (3185e071d2fb4715fc18f74b52d7e9f6db8aed9c) where we used hash on the input filename  and
#255 where we used pattern matching to extract file index from one full production

The definite solution was to fix the issue in the NEXUS itself making sure that one production has a unique event count, however since now (probably before too?) we need to be able to split a long NEXUS event into shorter buffer-size events we are back to the problem of having to re-index events after the NEXUS production and ensure their uniqueness. Maybe  @jacg and @jmbenlloch that were dealing with this problem in the past can give us some guidance?

mmkekic:
Since this is somewhat urgent and we lack a very appealing option, I am inclined to go with the simplest one for now, ie add one parameter in config file (either `start_event_id` or `file_id`) that will be used in the #752  too. If we opt for `file_id` we have to think of a unique pattern for `event_number`.  @andLaing  suggested something like  `int(f'{file_id}0{nexus_event_number}')`, thought we might need more 0s in between to be certain this is unique. Suggestions @paolafer , @jmalbos , @msorel , @jjgomezcadenas .. ?

andLaing:
In the current version the event numbers generated become `int(f'{nexus_seed_from_config}00{nexus_start_id}' + str(counter))`. Still some tests missing, higher level checks to make sure this format is enough needed too.

invisible_cities/io/mcinfo_io.py
```diff
@@ -296,6 +322,23 @@ def try_unique_evt_itr(group, itr):
     return np.unique(np.concatenate(evt_list)).astype(int)
 
 
+def load_eventnumbermap(file_name: str) -> pd.DataFrame:
```
mmkekic, line 337: Since you have the reader, maybe you should add a writer too instead of calling `df_writter` directly?

invisible_cities/io/mcinfo_io.py
```diff
@@ -108,6 +108,32 @@ def check_last_merge_index(h5out : tb.file.File) -> int:
     return -1
 
 
+def copy_nexus_eventmap(h5out   : tb.file.File,
```
mmkekic: Mabe add 'safe_copy ...` since in case the table is not present you don't copy and the name is a bit misleading

invisible_cities/io/mcinfo_io.py
```diff
@@ -108,6 +108,32 @@ def check_last_merge_index(h5out : tb.file.File) -> int:
...
+                        files_in: List[str]   ) -> np.ndarray:
+    """
+    Copies the nexus event map and returns the array
+    of nexus event_ids to be used to copy the MC tables.
+    """
+    nexus_evt = []
+    for fn in files_in:
+        evt_map = load_eventnumbermap(fn)
+        try:
+            evt_mask = evt_map.evt_number.isin(evt_arr)
```
mmkekic: This is a bit hard to understand. Can you add a comment that `evt_map` is None if the table is not present, and you are also protected against table present but with wrong columns?

invisible_cities/io/mcinfo_io.py
```diff
@@ -108,6 +108,32 @@ def check_last_merge_index(h5out : tb.file.File) -> int:
...
+            nexus_evt.extend(evt_map.nexus_evt[evt_mask].values)
+            df_writer(h5out, evt_map[evt_mask], 'Run', 'eventMap')
+        except AttributeError:
+            ## No copy and return the parts of the
+            ## original arrays found.
+            if check_mc_present(fn):
```
mmkekic: Why do you need this check? This function should work with MC files, so the table has to be present. The `copy_mc_info` should take care it raises appropriate error/warning in case the table is not present. In case there is no MC table you probably do not even want to do any event mapping

invisible_cities/cities/components.py
```diff
@@ -173,11 +173,12 @@ def append(l,e):
...
-                 h5out        : tb.File  ,
-                 event_numbers: List[int],
-                 db_file      :      str ,
-                 run_number   :      int ) -> None:
+def copy_mc_info(files_in     : List[str]  ,
+                 h5out        : tb.File    ,
+                 event_numbers: List[int]  ,
+                 db_file      :      str   ,
+                 run_number   :      int   ,
+                 copy_eventmap: bool = True) -> None:
```
mmkekic: Add new parameter to docstrings

invisible_cities/io/mcinfo_io.py
```diff
@@ -108,6 +108,32 @@ def check_last_merge_index(h5out : tb.file.File) -> int:
     return -1
 
 
+def copy_nexus_eventmap(h5out   : tb.file.File,
+                        evt_arr : np.ndarray  ,
+                        files_in: List[str]   ) -> np.ndarray:
+    """
+    Copies the nexus event map and returns the array
```
mmkekic, line 115: Add inputs and return to docstrings

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -104,6 +106,24 @@ def test_is_oldformat_file(ICDATADIR):
     assert not is_oldformat_file(file_in_new)
 
 
+def test_copy_nexus_eventmap(config_tmpdir, ICDATADIR):
+    file_in  = os.path.join(ICDATADIR                                  ,
+                            'nexus_new_kr83m_full.newformat.buffers.h5')
+    file_out = os.path.join(ICDATADIR, 'evtMapCopy.h5')
```
mmkekic: This should be a temporary path, not `ICDATADIR`

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -104,6 +106,24 @@ def test_is_oldformat_file(ICDATADIR):
...
+
+        assert len(nexus_evts) == len(expected_evts)
+        assert np.all(nexus_evts == expected_evts[:, 1])
+
+    with tb.open_file(file_in) as h5orig, tb.open_file(file_out) as h5out:
+        assert hasattr(h5out.root.Run, 'eventMap')
+        assert_tables_equality(h5out .root.Run.eventMap,
+                               h5orig.root.Run.eventMap)
+
+
```
mmkekic, line 128: Can you add a test tho show that the map is not copied and return event are correct in case the input file does not contain the table?

mmkekic (approved):
This PR adds a event splitting functionality keeping unique nexus event numbers across files by assuming a constant maximum number of splits per event. A new table that maps IC event number to original MC event number is added to Run group ensuring the code is compatible with the old formats of MC production. The code is documented and tested, good job!

PR #754, opened 2020-11-09
Update demoppdb

carmenromo:
The ratios in gains for demopp database have been fixed from min run = 1008744 according to these numbers:
- DB2/DB1 = 0.897
- DB3/DB1 = 0.949
- DB4/DB1 = 1.005


neuslopez (approved):
All tests passed and qmax corrections are correctly applied to each DB. PR approved!

Issue #749, opened 2020-10-27
Rethinking post-penthesilea production


mmkekic:
Since Beersheba hits are being used for track reconstructions, we need official code that runs paolina algorithm on Beersheba output. The code is already in Esmeralda and this is a matter of organizing different  _pipes_. 
As a reminder current algorithm  is:
 - Esmeralda : input uncorrected Penthesilea hits; output corrected naive hits (2 charge thresholds); tracks-related variables 
 - Beersheba : input corrected Esmeralda hits, output deconvoluted hits

The options that appeared at the meeting:

1. Add tracking part to Beersheba; leave Esmeralda as is. This gives us both naive reconstruction and tracks as well as deconvoluted hits and tracks
   
2.  Add energy correction to Beersheba, remove it from Esmeralda and run Esmeralda after Beersheba. Beersheba may or may not output naive hits reconstruction. This is more modular since the tracking is separated from hits reconstruction.

3. Have three independent cities : hits reconstruction (Beersheba), energy correction (??), tracking (Esmeralda)

@pnovella @msorel @paolafer @ausonandres @Aretno @jahernando @jjgomezcadenas @gonzaponte

mmkekic:
[![](https://api.gh-polls.com/poll/01ENNEQJHTYF3CM6WFPTBH4N6Y/1)](https://api.gh-polls.com/poll/01ENNEQJHTYF3CM6WFPTBH4N6Y/1/vote)
[![](https://api.gh-polls.com/poll/01ENNEQJHTYF3CM6WFPTBH4N6Y/2)](https://api.gh-polls.com/poll/01ENNEQJHTYF3CM6WFPTBH4N6Y/2/vote)
[![](https://api.gh-polls.com/poll/01ENNEQJHTYF3CM6WFPTBH4N6Y/3)](https://api.gh-polls.com/poll/01ENNEQJHTYF3CM6WFPTBH4N6Y/3/vote)

mmkekic:
I removed the poll since it was confusing, so please comment

msorel:
Hi, as I said in the meeting, 1 probably sufficient  in short term, as we are still applying  xyz energy corrections using pre-deconvolution hits. Up to upcoming bb2nu first paper I guess we will do this. 

[...]

PR #755, opened 2020-11-12
New tracking city

ausonandres:
According to issue #749, the aim of this PR is to add a city (to be run just after Beersheba, for the moment), that computes all the tracking information. Since the tracking-related functions will be shared between Esmeralda and this new city (so-called Isaura (credits to @jjgomezcadenas)), they will be moved to other places (`components.py`, in most of the cases).

There are two issues that I faced while doing the city. The first one was already pointed out in #749, and it is related to the kdst information. It would be interesting that the output of this tracking city contains kdst information (`nS2` variable is used for the posterior analysis), however current `Beersheba` version doesn't save this table. Therefore, for the time being, we can access to this information just opening the `cdst` files, although I believe that in the near futurue we should add this table in `Beersheba` and then in `Isaura`.
The second issue concerns the tracking algorithm. As you may know, it seems that `Paolina`, and particularly `reco/paolina_functions/find_extrema()` funcion is not deterministic. This is because in some events there are more than one possible extreme voxels for a given track, and it appear that in those cases they are chosen randomly. The solution that I propose (and add here, in order to check that the output was correct -event though maybe this is not the place to do it-) consists in save all the possibilities, and then choose the extreme where the sum of their energy takes the maximum value (in order to avoid, if possible, the usage of `paolina_functions/drop_end_point_voxels()` function). Nonetheless, there would be other solutions, but the choice of one or another will be a bit arbitrary at the end, in my opinion.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -66,7 +66,6 @@ from .. io     .pmaps_io          import               pmap_writer
 from .. io     .rwf_io            import             buffer_writer
 from .. io     .mcinfo_io         import            load_mchits_df
 from .. io     .dst_io            import                 df_writer
-from .. io     .hits_io           import               hits_writer
 from .. types  .ic_types          import                        xy
 from .. types  .ic_types          import                        NN
 from .. types  .ic_types          import                       NNN
@@ -1248,7 +1247,19 @@ def track_blob_info_creator_extractor(vox_size         : [float, float, float],
     return create_extract_track_blob_info
 
 
-def compute_and_write_tracks_info(paolina_params, h5out):
+def compute_and_write_tracks_info(paolina_params, h5out,
+                                  hit_type, filter_hits_table_name='hits_select',
+                                  write_paolina_hits=None):
+
+    filter_events_nohits = fl.map(lambda x : len(x.hits) > 0,
+                                      args = 'hits',
+                                      out  = 'hits_passed')
+    hits_passed          = fl.count_filter(bool, args="hits_passed")
+
+
+    copy_Efield          = fl.map(Efield_copier(hit_type),
+                                            args = 'hits',
+                                            out  = 'Ep_hits')
 
     # Create tracks and compute topology-related information
     create_extract_track_blob_info = fl.map(track_blob_info_creator_extractor(**paolina_params),
@@ -1271,18 +1282,22 @@ def compute_and_write_tracks_info(paolina_params, h5out):
     write_tracks          = fl.sink(   track_writer     (h5out=h5out)             , args="topology_info"      )
     write_summary         = fl.sink( summary_writer     (h5out=h5out)             , args="event_info"         )
     write_topology_filter = fl.sink( event_filter_writer(h5out, "topology_select"), args=("event_number", "topology_passed"    ))
-    write_hits_paolina    = fl.sink(    hits_writer     (h5out, group_name="CHITS", table_name="highTh")
-                                                                                  , args="paolina_hits"       )
+
+    write_no_hits_filter  = fl.sink( event_filter_writer(h5out, filter_hits_table_name), args=("event_number", "hits_passed"))
 
 
-    fn_list = (create_extract_track_blob_info              ,
+    fn_list = (filter_events_nohits                        ,
+               fl.branch(write_no_hits_filter)             ,
+               hits_passed.              filter            ,
+               copy_Efield                                 ,
+               create_extract_track_blob_info              ,
                filter_events_topology                      ,
                fl.branch(make_final_summary, write_summary),
                fl.branch(write_topology_filter)            ,
-               fl.branch(write_hits_paolina)               ,
+               write_paolina_hits                          ,
                events_passed_topology.   filter            ,
-               fl.branch(write_tracks)                    ,)
+               fl.branch(write_tracks)                     )
 
-    compute_tracks = pipe(*fn_list)
+    compute_tracks = pipe(*filter(None, fn_list))
 
     return compute_tracks
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -1009,18 +1009,21 @@ def calculate_and_save_buffers(buffer_length    : float        ,
     buffer_definition = pipe(*filter(None, find_signal_and_write_buffers))
     return buffer_definition
 
-
-def copy_Ec_to_Ep_hit_attribute_(hitc : HitCollection) -> HitCollection:
-    """
-    The functions copies values of Ec attributes into Ep attributes. Takes as input HitCollection and returns a copy.
-    """
-    mod_hits = []
-    for hit in hitc.hits:
-        hit = Hit(hit.npeak, Cluster(hit.Q, xy(hit.X, hit.Y), hit.var, hit.nsipm),
-                  hit.Z, hit.E, xy(hit.Xpeak, hit.Ypeak), s2_energy_c=hit.Ec, Ep=hit.Ec)
-        mod_hits.append(hit)
-    mod_hitc = HitCollection(hitc.event, hitc.time, hits=mod_hits)
-    return mod_hitc
+def Efield_copier(energy_type: HitEnergy):
+    def copy_Efield(hitc : HitCollection) -> HitCollection:
+        mod_hits = []
+        for hit in hitc.hits:
+            hit = Hit(hit.npeak,
+                      Cluster(hit.Q, xy(hit.X, hit.Y), hit.var, hit.nsipm),
+                      hit.Z,
+                      hit.E,
+                      xy(hit.Xpeak, hit.Ypeak),
+                      s2_energy_c=getattr(hit, energy_type.value),
+                      Ep=getattr(hit, energy_type.value))
+            mod_hits.append(hit)
+        mod_hitc = HitCollection(hitc.event, hitc.time, hits=mod_hits)
+        return mod_hitc
+    return copy_Efield
 
 
 def make_event_summary(event_number  : int              ,
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -30,6 +30,7 @@ from .. evm    .event_model       import                   KrEvent
 from .. evm    .event_model       import                       Hit
 from .. evm    .event_model       import                   Cluster
 from .. evm    .event_model       import             HitCollection
+from .. evm    .event_model       import                 HitEnergy
 from .. evm    .event_model       import                    MCInfo
 from .. evm    .pmaps             import                SiPMCharge
 from .. core                      import           system_of_units as units
@@ -50,6 +51,7 @@ from .. reco                      import            peak_functions as pkf
 from .. reco                      import           pmaps_functions as pmf
 from .. reco                      import            hits_functions as hif
 from .. reco                      import             wfm_functions as wfm
+from .. reco                      import         paolina_functions as plf
 from .. reco   .xy_algorithms     import                    corona
 from .. filters.s1s2_filter       import               S12Selector
 from .. filters.s1s2_filter       import               pmap_filter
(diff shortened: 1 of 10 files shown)


jacg:
This PR is almost 5 months old, and has received no attention whatsoever. It relates to an issue (#749) which solicited the wisdom of: @pnovella @msorel @paolafer @ausonandres @Aretno @jahernando @jjgomezcadenas @gonzaponte

7 tests fail, most of them for fairly basic reasons. Please fix. Then rebase, so that the tests run on our new CI infrastructure (GitHub Actions).

Please don't leave PRs with red crosses (failing tests) in the queue

+ If the failures are spurious, re run the CI jobs.
+ If the failures are not spurious, fix the problems.
+ If you do not know how to fix the problem, ask for help.

A PR queue full of red crosses encourages people to ignore PRs.

ausonandres:
Hi, sorry i's my fault... I was trying to solve the test failing issues, however some other urgent work that I had to attend to came up at that moment. After that, I completely forgot to come back to this PR. I'll do it as soon as I have some time.

jacg:
When it is ready for review, please explicitly ask someone to review it, using the buttons/links near the top-right of the PR page. If it is not yet ready for review, please mark it as a draft, using the link at the bottom of the reviewers section: 'Still in progress? convert to draft'.

invisible_cities/cities/components.py
```diff
@@ -589,6 +594,31 @@ def MC_hits_from_files(files_in : List[str], rate: float) -> Generator:
...
+        with tb.open_file(path, "r") as h5in:
+            try:
+                run_number  = get_run_number(h5in)
+            except (tb.exceptions.NoSuchNodeError, IndexError):
+                continue
+
+            event_info = load_dst(path, 'Run', 'events')
+            for evt in dhits_df.event.unique():
+                timestamp = event_info[event_info.evt_number==evt].timestamp.values[0]
+                dhits = hits_from_df(dhits_df.loc[dhits_df.event==evt])
```
gondiaz, line 615: This is not correct since the DECO hits for a single event can contain multiple peaks. Since no peak information is included in the final track information output, we can not directly check if a multiple-track event comes from a real spliting or by a multiple-S2 event. This is not a relevant issue as long as commonly used 1S2 and 1-track cut are performed, but it is biasing the efficiencies.

This PR also modifies the search of track extremes in order to be deterministic but it does not adapt the tests, also affecting esmeralda tests and making the PR more difficult to review. I think this modifications should be a PR appart. I will do a different PR implementing the same functionality of this city, essentially running paolina algorithm over deconvoluted hits, without modifying the track extreme search.

ausonandres: Sorry, I don't understand completely your plan. If you want to add a new PR that exclusively adds the city to run paolina over deconvoluted hits, what would you do with this PR? just ignore it? would you use this just for solving the extreme issue? There is another thing to take into account, this PR apart from doing both things, also moves some functions from `Esmeralda` to `components.py`, would you also add that functionality in your new PR?

In my opinion I believe that the best approach could be removing all the extreme-related issue from this PR, and then add that new PR that you commented to take it into account. So the goal of this PR will keep being to add the tracking city.

gondiaz: Yes, as you might see in #798 I just did another PR since I didn't aim to undo your changes related with the extreme finder. I didn't move the functions still from `esmeralda.py` to ease the review, but I intend to do it once the PR is reviewed. Then the main difference with this PR is that I left the paolina functions unmodified, which in my opinion should be done, if ever, in a different PR. It also changes the source to solve the issue with multipeak events, such that those events are filtered.

invisible_cities/cities/components.py
```diff
@@ -589,6 +595,31 @@ def MC_hits_from_files(files_in : List[str], rate: float) -> Generator:
                        timestamp    = timestamp(evt))
 
 
+def dhits_from_files(paths: List[str]) -> Iterator[Dict[str,Union[HitCollection, pd.DataFrame, MCInfo, int, float]]]:
```
mmkekic, line 597: It seems to me that 
`hits_and_kdst_from_files`, `cdst_from_files` and now a newly added `dhits_from_files` are all doing almost the same thing! 
They all read hits (from different tables) and optionally `kdst` and `summary_df`, but do we really need 3 different readers? Cant we just have one reader with hits- table and group name as parameters and optionally read `kdst` and `summary_df`? @gonzaponte , @Aretno

invisible_cities/cities/components.py
```diff
@@ -1002,3 +1033,272 @@ def calculate_and_save_buffers(buffer_length    : float        ,
...
+                      xy(hit.Xpeak, hit.Ypeak),
+                      s2_energy_c=getattr(hit, energy_type.value),
+                      Ep=getattr(hit, energy_type.value))
+            mod_hits.append(hit)
+        mod_hitc = HitCollection(hitc.event, hitc.time, hits=mod_hits)
+        return mod_hitc
+    return copy_energy_to_Ep_hit_attribute_
+
+
+types_dict_summary = OrderedDict({'event'     : np.int32  , 'evt_energy' : np.float64, 'evt_charge'    : np.float64,
```
mmkekic: This is a bit cumbersome, maybe is better to place `types_dict_...` in ic_types.py?

invisible_cities/cities/components.py
gonzaponte: Yes, all that needs some refactoring, but I would put it in another PR.
There is `dst_from_files` which is almost what you request, but there are some subtleties done in some of those readers that might not be easy to implement in a general reader.

invisible_cities/cities/components.py
```diff
@@ -1002,3 +1033,276 @@ def calculate_and_save_buffers(buffer_length    : float        ,
...
+                                            args = ('event_number', 'topology_info', 'paolina_hits', 'out_of_map'),
+                                            out  = 'event_info')
+
+
+    # Define writers and make them sinks
+    write_tracks          = fl.sink(   track_writer     (h5out=h5out)             , args="topology_info"      )
+    write_summary         = fl.sink( summary_writer     (h5out=h5out)             , args="event_info"         )
+    write_topology_filter = fl.sink( event_filter_writer(h5out, "topology_select"), args=("event_number", "topology_passed"    ))
+
+    if hits_type:
```
mmkekic: isnt it better to have 
`write_hits_paolina` be `None` or `fl.branch(fl.sink..))` and then filter the `fn_list` for `None`s?
`pipe(*filter(None, fn_list)`?

mmkekic: We could also put this writer definition  in esmeralda and pass it as argument to this function, if we are sure not to use this in other cities?

invisible_cities/cities/components.py
```diff
@@ -1002,3 +1033,276 @@ def calculate_and_save_buffers(buffer_length    : float        ,
...
+        group_table = hits_type.split("/")[0]
```
mmkekic: `hits_type` is very misleading name, isnt it better to simply have 2 parameters directly? It would also be less error prone since you wont depend on the user passing string with this explicit pattern.

invisible_cities/cities/components.py
```diff
@@ -1002,3 +1033,276 @@ def calculate_and_save_buffers(buffer_length    : float        ,
...
+        node_table  = hits_type.split("/")[1]
+        write_hits_paolina = fl.sink(hits_writer(h5out, group_name=group_table, table_name=node_table),
+                                     args="paolina_hits"                                              )
+    else: write_hits_paolina = fl.sink(lambda _: None)
+
+    fn_list = (create_extract_track_blob_info              ,
```
mmkekic: you should add the `E` or `Ec` to `Ep` copying here, it would simply need a `hit_type` as input and used in both cities.

you should also add hits filter (and filter writer) here since it is also shared between the cities

invisible_cities/io/hits_io.py
```diff
@@ -21,26 +21,34 @@ def hits_from_df (dst : pd.DataFrame, skip_NN : bool = False) -> Dict[int, HitCo
...
-        for i, row in hits_df.iterrows():
-            if skip_NN and row.Q == NN:
+        for i, row in df.iterrows():
+            if skip_NN and getattr(row,'Q', -1) == NN:
                 continue
             hit = Hit(row.npeak,
-                      Cluster(row.Q, xy(row.X, row.Y), xy(row.Xrms**2, row.Yrms**2),
-                              row.nsipm, row.Z, row.E,
-                              Qc = getattr(row, 'Qc', -1)),       # for backwards compatibility
+                      Cluster(getattr(row,'Q', row.E), xy(row.X, row.Y),
```
mmkekic: This is becoming a bit silly, we are missing half of the variables to construct our `HitCollection` type...
It is also not true in case of Xrms and Yrms squared - here you are setting them to 0.

I am not saying you should deal with all event model problems in this PR but maybe doing something a bit cleaner like

```
for i, row in df.iterrows():
    Q = getattr(row,'Q', -1)
    if skip_NN and Q == NN:
        continue
    if hasattr(row, 'Xrms):
         Xrms = row.Xrms
         Xrms2 = Xrms**2
    else:
        Xrms =  Xrms2 = -1
    if hasattr(row, 'Yrms):
    ....
    nsipm = getattr...
    Qc = ...
```
This is still very ugly but avoids calling `getattr` for parameters that are used more than once, and it also makes it a bit more clear where we are setting this default -1 value

invisible_cities/cities/components.py
ausonandres: ok, I agree

invisible_cities/cities/components.py
ausonandres: ok, I'll try that. Regarding your second comment, my intention here was to generalize this part of the function in order to allow writing the deconvoluted hits if this tracking city is run before Esmeralda's hit correction (maybe in the future). But I'll change it if you want.

invisible_cities/cities/components.py
ausonandres: As we have discussed offline, it doesn't make a lot of sense the possibility to save the deconvoluted hits, therefore I remove the option for allowing to change the name of the table. Now there is just a flag called `save_paolina_hits` that will be True, in case the function is executed inside Esmeralda (saving the hits in the table `CHITS/highTh`) and it will correspond to False, if the function is run inside Isaura.

invisible_cities/io/hits_io.py
ausonandres: Ok, I'll modify this to match your suggestion. It's completely true that this should be done just as a provisional solution: as you say, in the near future the **event model must be revisited and changed** in order to avoid these issues...

ausonandres: Ok, thanks, I've changed the function trying to be a bit clearer. 
Yes, I obviously agree with you, it has no sense using this `HitCollection` here, however I'd say that that issue is not related with this PR (since it's meant to adapt the tracking algorithm to the current code), so **it should be mandatory to revisit the `event_model`in the near future.**

invisible_cities/cities/components.py
```diff
@@ -1002,3 +1037,267 @@ def calculate_and_save_buffers(buffer_length    : float        ,
     # Filter out order_sensors if it is not set
     buffer_definition = pipe(*filter(None, find_signal_and_write_buffers))
     return buffer_definition
+
+def copy_E_or_Ec_to_Ep(energy_type: HitEnergy):
```
mmkekic: As  @gonzaponte  suggest, to keep our naming convention, this function can be called 
`Efield_copier` and the inner one `copy_Efield` (this later name can also be used in the pipe elements namings)

ausonandres: It sounds good, I've changed that.

invisible_cities/cities/esmeralda.py
```diff
@@ -429,37 +190,16 @@ def esmeralda(files_in, file_out, compression, event_range, print_mod,
 
     threshold_and_correct_hits_high = fl.map(hits_threshold_and_corrector(threshold_charge=cor_hits_params['threshold_charge_high'], **cor_hits_params_),
                                              args = 'hits',
-                                             out  = 'cor_high_th_hits')
+                                             out  = 'hits')
```
mmkekic: you can use 
`item = 'hits'`

ausonandres: Good to know, I didn't know about that option.

mmkekic (approved):
This PR adds a new city, Isaura,  that applies paolina tracking functions to the output of Beersheba. The functions that are shared between Esmeralda and Isaura are extracted into a single pipe, hence no code is copied. Esmeralda tests pass as before, and new tests are added for Isaura output. 
This city is a temporary step until we decide on final reconstruction + analysis order (as discussed in Issue #749).

PR #756, opened 2020-11-18
Cython functions for detsim

mmkekic:
This PR adds cython classes for light tables on sipms and pmts for S2 signal, and a loop over ionization electrons producing a waveform-shaped matrix of expected photon values.

### DIFF
--- a/invisible_cities/detsim/light_tables_c.pxd
+++ b/invisible_cities/detsim/light_tables_c.pxd
@@ -1,0 +1,8 @@
+cdef class LightTable:
+    cdef readonly:
+        double el_gap_width
+        double active_radius
+        int    num_sensors
+    cdef:
+        double [:] zbins_
+    cdef double* get_values_(self, const double x, const double y, const int sensor_id)
--- a/invisible_cities/detsim/light_tables_c.pyx
+++ b/invisible_cities/detsim/light_tables_c.pyx
@@ -1,3 +1,22 @@
+"""
+This is a module to read the light tables files. Since this depends on the format of the table itself,
+the table-specific class should inherit from the base LightTable class, providing:
+
+ - get_values_ method, that depends on the x, y position in the EL plane and the internal sensor_id
+   the output of the method is a pointer to array of all the values across EL gap for a given x, y bin
+   if there are no values for a given x, y, sensor the result is a NULL pointer
+
+ - zbins_  property,  is an  array of positions corresponding to the EL gap partitions
+
+ - num_sensors attribute,  total number of sensors
+
+
+LT_SiPM and LT_PMT are designed for currently availabel light tables for sipms and pmts;
+should be revisited once the format for the lt is fixed and standardized.
+
+"""
+
+
 import numpy  as np
 import pandas as pd
 
--- a/invisible_cities/detsim/light_tables_c.pyx
+++ b/invisible_cities/detsim/light_tables_c.pyx
@@ -127,10 +127,7 @@ cdef class LT_SiPM(LightTable):
         self.max_psf2    = self.max_psf**2
         self.num_sensors = len(sipm_database)
 
-    @cython.boundscheck(False)
     @cython.wraparound(False)
-    @cython.cdivision(True)
-    @cython.initializedcheck(False)
     cdef double* get_values_(self, const double x, const double y, const int sns_id):
         cdef:
             double dist
@@ -257,10 +254,7 @@ cdef class LT_PMT(LightTable):
                                       for column in columns],axis=-1)[..., None]).astype(np.double)
         return values_aux, (xmin, xmax), (ymin, ymax)
 
-    @cython.boundscheck(False)
     @cython.wraparound(False)
-    @cython.cdivision(True)
-    @cython.initializedcheck(False)
     cdef double* get_values_(self, const double x, const double y, const int sns_id):
         cdef:
             double*  values
--- a/invisible_cities/detsim/light_tables_c.pyx
+++ b/invisible_cities/detsim/light_tables_c.pyx
@@ -1,4 +1,17 @@
 import numpy  as np
+import pandas as pd
+
+cimport cython
+cimport numpy as np
+
+from libc.math cimport            sqrt
+from libc.math cimport           floor
+from libc.math cimport round as cround
+
+
+
+from ..         core import system_of_units as   units
+from  . light_tables import read_lighttable as read_lt
 
 cdef class LightTable:
     """
@@ -25,3 +38,119 @@ cdef class LightTable:
             return np.asarray(<np.double_t[:self.zbins_.shape[0]]> pointer)
         else:
             return np.zeros(self.zbins_.shape[0])
+
+
+def get_el_bins(el_pitch, el_gap):
+    """
+    Returns the array of bins position given the bin distance and the total gap
+    """
+    return np.arange(el_pitch/2., el_gap, el_pitch).astype(np.double)
+
+cdef class LT_SiPM(LightTable):
+    """
+    A class to handle reading of sipm distance-based light table. Inherits from base class LightTable
+
+    Attributes:
+    -----------
+       el_gap_width  : double
+             width of the EL gap
+       active_radius : double
+             active radius of full detector volume
+       num_sensors   : int
+             number of sipm sensors
+       snsx          : numpy array of doubles
+             x position of sensors
+       snsy          : numpy array of doubles
+             y position of sensors
+
+    Parameters (keyword):
+    -----------
+        fname         : string
+              filename of the light table
+        sipm_database : pandas dataframe
+              dataframe containing information about sensor positions
+        el_gap_width  : float
+              optionally set new EL gap width
+        active_radius : float
+              optionally set new active radius
+
+    """
+
+    cdef readonly:
+        double [:] snsx
+        double [:] snsy
+    cdef:
+        double [:, ::1] values
+        double psf_bin
+        double max_zel
+        double max_psf
+        double max_psf2
+        double inv_bin
+        double active_r2
+
+    def __init__(self, *, fname, sipm_database, el_gap_width=None, active_radius=None):
+        lt_df, config_df, el_gap, active_r = read_lt(fname, 'PSF', el_gap_width, active_radius)
+        lt_df.set_index('dist_xy', inplace=True)
+        self.el_gap_width  = el_gap
+        self.active_radius = active_r
+        self.active_r2 = active_r**2 # compute this once to speed up the get_values_ calls
+
+        el_pitch  = float(config_df.loc["pitch_z"].value) * units.mm
+        self.zbins_    = get_el_bins(el_pitch, el_gap)
+        self.values    = np.array(lt_df.values/len(self.zbins_), order='C', dtype=np.double)
+        self.psf_bin   = float(lt_df.index[1]-lt_df.index[0]) * units.mm #index of psf is the distance to the sensor in mm
+        self.inv_bin   = 1./self.psf_bin # compute this once to speed up the get_values_ calls
+
+        self.snsx        = sipm_database.X.values.astype(np.double)
+        self.snsy        = sipm_database.Y.values.astype(np.double)
+        self.max_zel     = el_gap
+        self.max_psf     = max(lt_df.index.values)
+        self.max_psf2    = self.max_psf**2
+        self.num_sensors = len(sipm_database)
+
+    @cython.boundscheck(False)
+    @cython.wraparound(False)
+    @cython.cdivision(True)
+    @cython.initializedcheck(False)
+    cdef double* get_values_(self, const double x, const double y, const int sns_id):
+        cdef:
+            double dist
+            double aux
+            unsigned int psf_bin_id
+            double xsipm
+            double ysipm
+            double tmp_x
+            double tmp_y
+            double*  values
+        if sns_id >= self.num_sensors:
+            return NULL
+        xsipm = self.snsx[sns_id]
+        ysipm = self.snsy[sns_id]
+        tmp_x = x-xsipm; tmp_y = y-ysipm
+        dist = tmp_x*tmp_x + tmp_y*tmp_y
+        if dist>self.max_psf2:
+            return NULL
+        if x*x+y*y>=self.active_r2:
+            return NULL
+        aux = sqrt(dist)*self.inv_bin
+        bin_id = <int> floor(aux)
+        values = &self.values[bin_id, 0]
+        return values
+
+    def get_values(self, const double x, const double y, const int sns_id):
+        """
+        Retrive values from the light tables for all z partitions.
+
+        Parameters:
+        -----------
+        x, y   : doubles
+            electron position at EL plane
+        sns_id : int
+            internal sensor id in range [0, num_sensors)
+            sensors are ordered by sipm_database raws
+
+        Returns:
+        --------
+        array of values over EL gap partitions
+        """
+        return super().get_values(x, y, sns_id)
--- a/invisible_cities/detsim/light_tables_c.pyx
+++ b/invisible_cities/detsim/light_tables_c.pyx
@@ -1,0 +1,27 @@
+import numpy  as np
+
+cdef class LightTable:
+    """
+    Base abstract class to be inherited from for all LightTables classes.
+    It needs get_values_ cython method implemented, as well as zbins_ and sensor_ids_ attributes.
+    """
+
+    cdef double* get_values_(self, const double x, const double y, const int sensor_id):
+        raise NotImplementedError
+
+    @property
+    def zbins(self):
+        """ Array of z positions """
+        return np.asarray(self.zbins_)
+
+    def get_values(self, const double x, const double y, const int sns_id):
+        """
+        Returns array of light table values over EL gap for x, y position
+        of the electron and internal sensor id
+        """
+        cdef double* pointer
+        pointer = self.get_values_(x, y, sns_id)
+        if pointer!=NULL:
+            return np.asarray(<np.double_t[:self.zbins_.shape[0]]> pointer)
+        else:
+            return np.zeros(self.zbins_.shape[0])
--- a/invisible_cities/detsim/light_tables_c.pyx
+++ b/invisible_cities/detsim/light_tables_c.pyx
@@ -154,3 +154,121 @@ cdef class LT_SiPM(LightTable):
         array of values over EL gap partitions
         """
         return super().get_values(x, y, sns_id)
+
+
+cdef class LT_PMT(LightTable):
+    """
+    A class to handle reading of PMTs light table. Inherits from base class LightTable
+
+    Attributes:
+    -----------
+       el_gap_width  : double
+             width of the EL gap
+       active_radius : double
+             active radius of full detector volume
+       num_sensors   : int
+             number of sipm sensors
+
+    Parameters (keyword):
+    -----------
+        fname         : string
+              filename of the light table
+        el_gap_width  : float
+              optionally set new EL gap width
+        active_radius : float
+              optionally set new active radius
+
+    """
+
+    cdef:
+        double [:, :, :, ::1] values
+        double max_zel
+        double max_psf
+        double max_psf2
+        double inv_binx
+        double inv_biny
+        double xmin
+        double ymin
+        double active_r2
+
+    def __init__(self, *, fname, el_gap_width=None, active_radius=None):
+        lt_df, config_df, el_gap, active_r = read_lt(fname, 'LT', el_gap_width, active_radius)
+        self.el_gap_width  = el_gap
+        self.active_radius = active_r
+        self.active_r2 = active_r**2 # compute this once to speed up the get_values_ calls
+
+        sensor = config_df.loc["sensor"].value
+        #remove column total from the list of columns
+        columns = [col for col in lt_df.columns if ((sensor in col) and ("total" not in col))]
+        el_pitch    = el_gap #hardcoded for this specific table
+        bin_x = float(config_df.loc["pitch_x"].value) * units.mm
+        bin_y = float(config_df.loc["pitch_y"].value) * units.mm
+
+        self.zbins_ = get_el_bins(el_pitch, el_gap)
+        values_aux, (xmin, xmax), (ymin, ymax)  = self.__extend_lt_bounds(lt_df, config_df, columns, bin_x, bin_y)
+        lenz = len(self.zbins)
+        # add dimension for z partitions (1 in case of this table)
+        self.values = np.asarray(np.repeat(values_aux, lenz, axis=-1), dtype=np.double, order='C')
+        self.xmin = xmin
+        self.ymin = ymin
+        # calculate inverse to speed up calls of get_values_
+        self.inv_binx    = 1./bin_x
+        self.inv_biny    = 1./bin_y
+        self.num_sensors = len(columns)
+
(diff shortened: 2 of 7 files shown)


mmkekic:
I have rebased on accepted #757 PR since they share some common functions and test files

jacg:
Are there any benchmarks that justify the use of Cython?

mmkekic:
The problem is that there are no reasonable alternatives that solve the exact problem : having a matrix of 
`(number_of_electrons (~100 000) X number_of_time_slices (10) ) X number_of_sensors (~3000)` to then be rebinned into 
`some_arbitrary_binning X number_of_sensors`.  Placing the original matrix in memory is impossible without the approximations (naive estimate is > 20GB).  @gondiaz  tried to follow this approach with **some approximations** managing ~5s per event and <1.5 GB in https://github.com/gondiaz/IC/tree/detsim_v0. However, this approach lowers our desired resolution and does not scale well with the type of event.
Looping over numpy array is never a good idea but that was also tried in https://github.com/gondiaz/IC/tree/detsim_loop_based, however again adding some approximations. The memory is <0.7 GB and time is ~20s per event. Without the unjustified approximation the time was measured in minutes (we needed to use slow `numpy.add.at` function). 
The cython one keeps the same memory consumption (is basically the memory of input and output arrays) and is ~6s per event, without using any approximation. However it is a bit ugly, in order to make it modular and flexible but still fast I had to use a c pointer directly, other approaches implied slicing of numpy memory view that was ~2x slower. Some cython trials are in https://github.com/mmkekic/detsim_checks repository (Warning: the code there is really just a messy draft).


> Are there any benchmarks that justify the use of Cython?

jacg:
Is there somewhere I can read a high-level description of the computational task?

jacg:
> Is there somewhere I can read a high-level description of the computational task?

Put another way: can you describe 

1. the structure and size of the data that go into the process

2. the structure and size of the data that come out

3. The relationship between 1. and 2.

mmkekic:
> Is there somewhere I can read a high-level description of the computational task?

I am afraid not, but I tried to prepare a document that explains the process. If there are unclear parts please ask.
[s2_description.pdf](https://github.com/next-exp/IC/files/5690763/s2_description.pdf)
Hope it can be useful to @jmalbos  as well.

jacg:
My current understanding is something like this:

# Input

1. `es`: Electrons ✕ 1.5 ✕ 10<sup>5</sup>

     `x` |  `y`  |  `t`  | `nph`
   ------|-------|-------|------
   `f64` | `f64` | `f64` | `u64`

   Total space: 4 ✕ 8 ✕ 1.5 ✕ 10<sup>5</sup> -> 5MB

   **Important question**: are these already ordered on `t`?

2. `LT(z, r) -> p`: Light table 

   AFAICT, this is a function: (f64, f64) -> f64 implemented as a lookup table
   with 10 different `z`-slices, and an unspecified number of `r`-slices, which
   therefore occupies

   `f64` ✕ 10 (`z`s) ✕ ? (`r`s) -> 1MB / 12 r-slices

# Output

`f64` ✕ *number of sensors* ✕ *number of time bins*

1. SiPM responses

   f64 ✕ 4000 ✕ 2000 -> 64 MB

2. PMT responses

   f64 ✕ 60 ✕ 200000 -> 96 MB

(You also mention applying a Poisson distribution, but it seems that it is not
part of *this* calculation.)

# Calculation

Very roughly, for SiPMs

``` 
dv = drift velocity
zs = z-positions of 10 uniformly distributed slices of EL plane
for each electron
    for each sensor
        r = electron-sensor distance in x-y plane
        ps = LT(r, zs) # 10 different probabilities
        for p,z in ps,zs
            time_bin = bin(t + z / dv)
            response_contribution = p * nph / 10 
            add response_contribution to appropriate (sensor, time_bin)
```

For PMTs it's almost the same, just that the z-distribution is collapsed to a
single point, for the purpose of the light table probabilities.

# Discussion

In your earlier comment you allude to memory being a problem because there are
10<sup>5</sup> electrons ✕ 10 time slices ✕ 3000 sensors.

I assume (trust?, hope!) that the detector spits out results in chronological
order. If this is the case, you only ever need to hold 10 time-slices
(corresponding to the EL-layer z-slices) in memory: as soon as an electron
appears whose `t` lies after the end of your earliest in-memory time-slice, that
time slice can be yielded, and a fresh one added to your in-memory time-slice
buffer.

You also mentioned *arbitrary binning* ✕ *number of sensors* in the context of
excessive memory use. Again, if the input is time-ordered, then your first
binning pass can produce a stream of time-ordered time slices, which do not need
to be kept in memory simultaneously. A second pass (or a second component just
downstream, in the same pass) can then regroup these into arbitrarily binned
slices.

My original question was about benchmarks which justify the use of Cython. That
means there should be a pure python implementation and an *equivalent* Cython
implementation, and their run-times (or memory usage) should be compared. I
don't see how the discussion of memory is pertinent to this question: Cython
doesn't magically make numpy arrays occupy less memory.

Fundamentally, you have to show that your Cython solution is significantly more
efficient than your *equivalent* pure Python solution. Without that[*], you
shouldn't take on the burden of replacing Python with a less flexible language.

[*] Or some other compelling advantage. For example, it might be reasonable to
rewrite Python code in Rust, even if the Rust version offers no significant
speed improvement, because of the safety guarantees and lower long-term cost of
development and maintenance of code written in Rust.
    
Cython offers very little advantage over Python, beyond speed of execution, but
it is significantly more painful to work in. So you really should make sure that
it is significantly faster, before choosing Cython over pure Python.

# Questions

1. Have you tried *streaming* your data, rather than keeping everything in
   memory at once?
   
2. What do you do with these values once they have been generated? (Write to
   file? Use immediately? If the latter, how?)

3. What is an acceptable run time, per event?

4. What benefits does Cython bring to this problem, that are not available in
   pure Python?

mmkekic:
Ah, ok, I think I was not clear enough. 

Looping over numpy is guaranteed to be slower than cython loop over numpy array but using fast (c-implemented) numpy operations on vectors is usually not. Hence the main goal would be to solve the problem using numpy operations on arrays directly, as was our first approach. This required putting number_of_electrons x number_of_partitions x number_of_sensors in memory and doing multiplications and histogrammings, and this is memory consuming (20+ Gb).
**The actual input/output array sizes are not a problem for now** (I agree that having all this array in memory is not optimal and could be a problem in the future, but this would need to be fixed in other parts of IC as well, for example irene/diomira where the full array is read and manipulated). 

Second approach was to do a loop over electrons and use numpy vector manipulations inside the first loop. This was order of **10 minutes** .  Doing all three loops explicitely is slower than that.

Then we tried to use cython.

I tried several other approaches, the most promising one was to first group electrons in voxels determined by the psf file (ie distance and EL time partitions), and then loop over sensors and do the calculations using numpy fast bincount. This was around **1.2 minutes** per event.

The conclusion:  we didnt manage to think of any pure python approach that can remotely come close to cython speed without the excessive memory usage.

jjgomezcadenas:
My take: While premature optimisation may be the root of evil, this is a case where a number of other alternatives have been tried and discarded, cython being the most effective solution and an accepted tool for IC. So I think it is OK, for now. 

Alternatives to write fast code based on rust could be attractive, and worth to explore when/if relevant.

jacg:
By all means accept this PR, and get on with your lives. This is the best thing
you can do in the short term. [You should probably stop reading this now, and
ignore what follows.]

However, as you know, I take the long-term view: Investing in some careful
thought now, pays handsome dividends in the future. Avoid technical debt.

In general, one shouldn't use Cython unless it solves a problem that cannot
adequately be solved in pure Python. On the evidence I have seen and understood
so far, I remain unconvinced that this is the case here.

Some technical details:

+ The only argument I have seen that sounds vaguely convincing, is that a
  hand-written loop over a numpy array is completely unviable in pure Python,
  and rewriting this loop in Cython makes it acceptable.

+ I suspect that the overall structure of the algorithm and the data structures
  used are misguided, which leads to the illusion chucking Cython at it, is a
  good solution. I wouldn't be surprised if completely rethinking the algorithm
  leads to a pure Python solution which significantly outperforms the proposed
  Cython solution. However, I am not (yet) in a position make a pronouncement on
  this.


(As for Rust, it offers many, *many* advantages which Cython does not, but it
also brings greater costs than Cython.)

jacg:
Are there data somewhere in the IC repo that can be fed in to these functions in order to make a benchmark?

mmkekic:
No, the data are too large. In abovementioned repository https://github.com/mmkekic/detsim_checks, in README there is a dropbox link to download the data that should be run through the full city, as well as a full light tables. 

The branch that has a working version of a full city is https://github.com/mmkekic/IC/tree/detsim_unification,  hence to make it work you would need to checkout the branch, compile cython and run standard 
`city detsim detsim.conf` , where `detsim.conf` file is also provided in the https://github.com/mmkekic/detsim_checks repository (or use `run_detsim.py` script that I used to run memory profiler). Note that here you run a full city flow and not just the desired function.

The other option is to give random x, y, time, number_of_photons arrays (of a size ~110k) but this might not describe the reality well if you plan to try to use  some grouping or voxelization.

jacg:
Why don't you add

1. the code in https://github.com/mmkekic/detsim_checks

2. the files in the Dropbox you mention

to the IC `detsim_unification` branch?

The first can be placed in an arbitrary convenient subdirectory of IC, the
second can be added via Git LFS. Both can be removed from the branch before the
branch is merged into `master` (if the branch is supposed to be merged).

As it stands, with everything scattered all over the place (across different
repositories and file-sharing services) the activation energy to get going with
this is unhelpfully huge, and the maintainability and feeding back of others'
contributions will also suffer greatly.

(Hard-wired absolute paths to one developer's local filesystem, don't help
either!)

mmkekic:
I added the files to the testfiles folder and detsim_benchmark folder has the corresponding config and run_detsim.py script, hence you need to checkout the branch `detsim_unification` and run the script (and so the typical setup IC + compile cython)
I havent added other trials from my scratch repository cause they were mostly used to test different implementations, all involving cython.
If you have issues in setting this up, or further questions about the flow please ask.

( I don't think it is a good practice to add unnecesarry files to our LFS server but I believe in this case is justified :) )

jacg:
Thanks. This should make it much easier to contribute.

I have some comments up-front:

> I don't think it is a good practice to add unnecesarry files to our LFS server

+ Benchmarks are *necessary* if you are contemplating adding Cython to the
  codebase (or indeed, doing any type of optimization).
  
+ These files are necessary for the benchmarks to run.

+ Ergo, these files are necessary and therefore should be added to the
  repository.

+ These files are too big to reasonably add as normal git files.

+ Ergo, these files must live on the LFS server.

It is incorrect to classify these files as unnecessary. It makes about as much
sense as saying that we shouldn't commit tests, because they're not necessary.
Benchmarks, like tests, give vital and independently-verifiable information
about the behaviour of the code. They are absolutely necessary.

Consequently, the commit title

> Add files for detsim benchmarks (DO NOT MERGE THIS COMMIT TO MASTER)
  
is wrong. This commit *must* be merged into master (if the code that it helps to
benchmark gets merged). The commit should be *reverted* before merging: these
files shouldn't pollute our working trees in day-to-day work, but we should be
able to retrieve them from history, so that we can run the benchmarks easily at
any time we need to, in the future.

mmkekic:
I agree partially, but the problem (as usual) is that we are doing development without a clear idea of what future flexibility we need. For example, the light tables format and dependencies can change, hence the light tables reading should be flexible (actually the format of the table did change during the development so we would need to commit them 2x if we uploaded the files to the server from the start). The event type can change, there could be events with more  hits (ie electrons) so whatever implementation should be at most linear with the number of electrons (both in time and memory consumption).
Basically I wouldn't take these files as an ultimate benchmark test, rather as a guidance for the current implementation since we need this code asap... 
(I also realized these files are ~100 MB so I really shouldn't make such a fuss about storing them on our server)

jacg:
I'm confused: Should I be able to find the function called `create_wfs` in this PR, anywhere in the `detsim_unification` branch?

jacg:
> I'm confused: Should I be able to find the function called create_wfs in this PR, anywhere in the detsim_unification branch?

OK, so it's approximately `electron_loop.pyx/electron_loop`.

jacg:
What is the relationship between the `detsim_v2.0` branch (this PR) and `detsim_unification`? Is it that the former aims to merge a subset of the work done in the latter, with the rest of the latter following in a later PR?

It's really annoying to be faced with two versions of the code which are similar-but-different. To review this PR it must be possible to run benchmarks for it. Having to switch to another branch and run a different version of the code in order to run the benchmarks, is not really an reasonable way of working!

mmkekic:
I agree! @gonzaponte  was against the idea of putting a full city in one PR so we divided it in a bunch of small ones. However they are not independent and were developing simultaniously so we ended up with a spaghetti code with merge conflicts. 
`detsim_unification` branch was a quick fix of the merge conflicts providing a working version but is not the final one since bunch of cosmetics changes were made in the 'official' branches. I am not sure what is the best way to proceed. 
The PRs needed for the full detsim city are #694, #695, #757 that should all be merged before this one. And the final city flow PR will depend on all of them.  The main issue is that we started developing before having a clear algorithm and the correlations between modules kept changing (@gondiaz  that started implementing the detsim has **14** detsim branches, this gives you the idea of the mess we have).

If it will make things easier I  (or @gondiaz  since he is supposed to open a final city body PR once all the rests are approved) can do (yet another!) branch that should resemble more a final version...

gonzaponte:
Originally the work was divided into different tasks that were (in principle) independent of each other. I am not really sure how we got here, but I guess there were not so independent after all. My involvement has been discontinuous and perhaps counterproductive since I couldn't really follow its evolution.

mmkekic:
@jacg , I have rebased `detsim_unification` branch on this PR (and other PRs). You would need to do a hard reset since there were conflicts. 
What is the status of this? Will anyone actually try some better approach or shell we proceed with the review accepting cython?

jacg:
I want to have a good look at this, but I'm not letting conda anywhere near my
machine, so either it will have to wait until we have merged the Nix branch, or
I'll have to juggle branches locally by rebasing on the Nix branch to work on
it, and rebasing back or cherry-picking anything I manage to contribute.

In short, I'm not going to get this done soon.

So, if it's urgent and it solves a problem for you, ignore me and proceed with
the review. Whether that's the best choice in the long term, I wouldn't know.

jjgomezcadenas:
It is urgent. Let’s approve this PR. If Jacek solves the problem in a better way we can always revisit the issue.

jacg:
> It is urgent. Let’s approve this PR

Good. I'm decoupling from this now, and will look at it again if/when we have Nix.

jmalbos:
An initial general comment, after reading the code a couple of times:  the `light_tables_c.pyx` file is a bit obscure, and there aren't a lot of comments documenting what's being done and why. For example, the SiPM and PMT LT-derived classes are significantly different, and it's hard to draw parallels between them. This is a potential future problem. 

I'll send further comments inline.

jacg:
I was going to decouple, but ... @jmalbos' comments above, reflect my initial impressions when looking at the code. (I'll leave it at that, lest I write something the size of *War and Peace* when my time would be better spent just rewriting the code.)

invisible_cities/detsim/light_tables_c.pxd
```diff
@@ -0,0 +1,9 @@
+cdef class LT:
```
jmalbos: I'd change the class name to `LightTable`.

invisible_cities/detsim/s2_waveforms_c.pyx
```diff
@@ -0,0 +1,88 @@
...
+def create_wfs(double [:] xs           ,
+               double [:] ys           ,
+               double [:] ts           ,
+               int    [:] phs          ,
+               LT         lt           ,
+               double     el_dv        ,
+               double     sns_time_bin ,
+               double     buffer_length,
+               double     tmin = 0    ):
+
```
jmalbos, line 63: Add docstring documenting the function.

invisible_cities/detsim/light_tables_c.pyx
```diff
@@ -0,0 +1,171 @@
...
+
+
+from ..         core import system_of_units as units
+from  . light_tables import                  read_lt
+
+cdef class LT:
+    """Base abstract class to be inherited from for all LightTables classes.
+    It needs get_values_ cython method implemented, as well as zbins_ and sensor_ids_ attributes.
+    """
+    cdef double* get_values_(self, const double x, const double y, const int sensor_id):
```
jmalbos, line 20: Is there any advantage in using a raw pointer to double instead of a Memoryview?

invisible_cities/detsim/light_tables_c.pyx
```diff
@@ -0,0 +1,171 @@
...
+        if pointer!=NULL:
+            return np.asarray(<np.double_t[:self.zbins_.shape[0]]> pointer)
+        else:
+            return np.zeros(self.zbins_.shape[0])
+
+
+def get_el_bins(el_pitch, el_gap):
+    return np.arange(el_pitch/2., el_gap, el_pitch).astype(np.double)
+
+cdef class LT_SiPM(LT):
```
jmalbos, line 43: The names of the class variables could be more explicit to improve readability.

invisible_cities/detsim/light_tables.py
```diff
@@ -1,16 +1,52 @@
...
 from functools import partial
-
+from .. core                import system_of_units as units
 from .. core.core_functions import binedges_from_bincenters
 
 from .. io.dst_io  import load_dst
 
 
-def create_lighttable_function(filename : str)->Callable:
+def read_lt(fname, group_name, el_gap=None, active_r=None):
```
jmalbos: Maybe this should be a member function of the `LightTable` class?

I'd rename it to `read_lighttable`.

invisible_cities/detsim/light_tables.py
```diff
@@ -1,16 +1,52 @@
...
+        group_name : group name under which table is saved
+        el_gap     : optional set of different el_gap
+                     a Warning is issued in case it differs from the config one
+        active_r   : optional set of different active_r
+                     a Warning is issued in case it differs from the config one
+    Returns a tuple
+        lt_df      : dataframe with light table
+        config_df  : dataframe with configuration parameters
+        el_gap     : input el_gap or el_gap read from file
+        el_gap     : input active_r or active_r read from file
```
jmalbos: `el_gap` => `active_r`

invisible_cities/detsim/light_tables.py
```diff
@@ -1,16 +1,52 @@
...
 from .. io.dst_io  import load_dst
 
 
-def create_lighttable_function(filename : str)->Callable:
+def read_lt(fname, group_name, el_gap=None, active_r=None):
+    """ A helper function to extract dataframes and configuration infro from files.
+    Inputs
+        fname      : string of file name
+        group_name : group name under which table is saved
+        el_gap     : optional set of different el_gap
```
jmalbos: This is a single number, right? The description ("set of different el_gap") seems to indicate one could specify several values.

invisible_cities/detsim/light_tables_c.pyx
```diff
@@ -0,0 +1,171 @@
...
+        lt_df, config_df, el_gap, active_r = read_lt(fname, 'PSF', el_gap, active_r)
+        lt_df.set_index('dist_xy', inplace=True)
+        self.el_gap    = el_gap
+        self.active_r  = active_r
+        self.active_r2 = active_r**2
+
+        el_pitch  = float(config_df.loc["pitch_z"].value) * units.mm
+        self.zbins_    = get_el_bins(el_pitch, el_gap)
+        self.values    = np.array(lt_df.values/len(self.zbins_), order='C', dtype=np.double)
+        self.psf_bin   = float(lt_df.index[1]-lt_df.index[0])
```
jmalbos: Add a comment like "pdf bin size inferred from the difference between the distances of the first two entries".

Should you be using the unit system there for consistency?

invisible_cities/detsim/light_tables_c.pyx
```diff
@@ -0,0 +1,171 @@
...
+        bin_x = float(config_df.loc["pitch_x"].value) * units.mm
+        bin_y = float(config_df.loc["pitch_y"].value) * units.mm
+        # extend min, max to go over the active volume
+        xmin, xmax = xmin_-np.ceil((self.active_r-np.abs(xmin_))/bin_x)*bin_x, xmax_+np.ceil((self.active_r-np.abs(xmax_))/bin_x)*bin_x
+        ymin, ymax = ymin_-np.ceil((self.active_r-np.abs(ymin_))/bin_y)*bin_y, ymax_+np.ceil((self.active_r-np.abs(ymax_))/bin_y)*bin_y
+        #create new centers
+        x          = np.arange(xmin, xmax+bin_x/2., bin_x).astype(np.double)
+        y          = np.arange(ymin, ymax+bin_y/2., bin_y).astype(np.double)
+        xx, yy     = np.meshgrid(x, y)
+        values_aux = (np.concatenate([griddata((xtable, ytable), lt_df[column], (yy, xx), method='nearest')[..., None]
```
jmalbos, line 253: This bit could be moved to a separate member function, adding a couple of lines documenting what you're doing (something like "Extend light table to fully cover the chosen radius extrapolating from the existing data.").

invisible_cities/detsim/light_tables_c.pxd
```diff
@@ -0,0 +1,9 @@
...
+    cdef readonly:
+        double el_gap
```
jmalbos: I'd go for more explicit names for these variables:
```
double el_gap_width
double active_radius
int    num_sensors
```

invisible_cities/detsim/light_tables_c.pyx
```diff
@@ -0,0 +1,171 @@
...
+        double max_psf2
+        double inv_bin
+        double active_r2
+
+    def __init__(self, *, fname, sipm_database, el_gap=None, active_r=None):
+        lt_df, config_df, el_gap, active_r = read_lt(fname, 'PSF', el_gap, active_r)
+        lt_df.set_index('dist_xy', inplace=True)
+        self.el_gap    = el_gap
+        self.active_r  = active_r
+        self.active_r2 = active_r**2
```
jmalbos: I'd add a comment to let future readers now that some of these quantities are defined to speed up the calculation. Something like
```self.active_r2 = active_r**2 # compute this once to speed up the get_values_ calls```

File: invisible_cities/detsim/light_tables_c.pyx [hunk 12]
```diff
@@ -0,0 +1,171 @@
...
+        self.nsensors = len(self.sensor_ids_)
+
+    @cython.boundscheck(False)
+    @cython.wraparound(False)
+    @cython.cdivision(True)
+    @cython.initializedcheck(False)
+    cdef double* get_values_(self, const double x, const double y, const int sns_id):
+        cdef:
+            double*  values
+            int xindx_, yindx_
```
jmalbos: Are the trailing underscores needed?

invisible_cities/detsim/light_tables.py
mmkekic: I wouldn't mix cython LightTable with this module, if we ever get rid of cython this could simply stay as is

jmalbos:
I think this is in much better shape now, but I'd suggest some commit squashing before merging.

mmkekic:
Commits are squashed and the branch is rebased on #757

jmalbos:
I approve this PR.

jmalbos (approved):

PR #757, opened 2020-11-19
Detsim S1 simulation

gondiaz:
This PR contains the detsim functions that simulate the S1 signal from the total number of photons using the S1 LightTable. It is a single module simulate_S1.py and its corresponding test module. Also, the S1 LightTable test file is added.

This PR, along with PR #756, overrides PRs #698, #699 and #734.


invisible_cities/detsim/simulate_S1.py
```diff
@@ -0,0 +1,194 @@
...
+
+    signaltype = Config.loc["signal_type"].value
+    sensor     = Config.loc["sensor"]     .value
+
+    lt = lt.set_index(["x", "y", "z"])
+    lt = lt.drop(sensor + "_total", axis=1) # drop total column
+
+    nsensors = lt.shape[-1]
+
+    if signaltype == "S1":
```
mmkekic: Can't you unite this two cases, there is a lot of repeated code. You'd probably need to assign `None` to `zbins` -related arrays in case of `S2` and make `get_lt_values` be able to work with that

invisible_cities/detsim/simulate_S1.py
```diff
@@ -0,0 +1,194 @@
...
+            hit position
+    Returns:
+        :pes: np.ndarray
+            photoelectrons at each PMT produced by all hits.
+    """
+    pes = photons[:, np.newaxis] * LT(xs, ys, zs)
+    pes = np.random.poisson(pes)
+    return pes.T
+
+tau1 = 4.5; c1 = 0.1
```
mmkekic: Is there a reason why you opted to make this global?

gondiaz: I am not sure if that is the best approax. We could define them as local variables inside `S1_minimize` or forget about defining the variables and put them as numbers directly.

invisible_cities/detsim/simulate_S1.py
gondiaz: A `None` type in bins does not work. The best approach I found is to include in the S2 LT a z column (let's say center of the EL gap). Then, the `get_lt_value` function would contain three arguments x, y, z in both, S1 and S2 LightTables. 

In my opinion this approach is good since we can greatly reduce the extent of this function, as well as to have symmetric LTs for S1 and S2 in the sense that they contain the same column indexes. I would go for it if does not colide with #756, that I think is not the case.

invisible_cities/detsim/simulate_S1.py
```diff
@@ -0,0 +1,188 @@
...
+    Parameters:
+        :LT: function
+            The Light Table in functional form
+        :photons: np.ndarray
+            The photons generated at each hit
+        :xs, ys, zs: np.ndarray
+            hit position
+    Returns:
+        :pes: np.ndarray
+            photoelectrons at each PMT produced by all hits.
```
mmkekic: Add output shape in the description, ie (`number_of_sensors X number_of_hits`)

invisible_cities/detsim/simulate_S1.py
mmkekic: You should place them in `S1_minimize`. Dont forget naming conventions, functions and variables should be lowercase  only with underscores between words.  Use more descriptive names than `n, a1, a2`, as well as a more descriptive name for the function itself with an underscore in front since it should be private to this module only. Alternatively you could even place it inside `generate_S1_times` (remove capitals). Add docstring, ie that this is inverse transform sampling and `S1_minimize` (without the P random number) is inverse cdf.

mmkekic: Also, `tau1, tau2` have time units (`ns`).

invisible_cities/detsim/simulate_S1.py
```diff
@@ -0,0 +1,188 @@
...
+    S1_times are histogramed in a waveform of given buffer_length and bin_width
+    Parameters
+        :S1_times: list
+            output of generate_S1_times_from_pes, list of size equal to the
+            number of pmts. Each element is an array with the times of the S1
+    Returns:
+        :wfs: np.ndarray
+            waveforms with buffer_length and bin_width
+    """
+    bins = np.arange(0, buffer_length + bin_width, bin_width)
```
mmkekic: `bias_time` should probably be called `start_time`, `t_start` or similar.  Then you can define bins in `[t_start, t_start+buffer_length)`

invisible_cities/detsim/simulate_S1.py
```diff
@@ -0,0 +1,188 @@
...
+
+    if lt.get("z") is None:
+        lt["z"] = 1 # add fake z
+        had_z = False
+    else: had_z = True
+
+    lt = lt.set_index(["x", "y", "z"])
+    nsensors = lt.shape[-1]
+
+    xbins=binedges_from_bincenters(np.unique(lt.index.get_level_values('x')))
```
mmkekic: There seems to be a numpy function that does this for you 
`histogram_bin_edges`, I believe you should call it with number of bins to be unique values +1, and specify the range you need.
However we need to decide what to do in case light table starts at `x_min> -r_active + x_pitch`, I think we should extend the lower and upper bin edge to be exactly on `r_active`. In that case we do need a custom function and `binedges_from_bincenters` can be easily adapted for this (ie adding arguments `left_edge, right_edge`). Eventually this function should then be moved to `core_functions.py`

invisible_cities/detsim/simulate_S1_test.py
```diff
@@ -0,0 +1,95 @@
...
+
+from pytest import raises
+
+from hypothesis import given
+from hypothesis import settings
+from hypothesis.strategies import integers
+
+
+@settings(deadline=1000)
+@given(integers(min_value=0, max_value=10000))
```
mmkekic: I think hypothesis is a total overkill for testing a size and a type of the output. If you do want to keep this test rename it to `test_generate_s1_time_output_type` and use some fixed size.

invisible_cities/detsim/simulate_S1.py
```diff
@@ -0,0 +1,188 @@
...
+        times = generate_S1_time(np.sum(pes_at_pmt))
+        S1_times.append(times + repeated_times)
+    return S1_times
+
+
+def S1_waveforms_creator(s1_lighttable, ws, wf_pmt_bin_width):
+    S1_LT = create_lighttable_function(os.path.expandvars(s1_lighttable))
+
+    def create_S1_waveforms_from_hits(x, y, z, time, energy, tmin, buffer_length):
+        s1_photons     = np.random.poisson(energy / ws)
```
mmkekic: Maybe this doesn't belong here, lets take it out in a 
```
def num_scintillation_photons (energy, ws):
    return np.random.poisson(energy, ws) 
```
with docstrings and type annotation of course :)

mmkekic (commented):
You have to review the naming, type annotation and docstrings for all functions. 
About tests, since we are dealing with distributions I guess we should use flaky and allow some failures, but I am never sure how to unit test random sampling. What are the standards for this @gonzaponte , @jacg ?

invisible_cities/detsim/simulate_S1.py
jacg: > I am never sure how to unit test random sampling. What are the standards for this @gonzaponte , @jacg ?

Sorry, I don't have any deep wisdom to offer on this.

gonzaponte:
> I am never sure how to unit test random sampling. What are the standards for this @gonzaponte

I don't have a good recipe. I would for sure try to check the hard limits of the output distribution. For example, if it doesn't make sense that the output is negative that would be a good test. Sometimes it makes sense to check that the output is of a given type (Poisson, binomial). Other tests may include checking extreme cases. For example, Poisson(0) should always return 0 or Gauss(mu, 0) should always be return mu. These kinds of tests are mostly deterministic so I would recommend adding as many of these as you can think of.

Statistical tests (such as checking the mean or std) are, in my experience, cumbersome. They fail every now and then even with flaky flags. I think a good replacement for them is for the functions that use them to have a fixed-seed test. For instance
```
@fix_seed(a_seed)
def test_drift_ie():
    initial_ie = [14, 19, 28]
    result = drift_ie(initial_ie)
    assert result == [3, 4, 6]
```
(`fix_seed` would need to be implemented). Of course, the hardcoded result would have to be approved based on the assumption that the function works correctly, which would be assessed by a human in some unspecified way. This gives you at least two things: if the random generators we use change from one version to another, we will probably notice it; and if someone decides to refactor the function at least there will be something to tell them whether they messed it up. It is not as robust as other things, but it is something.

invisible_cities/detsim/simulate_s1.py
```diff
@@ -0,0 +1,233 @@
...
+
+def generate_s1_times_from_pes(s1_pes_at_pmts : np.ndarray,
+                               hit_times      : np.ndarray)->list:
+    """Given the s1_pes_at_pmts, this function returns the times at which the pes
+    are be distributed (see generate_S1_time function).
+    It returns a list whose elements are the times at which the photoelectrons in that PMT
+    are generated.
+    Parameters:
+        :S1_pes_at_pmts: np.ndarray
+            the pes at each PMT generated by each hit
```
mmkekic, line 103: you are missing `hit_times`

invisible_cities/detsim/simulate_s1.py
```diff
@@ -0,0 +1,233 @@
...
+    s1_times = []
+    for pes_at_pmt in s1_pes_at_pmts:
+        repeated_times = np.repeat(hit_times, pes_at_pmt)
+        times = generate_s1_time(np.sum(pes_at_pmt))
+        s1_times.append(times + repeated_times)
+    return s1_times
+
+
+def s1_waveforms_creator(s1_lighttable, ws, wf_pmt_bin_width):
+    S1_LT = create_lighttable_function(os.path.expandvars(s1_lighttable))
```
mmkekic: Add type hints and docstrings.

invisible_cities/detsim/simulate_s1_test.py
```diff
@@ -0,0 +1,175 @@
...
+    idx = (np.abs(array - value)).argmin()
+    return array[idx]
+
+@settings(max_examples=500)
+@given(xs = arrays(np.float, 1, elements=floats(min_value=-500, max_value=500)),
+       ys = arrays(np.float, 1, elements=floats(min_value=-500, max_value=500)),
+       zs = arrays(np.float, 1, elements=floats(min_value=-500, max_value=500)))
+def test_get_lt_values_s1(ICDATADIR, xs, ys, zs):
+
+    s1_lighttable = os.path.join(ICDATADIR, "NEXT_NEW.energy.S1.PmtR11410.LightTable.h5")
```
mmkekic: This should be a fixture (something like e15d2063ac415291c68454695791233cd22523b9 , you can probably copy the code and add S1 table as well)

invisible_cities/detsim/simulate_s1.py
```diff
@@ -0,0 +1,233 @@
...
+        :binedges: np.ndarray
+            bin edges
+    """
+    if np.any(bincenters[:-1] >= bincenters[1:]):
+        raise Exception("unsorted or repeated bin centers")
+    if range is None:
+        range = (bincenters[0], bincenters[-1])
+    else:
+        if not (range[0]<range[1]):
+            raise Exception("lower edge must be slower than higher")
```
mmkekic: **S**lower (:

mmkekic (commented):
Suggested placing of the function:
- `binedges_from_bincenters` should go to `core_functions`
- `create_lighttable_function` should go to new `light_tables.py` module
- `s1_waveforms_creator` should go to `detsim_waveforms.py` 

and tests should be moved to the appropriate test modules.

Then a history should be rewritten as 'added a functionality', 'added tests' for each module (ie S1 functions that stay in `simulate_s1_module`can be commited together).

Finally, a comment about the finite times of photons to reach the pmts plane should be added at the beginning of the s1 module as a reminder that this might be a problem in the future (the distribution plots can be attached to this PR)

invisible_cities/detsim/simulate_s1.py
jacg: Please use a specific exception type, rather than the completely general and meaningless `Exception`!

invisible_cities/detsim/detsim_waveforms.py
```diff
@@ -0,0 +1,73 @@
...
+            waveforms with start_time, buffer_length and bin_width
+    """
+    bins = np.arange(start_time, start_time + buffer_length + bin_width, bin_width)
+    wfs  = np.stack([np.histogram(times, bins=bins)[0] for times in s1_times])
+    return wfs
+
+
+def s1_waveforms_creator(s1_lighttable   : str,
+                         ws              : float,
+                         wf_pmt_bin_width: float)->Callable:
```
mmkekic, line 40: Add to docstrings

invisible_cities/detsim/detsim_waveforms.py
```diff
@@ -0,0 +1,73 @@
...
+    Parameters:
+        :s1_lighttable: str
+            the s1 lighttable filename, it can contain env variables
+        :ws:
+            the inverse scintillation yield
+    Returns:
+        :create_s1_waveforms_from_hits: function
+            see function docstring
+    """
+    S1_LT = create_lighttable_function(os.path.expandvars(s1_lighttable))
```
mmkekic: I know is local but I prefer it to be a lower-case name

invisible_cities/detsim/light_tables.py
```diff
@@ -0,0 +1,65 @@
...
+    def get_lt_values(xs, ys, zs):
+        if len(zs) == 1:
+            zs = np.full(len(xs), zs)
+        if not (len(xs) == len(ys) == len(zs)):
+            raise Exception("input arrays must be of same shape")
+        sel = (np.sqrt(xs**2 + ys**2) <= act_r) & (zbins[0]<=zs) & (zs<=zbins[-1]) #inside bins
+        xindices = pd.cut(xs[sel], xbins, include_lowest=True, labels=xcenters)
+        yindices = pd.cut(ys[sel], ybins, include_lowest=True, labels=ycenters)
+        zindices = pd.cut(zs[sel], zbins, include_lowest=True, labels=zcenters)
+        indices = pd.Index(zip(xindices, yindices, zindices), name=("x", "y", "z"))
```
mmkekic: indentation

invisible_cities/detsim/light_tables.py
```diff
@@ -0,0 +1,65 @@
...
+        values  = np.zeros((len(xs), nsensors))
+        values[sel] = lt.loc[indices]
+        return values
+
+    if lt.get("z") is None:
+        lt.loc[:, "z"] = 1 # add fake z
+        had_z = False
+    else: had_z = True
```
mmkekic: put `had_z` in next line

invisible_cities/detsim/light_tables.py
```diff
@@ -0,0 +1,65 @@
...
+    xcenters = np.unique(lt.index.get_level_values('x'))
+    ycenters = np.unique(lt.index.get_level_values('y'))
+    zcenters = np.unique(lt.index.get_level_values('z'))
+
+    xbins = binedges_from_bincenters(xcenters, range=(-act_r, act_r))
+    ybins = binedges_from_bincenters(ycenters, range=(-act_r, act_r))
+    zbins = binedges_from_bincenters(zcenters)
+
+    if had_z: return get_lt_values
+    else:     return partial(get_lt_values, zs=np.array([1]))
```
mmkekic: indent

invisible_cities/detsim/simulate_s1.py
```diff
@@ -0,0 +1,110 @@
...
+        distributed with mean energy/ws
+    """
+    return np.random.poisson(energy / ws)
+
+
+def compute_s1_pes_at_pmts(xs      : np.ndarray,
+                           ys      : np.ndarray,
+                           zs      : np.ndarray,
+                           photons : np.ndarray,
+                           LT      : Callable  )->np.ndarray:
```
mmkekic: Use lower_case letters

invisible_cities/detsim/simulate_s1.py
```diff
@@ -0,0 +1,110 @@
...
+    tau1 = 4.5 * units.ns
+    tau2 = 100 * units.ns
+    p1 = 0.1
+
+    sel = np.random.uniform(0, 1, size=size)
+    size_1 = np.sum(sel<=p1)
+    size_2 = size - size_1
+    r1 = np.random.exponential(tau1, size_1)
+    r2 = np.random.exponential(tau2, size_2)
+    return np.concatenate([r1, r2])
```
mmkekic: Maybe shuffle the sample before returning?

invisible_cities/detsim/simulate_s1.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def generate_s1_times_from_pes(s1_pes_at_pmts : np.ndarray,
+                               hit_times      : np.ndarray)->list:
+    """Given the s1_pes_at_pmts, this function returns the times at which the pes
+    are distributed (see generate_S1_time function).
+    It returns a list whose elements are the times at which the photoelectrons in that PMT
+    are generated.
+    Parameters:
+        :S1_pes_at_pmts: np.ndarray
```
mmkekic: name is not capital

mmkekic (commented):
After the cosmetics changes and some squashing the PR will be ready to be approved

invisible_cities/detsim/simulate_s1_test.py
```diff
@@ -0,0 +1,68 @@
...
+
+def test_generate_s1_time():
+    np.random.seed(1234)
+    times = generate_s1_time(size=2)
+    expected = np.array([ 57.57690962, 153.87864745])
+
+    np.testing.assert_allclose(times, expected)
+
+
+def test_generate_s1_times_from_pes():
```
mmkekic, line 55: Can you add a test that generated times are always larger than hit times? You can set one photon per hit (even use only one pmt) and check the output times. Set largish differences in hit times.

gondiaz:
I have rewritted the commit history. 

Another relevant comment for this PR:
S1 photons in detsim are assumed to arrive to the PMTs instantaneously.  In detsim we compute the number of scintillation photons of each track-hit and then the pes (signal) arriving at each PMT using the S1 lighttable. Then we distribute the signal in the PMTs using the S1-times distribution (double exponential). In this proccess we assume that the photons are not time-shifter by the take it takes them to arrive to the sensors. This delay time can be seen in the MC simulation and can be as long as ~6 nano-seconds in NEXT100 (plot below). The delay time would scale with the detector size.

image

mmkekic (approved):
This PR adds functionalities for simulation of the S1 scintillation light needed for detsim city. The code is clear and tested, good job!

invisible_cities/detsim/light_tables.py
```diff
@@ -0,0 +1,103 @@
...
+
+from .. io.dst_io  import load_dst
+
+
+def read_lt(fname, group_name, el_gap=None, active_r=None):
+    """ A helper function to extract dataframes and configuration infro from files.
+    Inputs
+        fname      : string of file name
+        group_name : group name under which table is saved
+        el_gap     : optional set of different el_gap
```
mmkekic: review docstrings; ie should be 
```
param : type 
        explanation
```
`el_gap` and `active_r`  explanation should probably be  'optionally set a different el_gap' or similar

In a returns description `el_gap` is repeated

invisible_cities/detsim/light_tables.py
```diff
@@ -0,0 +1,103 @@
...
+from typing import Optional
+
+from functools import partial
+from .. core                import system_of_units as units
+from .. core.core_functions import binedges_from_bincenters
+
+from .. io.dst_io  import load_dst
+
+
+def read_lt(fname, group_name, el_gap=None, active_r=None):
```
mmkekic: @jmalbos  suggested longer `read_lighttable` name

invisible_cities/detsim/light_tables.py
```diff
@@ -0,0 +1,103 @@
...
+
+    lt = lt.set_index(["x", "y", "z"])
+    nsensors = lt.shape[-1]
+
+    xcenters = np.unique(lt.index.get_level_values('x'))
+    ycenters = np.unique(lt.index.get_level_values('y'))
+    zcenters = np.unique(lt.index.get_level_values('z'))
+
+    # the range for binedges is set to act_r if it is larger than the LT centers
+    range_x = (min(xcenters[0], -act_r), max(xcenters[-1], +act_r)) # centers are ordered
```
mmkekic: Id remove `+` in `max`

invisible_cities/detsim/light_tables.py
```diff
@@ -0,0 +1,103 @@
...
+        had_z = True
+
+    lt = lt.set_index(["x", "y", "z"])
+    nsensors = lt.shape[-1]
+
+    xcenters = np.unique(lt.index.get_level_values('x'))
+    ycenters = np.unique(lt.index.get_level_values('y'))
+    zcenters = np.unique(lt.index.get_level_values('z'))
+
+    # the range for binedges is set to act_r if it is larger than the LT centers
```
mmkekic: Can you expand the explanation, ie 
```
a trick to make the function work with arbitrary act_r,  extending the range if necessary.
Note that binedges_from_bincenters does not accept range smaller than centers range
```
or similar

PR #758, opened 2020-11-19
Bufferization pipe

andLaing:
moves some of the pipe from buffy to a compound component in components so it can be used by the parametrized detsim city (under development).

Also adds a couple of protections to the buffer functions so that they are safe for both cities.

### DIFF
--- a/invisible_cities/detsim/buffer_functions.py
+++ b/invisible_cities/detsim/buffer_functions.py
@@ -1,9 +1,10 @@
 import numpy  as np
 import pandas as pd
 
-from typing import  Callable
-from typing import      List
-from typing import     Tuple
+from typing import Callable
+from typing import     List
+from typing import    Tuple
+from typing import    Union
 
 from .. reco.peak_functions import indices_and_wf_above_threshold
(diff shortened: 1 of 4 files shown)


invisible_cities/detsim/buffer_functions.py
```diff
@@ -141,6 +141,9 @@ def position_signal(triggers   :       List                  ,
             pmt_q  = np.asarray(pmt_charge.tolist())
             sipm_q = np.empty((0,0))\
             if sipm_charge.empty else np.asarray(sipm_charge.tolist())
+        else:
```
mmkekic, line 144: Could you add a test for this, it should be very straightforward. Something similar to 
`test_buffer_calculator` flow but just asserting that the output is the same when using pandas and numpy array as input?

mmkekic (commented):
After a test and some squashing this is ready for approval

mmkekic (approved):
This PR adapts bufferization functions and extract the pipe that creates buffers from the buffy flow allowing it to be reused in the detsim city as well (#691, #700).  Tests are added to show that the functions work both with pandas and numpy inputs. Thanks for the effort!

PR #759, opened 2020-12-10
Update NEWDB

carmenromo:
NEWDB has been updated according to the last sensor calibration. PMT gains, SiPMs gains and pdfs have been added from run 8293.

Pedestal sigma for PMT9 seems to be decreased and it's under study, but it doesn't affect the calibration constants.

Comparison with previous runs are attached.

pmt_comparison

sipmRunDifferencePlots_test


jmbenlloch (approved):
Latest calibration properly added. Approved!

PR #761, opened 2020-12-12
Generate MC timestamp

### LINKED ISSUE #752: Generating MC timestamps
In the past we've tended to set the timestamp required information in the Run tables to zero for all events in MC runs. This is not ideal if we want to be able to study tagging of muons or other time related vetos in the future.

We should probably generate time stamps based on some user configured trigger rate to allow for more data-like vetoing in these cases. This should be simple enough to implement and, as long as the user doesn't try for a rate >~1 kHz leaves ample space for any long duration events that could be split into multiple triggers.

Also could be a nice easy first PR for someone.

MCruces-fz:
With the intention of adding a function that generates timestamps, this push is presented with the following changes:

- Updated `sensor_utils.py` with the create_timestamp function.
- Added `test_create_timestamp_greater_with_greater_arguments` to `sensor_utils_test.py` with the intention of testing whether with larger parameters said timestamp is greater.
- Updated `buffy.conf`, `buffy.py` and `components.py` to implement the function `create_timestamp`.

Fixes #752

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -42,6 +42,7 @@ from .. core   .configure         import          event_range_help
 from .. core   .random_sampling   import              NoiseSampler
 from .. detsim                    import          buffer_functions as  bf
 from .. detsim .sensor_utils      import             trigger_times
+from .. detsim .sensor_utils      import          create_timestamp
 from .. reco                      import           calib_functions as  cf
 from .. reco                      import          sensor_functions as  sf
 from .. reco                      import   calib_sensors_functions as csf
@@ -346,7 +347,8 @@ def check_lengths(*iterables):
 
 def mcsensors_from_file(paths     : List[str],
                         db_file   :      str ,
-                        run_number:      int ) -> Generator:
+                        run_number:      int ,
+                        rate      :    float ) -> Generator:
     """
     Loads the nexus MC sensor information into
     a pandas DataFrame using the IC function
@@ -361,7 +363,13 @@ def mcsensors_from_file(paths     : List[str],
                  Name of detector database to be used
     run_number : int
                  Run number for database
+    rate       : float
+                 Rate value in base unit (ns^-1) to generate timestamps
     """
+
+    if rate == 0:
+        warnings.warn("Zero rate is unphysical, using set period of 1 ms instead", 
+                      category=UserWarning)
 
     pmt_ids  = load_db.DataPMT(db_file, run_number).SensorID
 
@@ -371,11 +379,10 @@ def mcsensors_from_file(paths     : List[str],
                                                        db_file    = db_file   ,
                                                        run_no     = run_number)
 
-        ## MC uses dummy timestamp for now
-        ## Only in case of evt splitting will be non zero
-        timestamp     = 0
-
         for evt in mcinfo_io.get_event_numbers_in_file(file_name):
+
+            timestamp = create_timestamp(evt, rate) / units.ms
+
             try:
                 ## Assumes two types of sensor, all non pmt
                 ## assumed to be sipms. NEW, NEXT100 and DEMOPP safe
--- a/invisible_cities/cities/components_test.py
+++ b/invisible_cities/cities/components_test.py
@@ -211,8 +211,9 @@ def test_copy_mc_info_split_nexus_events(ICDATADIR, config_tmpdir):
 
 
 def test_mcsensors_from_file_fast_returns_empty(ICDATADIR):
+    rate = 0.5
     file_in = os.path.join(ICDATADIR, "nexus_new_kr83m_fast.newformat.sim.h5")
-    sns_gen = mcsensors_from_file([file_in], 'new', -7951)
+    sns_gen = mcsensors_from_file([file_in], 'new', -7951, rate)
     with warns(UserWarning, match='No binning info available.'):
         first_evt = next(sns_gen)
     assert first_evt[ 'pmt_resp'].empty
@@ -221,7 +222,7 @@ def test_mcsensors_from_file_fast_returns_empty(ICDATADIR):
 
 def test_mcsensors_from_file_correct_yield(ICDATADIR):
     evt_no         =    0
-    timestamp      =    0
+    rate           =    0.5
     npmts_hit      =   12
     total_pmthits  = 4303
     nsipms_hit     =  313
@@ -229,13 +230,13 @@ def test_mcsensors_from_file_correct_yield(ICDATADIR):
     keys           = ['event_number', 'timestamp', 'pmt_resp' , 'sipm_resp']
 
     file_in   = os.path.join(ICDATADIR, "nexus_new_kr83m_full.newformat.sim.h5")
-    sns_gen   = mcsensors_from_file([file_in], 'new', -7951)
+    sns_gen   = mcsensors_from_file([file_in], 'new', -7951, rate)
     first_evt = next(sns_gen)
 
     assert set(keys) == set(first_evt.keys())
 
     assert      first_evt['event_number']                 == evt_no
-    assert      first_evt[   'timestamp']                 == timestamp
+    assert      first_evt[   'timestamp']                 >= evt_no / rate
     assert type(first_evt[    'pmt_resp'])                == pd.DataFrame
     assert type(first_evt[   'sipm_resp'])                == pd.DataFrame
     assert  len(first_evt[    'pmt_resp'].index.unique()) == npmts_hit
--- a/invisible_cities/config/buffy.conf
+++ b/invisible_cities/config/buffy.conf
@@ -15,3 +15,5 @@ max_time          =  10 * ms
 buffer_length     = 800 * mus
 pre_trigger       = 100 * mus
 trigger_threshold =   0
+
+rate              =   0.5 * hertz
--- a/invisible_cities/detsim/sensor_utils.py
+++ b/invisible_cities/detsim/sensor_utils.py
@@ -5,10 +5,37 @@ import pandas as pd
 from typing import Callable
 from typing import     List
 from typing import    Tuple
+from typing import    Union
+
+from .. core import system_of_units as units
 
 from .. database.load_db   import            DataPMT
 from .. database.load_db   import           DataSiPM
 from .. io      .mcinfo_io import get_sensor_binning
+
+
+def create_timestamp(event_number: Union[int, float],
+                     rate        :            float ) -> float:
+    """
+    Calculates timestamp for a given Event Number and Rate.
+
+    Parameters
+    ----------
+    event_number : Union[int, float]
+                   ID value of the current event.
+    rate         : float
+                   Value of the rate.
+
+    Returns
+    -------
+    Calculated timestamp : float
+    """
+    try:
+        period = 1 / rate
+        timestamp = abs(event_number * period) + np.random.uniform(0, period)
+    except ZeroDivisionError:
+        timestamp = event_number * 1 * units.ms
+    return timestamp
 
 
 def trigger_times(trigger_indx: List[int] ,
--- a/invisible_cities/detsim/sensor_utils_test.py
+++ b/invisible_cities/detsim/sensor_utils_test.py
@@ -11,6 +11,7 @@ from . sensor_utils import          get_n_sensors
 from . sensor_utils import           sensor_order
 from . sensor_utils import pmt_and_sipm_bin_width
 from . sensor_utils import          trigger_times
+from . sensor_utils import       create_timestamp
 
 
 def test_trigger_times():
@@ -97,3 +98,20 @@ def test_pmt_and_sipm_bin_width(full_sim_file):
     pmt_binwid, sipm_binwid = pmt_and_sipm_bin_width(file_in)
     assert pmt_binwid  == expected_pmtwid
     assert sipm_binwid == expected_sipmwid
+
+
+def test_create_timestamp_greater_with_greater_eventnumber():
+    """
+    Value of timestamp must be always positive and 
+    greater with greater event numbers.
+    """
+    evt_no_1 = 10
+    rate_1   = 0.5
+    evt_no_2 = 100
+    rate_2   = 0.6
+
+    timestamp_1 = create_timestamp(evt_no_1, rate_1)
+    timestamp_2 = create_timestamp(evt_no_2, rate_2)
+    assert abs(timestamp_1) == timestamp_1
+    assert timestamp_1      <  timestamp_2
+    assert abs(timestamp_2) == timestamp_2
--- a/invisible_cities/cities/buffy.py
+++ b/invisible_cities/cities/buffy.py
@@ -45,7 +45,7 @@ from .  components import                  wf_binner
 @city
 def buffy(files_in     , file_out   , compression      , event_range,
           print_mod    , detector_db, run_number       , max_time   ,
-          buffer_length, pre_trigger, trigger_threshold):
+          buffer_length, pre_trigger, trigger_threshold, rate       ):
 
     npmt, nsipm       = get_n_sensors(detector_db, run_number)
     pmt_wid, sipm_wid = pmt_and_sipm_bin_width_safe_(files_in)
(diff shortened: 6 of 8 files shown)


invisible_cities/cities/components_test.py
```diff
@@ -149,15 +151,15 @@ def dummy_city(files_in, file_out, event_range):
 
 def test_hits_and_kdst_from_files(ICDATADIR):
     event_number = 1
-    timestamp    = 0.
+    # timestamp    = 0.
```
andLaing: This test is unrelated so shouldn't be changed

invisible_cities/cities/components_test.py
```diff
@@ -149,15 +151,15 @@ def dummy_city(files_in, file_out, event_range):
...
     num_hits     = 13
     keys = ['hits', 'kdst', 'run_number', 'event_number', 'timestamp']
     file_in     = os.path.join(ICDATADIR    ,  'Kr83_nexus_v5_03_00_ACTIVE_7bar_3evts.HDST.h5')
     generator = hits_and_kdst_from_files([file_in])
     output = next(generator)
     assert set(keys) == set(output.keys())
     assert output['event_number']   == event_number
-    assert output['timestamp']      == timestamp
+    # assert output['timestamp']      == timestamp
```
andLaing: See above

invisible_cities/cities/components_test.py
```diff
@@ -191,33 +193,50 @@ def test_copy_mc_info_repeated_event_numbers(ICDATADIR, config_tmpdir):
...
+    sns_gen = mcsensors_from_file([file_in], 'new', -7951, rate)
     first_evt = next(sns_gen)
     assert first_evt[ 'pmt_resp'].empty
     assert first_evt['sipm_resp'].empty
 
 
 def test_mcsensors_from_file_correct_yield(ICDATADIR):
     evt_no         =    0
-    timestamp      =    0
+    # timestamp      =    0
```
andLaing: Don't leave code commented out. remove the line if you don't need to use it anymore

invisible_cities/cities/components_test.py
```diff
@@ -191,33 +193,50 @@ def test_copy_mc_info_repeated_event_numbers(ICDATADIR, config_tmpdir):
...
     file_in   = os.path.join(ICDATADIR, "nexus_new_kr83m_full.newformat.sim.h5")
-    sns_gen   = mcsensors_from_file([file_in], 'new', -7951)
+    sns_gen   = mcsensors_from_file([file_in], 'new', -7951, rate)
     first_evt = next(sns_gen)
 
     assert set(keys) == set(first_evt.keys())
 
     assert      first_evt['event_number']                 == evt_no
-    assert      first_evt[   'timestamp']                 == timestamp
+    # assert      first_evt[   'timestamp']                 == timestamp
```
andLaing: Rather than removing this I'd do a check that it's `>= evt_no / rate`

invisible_cities/cities/components_test.py
```diff
@@ -191,33 +193,50 @@ def test_copy_mc_info_repeated_event_numbers(ICDATADIR, config_tmpdir):
...
+    rate           =    0.5
+    period         =    1 / rate
+    timestamp      = create_timestamp(event_number, rate)
+
+    file_inp = os.path.join(ICDATADIR, "nexus_new_kr83m_full.newformat.sim.h5")
+    sns_gen   = mcsensors_from_file([file_inp], 'new', -7951)
+    first_evt = next(sns_gen)
+
+    assert first_evt['timestamp'] > timestamp - period
+    assert first_evt['timestamp'] < timestamp + period
```
andLaing: Is this test to be added in? It seems out of date. I think if you include the rate argument in the source definition it should work.

invisible_cities/config/buffy.conf
```diff
@@ -14,4 +14,6 @@ run_number        = -6400
 max_time          =  10 * ms
 buffer_length     = 800 * mus
 pre_trigger       = 100 * mus
-trigger_threshold =   0
\ No newline at end of file
+trigger_threshold =   0
+
+rate              =   0.5
```
andLaing: I think we need units here.

`0.5 * hertz` should be valid, for example

invisible_cities/cities/components.py
```diff
@@ -341,7 +342,8 @@ def check_lengths(*iterables):
 
 def mcsensors_from_file(paths     : List[str],
                         db_file   :      str ,
-                        run_number:      int ) -> Generator:
+                        run_number:      int ,
+                        rate      :    float ) -> Generator:
     """
     Loads the nexus MC sensor information into
     a pandas DataFrame using the IC function
```
andLaing, line 354: Add the new parameter to the docstring.

invisible_cities/cities/components.py
```diff
@@ -366,11 +368,10 @@ def mcsensors_from_file(paths     : List[str],
...
                                                        run_no     = run_number)
 
-        ## MC uses dummy timestamp for now
-        ## Only in case of evt splitting will be non zero
-        timestamp = 0
-
         for evt in mcinfo_io.get_event_numbers_in_file(file_name):
+
+            timestamp = create_timestamp(evt, rate)
+
```
andLaing, line 385: The timestamp that's in the files from data use milliseconds so we should probably be consistent.

Here we'd want to do `timestamp = creat_timestamp(evt, rate) / units.ms`

This solution relies on the correct use of the units in the configuration of the job (I've put a comment to add units to the configuration)

invisible_cities/detsim/sensor_utils.py
```diff
@@ -11,6 +11,20 @@
 from .. io      .mcinfo_io import get_sensor_binning
 
 
+def create_timestamp(event_number: int or float,
```
andLaing: Think the standard way to do this type hint is `Union[int, float]`

andLaing (changes requested):
Looks good. There's a few stylistic things to sort and the use of the `system_of_units` so that we get the correct final units in the output.

invisible_cities/cities/components_test.py
MCruces-fz: Okay, I had assumed that this test would fail and I did not get to check it.

invisible_cities/cities/components_test.py
MCruces-fz: Thank you, good idea

invisible_cities/cities/components_test.py
MCruces-fz: Yes, it is a bad test. It is better to remove it because it is unnecessary.

invisible_cities/cities/components.py
MCruces-fz: Oh, I forgot about that

invisible_cities/cities/components.py
MCruces-fz: Since I'm not used to it, I didn't even think about adding units. I'll keep it in mind next times.

invisible_cities/cities/components.py
```diff
@@ -356,6 +358,8 @@ def mcsensors_from_file(paths     : List[str],
                  Name of detector database to be used
     run_number : int
                  Run number for database
+    rate       : float
+                 Rate value in Hertz to generate timestamps
```
andLaing: So the number (if we use the units correctly) won't be in Hertz. It can be a bit confusing at first but units in this sense convert everything into a base unit. In the case of time we don't define seconds as the base unit (`=1.0`) so Hertz isn't the base unit for rate/frequency either.

The line in the configuration `0.5 * hertz` actually multiplies 0.5 by `1e-9` since the time base unit is `ns`.

I think the least confusing comment might be something like "rate in base unit (ns^-1) but maybe @mmkekic or @jmalbos have an opinion.

invisible_cities/cities/components_test.py
```diff
@@ -191,30 +193,31 @@ def test_copy_mc_info_repeated_event_numbers(ICDATADIR, config_tmpdir):
...
     file_in   = os.path.join(ICDATADIR, "nexus_new_kr83m_full.newformat.sim.h5")
-    sns_gen   = mcsensors_from_file([file_in], 'new', -7951)
+    sns_gen   = mcsensors_from_file([file_in], 'new', -7951, rate)
     first_evt = next(sns_gen)
 
     assert set(keys) == set(first_evt.keys())
 
     assert      first_evt['event_number']                 == evt_no
-    assert      first_evt[   'timestamp']                 == timestamp
+    assert      first_evt[   'timestamp']                 == evt_no / rate
```
andLaing: Maybe it's safer as `>=`?

invisible_cities/detsim/sensor_utils.py
```diff
@@ -11,6 +11,27 @@
 from .. io      .mcinfo_io import get_sensor_binning
 
 
+def create_timestamp(event_number: Union[int, float],
```
andLaing, line 17: I think you'll need to include `Union` from `typing`, I'm getting a fail on the tests at least.

invisible_cities/cities/components.py
MCruces-fz: Right, by reading `system_of_units.py` I have understood the concept about units. Thank you for the explanation

I can wait for other comments and then change the docstring.

invisible_cities/cities/components_test.py
MCruces-fz: True, that was due to lack of attention.

invisible_cities/detsim/sensor_utils.py
MCruces-fz: Yes, I need to
`from typing import Union`

Now the pytest does'nt break.

MCruces-fz:
I am sorry, I will try to amend this mess of commits

andLaing:
We can sort the history a bit once we're happy with the code. No rush.

invisible_cities/detsim/sensor_utils_test.py
```diff
@@ -97,3 +98,20 @@ def test_pmt_and_sipm_bin_width(full_sim_file):
     pmt_binwid, sipm_binwid = pmt_and_sipm_bin_width(file_in)
     assert pmt_binwid  == expected_pmtwid
     assert sipm_binwid == expected_sipmwid
+
+
+def test_create_timestamp_greater_with_greater_arguments():
```
andLaing: I'd maybe change `arguments` to `eventnumber` both in the name of the test and the docstring

andLaing (commented):
This is almost ready. Rebase on the master and fix any conflicts then we can see about tidying the history a bit, though probably we can reduce to two commits: the new function and its test added in one and the adjustments to the other functions and tests to accommodate it in another.

andLaing:
The history seems to have gone a bit mad, how did you do the rebase?

Think we'll have to sort it since at the moment some of the more recently merged commits look like they're part of this PR. I'd be tempted to stash your commits and do a hard rest to the current master followed by reapplying the commits for this PR. Maybe there's a more clever way, @gonzaponte , @mmkekic ?

MCruces-fz:
I did an interactive rebase trying to fix the conflicts, but I did worse. I didn't realize the mess until I did a push, and now I still haven't found a way to reverse it.

I think if I started PR again I would be able to do better, but if we have a better way to do it, great.

MCruces-fz:
There are conflicts in `invisible_cities/cities/components_test.py`
```Python
 def test_mcsensors_from_file_fast_returns_empty(ICDATADIR):
    rate = 0.5
    file_in = os.path.join(ICDATADIR, "nexus_new_kr83m_fast.newformat.sim.h5")
<<<<<<< add_mc_timestamp
    sns_gen = mcsensors_from_file([file_in], 'new', -7951, rate)
    first_evt = next(sns_gen)
=======
    sns_gen = mcsensors_from_file([file_in], 'new', -7951)
    with warns(UserWarning, match='No binning info available.'):
        first_evt = next(sns_gen)
>>>>>>> master
    assert first_evt[ 'pmt_resp'].empty
    assert first_evt['sipm_resp'].empty
```
So that I think that I need to do this:
```Python
def test_mcsensors_from_file_fast_returns_empty(ICDATADIR):
    rate = 0.5
    file_in = os.path.join(ICDATADIR, "nexus_new_kr83m_fast.newformat.sim.h5")
    sns_gen = mcsensors_from_file([file_in], 'new', -7951, rate)
    with warns(UserWarning, match='No binning info available.'):
        first_evt = next(sns_gen)
    assert first_evt[ 'pmt_resp'].empty
    assert first_evt['sipm_resp'].empty
```

Also, there are tests that are now failing in `buffy_test.py` and I don't understand why...

andLaing:
The failing test is `test_buffy_exact_result`. Since you've changed the timestamp that gets saved the comparison with the saved file fails on the events table.

MCruces-fz:
Yes, I've been reading the log and I saw that test fail, but I didn't know what to change. Should I mark that test as `@xfail`?

andLaing:
No, it's quite an important test. Since this is adding to the info in the output it's one of the situations where a change of the 'exact result' file is valid. I think adding another commit with the changed file to show that the test is fixed is the right option.

@mmkekic will keep you right but you need to update the file `nexus_new_kr83m_full.newformat.buffers.h5` so that the events table matches the new expectation.

mmkekic:
I think it is ok to change the exact test file (I think you will have to fix the random seed for that test).
We also realized that `rate =0 ` will end with `ZeroDivisionError` so maybe it should be some protection against it?

invisible_cities/detsim/sensor_utils.py
```diff
@@ -34,7 +34,7 @@ def create_timestamp(event_number: Union[int, float],
         period = 1 / rate
         timestamp = abs(event_number * period) + np.random.uniform(0, period)
     except ZeroDivisionError:
-        ## Maybe a warning here
+        print(f"WARNING: rate = {rate}")
```
andLaing: Now that I see it in place I think that putting a warning here (even a proper one) might be unnecessarily verbose. Maybe it's a warning that should go in buffy itself or in the component so it's only sent once per job for the records.

invisible_cities/cities/components_test.py
```diff
@@ -29,6 +29,8 @@
 
 from .. dataflow   import dataflow as fl
 
+from .. detsim.sensor_utils import create_timestamp
+
```
andLaing: This doesn't get used in this module, I'd remove it.

andLaing:
OK, so we're almost there. Just need to remove the unnecessary import form `components_test.py` and do a bit of history cleaning.

I'd just remove f361b3a18b669131de1f049b1b801590c58fff75 and c5d5e25472f9c4d245078736b393e264d609a85d as they basically undo each other. Then move and squash the other commits so we have the 2 that we talked about.

andLaing (approved):
This PR adds functionality allowing a user to set a timestamp for Monte Carlo data more similar to that present in data. The function is stand alone so could be used to later change the timestamp in existing MC files if required for a specific analysis.

Tests added and updated where necessary and warning added for unphysical values of the new configuration parameter.

Approved pending integrated testing issue (see #768).

andLaing:
How should we proceed re: travis, @jmbenlloch, @mmkekic? The tests all pass but for occasional flakys on my local machine but we need the pass online for merging.

jmbenlloch:
Can you rebase on top of the new master? Then the tests should run in Github automatically.

andLaing:
You seem to have done a merge rather than a rebase... it affects the history in a bit of a weird way.

jacg:
You can easily undo this be resetting your branch (locally) to a commit id you can find in your reflog. Then perfom a rebase, then force-push (*with* lease). I can offer assistance if needed.

MCruces-fz:
I have done that merge because I did not expect that touching the `update` button in the web browser would happen that
I can rebase locally and force push (but _with_ or _without lease_?)

jacg:
> with or without lease?

Force pushing without lease essentially means:

> I don't care what has happened on the server since I last communicated with it, so if there's something there that conflicts with what I'm pushing, just throw that other stuff away and don't bother warning me about it.

Please *don't* do that. *Ever!*

If you are going to force-push, *always* do it with lease. If you think that you need to force push without lease, first discuss it with at least 2 people who know their way around Git.

MCruces-fz:
Okay, thank you, only `git push --force-with-lease`

I will force push with lease (with `ff34bd4` as last commit in my branch) to be sure, and then I will rebase `upstream/master` to get the updates (the commit `31fa0002`)

mmkekic:
Can we merge this?

Issue #762, opened 2020-12-16
mark.feetest warning


jacg:
The tests currently always issue a warning 

```
IC/invisible_cities/sierpe/fee_test.py:123: PytestUnknownMarkWarning: Unknown pytest.mark.feetest - is this a typo?
```

Does anyone know what the purpose of this mark is?

In any case, this should be corrected or removed.

PR #764, opened 2020-12-23
Filter warnings

andLaing:
Suppresses expected warnings in MC writer and tests as well as tidying imports.


jacg (commented):
Please add `closes #762` to the commit message of this commit.

jacg:
I trust that we agree that, wherever possible, rather than suppressing or ignoring the warning, the cause of the warning should be removed.

In some cases (for example, in a test that checks the behaviour of the code in precisely the situation that is being warned about) it's obvious that the warning should simply be ignored. In other cases, this isn't obvious.

Can you give some justifications of the cases where warnings are being ignored rather than their causes being removed? Ideally these would appear in the commit messages of the relevant commit (or a comment at the point of suppression).

jacg:
This one still shows up on Travis:

``` 
invisible_cities/reco/tbl_functions.py:92
  /home/travis/build/next-exp/IC/invisible_cities/reco/tbl_functions.py:92: DeprecationWarning: invalid escape sequence \d
    pattern = re.compile('NEXT_v\d[_\d+]+[_\w]+?_(?P<fnumber>\d+)_[_\w]+?_(?P<nevts>\d+)\..*',
```

I think that preceding the string with `r` (i.e. `r'NEXT_v\d...'`) should do the trick.

invisible_cities/sierpe/fee_test.py
```diff
@@ -120,7 +120,6 @@ def test_spe_to_adc():
     assert 24 < adc_to_pes_lpf < 24.1
 
 @flaky(max_runs=3, min_passes=1)
-@pytest.mark.feetest
```
jacg, line 123: ... this is the commit where `closes #762` belongs.

andLaing:
> I trust that we agree that, wherever possible, rather than suppressing or ignoring the warning, the cause of the warning should be removed.

Yes, I think they are justified although in some cases there might be alternatives. The `OptimizeWarning` for example could maybe be fixed but since there's a random generator involved maybe it would just be fixed for certain machines. I've added to the commit messages and put comments where possible to explain the motivation.

> This one still shows up on Travis:

Should be we be worried that it only shows up on Travis? Especially since it's a `DeprecationWarning`, doesn't that mean that we have some version differences between what Travis uses and what's downloaded by the scripts? Seem unstable.

andLaing:
Now two new warnings hace appeared, there might be something going on behind the scenes

jacg:
Apologies for the silence. I am afraid that I will be unavailable for some time.

invisible_cities/cities/components_test.py
```diff
@@ -209,11 +211,12 @@ def test_copy_mc_info_split_nexus_events(ICDATADIR, config_tmpdir):
             assert_tables_equality(got, expected)
 
 
-
 def test_mcsensors_from_file_fast_returns_empty(ICDATADIR):
     file_in = os.path.join(ICDATADIR, "nexus_new_kr83m_fast.newformat.sim.h5")
     sns_gen = mcsensors_from_file([file_in], 'new', -7951)
-    first_evt = next(sns_gen)
+    with warnings.catch_warnings():
```
mmkekic: Here you do expect `No binning info available.` so maybe you can use 
```
with pytest.warns(UserWarning, match='No binning info available.'):
        first_evt = next(sns_gen)
```

invisible_cities/io/dst_io.py
```diff
@@ -109,7 +109,7 @@ def df_writer(h5out              : tb.file.File ,
 
     data_types = table.dtype
     if len(arr) == 0:
-        warnings.warn(f'dataframe is empty', UserWarning)
+        warnings.warn('dataframe is empty', UserWarning)
```
mmkekic: I know that this is not strictly related to this PR but I wonder if we really need this warnings. For example, in pandas `to_hdf` no warning is raised if dataframe is empty, so maybe we should just keep the same behavior, ie removing the warning completely.  Especially since it is almost never guaranteed to have non-empty df in our flows...

mmkekic (commented):
Probably should also look into `filterwarnings` of pytest.ini file. One (import error of collections) is fixed in 06f09f98d860df21f246ae129ec412afa824e4a6 so maybe you can just cherry-pick it.

jacg:
@andLaing You wrote 'closes issue #762' in the commit message, but I think that the presence of 'issue' between 'closes' and '#762' interferes with the pattern that GitHub recognizes: https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

andLaing:
Ah, ok, it seemed more natural to specify but I should have checked the standard. My bad, I'll edit the commit message now

jacg:
A fresh installation on a pristine Debian 10 with both Python 3.7 and 3.8 passes all tests with zero warnings.

The latest Travis run produces this single warning:

```
/home/travis/miniconda/envs/IC-3.7-2020-06-16/lib/python3.7/site-packages/pandas/core/dtypes/missing.py:478
  /home/travis/miniconda/envs/IC-3.7-2020-06-16/lib/python3.7/site-packages/pandas/core/dtypes/missing.py:478: RuntimeWarning: invalid value encountered in equal
    return ((left == right) | (isna(left) & isna(right))).all()
```

Anyone recognize this?

(The Mac OS run fails, but we'll deal with that in a separate PR.)

jacg:
Some of the commit messages contain justifications/explanations of why warnings are suppressed/ignored. These should probably be added as comments in the code itself.

jacg:
Minor niggle: in 'Change string to r-string in tbl_functions_test' the commit message contains 'Suppress DeprecationWarning', but the warning isn't being suppressed (which would require justification) bit it's being heeded (which requires no further explanation).

jacg:
We should also do a review of all the skipped and xfail tests. Probably in a separate PR, unless it can be done quickly.

jacg:
Latest Travis run shows no warnings.

Add justifications (from commit messages) as comments to source code, and I'm ready to approve.

invisible_cities/io/dst_io.py
jacg: Anybody care to share an opinion on this?

Does anyone ever notice or appreciate this warning appearing? Can anyone remember an instance when this warning helped them? If not, it should probably go.

andLaing: Marija and I discussed it and we decided to remove it in the final version of this PR. That's why it shows as outdated above.

jacg:
> Should be we be worried that it only shows up on Travis? Especially since it's a `DeprecationWarning`, doesn't that mean that we have some version differences between what Travis uses and what's downloaded by the scripts?

In theory, our `manage.sh` (or Nix, or whatever we choose to use) should give us identical environments on test, production and development machines. It might be something to do with the Travis test profile asking Hypothesis to produce more cases, and thus dodgy cases (which raise warnings) being found more frequently on Travis. In any case, I agree that it's probably flaky, and that we can probably live with it appearing every now and then.

invisible_cities/io/dst_io.py
jacg: I think that it shows as outdated, because you have rebased, and the commit id has changed. In my latest checkout of the branch, I can still see the warning being raised (strangely enough, with the superfluous 'f' which was removed here!)

andLaing: Strange. It's changed in the first commit of the current history and I can't find anywhere in the other commits where it could have been accidentally reintroduced.

jacg: Apologies. I failed to notice that a Git LFS error prevented me from looking at the right branch. I confirm that it's gone from the current version.

jacg (approved):
All stable warnings have been eliminated, either by addressing the issue they point out, or, where appropriate, ignoring them explicitly. In the latter cases, justifications have been added to both the commit messages and as comments to the code.

This eliminates the deluge of warnings that accompanied each run of the test suite, causing all warnings to be ignored by any humans looking at the test output.

We should go through a similar exercise for skipped and xfailing tests in a separate PR.

PR #765, opened 2021-01-12
Update NEWDB

carmenromo:
NEWDB has been updated according to the last sensor calibration. PMT gains, SiPMs gains and pdfs have been added from run 8369.

Comparison with previous runs are attached.

pmtRunDifferencePlots

sipmRunDifferencePlots_test


jmbenlloch (approved):

PR #766, opened 2021-01-13
Github action to check database status after calibration updates

jmbenlloch:
Whenever a new calibration is done some tables are changed in our MySQL server, that are later reflected in the SQLite file uploaded to this repository (examples in #765, #759, #754, etc.). Usually I'm the one responsible of reviewing those PRs. Besides checking all tests passed in Travis there is an extra step to ensure everything is ok and there was no human error in the process of creating the `sqlite` copy. This PR shows a way to do it automatically.

The calibration in the database affects the tables `ChannelGain`, `ChannelMask` and `SipmNoisePDF`. All of them are indexed using a `MinRun` and a `MaxRun`, delimiting the range of runs to which those values are applicable. The set of (`MinRun`, `MaxRun`) for each of those tables must be the same in our MySQL server and in the `sqlite3` file included in the corresponding PR.

So far, I have been doing that step manually. This PR includes a Github action to do that check automatically. It will be triggered only for PR's modifying some `*.sqlite3` file. This is what it does:

- Detect which is the updated database by parsing the name of the updated file. So it is useful for NEW, DEMOPP, etc.
- Retrieve the (`MinRun`, `MaxRun`) set for the relevant tables and checks the SQLite file contains the same thing both for SiPMs and PMTs (separately).
- It does NOT check all the values (too much data), but only that there is consistency between the latest calibration run number in both databases.

As illustration I have already included the code in this PR into [my fork's master](https://github.com/jmbenlloch/IC-1) and created several PR's two show you how it works:

- [A PR updating DEMOPP database](https://github.com/jmbenlloch/IC-1/pull/3) without including the latest calibration in the MySQL server.
  - This PR passed all Travis tests but [did NOT passed the new test](https://github.com/jmbenlloch/IC-1/pull/3/checks?check_run_id=1696388822) done by the Github action included in this PR.
image

- [Another PR with the same idea for the NEW database](https://github.com/jmbenlloch/IC-1/pull/4).
  - The code autodetects the proper database to check and [shows the error](https://github.com/jmbenlloch/IC-1/pull/4/checks?check_run_id=1696737387).

- [One last PR modifying the DEMOPP sqlite](https://github.com/jmbenlloch/IC-1/pull/5) file but keeping the latest calibration in place.
  - [All tests passed](https://github.com/jmbenlloch/IC-1/pull/5/checks?check_run_id=1696446530) in this case.
image

I think this automation will simplify the reviewing of calibration PR's (which are common) and can also be useful as a proof of concept for tests/tasks that we want to run occasionally and do not fit well in the pytest scheme. This can be the first Github action included in IC, but there may be more in the future.

PS: In the sample images you can see a check called `Python package / build (3.7) (pull_request)`, I have updated that name to `Calibration check / build (3.7) (pull_request)`

### DIFF
--- a/invisible_cities/database/check_gain_update.py
+++ b/invisible_cities/database/check_gain_update.py
@@ -1,0 +1,54 @@
+import sys
+import sqlite3
+import pymysql
+pymysql.install_as_MySQLdb()
+import numpy as np
+
+# Absolute imports to allow usage as standalone program:
+# python invisible_cities/database/check_gain_update.py
+from invisible_cities.database.db_connection import connect_sqlite
+from invisible_cities.database.db_connection import connect_mysql
+
+
+def check_minrun_maxrun(table, filters):
+    sql = f'select distinct(Minrun), maxrun from {table} where {filters} ORDER BY MinRun,MaxRun;'
+
+    cursor_sqlite.execute(sql)
+    cursor_mysql .execute(sql)
+
+    data_mysql  = cursor_mysql .fetchall()
+    data_sqlite = cursor_sqlite.fetchall()
+
+    # Convert to np float64 arrays to replace None's vy np.nan's
+    array_mysql  = np.array(data_mysql , dtype=np.float64)
+    array_sqlite = np.array(data_sqlite, dtype=np.float64)
+
+    np.testing.assert_allclose(array_mysql, array_sqlite)
+
+    try:
+        latest_run = data_sqlite[-1][0]
+    except IndexError:
+        latest_run = None
+
+    return latest_run
+
+
+# Get DB name from the list of updated files in the commit
+dbfile = next(filter(lambda f: f.endswith('sqlite3'), sys.argv))
+dbname = dbfile.split('.')[-2]
+
+# Connect to MySQL server and open SQLite copy
+_, cursor_sqlite = connect_sqlite(dbfile)
+_, cursor_mysql  = connect_mysql (dbname)
+
+# Check tables
+tables = ['ChannelGain', 'ChannelMask', 'SipmNoisePDF']
+pmts  = 'SensorID < 100'
+sipms = 'SensorID > 100'
+
+# Print latest run available in each case
+for table in tables:
+    latest_run_sipms = check_minrun_maxrun(table, sipms)
+    latest_run_pmts  = check_minrun_maxrun(table, pmts)
+    print(f'{table}: latest SiPM run {latest_run_sipms}')
+    print(f'{table}: latest PMT  run {latest_run_pmts}')
--- a/.github/workflows/check_db_calibration.yml
+++ b/.github/workflows/check_db_calibration.yml
@@ -1,0 +1,42 @@
+# This workflow will check the consistency between MySQL and SQLite databases
+
+name: Calibration check
+
+on:
+  push:
+    branches: [ master ]
+    paths:
+        - '**.sqlite3'
+  pull_request:
+    branches: [ master ]
+    paths:
+        - '**.sqlite3'
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.7]
+
+    steps:
+    - uses: actions/checkout@v2
+    - id: files
+      uses: jitterbit/get-changed-files@v1
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Get LFS files
+      run: |
+        git lfs pull
+    - name: Install dependencies
+      run: |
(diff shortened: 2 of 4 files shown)


carmenromo:
Amazing @jmbenlloch!!! Thank you!!!

> Retrieve the (MinRun, MaxRun) set for the relevant tables and checks the SQLite file contains the same thing both for SiPMs and PMTs (separately).

In the case of DEMO we do not calibrate always both type of sensors. Here you mean that they are independent, right? Or does it check that the run numbers match for SiPMs and PMTs?

jmbenlloch:
> In the case of DEMO we do not calibrate always both type of sensors. Here you mean that they are independent, right? Or does it check that the run numbers match for SiPMs and PMTs?

The code will extract from the database the (`MinRun`, `MaxRun`) set first for the SiPMs (`SensorID > 100`) and then for the PMTs (`SensorID < 100`). In this way there shouldn't be any problem if you have updated both SiPMs and PMTs or only one of them.

invisible_cities/database/check_gain_update.py
```diff
@@ -0,0 +1,65 @@
...
+
+    try:
+        latest_run = data_sqlite[-1][0]
+    except IndexError:
+        latest_run = None
+
+    return latest_run
+
+
+if __name__ == '__main__':
```
jacg: This guard serves no purpose. Please remove.

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,40 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
```
jacg: Is this comment accurate? Specifically, is there any linting going on here?

invisible_cities/database/check_gain_update.py
```diff
@@ -0,0 +1,65 @@
...
+    conn_sqlite   = sqlite3.connect(dbfile)
+    cursor_sqlite = conn_sqlite.cursor()
+    return cursor_sqlite
+
+
+def connect_mysql(dbname):
+    conn_mysql  = pymysql.connect(host="neutrinos1.ific.uv.es",
+                                  user='nextreader',passwd='readonly', db=dbname)
+    cursor_mysql  = conn_mysql .cursor()
+    return cursor_mysql
```
jacg: Do these two functions not already exist somewhere else in our code?

invisible_cities/database/check_gain_update.py
jmbenlloch: Done.

.github/workflows/python-package.yml
jmbenlloch: You are right. That comment was from the template I took for the yml file. I have replaced it with an actual description of what is done.

invisible_cities/database/check_gain_update.py
jmbenlloch: Not really. [Here](https://github.com/next-exp/IC/blob/e3383b08c72e7aa6b9f579a75b824b0ffa314d9e/invisible_cities/database/download.py#L61) two connections with their cursors are created but they are not in a separate function.

I will see if I can manage to refactor those lines and call them from the Github action.

jmbenlloch: I have refactored those functions and tested again the code in my fork. The new version works well too.
One minor detail: in the files `invisible_cities/database/download.py` and `invisible_cities/database/check_gain_update.py` I have not used relative imports because we want to be able to run those scripts independently as:

 `python invisible_cities/database/download.py`
 `python invisible_cities/database/check_gain_update.py`

jacg: It's probably a good idea to explain the inconsistent-with-the-rest-of-IC choice of absolute imports, in a comment in the code itself, lest someone wonder later on, and waste time on refactoring (and breaking) it.

jacg:
I doubt I'll have much more to say on this, but I think that someone who is familiar with the task that is being automated, should have a look at exactly what is being verified, whether it makes sense, and whether anything has been overlooked.

AFAIAC: LGTM.

invisible_cities/database/check_gain_update.py
jmbenlloch: You are right, I have added comments about that to both files.

jmbenlloch:
> I doubt I'll have much more to say on this, but I think that someone who is familiar with the task that is being automated, should have a look at exactly what is being verified, whether it makes sense, and whether anything has been overlooked.

Thanks @jacg. I have squashed all the commits. Maybe @carmenromo can have a look a give the final review.

carmenromo (approved):
I suggested to @jmbenlloch to make another test to check that there are no gaps between the run numbers. However, given the large number of required exceptions we agreed on leaving it for the future, after reviewing the database itself.

I have reviewed the function that checks the gain update and it does its work, so I approve!

PR #767, opened 2021-01-25
Update PMT xpos in DEMOPPDB

carmenromo:
DEMO++ database has been updated with the x values of the PMTs. The three of them have been changed adding a `-`:

- `PMT2` --> `x_old = 41.14mm`, `x_new = -41.14mm`

- `PMT4` --> `x_old = -69.62mm`, `x_new = 69.62mm`

- `PMT3` --> `x_old = 28.47mm`, `x_new = -28.47mm`

PR #768, opened 2021-01-28
Replace Travis with Github Actions

jmbenlloch:
Some of the latests PRs (#767, #761) are not running the tests in Travis. I have checked our travis account and it seems that there is a time limit for free accounts in travis-ci.com. Each free accounts gets 10000 "credits" but we have used all of them:

image

An error message appears in our travis account: 

>Builds have been temporarily disabled for private and public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance.

Since Github Actions does not impose any time limit for free accounts, I think we should move now to github and stop using Travis. This PR includes a first idea on how to implement our tests using a Github Action.

There are several issues:

- Tests are failing for different reasons: https://github.com/jmbenlloch/IC-1/actions/runs/517922412
  -  In linux, sometimes there is an [error in the database tests](https://github.com/jmbenlloch/IC-1/runs/1784584691?check_suite_focus=true). Ocasionally, there is a timeout error when connecting to our MySQL server at IFIC. I don't know the reason yet, especially because it is not deterministic. Here is a sample where it has run [successfully](https://github.com/jmbenlloch/IC-1/runs/1777615906?check_suite_focus=true). It might be related to the OS version used in the container that runs the jobs. Apparently `ubuntu-latest` is transitioning from 18.04 to 20.04. I'll try later to fix the version to 18.04 to check it.
  - In MacOS I have never seen a successful run. There are [some errors related to hypothesis](https://github.com/jmbenlloch/IC-1/runs/1784584730?check_suite_focus=true) and the timing of the tests. I do not know now how to address this problem.

DO NOT MERGE yet until we sort out the details. We need a reliable system to substitute Travis and for some reason we are not quite there yet. I think this is somehow urgent because, currently, we do not have any working CI system.

### DIFF
--- a/.github/workflows/test_suite.yml
+++ b/.github/workflows/test_suite.yml
@@ -1,0 +1,36 @@
+# This workflow will install IC and run all the tests with pytest
+
+name: Test suite
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  build:
+
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.7]
+        #platform: [ubuntu-18.04, macos-latest]
+        platform: [ubuntu-18.04]
+    runs-on: ${{ matrix.platform }}
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Get LFS files
+      run: git lfs pull
+    # - name: Fix Conda permissions on macOS
+    #   run: sudo chown -R $UID $CONDA
+    #   if: runner.os == 'macOS'
+    - name: Install IC
+      run: |
+        source $CONDA/etc/profile.d/conda.sh
+        source manage.sh work_in_python_version_no_tests ${{ matrix.python-version }}
+    - name: Run tests
+      run: |
+        source $CONDA/etc/profile.d/conda.sh
+        source manage.sh work_in_python_version_no_tests ${{ matrix.python-version }}
+        HYPOTHESIS_PROFILE=travis-ci bash manage.sh run_tests_par
+


jmbenlloch:
I have uploaded a newer version of the action excluding the MacOS runner since I do not know why it fails. I don't think it is very important since we had already marked MacOS as an allowed-failure in Travis.

Regarding Ubuntu I have fixed the version to 18.04 and after rerunning several times I have not seen the MySQL timeout error again. Since we have several PRs already waiting I would approve this PR and leave for the future possible improvements.

mmkekic:
From a fast scroll of failed MacOS run it seems all of them fail due to `Unreliable test timings!`. This is something that happened on Ubuntu as well form time to time in Travis, and is related to the Hypothesis settings `travis-ci` (in `conftest.py`) that sets the deadline to 300ms. Do you think that lifting the limit to something like 1000ms makes sense? Or we should try to remove the limits of the slow tests? 

In any case the problem seems not to be related to this PR so I believe this can be approved.

I am a bit confused by `Some checks haven't completed yet`, can it be that you added GitHub Actions but havent removed Travis yet?

jmbenlloch:
> From a fast scroll of failed MacOS run it seems all of them fail due to Unreliable test timings!. This is something that happened on Ubuntu as well form time to time in Travis, and is related to the Hypothesis settings travis-ci (in conftest.py) that sets the deadline to 300ms. Do you think that lifting the limit to something like 1000ms makes sense? Or we should try to remove the limits of the slow tests?

What bothers me is that it ALWAYS happens in the MacOS runner. Maybe they are using different machines to run Linux and Mac and the latter for some reason are slower. If we want to run the tests on MacOS it could be a good idea to increase the limit, although I do not know how much time would take the tests the end in that case.

> In any case the problem seems not to be related to this PR so I believe this can be approved.

Agree

> I am a bit confused by Some checks haven't completed yet, can it be that you added GitHub Actions but havent removed Travis yet?

You are right, I have not changed it yet. I'll do it now.

mmkekic (approved):
This PR allows the switch from Travis CI to GitHub Actions CI, needed after Travis changed the free plan policy.
Thanks for the effort and rapidity!

mmkekic:
Since you have left commented the lines about MacOS installation maybe you can squash the history?

jmbenlloch:
> Since you have left commented the lines about MacOS installation maybe you can squash the history?

Done. I'll merge the PR after the tests pass.

PR #769, opened 2021-02-03
Detsim city flow

gondiaz:
This is the last of the series of PRs #757 #756 #695  #694.

It contains detsim city flow module that performs a fast detector simulation.

From MC hits, it simulates inonization electron creation, drift and diffusion. Then it simulates the photon propagation based on previously computed light-tables, which depend on detector geometry. The final output are the signal waveforms at sensors (without electronic simulation).

### DIFF
--- a/invisible_cities/cities/detsim.py
+++ b/invisible_cities/cities/detsim.py
@@ -1,0 +1,225 @@
+"""
+-----------------------------------------------------------------------
+                              Detsim
+-----------------------------------------------------------------------
+From Detector Simulation. This city reads energy deposits (hits) and simulates
+S1 and S2 signals. This is accomplished using light tables to compute the signal
+generated in the sensors. The input of this city is nexus output containing hits.
+This city outputs:
+    - pmtrd : PMT  waveforms
+    - sipmrd: SIPM waveforms
+    - MC  info
+    - Run info
+    - Filters
+"""
+
+import os
+import numpy  as np
+import tables as tb
+
+from . components import city
+from . components import print_every
+from . components import copy_mc_info
+from . components import collect
+from . components import calculate_and_save_buffers
+from . components import MC_hits_from_files
+
+from .. reco     import tbl_functions as tbl
+from .. database import load_db       as db
+from .. dataflow import dataflow      as fl
+
+from .. io.event_filter_io  import event_filter_writer
+
+from .. detsim.simulate_electrons import generate_ionization_electrons
+from .. detsim.simulate_electrons import drift_electrons
+from .. detsim.simulate_electrons import diffuse_electrons
+from .. detsim.light_tables_c     import LT_SiPM
+from .. detsim.light_tables_c     import LT_PMT
+from .. detsim.s2_waveforms_c     import create_wfs
+from .. detsim.detsim_waveforms   import s1_waveforms_creator
+
+
+def hits_selector(active_only: bool=True):
+    """
+    Filtering function that selects hits. (see :active_only: description)
+
+    Parameters:
+        :active_only: bool
+            if True, returns hits in ACTIVE
+            if False, returns hits in ACTIVE and BUFFER
+    Returns:
+        :select_hits: Callable
+            function that select the hits depending on :active_only: parameter
+    """
+    def select_hits(x, y, z, energy, time, label):
+        sel = (label == "ACTIVE")
+        if not active_only:
+            sel =  sel | (label == "BUFFER")
+        return x[sel], y[sel], z[sel], energy[sel], time[sel], label[sel]
+    return select_hits
+
+
+def ielectron_simulator(*, wi: float, fano_factor: float, lifetime: float,
+                        transverse_diffusion: float, longitudinal_diffusion: float, drift_velocity:float,
+                        el_gain: float, conde_policarpo_factor: float):
+    """
+    Function that simulates electron creation, drift, diffusion and photon generation at the EL
+
+    Parameters: floats
+        parameter names are self-descriptive.
+    Returns:
+        :simulate_ielectrons:
+            function that returns the positions emission times and number of photons at the EL
+    """
+    def simulate_ielectrons(x, y, z, time, energy):
+        nelectrons = generate_ionization_electrons(energy, wi, fano_factor)
+        nelectrons = drift_electrons(z, nelectrons, lifetime, drift_velocity)
+        dx, dy, dz = diffuse_electrons(x, y, z, nelectrons, transverse_diffusion, longitudinal_diffusion)
+        dtimes = dz/drift_velocity + np.repeat(time, nelectrons)
+        nphotons = np.random.normal(el_gain, np.sqrt(el_gain * conde_policarpo_factor), size=nelectrons.sum())
+        nphotons = np.round(nphotons).astype(np.int32)
+        return dx, dy, dz, dtimes, nphotons
+    return simulate_ielectrons
+
+
+def buffer_times_and_length_getter(pmt_width, sipm_width, el_gap, el_dv, max_length):
+    """
+    Auxiliar function that computes the signal absolute starting-time and an estimated buffer_length
+    """
+    max_sensor_bin = max(pmt_width, sipm_width)
+    def get_buffer_times_and_length(time, times_ph):
+        start_time = np.floor(min(time) / max_sensor_bin) * max_sensor_bin
+        el_traverse_time = el_gap / el_dv
+        end_time   = np.ceil((max(times_ph) + el_traverse_time)/max_sensor_bin) * max_sensor_bin
+        buffer_length = min(max_length, end_time-start_time)
+        return start_time, buffer_length
+    return get_buffer_times_and_length
+
+
+def s2_waveform_creator(sns_bin_width, LT, el_drift_velocity):
+    """
+    Same function as create_wfs in module detsim.s2_waveforms_c with poissonization.
+    See description of the refered function for more details.
+    """
+    def create_s2_waveform(xs, ys, ts, phs, tmin, buffer_length):
+        waveforms = create_wfs(xs, ys, ts, phs, LT, el_drift_velocity, sns_bin_width, buffer_length, tmin)
+        return np.random.poisson(waveforms)
+    return create_s2_waveform
+
+
+def bin_edges_getter(pmt_width, sipm_width):
+    """
+    Auxiliar function that returns the waveform bin edges
+    """
+    def get_bin_edges(pmt_wfs, sipm_wfs):
+        pmt_bins  = np.arange(0, pmt_wfs .shape[1]) * pmt_width
+        sipm_bins = np.arange(0, sipm_wfs.shape[1]) * sipm_width
+        return pmt_bins, sipm_bins
+    return get_bin_edges
+
+
+@city
+def detsim(*, files_in, file_out, event_range, print_mod, compression,
+           detector_db, run_number, s1_lighttable, s2_lighttable, sipm_psf,
+           buffer_params, physics_params):
+
+    physics_params_ = physics_params.copy()
+
+    ws    = physics_params_.pop("ws")
+    el_dv = physics_params_.pop("el_drift_velocity")
+
+    # derived parameters
+    datapmt  = db.DataPMT (detector_db, run_number)
+    datasipm = db.DataSiPM(detector_db, run_number)
+    lt_pmt   = LT_PMT (fname=os.path.expandvars(s2_lighttable))
+    lt_sipm  = LT_SiPM(fname=os.path.expandvars(sipm_psf), sipm_database=datasipm)
+    el_gap   = lt_sipm.el_gap_width
+
+    select_s1_candidate_hits = fl.map(hits_selector(False),
+                                item = ('x', 'y', 'z', 'energy', 'time', 'label'))
+
+    select_active_hits = fl.map(hits_selector(True),
+                                args = ('x', 'y', 'z', 'energy', 'time', 'label'),
+                                out = ('x_a', 'y_a', 'z_a', 'energy_a', 'time_a', 'labels_a'))
+
+    filter_events_no_active_hits = fl.map(lambda x:np.any(x),
+                                          args= 'energy_a',
+                                          out = 'passed')
+    events_passed_active_hits = fl.count_filter(bool, args='passed')
+
+    simulate_electrons = fl.map(ielectron_simulator(**physics_params_),
+                                args = ('x_a', 'y_a', 'z_a', 'time_a', 'energy_a'),
+                                out  = ('x_ph', 'y_ph', 'z_ph', 'times_ph', 'nphotons'))
+
+    get_buffer_info = buffer_times_and_length_getter(buffer_params["pmt_width"],
+                                                     buffer_params["sipm_width"],
+                                                     el_gap, el_dv,
+                                                     buffer_params["max_time"])
+    get_buffer_times_and_length = fl.map(get_buffer_info,
+                                         args = ('time', 'times_ph'),
+                                         out = ('tmin', 'buffer_length'))
+
+    create_pmt_s1_waveforms = fl.map(s1_waveforms_creator(s1_lighttable, ws, buffer_params["pmt_width"]),
+                                     args = ('x_a', 'y_a', 'z_a', 'time_a', 'energy_a', 'tmin', 'buffer_length'),
+                                     out = 's1_pmt_waveforms')
+
+    create_pmt_s2_waveforms = fl.map(s2_waveform_creator(buffer_params["pmt_width"], lt_pmt, el_dv),
+                                     args = ('x_ph', 'y_ph', 'times_ph', 'nphotons', 'tmin', 'buffer_length'),
+                                     out = 's2_pmt_waveforms')
+
+    sum_pmt_waveforms = fl.map(lambda x, y : x+y,
+                               args = ('s1_pmt_waveforms', 's2_pmt_waveforms'),
+                               out = 'pmt_bin_wfs')
+
+    create_pmt_waveforms = fl.pipe(create_pmt_s1_waveforms, create_pmt_s2_waveforms, sum_pmt_waveforms)
+
+    create_sipm_waveforms = fl.map(s2_waveform_creator(buffer_params["sipm_width"], lt_sipm, el_dv),
+                                   args = ('x_ph', 'y_ph', 'times_ph', 'nphotons', 'tmin', 'buffer_length'),
+                                   out = 'sipm_bin_wfs')
+
+    get_bin_edges  = fl.map(bin_edges_getter(buffer_params["pmt_width"], buffer_params["sipm_width"]),
+                            args = ('pmt_bin_wfs', 'sipm_bin_wfs'),
+                            out = ('pmt_bins', 'sipm_bins'))
+
+    event_count_in = fl.spy_count()
+    evtnum_collect = collect()
+
+    with tb.open_file(file_out, "w", filters = tbl.filters(compression)) as h5out:
+        buffer_calculation = calculate_and_save_buffers(buffer_params["length"]     ,
+                                                        buffer_params["max_time"]   ,
+                                                        buffer_params["pre_trigger"],
+                                                        buffer_params["pmt_width"]  ,
+                                                        buffer_params["sipm_width"] ,
+                                                        buffer_params["trigger_thr"],
+                                                        h5out        ,
+                                                        run_number   ,
+                                                        len(datapmt) ,
+                                                        len(datasipm),
+                                                        int(buffer_params["length"] / buffer_params["pmt_width"]),
+                                                        int(buffer_params["length"] / buffer_params["sipm_width"]))
+
+        write_nohits_filter = fl.sink(event_filter_writer(h5out, "active_hits")   , args=("event_number", "passed"))
+        result = fl.push(source= MC_hits_from_files(files_in),
+                         pipe  = fl.pipe(fl.slice(*event_range, close_all=True),
+                                         event_count_in.spy    ,
+                                         print_every(print_mod),
+                                         select_s1_candidate_hits,
+                                         select_active_hits,
+                                         filter_events_no_active_hits,
+                                         fl.branch(write_nohits_filter),
+                                         events_passed_active_hits.filter,
+                                         simulate_electrons,
+                                         get_buffer_times_and_length,
+                                         create_pmt_waveforms,
+                                         create_sipm_waveforms,
+                                         get_bin_edges,
+                                         fl.branch("event_number"     ,
+                                                   evtnum_collect.sink),
+                                         buffer_calculation),
+                         result = dict(events_in     = event_count_in.future,
+                                       evtnum_list   = evtnum_collect.future))
+
+        copy_mc_info(files_in, h5out, result.evtnum_list,
+                     detector_db, run_number)
+
+        return result
--- a/invisible_cities/cities/detsim_test.py
+++ b/invisible_cities/cities/detsim_test.py
@@ -1,0 +1,221 @@
+import os
+import shutil
+import numpy  as np
+import pandas as pd
+import tables as tb
+
+from .. cities.detsim  import detsim
+from .. core           import system_of_units as units
+from .. core.configure import configure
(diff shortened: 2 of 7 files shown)


invisible_cities/cities/detsim.py
```diff
@@ -0,0 +1,216 @@
...
+    def get_bin_edges(pmt_wfs, sipm_wfs):
+        pmt_bins  = np.arange(0, pmt_wfs .shape[1]) * wf_pmt_bin_width
+        sipm_bins = np.arange(0, sipm_wfs.shape[1]) * wf_sipm_bin_width
+        return pmt_bins, sipm_bins
+    return get_bin_edges
+
+
+# @profile
+@city
+def detsim(files_in, file_out, event_range, detector_db, run_number, s1_lighttable, s2_lighttable, sipm_psf,
```
mmkekic: you should probably group input parameters in dictionaries

invisible_cities/cities/detsim.py
```diff
@@ -0,0 +1,216 @@
...
+    """
+    def select_hits(x, y, z, energy, time, label):
+        sel = (label == "ACTIVE")
+        if not active_only:
+            sel =  sel | (label == "BUFFER")
+        return x[sel], y[sel], z[sel], energy[sel], time[sel], label[sel]
+    return select_hits
+
+
+def ielectron_simulator(wi: float, fano_factor: float, lifetime: float,
```
mmkekic: Maybe make them keyword-only arguments to avoid possible misuse?

invisible_cities/cities/detsim.py
```diff
@@ -0,0 +1,216 @@
...
+    Function that simulates electron creation, drift, diffusion and photon generation at the EL
+
+    Parameters: floats
+        parameter names are self-descriptive.
+    Returns:
+        :simulate_ielectrons:
+            function that returns the positions emission times and number of photons at the EL
+    """
+    def simulate_ielectrons(x, y, z, time, energy):
+        electrons = generate_ionization_electrons(energy, wi, fano_factor)
```
mmkekic: How about `num_electrons`, or `nelectrons` so it is clear you are getting the number of electrons?

invisible_cities/cities/detsim.py
```diff
@@ -0,0 +1,216 @@
...
+    def get_buffer_times_and_length(time, times_ph):
+        start_time = np.floor(min(time) / max_sensor_bin) * max_sensor_bin
+        el_traverse_time = el_gap / el_dv
+        end_time   = np.ceil((max(times_ph) + el_traverse_time)/max_sensor_bin) * max_sensor_bin
+        buffer_length = min(max_length, end_time-start_time)
+        return start_time, buffer_length
+    return get_buffer_times_and_length
+
+
+def s2_waveform_creator (sns_bin_width, LT, EL_drift_velocity):
```
mmkekic: use lowercase for variables (uppercase for classes instances are ok)

invisible_cities/cities/detsim.py
```diff
@@ -0,0 +1,216 @@
...
+    Auxiliar function that returns the waveform bin edges
+    """
+    def get_bin_edges(pmt_wfs, sipm_wfs):
+        pmt_bins  = np.arange(0, pmt_wfs .shape[1]) * wf_pmt_bin_width
+        sipm_bins = np.arange(0, sipm_wfs.shape[1]) * wf_sipm_bin_width
+        return pmt_bins, sipm_bins
+    return get_bin_edges
+
+
+# @profile
```
mmkekic: remove

mmkekic (commented):
First round of mostly style comments

invisible_cities/cities/detsim.py
jacg: If we are used to seeing it upper case in our standard notation outside of source code, I would recommend sticking to uppercase even in source code.

But we do follow PEP-8 on *no-spaces-before-parentheses*, unless there's an overriding reason (usually vertical alignment) to do otherwise: I don't see such a reason here.

invisible_cities/cities/detsim.py
gondiaz: I grouped in `physics_parameters` and `buffer_parameters`. Maybe we can also group ligth tables paths.

invisible_cities/cities/detsim.py
gondiaz: Maybe this is solved by the` physics_parameters` grouping.

invisible_cities/cities/detsim.py
```diff
@@ -98,39 +98,29 @@ def create_s2_waveform(xs, ys, ts, phs, tmin, buffer_length):
...
         return pmt_bins, sipm_bins
     return get_bin_edges
 
 
 @city
 def detsim(files_in, file_out, event_range, print_mod, compression,
            detector_db, run_number, s1_lighttable, s2_lighttable, sipm_psf,
-           buffer_parameters  = dict(),
-           physics_parameters = dict()):
+           buffer_params  = dict(),
```
jacg: Eeeeek! Don't use mutable objects as default arguments!

(Do you know how to deal with this, or shall I explain?)

gondiaz: Please explain. Is not the same case in [dorothea](https://github.com/next-exp/IC/blob/master/invisible_cities/cities/dorothea.py), [penthesilea](https://github.com/next-exp/IC/blob/master/invisible_cities/cities/penthesilea.py) and [esmeralda](https://github.com/next-exp/IC/blob/master/invisible_cities/cities/esmeralda.py)?

jacg: The fundamental problem is that (in Python) default arguments are evaluated only once, at the time the function is defined. Consequently, the same *single object* is shared by all calls to the function which do not provide an overriding argument. If this object is mutated in any of those calls, then all subsequent calls use a different default value from the one visible in the function's parameter list. When this happens, the resulting observations (if you spot them; you might not, and then you're producing incorrect results) tend to be so freakishly weird, and waste so much debugging time, that we simply want to make sure that we never get to that stage. So, we studiously avoid *ever* using mutables as default arguments, directly.

Try this simple experiment to see it in action:

``` python
def foo(x, a=[]):
    a.append(x)
    return max(x, *a)

foo(1)   # 1
foo(666) 
foo(1)   # 666
```

It's a subtle manifestation of the scourge of mutable shared state.

A simple solution would be to use a frozen dict. Unfortunately a PEP proposing frozen dicts was rejected some years ago, leaving us with the bizarre situation that `frozenset` is a builtin, while its dictionary brother was not allowed in.

This leaves two obvious options:

1. Use the frozen dict package. Web search left as an exercise. I've no experience with it, so I can't vouch for it.

2. Use the standard idiom for dealing with mutable default arguments:

   ```python
   def fn(parameter=None):
       if parameter is None:
           parameter = <create *NEW INSTANCE* of mutable default value here>
   ```

   which is a bit annoying, but always works.
   
> Is not the same case in dorothea, penthesilea and esmeralda?

If so, then it should be fixed there too!

Not necessarily in this PR. If there is someone new to Next/IC/Git/PRs/our-workflow, nervously waiting in the wings, then fixing this problem in those other places could be an excellent introduction to submitting a PR in our environment.

jacg: Oh ...

> Is not the same case in dorothea, penthesilea and esmeralda?

... let this act as an illustration of why the practice of copy pasting stuff without understanding what it means and why it is done the way it is, has significant costs and is thoroughly discouraged.

Invest time and effort in understanding things deeply, and it will pay dividends in the long term.

gonzaponte: But this is a city, i.e. this function will never be called twice in the same program. At least it is not meant to. Isn't that a good motivation to reduce the noise introduced by initializing them to `None`? On the other hand, I don't know why these are the only parameters that are initialized to "nothing" as if they were optional, aren't they essential to the running of the city?

gondiaz: Yes, they are essential. Those parameters being an empty dictionary as they are by default won't work. The config file must contains all the parameters defined.

jacg: > this function will never be called twice in the same program. 

You hope ...

> At least it is not meant to.

... there you go: things than are not meant to happen, tend to have a habit of happening anyway, every now an then at least.

> Isn't that a good motivation to reduce the noise introduced by initializing them to None?

It may be.

A competent Python programmer should get very nervous at seeing mutable default values, and will then have to waste time wondering about the safety, or fixing the (non-)issue. So, if it stays, it must be accompanied by comments as to why it is safe (which the reader may, or may not, trust) ... at which point it might be less noisy simply to do the `None` trick.

jacg: > Those parameters being an empty dictionary as they are by default won't work. 

So the default value defined in the function must *always* be overridden for the function to stand a chance of working?

That's so wrong!

Why not turn them into keyword-only args?

jacg: > Why not turn them into keyword-only args?

... in fact, IIRC, it would simply make sense to make all parameters of all cities keyword-only.

But, even without that, just remove the default values, and everything should work, shouldn't it?

gonzaponte: This is something inherent to all cities in IC. All arguments are actually handled as keyword arguments by the `city` decorator (IIRC). We could turn them into keyword-only arguments, and maybe we should as I don't think it is of any use to have positional arguments. Somehow it bothers me to have that `*` at the beginning of the function signature, but it is definitely something irrelevant for the decision.

jacg: My hypotheses are

1. Adding `*` to the front of each city's signature should break nothing.
2. Removing all `= dict()` from all cities' signatures should break nothing.

I'm pretty confident on the first: that's by design of the whole liquid city thing. If anything breaks, it will be because, someone, somewhere is abusing something. Adding the `*` reduces the chances of someone abusing something in the future.

I'm less confident on the second: these are lower-level details so I'd have to look into the details case-by-case.

I propose removing the `= dict()`s in this city, and running all the tests. If that works, the obvious followups are the two points listed above, in reverse order.

gonzaponte: I agree. The first point will work for sure. The second one might break some random tests that don't care about the dict-like arguments to run, but it won't break any realistic usage of the city.

invisible_cities/cities/detsim_test.py
```diff
@@ -0,0 +1,103 @@
...
+
+def test_detsim_filter_active_hits(ICDATADIR, config_tmpdir):
+
+    PATH_IN  = os.path.join(ICDATADIR    , "nexus_new_kr83m_fast.newformat.sim.h5")
+    PATH_OUT = os.path.join(config_tmpdir, "detsim_test.h5")
+    conf = configure('detsim $ICTDIR/invisible_cities/config/detsim.conf'.split())
+
+    # modify labels for event_id = 0 in test file
+    # all its hits labels are set to "NO_ACTIVE"
+    modified_testfile = os.path.join(config_tmpdir, "modified_testfile.h5")
```
mmkekic: the files are less than 100KB, is probably ok to include them in the standard test suite

invisible_cities/cities/detsim_test.py
```diff
@@ -0,0 +1,103 @@
...
+from invisible_cities.core  import system_of_units as units
+
+
+def test_detsim_contains_all_tables(ICDATADIR, config_tmpdir):
+
+    PATH_IN  = os.path.join(ICDATADIR    , "nexus_new_kr83m_fast.newformat.sim.h5")
+    PATH_OUT = os.path.join(config_tmpdir, "detsim_test.h5")
+    conf = configure('detsim $ICTDIR/invisible_cities/config/detsim.conf'.split())
+
+    buffer_length     = 800 * units.mus
```
mmkekic: You seem to be reusing `conf['buffer_length']` rather than changing all its values, so you can either update it directly or create a local reference to it and update it since those are all mutable objects

invisible_cities/cities/detsim_test.py
```diff
@@ -0,0 +1,103 @@
...
+import pandas as pd
+import tables as tb
+
+from .. cities.detsim  import detsim
+from .. core.configure import configure
+
+from invisible_cities.core  import system_of_units as units
+
+
+def test_detsim_contains_all_tables(ICDATADIR, config_tmpdir):
```
mmkekic: probably you want to use `output_tmpdir` (yes, i know in other city tests is usually all placed in config one); also is better to name the output files differently so you are not overwriting the same file in one pytest run

invisible_cities/cities/detsim_test.py
jacg: > create a local reference to it and update it since those are all mutable objects

Didactic point ... I think that 'local reference' is an unfortunate choice of words: you should have said 'local copy'. 'Local reference' is not strictly incorrect, but it's ambiguous and fails to convey the important point. I'd go as far as saying that its most obvious interpretation is the exact opposite of what you mean!

It would be good if you all developed a habit of getting nervous when something is mutated, especially something that you got from somewhere else. Mutable shared state is the root of all evil in programming. Sometimes we need to mutate for pragmatic reasons, but do avoid mutation wherever pragmatically possible. In this case, we're talking about copying a smallish dictionary in a test that does IO (hint: IO is veeeeeeery slow compared to in-memory operations).

If you make a local copy of this dict, you can mutate it to your heart's content, and nobody ever has to worry that you're breaking things in some other part of the code which holds another reference to the thing you are mutating.

jacg:
> Correct bug

Please give *some* indication of the nature of the bug, or the correction, even if the indication is absolutely minimal. In this case, maybe "Fix bug: pmt_width -> sipm_width". (Don't bother rebasing just to fix it, but bear it in mind for the future.)

invisible_cities/cities/detsim_test.py
```diff
@@ -0,0 +1,277 @@
...
+from .. core.testing_utils   import assert_tables_equality
+
+
+def test_detsim_contains_all_tables(ICDATADIR, output_tmpdir):
+
+    PATH_IN  = os.path.join(ICDATADIR    , "nexus_new_kr83m_fast.newformat.sim.h5")
+    PATH_OUT = os.path.join(output_tmpdir, "detsim_test.h5")
+    conf = configure('detsim $ICTDIR/invisible_cities/config/detsim.conf'.split())
+
+    buffer_params = dict()
```
mmkekic: why not 
`dict(length=800*mus, pmt_width=100*units.ns`...?

invisible_cities/cities/detsim_test.py
```diff
@@ -0,0 +1,277 @@
...
+def test_detsim_filter_active_hits(ICDATADIR, output_tmpdir):
+
+    PATH_IN  = os.path.join(ICDATADIR    , "nexus_new_kr83m_fast.newformat.sim.h5")
+    PATH_OUT = os.path.join(output_tmpdir, "detsim_test.h5")
+    conf = configure('detsim $ICTDIR/invisible_cities/config/detsim.conf'.split())
+
+    # modify labels for event_id = 0 in test file
+    # all its hits labels are set to "NO_ACTIVE"
+    modified_testfile = os.path.join(output_tmpdir, "modified_testfile.h5")
+    shutil.copyfile(PATH_IN, modified_testfile)
```
mmkekic: Just add the modified file to the test files

invisible_cities/cities/detsim_test.py
```diff
@@ -0,0 +1,277 @@
...
+        np.testing.assert_array_equal(filters["passed"], [False, True])
+
+
+def test_detsim_empty_input_file(ICDATADIR, output_tmpdir):
+
+    PATH_IN  = os.path.join(ICDATADIR    , "nexus_new_kr83m_fast.newformat.sim.h5")
+    PATH_OUT = os.path.join(output_tmpdir, "detsim_test.h5")
+    conf = configure('detsim $ICTDIR/invisible_cities/config/detsim.conf'.split())
+
+    # remove test file rows
```
mmkekic: again, add empty file to the test files

invisible_cities/cities/detsim_test.py
```diff
@@ -0,0 +1,277 @@
...
+
+
+def test_detsim_exact(ICDATADIR, output_tmpdir):
+
+    PATH_IN   = os.path.join(ICDATADIR    , "nexus_new_kr83m_fast.newformat.sim.h5")
+    PATH_OUT  = os.path.join(output_tmpdir, "detsim_test.h5")
+    PATH_TRUE = os.path.join(ICDATADIR    , "detsim_new_kr83m_fast.newformat.sim.h5")
+    conf = configure('detsim $ICTDIR/invisible_cities/config/detsim.conf'.split())
+
+    buffer_params = dict()
```
mmkekic: No need to initialize empty dictionary, just create the dictionary with all the values you need.  You could **copy** the  `conf['buffer_params']`already existing dictionary and modify it since the only modification seems to be `pmt_width`

invisible_cities/cities/detsim_test.py
```diff
@@ -0,0 +1,277 @@
...
+                expected = getattr(true_output_file.root, table)
+                assert_tables_equality(got, expected)
+
+
+def test_detsim_exact_time_translation(ICDATADIR, output_tmpdir):
+
+    PATH_IN   = os.path.join(ICDATADIR    , "nexus_new_kr83m_fast.newformat.sim.h5")
+    PATH_OUT  = os.path.join(output_tmpdir, "detsim_test.h5")
+
+    physics_params = dict()
```
mmkekic: Again, copy already existing dictionary inside `conf` and only modify needed values

invisible_cities/cities/detsim_test.py
```diff
@@ -0,0 +1,277 @@
...
+    buffer_params = dict()
+    buffer_params["length"]      = 800 * units.mus
+    buffer_params["pmt_width"]   = 100 * units.ns
+    buffer_params["sipm_width"]  =   1 * units.mus
+    buffer_params["max_time"]    =  10 * units.ms
+    buffer_params["pre_trigger"] = 100 * units.mus
+    buffer_params["trigger_thr"] = 0
+
+    # copy to avoid mutability
+    physics_params_ = physics_params.copy()
```
mmkekic: I dont see that you later plan to modify `physics_params` so why copying it again?

invisible_cities/cities/detsim_test.py
```diff
@@ -0,0 +1,277 @@
...
+    buffer_params_  = buffer_params .copy()
+
+    # modify test file
+    modified_testfile = os.path.join(output_tmpdir, "modified_testfile.h5")
+    shutil.copyfile(PATH_IN, modified_testfile)
```
mmkekic: Add the `modified_testfile.h5`to test files (and rename it)

invisible_cities/cities/detsim_test.py
```diff
@@ -0,0 +1,277 @@
...
+    with tb.open_file(PATH_OUT) as output_file:
+        with tb.open_file(modified_file_out) as modified_output_file:
+            for table in tables:
+                assert hasattr(output_file.root, table)
+                got      = getattr( modified_output_file.root, table)
+                expected = getattr(          output_file.root, table)
+                assert_tables_equality(got, expected)
+
+
+def test_detsim_buffer_times(ICDATADIR, output_tmpdir):
```
mmkekic, line 151: Add a comment on what this test does, for example the nonphysical `el_drift_velocity`is confusing. Also, again,  you can copy dictionaries from conf and update the changed parameters

invisible_cities/cities/detsim_test.py
```diff
@@ -0,0 +1,277 @@
...
+    buffer_params["max_time"]    =  10 * units.ms
+    buffer_params["pre_trigger"] = 100 * units.mus
+    buffer_params["trigger_thr"] = 0
+
+    # run over test file
+    conf = configure('detsim $ICTDIR/invisible_cities/config/detsim.conf'.split())
+    conf.update(dict(files_in      = PATH_IN,
+                     file_out      = PATH_OUT,
+                     run_number    = 0,
+                     event_range   = (1, 2),
```
mmkekic, line 174: Can you add a comment why you skip the first event from the file? I understand it is because Z is very small for the first event in the file

invisible_cities/cities/detsim_test.py
```diff
@@ -0,0 +1,277 @@
...
+    # compute max signal time
+    hits = load_dst(PATH_IN, "MC", "hits")
+    hits = hits[hits["event_id"]==1]
+    z, time  = hits.z.values, hits.time.values
+    s2_times = z/physics_params["drift_velocity"] + time
+
+    s2_tmax = max(s2_times) + buffer_params["pre_trigger"]
+
+    with tb.open_file(PATH_OUT) as h5out:
+        pmtwf  = h5out.root.pmtrd.read()[0].sum(axis=0)
```
mmkekic: Can you also read sipmwfs (so we check timings there too)

invisible_cities/cities/detsim_test.py
```diff
@@ -0,0 +1,277 @@
...
+
+    signal_bins = np.argwhere(pmtwf>0).flatten()
+
+    # start s1 time
+    tstart = signal_bins[0]*buffer_params["pmt_width"]
+    assert tstart == buffer_params["pre_trigger"]
+
```
mmkekic, line 199: Maybe now you can simply assume that s2 will be after `pre_trigger + 2*mus` so set to s2s indices whatever is after that time. Then you can check that both start and end make sense. And do the same for sipms waveforms (without the s1 of course)

invisible_cities/cities/detsim.py
```diff
@@ -0,0 +1,214 @@
...
+    el_dv = physics_params.pop("el_drift_velocity")
+
+    # derived parameters
+    datapmt  = db.DataPMT (detector_db, run_number)
+    datasipm = db.DataSiPM(detector_db, run_number)
+    lt_pmt  = LT_PMT (fname=os.path.expandvars(s2_lighttable))
+    lt_sipm = LT_SiPM(fname=os.path.expandvars(sipm_psf), sipm_database=datasipm)
+    el_gap = lt_sipm.el_gap_width
+
+    select_s1_candidate_hits = fl.map(hits_selector(False),
```
mmkekic, line 138: This is not even used in the flow!

invisible_cities/cities/detsim.py
```diff
@@ -0,0 +1,214 @@
...
+from .. io.event_filter_io  import event_filter_writer
+
+from .  components                import MC_hits_from_files
+from .. detsim.simulate_electrons import generate_ionization_electrons
+from .. detsim.simulate_electrons import drift_electrons
+from .. detsim.simulate_electrons import diffuse_electrons
+from .. detsim.light_tables_c     import LT_SiPM
+from .. detsim.light_tables_c     import LT_PMT
+from .. detsim.s2_waveforms_c     import create_wfs
+from .. detsim.detsim_waveforms   import s1_waveforms_creator
```
mmkekic, line 39: Remove all unnecessary imports (use pyflakes to discover which ones to remove)

invisible_cities/cities/detsim_test.py
gondiaz: `physics_params_` is mutted inside detsim city (`ws` and `el_drift_velocity`), therefore the second city call can not be done with the same dictionary. They will take the values from **detsim.conf** from now on.

mmkekic (commented):
After some commits squashing I am ready to approve this PR

mmkekic (approved):
This PR is the last piece of the long lasting effort to implement fast detector light simulation in IC. The PR introduces the final city flow that is using already approved functionalities. Code is readable and appropriate tests are added. Good job!

gondiaz:
Validation plots for detsim city for 0nubb events. The plots compare the results obtained using this city and the full simulation results from nexus for the same events. The plots show several higher level variables, namely S2 width, height and PMT energy (S2w, S2h, S2e). Also the SIPM charge (S2q) for events at low radious (*). The right plots represent the quotient between detsim and fullsim corresponding variables.

Plots taken from: https://next.ific.uv.es/DocDB/0011/001138/001/DETSIM_Dec2020.pdf

0nubb_detsim_fullsim_0

0nubb_detsim_fullsim_1

(*) The reconstruction bias affects the R (radius) assigned to the event by penthesilea. PSF is assumed to be radial independent in detsim (but easily implementable), which is not true in fullsim or data. Physics results are not expected to be affected since TP signal is used for transverse positioning.

PR #770, opened 2021-02-03
Update demopp database

carmenromo:
Demo++ database has been updated with the new constants for the SiPMs from run 9523. Channel `16028` may be dead so it has been masked and checked in the next calibrations. 

Some plots are attached comparing the current runs with the last calibration, taken in September. The gains apparently have been increased slightly and the 1pe sigma decreased...

sipmRunDifferencePlots


jmbenlloch (approved):
Latest calibration properly added. All tests passed, including the action specific for this cases.

Issue #771, opened 2021-02-05
Get Event Numbers from MC files


jmunozv:
Currently, this action is implemented in `invisible_cities.io.mcinfo_io.get_event_numbers_in_file(...)`

For simulations with the new nexus persistency format, it is extremely slow. As an example, to get the list of the 4 event_ids contained in a NEXT100 bb0nu full simulation file takes almost 1 second. The reason is that this function checks 'hits', 'particles' and 'sns_response' tables to look for unique event_ids in them.

The idea is to simplify and speed up this function given that:

- There are no events with hits and no particles. In Geant4 all hits belong to a particle, and these particles are aways stored in nexus.
- Something similar happens with sns_response, although in this particular case, optical photons (the only particles seen by our sensors) can be stored in the simulation or not.

If these statements are correct (please experts in nexus simulations @jmalbos and @paolafer confirm) my claim is that the function should:
1. If there are stored particles, return the unique event_ids from this table.
2. If not, return the unique event_ids from sns_response.
 
Some preliminary checks yield a time consumption reduction higher than a factor 100 for previous example.

paolafer:
I think that this issue can be closed, since it was solved in PR #772.

gonzaponte:
Closing as solved in #772.

PR #772, opened 2021-02-09
Speed up the process of getting event numbers from mc files

jmunozv:
This PR addresses issue #771 speeding up the process of getting the event numbers from nexus files stored with the new persistency.

To illustrate the gain, I have profiles the process of getting the event numbers from 10 different full simulation files containing 4 bb0nu events each one of the files.

Current function yields an average time of:  833356.9 microseconds
This PR function yields an average time of:       7693.2 microseconds

All tests pass succesfully.

### DIFF
--- a/invisible_cities/io/mcinfo_io.py
+++ b/invisible_cities/io/mcinfo_io.py
@@ -128,7 +128,7 @@ def safe_copy_nexus_eventmap(h5out  : tb.file.File,
 
     returns
     -------
-    Mapping : 
+    Mapping :
             'evt_number': the evt_number copied to the output;
             'nexus_evt' : the nexus event_ids copied to the output.
     """
@@ -325,12 +325,20 @@ def get_event_numbers_in_file(file_name: str) -> np.ndarray:
 
 
 def _get_list_of_events_new(h5in : tb.file.File) -> np.ndarray:
+    """
+    This function relies on the assumption that, if the particles table exists,
+    it includes all the events of the file. At the same time,
+    if that table doesn't exists, all the events are assumed to have
+    generated some sensor response, therefore to appear in the sns_response table.
+    Events that don't generate sensor response and don't have any saved particle
+    are assumed to be irrelevant for the analysis.
+    """
     def get_event_ids_table(tablename):
         try:
             evt_list = getattr(h5in.root.MC, tablename).cols.event_id
         except tb.exceptions.NoSuchNodeError:
             raise AttributeError('Trying to get event number from MC corrupted file.')
-        return np.unique(evt_list).astype(int)   
+        return np.unique(evt_list).astype(int)
     evt_list = get_event_ids_table('particles')
     if len(evt_list):
         return evt_list
--- a/invisible_cities/io/mcinfo_io.py
+++ b/invisible_cities/io/mcinfo_io.py
@@ -325,19 +325,16 @@ def get_event_numbers_in_file(file_name: str) -> np.ndarray:
 
 
 def _get_list_of_events_new(h5in : tb.file.File) -> np.ndarray:
-    mc_tbls  = ['hits', 'particles', 'sns_response']
-    def try_unique_evt_itr(group, itr):
-        for elem in itr:
-            try:
-                yield np.unique(getattr(group, elem).cols.event_id)
-            except tb.exceptions.NoSuchNodeError:
-                pass
-
-    evt_list = list(try_unique_evt_itr(h5in.root.MC, mc_tbls))
-    if len(evt_list) == 0:
-        raise AttributeError("At least one of MC/hits, MC/particles, \
-        MC/sns_response must be present to use get_list_of_events.")
-    return np.unique(np.concatenate(evt_list)).astype(int)
+    def get_event_ids_table(tablename):
+        try:
+            evt_list = getattr(h5in.root.MC, tablename).cols.event_id
+        except tb.exceptions.NoSuchNodeError:
+            raise AttributeError('Trying to get event number from MC corrupted file.')
+        return np.unique(evt_list).astype(int)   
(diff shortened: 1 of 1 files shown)


jacg:
> I have profiles the process of getting the event numbers from 10 different full simulation files containing 4 bb0nu events each one of the files.
>
> Current function yields an average time of: 833356.9 microseconds
> This PR function yields an average time of: 7693.2 microseconds

Those look like benchmarking, not profiling, results.

(For anyone else getting lost in the noisy digit count, it's a x100 improvement.)

invisible_cities/io/mcinfo_io.py
```diff
@@ -325,19 +325,19 @@ def get_event_numbers_in_file(file_name: str) -> np.ndarray:
...
-            except tb.exceptions.NoSuchNodeError:
-                pass
-
-    evt_list = list(try_unique_evt_itr(h5in.root.MC, mc_tbls))
-    if len(evt_list) == 0:
-        raise AttributeError("At least one of MC/hits, MC/particles, \
-        MC/sns_response must be present to use get_list_of_events.")
-    return np.unique(np.concatenate(evt_list)).astype(int)
+
+    def _get_event_ids_table(tablename):
```
jacg: Leading underscores in function local scopes don't really make much sense.

jmunozv: Changed

invisible_cities/io/mcinfo_io.py
```diff
@@ -325,19 +325,19 @@ def get_event_numbers_in_file(file_name: str) -> np.ndarray:
...
-        MC/sns_response must be present to use get_list_of_events.")
-    return np.unique(np.concatenate(evt_list)).astype(int)
+
+    def get_event_ids_table(tablename):
+        try:
+            evt_list = getattr(h5in.root.MC, tablename).cols.event_id
+        except tb.exceptions.NoSuchNodeError:
+            raise AttributeError('Trying to get event number from MC corrupted file.')
+        return np.unique(evt_list).astype(int)
+
```
jacg: It would be good if we could *all* please instruct our editors to remove trailing whitespace whenever we save source files. It's not a big deal, but it helps us keep irrelevant noise out of diffs. (And the most annoying noise usually appears many commits later, and is thrown in the face of someone other than the party that was guilty of introducing the spurious whitespace in the first place ... so it's a matter of courtesy to our colleagues!)

jacg:
The tests are timing out with [the error](https://github.com/next-exp/IC/pull/772/checks?check_run_id=1862745384#step:5:244):

> pymysql.err.OperationalError: (2003, "Can't connect to MySQL server on 'neutrinos1.ific.uv.es' (timed out)")

jacg:
Do you have any insight into *why* this gives a x100 speedup? 

IIUC, there was a `try` inside a loop, but the loop only considered 3 elements (`hits`, `particles`, `sns_response`). But the results for all 3 of these were combined and uniquified, whereas the new version only considers `particles` or `sns_response`, and the latter only if the former produces an empty result set.

There is a significant semantic difference here between new and old which is beyond my understanding of the high-level point of this code, and the tests are blocked by a timeout ...

invisible_cities/io/mcinfo_io.py
jmunozv: Changed

jmunozv:
I tried to explain all the related stuff in issue #771 as commented in the PR message.

andLaing:
The function that @jmunozv changes in the PR was introduced as part of a larger update of the readers and writers for MC (PR #693 ) where we tended to air on the side of caution/generalisation. The reason we went with the check of the three tables and uniques and overlaps, if I remember correctly, was because we saw that some files had only a subset of the tables present and that there wasn't always 1-to-1 overlap of the event numbers present. That being said we didn't do a lot of benchmarking or optimisation so it's possible that we were overly cautious.

I'd be careful to make sure we're getting the full set of event numbers with the new version but then maybe an argument can be made that certain event numbers are not relevant further upstream but that's a wider question that'd have to be discussed by the users of the MC data.

jmunozv:
I agree with Andrew's comment https://github.com/next-exp/IC/pull/772#issuecomment-778596012. We should not approve this PR until we are 100% sure about we don't loose any event number in the process.

Until now, I've not been able to imagine any situation in which the formula implemented in this PR could fail, and I appreciate that if someone can, or have a real case in which it does, they bring it up to the current discussion.

jacg:
On a completely frivolous tangent, I was tickled by the observation that this typo:

> to air on the side of caution

only makes sense in a Scottish accent :-)

paolafer (approved):
I cannot think of any user case where this approach wouldn't work. I assume that the case of corrupt files or files that are not closed properly is dealt with in a different way. To prevent our future selves from making changes to nexus that could break this unwillingly, a comment can be added declaring explicitly that this works under the conditions that any event with saved hits has saved particles as well and that for a given simulation no events with sensor response only can be generated mixed with others with saved particles. 

I would approve this PR.

paolafer:
I was thinking, just to be overcautious, why don't we run a check on a large number of files of a typical simulation, such as background model, ensuring that the event IDs in the `particles` table are the same as in the `hits` table?

jmunozv:
> I was thinking, just to be overcautious, why don't we run a check on a large number of files of a typical simulation, such as background model, ensuring that the event IDs in the `particles` table are the same as in the `hits` table?

This is a good idea but, currently we don't have any set of nexus simulations with the new persistency to run the test over.

Considering that we are not in hurry to approve this PR, I suggest to wait for having such a set of simulations, and be able to run the the suggested test.

paolafer:
I added a comment to clarify the assumptions underlying the new way of retrieving event numbers. From my side, the PR can be merged.

PR #775, opened 2021-02-25
Upload NEW database

carmenromo:
This PR includes the latest sensor calibrations from run 8480. Database has been uploaded with the PMT and SiPMs gains and the pdf values. Sensor 15060 is seeing no light again, so it has been masked.

The poisson mu of the PMT9 has slightly decreased to its previous value.

Plots comparing with other stable runs are attached.

pmtRunDifferencePlots2

sipmRunDifferencePlots_test


jmbenlloch (approved):

PR #776, opened 2021-02-26
Add protection for empty waveforms

gondiaz:
This PR add a protection for empty waveforms in signal finder. 
I realized about this issue throught detsim simulation of background events close to the chamber edges at radious higher than the active radious. 
In such cases the waveforms created by detsim are empty, ie they are a list of zeros. Current implementation is not prepared to deal this case. Also, note that a `trigger_thr` parameter high enought, no signals are found and the same problem arises. This issue also affects buffy city.


mmkekic:
We haven't thought about this possibility indeed. I think the most sensible is to filter out events with empty waveforms, instead of this hacky solution of tricking signal finding. 
I suggest you :
1. provide a test ( and a 1-event testfile ) where simply running a detsim will fail
2. add a filter to city flow to filter out events with empty waveforms
3. update the test to check that the events are filtered

jacg:
In 3. you mean *rerun* the test?

andLaing:
I think there's even a filter function in `buffy.py` maybe worth moving it to components or something so it's generally useful.

mmkekic:
> I think there's even a filter function in buffy.py maybe worth moving it to components or something so it's generally useful.

Yeah, I think it can go to commonly used `calculate_and_save_buffers` 

> In 3. you mean rerun the test?

Well notice that we do 2 things here, fix a bug so the 'weird' file doesn't produce the error, but also add a new filter table that we  want to test that it contains the 'weird' event. My suggestion was to, in the first commit, introduce a test that will just run the city over the file (and fail), ie lines of a type
```
conf.update(dict(files_in='weird_file.h5', file_out='tmp_file.h5'))
detsim(**conf)
```
and in step 3 add 
check Filter table inside `tmp_file.h5`

We could, of course, do this in 2 separate tests, one that is simply showing the bug fix and the other that is testing the filtering, in which case we run over the same file 2x instead of reusing the output of one run only.

jacg:
I'm sorry if I haven't got the time to carefully understand the specific details of this case.

In general, I recommend the following procedure:

1. Commit a test that points out a problem. This test should fail.
2. Commit a solution to the problem.
3. The same, *unmodified* test introduced in the first step should now pass, proving that

   - there was a problem that needed fixing
   - the problem has been fixed
   - the latest commit is the solution to that problem.
   
Now, as for this specific case: I guess that I don't understand sufficiently what is meant exactly by 'filter table' and the weird event being in it.

Is the bug fixed by filtering out the weird event, or by keeping the event and adding tolerance for such weird events to the implementation?

mmkekic:
> Is the bug fixed by filtering out the weird event, or by keeping the event and adding tolerance for such weird events to the implementation?

It is fixed by filtering out the event, it is an empty event hence we do not want to keep it. So, in addition to code not breaking with unexpected error, we also want to test that the particular event number is saved in the Filters table in the output file (filtered out events are saved in Filters/filter_name tables).

gondiaz:
My proposed solution requires to add a filter for empty signals in `calculate_and_save_buffers ` function. In order to maintaint the `evtnum_collect` the city pipes of detsim and buffy must be rewritted.

Note also that the simplistic solution of filter empty waveforms directly in the city flow is somehow buggy, since those waveforms can contain signal which could not be found by `find_signal_start`. Now, if no signal is found by this function it returns an empty list that is filtered afterwards in `calculate_and_save_buffers `.

invisible_cities/cities/detsim_test.py
```diff
@@ -59,6 +59,20 @@ def test_detsim_filter_active_hits(ICDATADIR, output_tmpdir):
         np.testing.assert_array_equal(filters["passed"], [False, True])
 
 
+def test_detsim_filter_empty_waveforms(ICDATADIR, output_tmpdir):
+    # the first event radious is slighty above NEW active radious of 208.0 mm
```
jacg: Typo: radious -> radius (twice)

invisible_cities/detsim/buffer_functions.py
```diff
@@ -55,6 +55,7 @@ def find_signal_start(wfs          : Union[pd.Series, np.ndarray],
         eng_sum = wfs.sum()
     indices = indices_and_wf_above_threshold(eng_sum,
                                              bin_threshold).indices
+    if len(indices) == 0: return []
     ## Just using this and the stand_off for now
     ## taking first above sum threshold.
     ## !! To-do: make more robust with min int? or similar
```
jacg: It's really not worth opening a separate PR for it, so, while we're touching this file, could you please add a single commit which fixes all the `to-do`s in this file (and any others that you are already touching in this PR), changing them to this *precise* spelling: `TODO` (exactly four characters, all uppercase, no spaces, no dashes). This is because many development tools recognize `TODO` and can help managing them.

invisible_cities/cities/components.py
```diff
@@ -878,6 +878,12 @@ def calculate_and_save_buffers(buffer_length    : float        ,
                                args = "pmt_bin_wfs"                 ,
                                out  = "pulses"                      )
 
+    filter_events_signal = fl.map(lambda x: len(x) > 0,
+                                          args= 'pulses',
+                                          out = 'passed_signal')
```
jacg: Fix indentation

invisible_cities/cities/components.py
```diff
@@ -902,10 +908,13 @@ def calculate_and_save_buffers(buffer_length    : float        ,
                                      saved_buffers                ))
 
     fn_list = (find_signal      ,
```
jacg: `fn_list` is a very unhelpful name: please use a name that summarizes what the sub-pipeline does.

invisible_cities/cities/components.py
```diff
@@ -902,10 +908,13 @@ def calculate_and_save_buffers(buffer_length    : float        ,
...
+               filter_events_signal,
+               fl.branch(write_signal_filter),
```
jacg: A `branch` containing a function wrapped in `sink` is a common pattern, which is why `spy` exists as an abstraction:

```python
spy(fn)
```

is a shorthand for 

```python
fn_wrapped_in_sink = sink(fn)
branch(fn_wrapped_in_sink)
```
or, more simply

```python
branch(sink(fn))
```

Apart from being slightly shorter to type, it sends a strong signal to the human reader about the simplicity of what is going on.

invisible_cities/cities/components.py
```diff
@@ -902,10 +908,13 @@ def calculate_and_save_buffers(buffer_length    : float        ,
...
+               events_passed_signal.filter,
                event_times      ,
                calculate_buffers,
                order_sensors    ,
-               buffer_writer_   )
+               fl.branch(buffer_writer_))
```
jacg: Again, you could use `spy` here, and remove the `sink` wrapper in the definition of `buffer_writer_`

invisible_cities/cities/detsim.py
```diff
@@ -213,9 +213,10 @@ def detsim(*, files_in, file_out, event_range, print_mod, compression,
                                          create_pmt_waveforms,
                                          create_sipm_waveforms,
                                          get_bin_edges,
+                                         buffer_calculation,
                                          fl.branch("event_number"     ,
                                                    evtnum_collect.sink),
-                                         buffer_calculation),
+                                         fl.sink(lambda x: x)),
```
jacg: What is the point of this? Is it to have a `/dev/null` sink? (One that just throws the data away?) 

If so, then simply don't branch off at the previous step!

invisible_cities/cities/buffy.py
```diff
@@ -112,9 +112,10 @@ def buffy(files_in     , file_out   , compression      , event_range,
                                           extract_tminmax               ,
                                           bin_pmt_wf                    ,
                                           bin_sipm_wf                   ,
+                                          buffer_calculation            ,
                                           fl.branch("event_number"      ,
                                                     evtnum_collect.sink),
-                                          buffer_calculation            )   ,
+                                          fl.sink(lambda x: x))         ,
```
jacg: Again, you won't need a cryptic `/dev/null` sink if you do the previous step without branching off.

invisible_cities/cities/components.py
jacg: Generally speaking, lining the commas up vertically helps to avoid mistakes and misunderstandings, and this can be especially important in dataflow pipelines. Now, adjusting the commas at the end of the line is a PITA, and adds unnecessary noise to the diffs.

I recommend that we adopt the following style. Rather than

``` python
fn(foo,
   baaaaaaaa,
   baz,
   quux,
   quuuuuux)
```
where comma-errors or surprises are difficult to spot, or 
``` python
fn(foo      ,
   baaaaaaaa,
   baz      ,
   quux     ,
   quuuuuux )
```
which is a PITA to keep aligned, do this:
``` python
fn( foo
  , baaaaaaaa
  , baz
  , quux
  , quuuuuux)
```
This will really freak out the PEP8-ists, (and many of you) but, turst me, it's the best way forward :-)

gondiaz: The `sink`ed function, namely `write_signal_filter`, requires `args` which are not admitted in `spy`. I don't see how we can deal with it.

invisible_cities/cities/detsim.py
gondiaz: We could sink the pipe inside `calculate_and_save_buffers`, but I wanted to keep the `evtnum_collect` counter. I dont know how to `sink` the main pipe afterwards so I added this dummy `sink` which is irrelevant for the city itself.

invisible_cities/cities/components.py
jacg: Ah, OK, `sink` is one of the privileged few that have been endowed with argument picking. (Which is why [liquidata](https://github.com/jacg/liquidata) implements argument picking and result setting as independent, orthogonal and universally applicable utilities).

So, yes, in this case the sink-in-branch approach is probably the cleanest.

invisible_cities/cities/detsim.py
jacg: I don't follow. At the moment you have

```
buffer_calculation ----> /dev/null
                  \
                    ---> pick event_number -> evtnum_collect.sink
```
Why can't you do the following instead?
```
buffer_calculation ---> pick event_number -> evtnum_collect.sink
```

gondiaz: Yes, you are right. My bad because didn't really understand what `"event_number"` did inside a pipe.

invisible_cities/cities/components.py
```diff
@@ -898,24 +898,24 @@ def calculate_and_save_buffers(buffer_length    : float        ,
...
     max_subevt    =  max_time // buffer_length + 1
-    buffer_writer_    = sink(buffer_writer(h5out                  ,
-                                           run_number = run_number,
-                                           n_sens_eng = npmt      ,
-                                           n_sens_trk = nsipm     ,
-                                           length_eng = nsamp_pmt ,
-                                           length_trk = nsamp_sipm,
-                                           max_subevt = max_subevt),
+    buffer_writer_    = sink(buffer_writer(  h5out
+                                           , run_number = run_number
```
jacg: I'm partial to aligning the commas directly under the opening bracket (and I'm pretty sure that is de-facto standard for this style), but I won't kick up a fuss if you leave it as it is.

gondiaz:
If there are no more issues/comments, this PR should be ready to approval @jacg

jacg:
I'm tied up elsewhere at the moment. Apart from the stylistic comments I've already made, I wanted to have a look at the semantics, but I won't have time for that in the next 24h, at least. Maybe someone else can have a look.

mmkekic:
I had a look and it seems like a minimal change that address the issue : if no pulses are found, filter the event. 
This is a generalization of the previous filter in buffy where we filtered out waveforms that had no positive signal, hence implicitly assuming that the threshold for pulses will always be 0.  
The main change is in `find_signal_start`in `buffer_functions.py` where an empty list is returned in case there are no pulse candidates, and the change in pipe `calculate_and_save_buffers` in `components.py` where the filter itself is added. If nobody complains I would approve this PR and the only thing left is some history squashing.

mmkekic (approved):
This PR adds a filter in case no signal pulses were found in the waveforms. New filter is used both in detsim and buffy (generalizing  the previous buffy filter that was triggered for only all-0s waveforms). The test to show unwanted behavior in detsim is added - it fails before the changes and passes afterwards.
The code changed is simple and clean. Thanks!

PR #777, opened 2021-03-04
Fix Xpeak & Ypeak in Penthesilea

### LINKED ISSUE #730: XYpeak in Penthesilea
The hDST contains two columns `Xpeak` and `Ypeak` that are meant to give a center of gravity of a given S2 signal in XY. However, the current implementation is not very helpful. The algorithm used is the barycenter, without any thresholds. @Aretno claims that this configuration is not reliable as many spurious SiPM hits shift the reconstruction away from the "true" value. On the other hand we have the Dorothea branch of the processing running in the same city that performs this operation with a more suitable reconstruction configuration, but it writes it to a different table.

We have a couple of options:
- Add an argument that allows a full configuration of the algorithm, probably using the same one as in the Dorothea branch.
- Remove `Xpeak` and `Ypeak` from the output and optionally attach those columns to the table when loading from file.

gonzaponte:
This PR addresses issue #730. The computation of the `Xpeak` and `Ypeak` columns used hardwired parameters which included `Qthr = 0`. This results in a biased computation of those columns. The solution includes the usage of the `global_reco_params` city argument (currently used for the kdst side of the pipeline) for the calculation of `Xpeak` and `Ypeak`.

A test together with a new test file has been added. The test file contains a single event with an artificial pmap made to illustrate the problem. The S2 of this pmap contains a single sipm with charge over threshold and a bunch of sipms with charge below threshold (threshold = 5 pes). With the previous configuration, no threshold is applied and therefore all the sipms are taken into account, resulting in a biased result. However, if the threshold is applied (this PR), only 1 sipm survives, and the reconstructed position matches that of the sipm.

Fixes  #730.

### DIFF
--- a/invisible_cities/cities/penthesilea_test.py
+++ b/invisible_cities/cities/penthesilea_test.py
@@ -373,11 +373,12 @@ def test_penthesilea_global_xyreco_bias(config_tmpdir, ICDATADIR):
     PATH_OUT = os.path.join(config_tmpdir, 'fake_hdst_nothreshold_barycenter_bias.h5')
 
     conf = configure('dummy invisible_cities/config/penthesilea.conf'.split())
-    conf.update(dict(files_in    = PATH_IN ,
-                     file_out    = PATH_OUT,
-                     rebin       =      100,
-                     run_number  =     7000,
-                     s2_nsipmmax =      200))
+    conf.update(dict( files_in    = PATH_IN
+                    , file_out    = PATH_OUT
+                    , rebin       =  100
+                    , run_number  = 7000
+                    , s2_nsipmmax =  200
+                    ))
 
     conf["global_reco_params"].update(dict(Qthr=5))
 
--- a/invisible_cities/cities/penthesilea_test.py
+++ b/invisible_cities/cities/penthesilea_test.py
@@ -359,3 +359,30 @@ def test_penthesilea_empty_input_file(config_tmpdir, ICDATADIR):
     with warnings.catch_warnings():
         warnings.simplefilter("ignore", category=UserWarning)
         penthesilea(**conf)
+
+
+
+def test_penthesilea_global_xyreco_bias(config_tmpdir, ICDATADIR):
+    # Make sure the global_reco_params are used for the columns Xpeak/Ypeak
+    # This is done by running over an event that has many spurious sipm hits
+    # and only one sipm with signal over the threshold. If the threshold
+    # is used, the reconstructed position will match the sipm position.
+    # If the threshold is not used the position will be biased.
+
+    PATH_IN  = os.path.join(ICDATADIR    , 'fake_pmap_nothreshold_barycenter_bias.h5')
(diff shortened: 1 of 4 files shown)


invisible_cities/cities/penthesilea_test.py
```diff
@@ -359,3 +359,30 @@ def test_penthesilea_empty_input_file(config_tmpdir, ICDATADIR):
...
+    # This is done by running over an event that has many spurious sipm hits
+    # and only one sipm with signal over the threshold. If the threshold
+    # is used, the reconstructed position will match the sipm position.
+    # If the threshold is not used the position will be biased.
+
+    PATH_IN  = os.path.join(ICDATADIR    , 'fake_pmap_nothreshold_barycenter_bias.h5')
+    PATH_OUT = os.path.join(config_tmpdir, 'fake_hdst_nothreshold_barycenter_bias.h5')
+
+    conf = configure('dummy invisible_cities/config/penthesilea.conf'.split())
+    conf.update(dict(files_in    = PATH_IN ,
```
Aretno: Maybe you want to change the comma alignment following https://github.com/next-exp/IC/pull/776#discussion_r584109590

Aretno (approved):
Basic but long needed fix in penthesilea. With this the mean position in the hdst matches that of the kdsts, which was the desired behavior. This will fix a PSF related issue where the user had to recalculate the barycenter in order to use proper positioning. Code is nice and clean and aside form a minor style issue I'd say it's ready to go.

invisible_cities/cities/penthesilea_test.py
gonzaponte: Is it agreed to change to this alignment style?

jacg: Given that we have

+ In favour:
   1. my original suggestion
   2.  @gondiaz' heart emoji in response to said suggestion
   3.  @Aretno's suggestion that you use the style
+ against:
   ∅

that is to say, the ratio of (implicit or explicit) favourable to unfavourable opinions is ∞, I'm tempted (in the absence of vehement opposition in the next 24 hours) to conclude that we've 'agreed' to change to this style. Do you have an opinion on it?

This is a choice which has no profound and irreversible consequences. No point in waiting on a decision. Just Do It! (Unless you hate it.)

gonzaponte: > Do you have an opinion on it?

I'm in 100% favor of this alignment style. I was just wondering whether it was official or not. I am guessing we will not make the effort of realigning existing code, but rather request new code to follow this rule, right?

jacg: > I'm in 100% favor of this alignment style.

Quadruple infinity. I think that settles it.

> I was just wondering whether it was official or not.

AFAIAC, it is now.

> I am guessing we will not make the effort of realigning existing code, but rather request new code to follow this rule, right?

Exactly.

We can also adapt existing code when we are changing it for some other purpose.

Aretno:
Now this is fully ready to be merged. Good job!

PR #780, opened 2021-04-02
Add flatmap operator

gonzaponte:
This PR adds the flatmap operator to dataflow, which will be needed in an upcoming PR.

This is meant to be a temporary solution while we get to the point of replacing dataflow with its beefed-up version, [liquidata](https://github.com/jacg/liquidata), which already provides the functionality.

I intend to pay the debt soon(ish).

### DIFF
--- a/invisible_cities/dataflow/dataflow_test.py
+++ b/invisible_cities/dataflow/dataflow_test.py
@@ -68,73 +68,6 @@ def test_map():
     assert result == list(map(the_operation, the_source))
 
 
-
-def test_flatmap():
-
-    # Flatmap accepts a function which returns an iterable.
-    # While using map with such a function would result in entire
-    # iterables being sent downstream as separate elements,
-    # flatmap concatenates the contents of these iterables producing one
-    # long stream of the individual elements contained in those iterables.
-    # Equivalent to
-    # for incoming_item in upstream:
-    #     for outgoing_item in make_sequence(incoming_item):
-    #         send_downstream(outgoing_item)
-
-    the_operation = range
-    ranges = df.flatmap(the_operation)
-
-    the_source = list(range(1,11))
-
-    result = []
-    the_sink = df.sink(result.append)
-
-    df.push(source = the_source,
-            pipe   = ranges(the_sink))
-
-    expected = []
-    for s in the_source:
-        for item in the_operation(s):
-            expected.append(item)
-
-    assert result == expected
-
-
-
-def test_flatmap_args_out():
-
-    # A more complex case, taking and producing multiple values
-
-    def repeat_numbers(a, b):
-        return [(b, str(b)*i) for i in range(1, a+1)]
-
-    the_operation = df.flatmap( repeat_numbers
-                              , args = list("ab")
-                              , out  = list("cd")
-                              )
-
-    the_source = [ dict(a=1, b=4)
-                 , dict(a=2, b=5)
-                 , dict(a=3, b=6)
-                 ]
-
-    result_a = []
-    result_b = []
-    def collect(a, b):
-        result_a.append(a)
-        result_b.append(b)
-
-    the_sink = df.sink(collect, args=list("cd"))
-
-    df.push(source = the_source,
-            pipe   = the_operation(the_sink))
-
-    expected_a = [4, 5, 5, 6, 6, 6]
-    expected_b = ["4", "5", "55", "6", "66", "666"]
-    assert result_a == expected_a
-    assert result_b == expected_b
-
-
 def test_pipe():
 
     # The basic syntax requires any element of a pipeline to be passed
@@ -194,6 +127,71 @@ def test_filter():
             pipe   = df.pipe(odd, the_sink))
 
     assert result == list(filter(the_predicate, the_source))
+
+
+def test_flatmap():
+
+    # Flatmap accepts a function which returns an iterable.
+    # While using map with such a function would result in entire
+    # iterables being sent downstream as separate elements,
+    # flatmap concatenates the contents of these iterables producing one
+    # long stream of the individual elements contained in those iterables.
+    # Equivalent to
+    # for incoming_item in upstream:
+    #     for outgoing_item in make_sequence(incoming_item):
+    #         send_downstream(outgoing_item)
+
+    the_operation = range
+    ranges = df.flatmap(the_operation)
+
+    the_source = list(range(1,11))
+
+    result = []
+    the_sink = df.sink(result.append)
+
+    df.push(source = the_source,
+            pipe   = ranges(the_sink))
+
+    expected = []
+    for s in the_source:
+        for item in the_operation(s):
+            expected.append(item)
+
+    assert result == expected
+
+
+def test_flatmap_args_out():
+
+    # A more complex case, taking and producing multiple values
+
+    def repeat_numbers(a, b):
+        return [(b, str(b)*i) for i in range(1, a+1)]
+
+    the_operation = df.flatmap( repeat_numbers
+                              , args = list("ab")
+                              , out  = list("cd")
+                              )
+
+    the_source = [ dict(a=1, b=4)
+                 , dict(a=2, b=5)
+                 , dict(a=3, b=6)
+                 ]
+
+    result_a = []
+    result_b = []
+    def collect(a, b):
+        result_a.append(a)
+        result_b.append(b)
+
+    the_sink = df.sink(collect, args=list("cd"))
+
+    df.push(source = the_source,
+            pipe   = the_operation(the_sink))
+
+    expected_a = [ 4 ,  5 ,   5 ,  6 ,   6 ,    6 ]
+    expected_b = ["4", "5", "55", "6", "66", "666"]
+    assert result_a == expected_a
+    assert result_b == expected_b
 
 
 def test_count():
--- a/invisible_cities/dataflow/dataflow_test.py
+++ b/invisible_cities/dataflow/dataflow_test.py
@@ -68,6 +68,73 @@ def test_map():
     assert result == list(map(the_operation, the_source))
 
 
+
+def test_flatmap():
+
+    # Flatmap accepts a function which returns an iterable.
+    # While using map with such a function would result in entire
+    # iterables being sent downstream as separate elements,
+    # flatmap concatenates the contents of these iterables producing one
+    # long stream of the individual elements contained in those iterables.
+    # Equivalent to
+    # for incoming_item in upstream:
+    #     for outgoing_item in make_sequence(incoming_item):
+    #         send_downstream(outgoing_item)
+
+    the_operation = range
+    ranges = df.flatmap(the_operation)
+
+    the_source = list(range(1,11))
+
+    result = []
+    the_sink = df.sink(result.append)
+
+    df.push(source = the_source,
+            pipe   = ranges(the_sink))
+
+    expected = []
+    for s in the_source:
+        for item in the_operation(s):
+            expected.append(item)
+
+    assert result == expected
+
+
+
+def test_flatmap_args_out():
+
+    # A more complex case, taking and producing multiple values
+
+    def repeat_numbers(a, b):
+        return [(b, str(b)*i) for i in range(1, a+1)]
+
+    the_operation = df.flatmap( repeat_numbers
+                              , args = list("ab")
+                              , out  = list("cd")
+                              )
+
+    the_source = [ dict(a=1, b=4)
+                 , dict(a=2, b=5)
+                 , dict(a=3, b=6)
+                 ]
+
+    result_a = []
+    result_b = []
+    def collect(a, b):
+        result_a.append(a)
+        result_b.append(b)
+
+    the_sink = df.sink(collect, args=list("cd"))
+
+    df.push(source = the_source,
+            pipe   = the_operation(the_sink))
+
+    expected_a = [4, 5, 5, 6, 6, 6]
+    expected_b = ["4", "5", "55", "6", "66", "666"]
+    assert result_a == expected_a
+    assert result_b == expected_b
+
+
 def test_pipe():
 
     # The basic syntax requires any element of a pipeline to be passed
--- a/invisible_cities/dataflow/dataflow.py
+++ b/invisible_cities/dataflow/dataflow.py
@@ -82,6 +82,43 @@ def map(op=None, *, args=None, out=None, item=None):
     return coroutine(map_loop)
 
 
+def flatmap(op=None, *, args=None, out=None, item=None):
+    if item is not None:
+        if args is not None or out is not None:
+            raise ValueError("dataflow.flatmap: use of `item` parameter excludes both `args` and `out`")
+        assert args is None and out is None
+        args = out = item
+
+    if args is None and out is None:
+        def flatmap_loop(target):
+            with closing(target):
+                while True:
+                    for result in op((yield)):
+                        target.send(result)
+    else:
+        if _exactly_one(args):
+            args = args,
+
+        merged_output = _exactly_one(out)
+        if merged_output:
+            out = out,
+
+        def flatmap_loop(target):
+            with closing(target):
+                while True:
+                    data   = yield
+                    values = (data[arg] for arg in args)
+                    for result in op(*values):
+                        if merged_output:
+                            result = result,
+
+                        for name, value in zip(out, result):
+                            data[name] = value
+                        target.send(data)
+
+    return coroutine(flatmap_loop)
+
+
 def filter(predicate, *, args=None):
     if args is None:
         def filter_loop(target):


gonzaponte:
If you can, @jacg, please take a look.

invisible_cities/dataflow/dataflow_test.py
```diff
@@ -68,6 +68,50 @@ def the_operation(n): return n*n
     assert result == list(map(the_operation, the_source))
 
 
+
+def test_flatmap():
+
+    # Flatmap pushes downstream each element
```
jacg: Spanglish.

And a bit too cryptic.

How about something like

> flatmap accepts a function which returns an iterable. While using map with such a function would result in entire iterables being sent downstream as separate elements, flatmap concatenates the contents of these iterables producing one long stream of the individual elements contained in those iterables.

?

Maybe even explain it with code:

```python
for incoming_item in upstream:
    for outgoing_item in make_sequence(incoming_item):
        send_downstream(outgoing_item)
```

or

```python
for sequence in map(make_sequence, incoming_stream):
    for inner_item in sequence:
        send_downstream(inner_item)
```

invisible_cities/dataflow/dataflow_test.py
```diff
@@ -68,6 +68,50 @@ def the_operation(n): return n*n
...
+
+    the_source = list(range(1,11))
+
+    result = []
+    the_sink = df.sink(result.append)
+
+    df.push(source = the_source,
+            pipe   = ranges(the_sink))
+
+    assert result == list(i for x in map(the_operation, the_source) for i in x)
```
jacg: This line is pretty cryptic, making the whole test very difficult to understand.

Given that the tests are, at the moment, the *only* documentation of `dataflow`, I think that we would need to precede this test with a much clearer one, probably with hard-wired, easy to follow values, so that Joe HEP Programmer can look at it, and have the coin drop in under 30 seconds. As it stands, reading this test will melt Joe's brain.

The same critique probably applies to the other tests already present in `dataflow`, but as I wrote them, it's less obvious to me.

jacg (commented):
I would like to point out to *everyone* how immensely helpful and comforting it is, when reviewing a PR, to have the first commit contain tests (and nothing but the tests) for the functionality being added in subsequent commits. This not only gives a clear idea of what it is that the new code is *supposed* to achieve (it's an advertising brochure for the code), but running the tests also gives confidence in the code being up to the task (so it's advertising you can believe in!), thereby removing most of the paranoia about the code failing to do what it should. As such, these tests almost turn the PR into a self-reviewing one! The reviewer's job then is mostly about issues of style, design and efficiency, rather than fundamental meaning or fitness-for-purpose. This saves significant amounts of time, effort and stress.

In this PR, the third commit proposes an alternative implementation with a slightly different interface. The fact that the diff of the implementation, and the corresponding diff of the tests are there side-by-side, makes it really easy to see what the point and consequences are, while the explicit commit message gives vital motivation and explanation.

Thank you @gonzaponte !

As for the code itself ...

I appreciate the motivation for the second implementation, but it's *not* what one expects from something called `flatmap`. I would add it as a *separate* utility. Not sure what it should be called.

Speaking of names, `dataflow` is unlikely ever to be used anywhere outside of IC (its reincarnation in the form of `liquidata` might, but not `dataflow` itself), so perhaps we should choose a name that makes sense to the denizens of IC. OTOH, perhaps we should choose one of the names that tend to be used 'out there', so that people who first meet the idea here will recognize the name when they see it out in the wild. So here's a bunch of names, some of which are from 'out there' and some of which might be naturally meaningful to 'us':

`flatmap`, `join`, `chain`, `concatenate`, `sequence`, `concatmap`, `bind`, `>>=`, `=<<`, etc.

jacg:
> while we get to the point of replacing dataflow with its beefed-up version, liquidata, 

You think this might happen within our lifetimes?

(Hmm ... I think you're the only person who might make it happen ... are you up to something?)

> I intend to pay the debt soon(ish).

Which debt? You probably take out less debt here than anyone else!

gonzaponte:
> I appreciate the motivation for the second implementation, but it's not what one expects from something called flatmap. I would add it as a separate utility. Not sure what it should be called.

I am not sure what you mean by this. Would you call it differently but keep the second implementation or keep the first implementation and call it flatmap?

As I tried to explain in the commit message, I find this operator more useful and more versatile when it produces an iterable of sequences rather than a sequence of iterables. In the example depicted in the commit message, the pair of items returned cannot be decoupled easily and this is a pattern that I've encountered a few times already.

> You think this might happen within our lifetimes?
> are you up to something?
> Which debt?

I don't know if "debt" is the correct term... If we are indeed going to replace dataflow with liquidata, adding stuff to the former that is already provided by the latter, seems a touch pointless. I meant that I intend to actually work on the transition in the not-too-distant-future, which would settle the debt acquired in this PR.

jacg:
The name `flatmap` corresponds to the first implementation. If you add something called `flatmap` it should not be the second implementation. If the second implementation goes in, it should not be called `flatmap`.

On rereading the examples in the commit message, I find myself confused: `flatmap(f)` would produce a stream containing all the elements of `a` followed by all the elements of `b`: as the `a`s and `b`s are quite different things, I don't see when this would make any sense. I don't see why you would ever want to stick something like your `f` into `flatmap` Do you have a real-world example?

Re technical debt: I forgot that in `dataflow` the `args`/`out`/`item` song and dance has to be implemented over and over again for every utility, while in `liquidata` the `get`/`put`/`*`/`>>` are designed to compose with any utility that needs them. So having to do it the long-winded way, repeatedly, in `dataflow` is paying interest on the technical debt that is inherent in this shortcoming of `dataflow`. Whether or not one likes the cryptic-but-concise syntax of `liquidata`, its reusable arg/output picking/setting is a *major* advantage over `dataflow`.

Another point: the `dataflow` tests are essentially a tutorial. As such, perhaps `flatmap` shouldn't feature that early on, as it's a slightly esoteric feature. OTOH, it does make sense to have it close to `map`.

jacg:
> I meant that I intend to actually work on the transition in the not-too-distant-future, which would settle the debt acquired in this PR.

More significantly, it would settle the debt acquired by the non-reusable args/out feature in `dataflow` itself.

gonzaponte:
> On rereading the examples in the commit message, I find myself confused: flatmap(f) would produce a stream containing all the elements of a followed by all the elements of b: as the as and bs are quite different things, I don't see when this would make any sense. I don't see why you would ever want to stick something like your f into flatmap 

Sorry, I wasn't clear enough about this. I was assuming that `f` would be used with the  `args-out` syntax of dataflow like `flatmap(f, args="n", out=("a", "b"))`. In that case and using the first implementation of `flatmap`, the object passed to the next pipe downstream would be `dict(a=element_of_a, b=element_of_b)` as the output of `f` is zipped within flatmap. With the second implementation one would use `g` as `flatmap(g, args="n", out=("a", "b"))` and the result would be the same.

> Do you have a real-world example?

I do, I will open a PR later today (I hope). Here is a summary: a dataset is grouped and split into N groups according to some variable. Each of the subdatasets can be processed independently. For each subdataset, we need to provide both the subdataset and a number of values. Therefore this function that splits the dataset will produce M separate items for each of the subdatasets.

Now, with the first implementation, we would need to produce `Iterable[Sequence]` (M items of length N) (not `Iterable[Iterable]`, because it is unpractical), while with the second implementation we can produce `Iterable[Iterable]` (N items of length M) directly.
In other words, the first implementation requires computing all the subdatasets beforehand and keep them in memory while we iterate through them. The second one allows you to throw away the subdatasets are you process them.

jacg:
The explanation of the subgroups is too abstract for me to extract meaning from it. Can you provide a toy example showing the shape of the data going in and coming out? Specifically, what do you expect to do with them downstream?

gonzaponte:
For a real-world example please check the functions `split_in_z` and `split_in_xy` in [2ce5f862](https://github.com/next-exp/IC/pull/781/commits/2ce5f862b7bc88bf4560f55b8381f436e1c1bc0d) - Add create_psf_city

For instance, for `split_in_z`, the input is a `DataFrame` (the last argument). The function will go over a sequence to produce a pair of items per iteration. What I expect when I `flatmap` this function is that each of the pairs goes downstream separately with each item in the pair having its own name. This would have happened with any of the two implementations I provided for `flatmap`, but it requires slightly different implementations for `split_in_z`.

The difference comes in on how to handle that with `args-out`. For the first implementation, `split_in_z` would be written in such a way that it returns `(zs, slices)` and `flatmap` would take care of zipping them and assigning the `out` names. For me, this looks counterintuitive, as one would expect `flatmap` to concatenate `zs` with `slices`. That's why I switched to the second implementation, in which `split_in_z` returns a single item that is iterable (`zip(zs, slices)`). `flatmap` then only needs to assign names at each iteration.

Currently, these items are stored in lists and then zipped at the end to produce a sequence of pairs, but the point is that it can be done more efficiently with a generator.

jacg:
[Aside to everyone involved in developing software on our projects: *I* am the author of `dataflow` and I am saying that it is miserably awful and needs replacing: (even though it is a huge improvement on what it replaced) it contains *major* flaws. It is important to recognize this and try to improve the situation. Mea culpa! We all make mistakes. We should learn from them rather than allowing them to hold us back.]

Ugh, looking at the dataflow implementation reminds me how awful it is in terms of the need to copy-paste the implementation of the cross-cutting args/out/item concern, into *each* and every component that might ever pick args or outputs. This completely drowns out the signal of the component's fundamental behaviour, and turns the whole thing into a maintenance nightmare. I must emphasize once more: this is an *immense* advantage that `liqudata` offers over `dataflow`, and probably the major reason that drove me to writing `liquidata`. Just look at the implementation of `map` and `flat`(map) in `liquidata`:

``` python
@component
def _Map(fn):
    def map_loop(downstream):
        with closing(downstream):
            while True:
                downstream.send((fn(*(yield)),))
    return map_loop


@component
def flat(fn):
    def flat_loop(downstream):
        with closing(downstream):
            while True:
                for item in fn(*(yield)):
                    downstream.send((item,))
    return flat_loop
```

and compare it to the awful mess in dataflow: https://github.com/next-exp/IC/blob/master/invisible_cities/dataflow/dataflow.py#L50-L82 : try to find the signal among the noise in that!

In `liquidata` the implementations of `map` *and* `flat` take 15 lines in total (still about twice as long as I'd like them to be!); in `dataflow`, `map` on its own is 33 lines! And it contains more than one loop expressing the fundamental meaning of `map` (because it mixes in the concerns of picking/setting arguments/results).

In the `liquidata` version, it's pretty easy to find the salient difference between `map` and `flat`:

``` python
downstream.send((fn(*(yield)),))
```
vs.
``` python
for item in fn(*(yield)):
    downstream.send((item,))
```

In the latter, there is an inner loop which separates the elements of whatever `fn` returns, sending them downstream separately. Good luck understanding that in `dataflow`.

As for this PR: I am of the strong opinion that `flatmap` should do the equivalent of `liquidata`'s '`flat`. If you need some adjustment of how inputs/outputs are grouped/named, I'm pretty sure that should happen in some orthogonal utility, and not in `flatmap` itself.

Because of `dataflow`'s miserably awful design in this area, it is extremely difficult to understand the consequences of your juggling around with `zip` in your two alternative implementations, or indeed to understand why there's a `zip` in `map` in the first place.

Have you tried expressing your need in `liquidata`? I suspect that valuable lessons could be learned by doing so, because the far greater clarity of both framework and client code should make all sorts of things about the problem easier to understand.

I think that further discussion of this would require a more interactive medium. I am available today.

I see that #781 adds some argument picking aspects to `dataflow.reduce`. `liquidata`'s orthogonal design ensures this feature is available without anyone having to do any work at all! I think that this is the universe's way of telling us that it's time to leave `dataflow` behind. Does #781 re-use much existing `dataflow` client code? If not, perhaps it should just be written in `liquidata` in the first place?

Unlike the transition from solid to liquid cities, moving from `dataflow` to `liquidata` does not need to be done in one big seismic shift. Perhaps this is the perfect opportunity to start.

gonzaponte:
> I am of the strong opinion that flatmap should do the equivalent of liquidata's 'flat. If you need some adjustment of how inputs/outputs are grouped/named, I'm pretty sure that should happen in some orthogonal utility, and not in flatmap itself.

Me too! and I think I achieved that with the second implementation, but as far as I understand, the architecture of dataflow requires the name management to be coded into the component.

> Have you tried expressing your need in liquidata?

No, I will have to revisit the package and refresh my memory. Shouldn't take long, but it's something I have to do. [...] After skimming through the `liquidata` tests I think it would simply be written as `get.dst * flat(split_in_z) >> put.z.dst_z`, without any complications.

> I see that #781 adds some argument picking aspects to dataflow.reduce

Yes, more debt to pay...

> Does #781 re-use much existing dataflow client code?

No, every component used is for this city only.

> If not, perhaps it should just be written in liquidata in the first place?

I am not sure about how to proceed here. It would probably make sense to code it in `liquidata`, but who will review it? I guess we need two separate reviews: one for the algorithmic part and another one for the design part. What do you think?

Moreover, I would like this city to be ready on a time scale of 1 week. maybe 2. I am pretty sure it can be achieved by acquiring a lot of technical debt, but I am not so sure if we switch to `liquidata` directly. On the other hand...

>  Unlike the transition from solid to liquid cities, moving from dataflow to `liquidata` does not need to be done in one big seismic shift. Perhaps this is the perfect opportunity to start.

This could serve as a showcase for a city written in `liquidata`. I think this is important for everybody to get a feeling about `liquidata` and make the transition as smooth as possible.

jacg:
>> Does #781 re-use much existing dataflow client code?

> No, every component used is for this city only.

In which case ...

> This could serve as a showcase for a city written in liquidata. I think this is important for everybody to get a feeling about liquidata and make the transition as smooth as possible.

... if you have the time, I suggest that you translate it to `liquidata` in this PR. If we don't like it and want to stick with `dataflow`, then the `liquidata` version should be reverted, but kept in history.

invisible_cities/cities/psf_liquidata.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+    for path in paths:
+        try:
+            df = load_dst(path, group, node)
+        except tb.exceptions.NoSuchNodeError:
+            continue
+
+        yield Namespace(dst = df)
+
+def column_dropper(columns, df):
```
jacg: This (and the next 4-ish functions) seem to be written for use just here, and they are all used with `partial`. You should probably curry them.

invisible_cities/cities/psf_liquidata.py
```diff
@@ -0,0 +1,193 @@
...
+    sectors = xsectors, ysectors
+
+    nbin      = (np.diff(ranges, axis=1) / bin_size_xy).astype(int).flatten()
+    bin_edges = [np.linspace(*limits, n+1) for limits, n in zip(ranges, nbin)]
+
+    columns_to_drop     = "Xrms Yrms Qc Ec".split()
+    drop_unused_columns = get.dst    *      partial(column_dropper, columns_to_drop)   >> put.dst
+    create_slices_z     = get.dst    * flat(partial(split_in_z, zbins, ranges, sipms)) >> put.z.dst_z
+    create_sectors_xy   = get.dst_z  * flat(partial(split_in_xy, sectors, bin_edges))  >> put.x.y.dst_xy
+    compute_psf         = get.dst_xy *      partial(psf_builder, bin_edges)            >> put.factors.entries.centers
```
jacg: `liquidata`s `*` (and `>>`) could be very confusing, especially if liquidata code is interleaved with non-liquidata code using these operators in their 'normal' sense. *Perhaps* we should consider a style where the `liquidata` names (especially `get` and `put`) are imported uppercased, to give a visual warning that something unusual is happening.

You can see this style being used [here](https://jacg.github.io/liquidata/#still-with-me).

We might even add these names as synonyms inside `liquidata` itself.

invisible_cities/cities/psf_liquidata.py
```diff
@@ -0,0 +1,193 @@
...
+    pipeline = pipe( drop_unused_columns
+                   , create_slices_z
+                   , create_sectors_xy
+                   , compute_psf
+                   , build_dataframe
+                   , get.psf
+                   , out(combine_psfs, None)
+                   )
+
+    result = pipeline(dst_from_files(files_in, "RECO", "Events"))
```
jacg: Perhaps a name more descriptive than `pipeline` might be good here. OTOH, given that this is the first time many people will ever see a `liquidata` pipeline, emphasizing that it's a pipeline is a good idea. But then `pipe` already does that to a sufficient extent?

invisible_cities/cities/psf_liquidata.py
gonzaponte: I kinda like partial, but ok.

What are your thoughts on adding a `curryable` decorator? Something like
```
from inspect import signature

def curryable(f):
    total_args = len(signature(f).parameters)
    def g(*args):
        if len(args) < total_args:
            return curryable(partial(f, *args))
        else:
            return f(*args)
    returng g
```

I am afraid it might be a bit obscure to find a function _in python_ that is called with too few parameters, but it would avoid writing these functions that essentially implement a partial application.

invisible_cities/cities/psf_liquidata.py
gonzaponte: I am ok with both options. I am already familiar with `get` and `put`, perhaps we should get the view from someone less involved with this code.

invisible_cities/cities/psf_liquidata.py
gonzaponte: Since the city will eventually have a more flowery (aka cryptic) name, I would opt to rename `pipeline` such that it really indicates what is meant to be done here.

invisible_cities/cities/psf_liquidata.py
jacg: I dislike `partial` because

+ It wastes a lot of precious space
+ It puts the callable along with its arguments inside an argument list. This sends a very misleading message, and it takes careful conscious effort to suppress one's natural parsing instincts.

`curryable` looks OKish.

+ `@wraps(f)` is missing
+ I'm always wary of `inspect`, because it's easy to make assumptions which don't work out in the general case. This version completely ignores the possibility of keyword arguments. We'd need to think to what extent it is worth the effort to cover all cases.
+ `partially_applicable` would be a more correctly-descriptive name.
+ Stick it in an inner loop, and it might degrade performance noticeably.

> I am afraid it might be a bit obscure to find a function in python that is called with too few parameters

?

[Aside, point of pedantic nomenclature: '... called with too few *arguments*'. Parameters are *names*; arguments are *values*. I know that people tend to use them interchangeably, and in 70% of cases it doesn't matter (because parameters are bound to arguments and therefore represent or 'contain' them). But in the 30% of the cases where it does matter, it really helps if everyone is already in the habit of using the correct terminology.]

Anyway ... I guess you're contrasting it with 

+ `partial` which is in-your-face at the point of use
+ curried-by-hand functions (e.g. `operator.itemgetter`) which strictly separate early and late arguments

so you're worried that `curryable` (or `partially_applicable`) will confuse people because, at point of use,

+ it's excessively flexible in what it accepts
+ whether it runs or not is implicitly determined by how many arguments you pass.

invisible_cities/cities/psf_liquidata.py
jacg: Yes, I'd love to get some input from someone less involved with the code. I'd love to get any indication that we are not alone in this universe.

invisible_cities/cities/psf_liquidata.py
```diff
@@ -0,0 +1,193 @@
...
+import pandas as pd
+
+from invisible_cities.reco.psf_functions import create_psf
+from invisible_cities.reco.psf_functions import hdst_psf_processing
+
+from liquidata import flat
+from liquidata import get
+from liquidata import put
+from liquidata import pipe
+from liquidata import out
```
jacg: I don't see how you provide `liquidata`.

There are two options:

+ pip in the IC conda environment
+ add its single-file implementation to the IC repo. (The implementation has been kept in a self-contained single file, for now, exactly for the purpose of making it trivial to install in this way.)

If we were using Nix, I would recommend the `pip` approach: Nix (and `direnv`) would automatically take care of ensuring that it is installed wherever it is needed, and absent where it is not needed. Without Nix, the user has to take a much more active role in managing the conda environment. Given that the IC conda environment hasn't changed for over 18 months (?), the users have probably forgotten how to do this. (I certainly have.)

invisible_cities/cities/psf_liquidata.py
Aretno: Not sure if i'm fully eligible as 'less involved' as I already know what the code should do (but have no idea about `liquidata`) but this seems pretty straightforward and readable. I get the confusing part, though, so it could be useful to have the uppercase hint (although I guess that in a liquidata world one could infere the use of the characters once you have been warned with the `@city` decorator).

invisible_cities/cities/psf_liquidata.py
gonzaponte: > I dislike partial because [...]

I agree on both points.

> I'm always wary of inspect, because it's easy to make assumptions which don't work out in the general case. This version completely ignores the possibility of keyword arguments. We'd need to think to what extent it is worth the effort to cover all cases.

Indeed. This was just a draft to get your view on the mere idea of adding this kind of function. But also...

> Stick it in an inner loop, and it might degrade performance noticeably.

I was thinking of using this strictly for city components, so we get rid of the pattern

```
def outer(*init_args):
    ...
    def inner(*late_args):
        ...
    return inner
```

which, IMO, becomes a bit noisy. Also notice that it might not be always optimal. For instance, if `init_args` need to be massaged into something more useful, it doesn't make sense to do it for every call of the inner function.

> it really helps if everyone is already in the habit of using the correct terminology.

yeah.. I'm sorry, I've always had trouble using the correct terminology... as you probably noticed :) I'll keep trying

> ?
> so you're worried that curryable (or partially_applicable) will confuse people because, at point of use,
>
>    it's excessively flexible in what it accepts
>    whether it runs or not is implicitly determined by how many arguments you pass.

Yes, mostly that. Also, it is very easy for people (particularly those less interested in code development) to ignore the decorator and think that they don't understand the code.

invisible_cities/cities/psf_liquidata.py
gonzaponte: I didn't add it anywhere yet. For me, I just added the top directory to PYTHONPATH to be able to run the tests. For the actual repo we need to add it to `manage.sh`. I will do it now.

invisible_cities/cities/psf_liquidata.py
jacg: I don't think that `@city` is enough warning, because within city code one can find uses of `*` which mean multiplication, so inside city code we might end up having such lines interspersed with lines on which `*` means `liquidata` argument selection.

It needs to be clear to the reader that we're in a `liquidata` context rather than an arithmetic (or other normal) context.

But I have no strong opinions on whether uppercasing `liquidata` names helps.

jacg:
>  we need to add it to manage.sh. I will do it now.

... and if we find something missing in `liquidata` this would require cutting a new release of `liquidata`, publishing it on PyPI and changing `manage.sh` again, with its concomitant need to create yet another conda env. I'd go with including the `liquidata` source in the IC repo, for the time being.

invisible_cities/cities/psf_liquidata.py
jacg: > I was thinking of using this strictly for city components, so we get rid of the pattern
> 
> ``` python
> def outer(*init_args):
>     ...
>     def inner(*late_args):
>         ...
>     return inner
> ```
> 
> which, IMO, becomes a bit noisy.

Agreed: this pattern is a noisy PITA, and it would be good to abstract it. If only we had Lisp macros. Which we do in Julia.

> it is very easy for people (particularly those less interested in code development) to ignore the decorator and think that they don't understand the code.

And you think that the nested-def pattern is better in this respect?

gonzaponte:
> I'd go with including the liquidata source in the IC repo, for the time being.

... and make the necessary changes to liquidate here instead of the main repo?

jacg:
Exactly. This allows us to experiment, test, modify, and adapt `liquidata` easily. Once we have changes that we believe in, they can be copied into the official `liquidata` repo. Otherwise, *every single* change to `liquidata` that we want to test in IC, would require

1. Release new version of `liquidata`
2. Publish it on PyPI
3. Modify IC's `manage.sh`
4. Create new IC conda env

invisible_cities/cities/psf_liquidata.py
gonzaponte: In my experience, nested functions seem to create some confusion as well. In practice, if someone wants to write a city, they will probably get some IC 101 course anyway so perhaps it doesn't matter. I think a simple demonstration of how `partially_applicable` works is as good as explaining nested-defs.

gonzaponte:
ok. How about this:
1. We proceed with dataflow in the current PRs.
2. We add a new branch to IC where we remove dataflow and include the `liquidata` source file. Here we begin translating the cities. If we find the need to modify `liquidata` in the process, we do it there.
3. Once we are happy with the full translation and the branch can be merged into master, we make a new release of `liquidata` (if necessary) and replace the source file with a `pip` installation.
4. We merge the branch and we live happily ever after.

jacg:
Sounds like an excellent plan.

gonzaponte:
Ok, so for this PR... Is there any point in adding `liquidata` for the demonstration? I would simply revert the implementation of the city, keeping it in the commit history. That aside, is there anything left to be done?

jacg:
I don't think that the liquidata implementation of the psf city should be on this branch at all! Not even reverted. Here we are simply adding some capabilities to `dataflow`.

If you're going to stick a reverted `liquidata` version of the city anywhere (other than the let's-rewrite-everything-in-liquidata branch), it should be on that city's branch (which is why I was so confused by failing to find it there).

BTW, if you do put it somewhere, please note that `liquidata` provides `on` as a shorthand for getting and putting the same name (a bit like `dataflow`'s `item` parameter in `map` etc.). Thus, the following two lines are equivalent

``` python
 on.dst(  fn)
get.dst * fn >> put.dst
```

Whilst we're at it, let's note that `liquidata` tries to be very flexible about letting you emphasize what you want, so the following all pass the same test

``` python
get.a.b   *  hard_work >> put.x.y
hard_work *  get.a.b   >> put.x.y
put.x.y   << hard_work *  get.a.b
```

As for finishing this PR:

+ We agreed (offline) that your second implementation is the only one that makes sense. So remove the first implementation.
+ You've cleaned up the explanations/comments/tests.
+ I can't see anything else outstanding and relevant above (I don't appear to have any 'resolve conversation' buttons).

invisible_cities/dataflow/dataflow_test.py
```diff
@@ -68,6 +68,73 @@ def the_operation(n): return n*n
...
+    result_a = []
+    result_b = []
+    def collect(a, b):
+        result_a.append(a)
+        result_b.append(b)
+
+    the_sink = df.sink(collect, args=list("cd"))
+
+    df.push(source = the_source,
+            pipe   = the_operation(the_sink))
```
jacg, line 189: Why not `pipe(the_operation, the_sink)`? Isn't that more natural for `dataflow` users?

Oh, it's because `pipe` hasn't been introduced yet, at this point in the file? Hmm, another reason why `flatmap` shouldn't be discussed this soon. How about just after `test_filter`?

invisible_cities/dataflow/dataflow_test.py
```diff
@@ -68,6 +68,73 @@ def the_operation(n): return n*n
...
+
+    expected_a = [4, 5, 5, 6, 6, 6]
+    expected_b = ["4", "5", "55", "6", "66", "666"]
```
jacg: Align the a-elements above their corresponding b-elements.

jacg:
CI tests fail because of MySQL server timeout, again ... this is getting veeeeery irksome.

gonzaponte:
This issue has been mentioned in a software meeting and @jmbenlloch is going to try and increase timeouts and things like this to solve the problem. If this doesn't work I'm afraid that we will have to skip the test. Right now this is not critical. If there is an issue with the database connection we will notice it by other means.

jacg:
I'm probably ready to approve and merge. Do you want to do anything before I go ahead?

gonzaponte:
If you want I can squash the last two commits into the first one. Otherwise, go ahead.

jacg:
I'll do it.

jacg (approved):
The conceptually difficult part of this PR was discussed con peras, offline.

A `flatmap` utility is added to `dataflow`. Ideally, rather than adding this to `dataflow`, we would switch from `dataflow` to `liquidata` which already provides the feature. But we took this route to expedite the inclusion of #781.

PR #781, opened 2021-04-04
Add city to build psfs

gonzaponte:
Introduce a city to build psfs. Original code from @Aretno, I just reshaped it into a city.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -549,6 +549,27 @@ def hits_and_kdst_from_files(paths: List[str]) -> Iterator[Dict[str,Union[HitCol
                            timestamp = timestamp)
 
 
+def dst_from_files(paths: List[str], group: str, node:str) -> Iterator[Dict[str,Union[pd.DataFrame, int, np.ndarray]]]:
+    """
+    Reader for a generic dst.
+    """
+    for path in paths:
+        try:
+            df = load_dst(path, group, node)
+            with tb.open_file(path, "r") as h5in:
+                run_number  = get_run_number(h5in)
+                evt_numbers = get_event_info(h5in).col("evt_number")
+                timestamps  = get_event_info(h5in).col("timestamp")
+        except (tb.exceptions.NoSuchNodeError, IndexError):
+            continue
+
+        yield dict( dst           = df
+                  , run_number    = run_number
+                  , event_numbers = evt_numbers
+                  , timestamps    = timestamps
+                  )
+
+
 def MC_hits_from_files(files_in : List[str], rate: float) -> Generator:
     timestamp = create_timestamp(rate)
     for filename in files_in:
--- a/invisible_cities/cities/eutropia.py
+++ b/invisible_cities/cities/eutropia.py
@@ -1,0 +1,221 @@
+"""
+-----------------------------------------------------------------------
+                                Eutropia
+-----------------------------------------------------------------------
+
+From ancient Greek, ευτροπία: docile, of good manners, versatile, variable.
+
+This city processes hits to produce a Point Spread Function (PSF).
+The PSF is obtained by mapping the spatial charge distribution
+relative to the barycenter (center of gravity of SiPM response)
+for pointlike events collapsed along the drift coordinate. This
+charge distribution is averaged over a large number of events.
+The PSF can be obtained for different sections of the active
+volume independently.
+
+The tasks performed are:
+    - Load hits processed for psf configuration
+    - Split data in z-sections
+    - Process each slice (charge normalization)
+    - Split each slice in xy-sectors
+    - Create PSF for each sector (binned charge average)
+    - Accumulate the PSFs to produce an overall PSF
+    - Write the PSF
+
+The overall PSF is table that describes a collection of discrete
+matrices organized according to these variables:
+    - x, y, z: coordinates of the center of each xy-sector and slice
+
+Each triplet of values is a PSF on its own. A PSF is a table containing:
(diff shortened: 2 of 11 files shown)


jacg:
Erm ... commit 237e32a, the one which adds the liquidata version ... *where* is it? I cannot find it on the `gonzaponte:psfcity` branch, be it via Magit, git CLI or the GitHub web interface.

gonzaponte:
It's in gonzaponte:flatmap. This banch hasn't been rebased yet on top of that. From what I understood [here](https://github.com/next-exp/IC/pull/780#issuecomment-813504031) this is what you wanted.

jacg:
... and it appears *here* because it references this PR. OK. And as it's an implementation of a psf city, the `psfcity` branch seemed the logical place for it to be. Hmm.

invisible_cities/config/create_psf_city.conf
```diff
@@ -0,0 +1,14 @@
+files_in    = '$ICDIR/database/test_data/kr_hits_for_psf.h5'
+file_out    = '/tmp/psf.h5'
+compression = 'ZLIB4'
+
+run_number    = 8000
+detector_db   = 'new'
```
Aretno: I don't see any reason to have that many empty spaces before `=`, dunno if i'm missing something.

invisible_cities/cities/create_psf.py
```diff
@@ -0,0 +1,181 @@
...
+                                      , accumulate_psf.sink
+                                      )
+                    , result = accumulate_psf.future
+                    )
+
+    with tb.open_file(file_out, "w") as outfile:
+        write_psf = psf_writer(outfile, compression=compression)
+
+        # TODO: Probably better to ignore this writer
+        # and use df_writer directly
```
Aretno: Agreed, as you already have the dataframe created in `combine_psfs` you might as well use the dataframe writer.

invisible_cities/cities/create_psf.py
```diff
@@ -0,0 +1,181 @@
...
+    return split_in_z
+
+
+def xy_splitter(sectors, bin_edges):
+    def split_in_xy(dst):
+        xsectors, ysectors = sectors
+        for xmin, xmax in zip(xsectors, xsectors[1:]):
+            # TODO: replace floor with something more robust
+            xpsf = np.floor((xmin + xmax) / 2)
+            selx = in_range(dst.X, xmin, xmax)
```
Aretno: Selection should be made based on Xpeak (and Ypeak a couple lines later)

invisible_cities/cities/create_psf.py
```diff
@@ -0,0 +1,181 @@
...
+                     , factor
+                     , entries
+                     )
+
+    return result
+
+
+def column_dropper(columns):
+    def dropper(df):
+        # TODO: is reset_index really needed?
```
Aretno: I think you are right and it's not needed.

invisible_cities/cities/create_psf.py
```diff
@@ -0,0 +1,181 @@
...
+        return df.drop(columns=columns).reset_index(drop=True)
+    return dropper
+
+
+def z_splitter(zbins, ranges, sipms):
```
Aretno: I would change the name of this function, as it not only splits the data but also does some processing. Actually, `hdst_psf_processing` does not need to necessarily happen post z-splitting so you could have completely separate function doing the processing before doing the splitting. The caveat was that hdst_psf_processing uses groupby and memory comsumption could be delicate. Still as in a city this is executed file by flile it should be ok. I'm more inclined of having separate functions but at the very least go for another name.

invisible_cities/cities/create_psf.py
```diff
@@ -0,0 +1,181 @@
...
+    def split_in_z(dst):
+        for zmin, zmax in zip(zbins, zbins[1:]):
+            zpsf = (zmin + zmax) / 2
+
+            sel  = in_range(dst.Z, zmin, zmax)
+            if not np.any(sel): continue
+
+            processed = hdst_psf_processing(dst[sel], ranges, sipms)
+
+            # TODO: is this really needed?
```
Aretno: I don't think it's needed, nope.

Aretno (changes requested):
Mostly ready to go, aside from some minor fixes (many derived from the original code) I think it would be good to store the run info (or MC info in case of a MC production) in the output file. Additionally, a test checking that two files are properly combined would also be desireable.

gonzaponte:
> I think it would be good to store the run info (or MC info in case of a MC production)

We agreed offline that since this is not done (yet) for any other city and there are no tools to do it, we leave it for another time.

> a test checking that two files are properly combined would also be desireable.

I've modified the exact-result test so it takes a case in which we split in all three coordinates and that runs over three files to make sure we check every possible feature.

I've marked all the other requests as resolved.

Aretno (commented):
All good, just name the city (I'm inclined to go with Eutropia) and this is ready to go.

Aretno (approved):
This PR marks the, long overdue, formalization as a city, Eutropia, of the PSF extraction procedure. It takes penthesilea output (with a big enough rebinning so each sensor charge is fully integrated over time) to produce a set of PSFs, each one corresponding to a different region in the detector. Such regions are configurable. The PSFs are composed of their region position, the relative position sensor-event, the charge fraction seen by the sensor and the number of events used in the computation of such fraction.

Great job adapting the code @gonzaponte.

gonzaponte:
I've just added some type hints and improved alignment here and there. Please merge when we get the green flag from the GHA.

PR #784, opened 2021-04-06
Skip database test which causes too many spurious failures

jacg:
This test has been responsible for far too many spurious failures of CI runs, thereby significantly increasing the workload of managing PRs.

Here we skip the test, but this should be reversed as soon as the problem is fixed at the server.

### DIFF
--- a/invisible_cities/database/db_connection.py
+++ b/invisible_cities/database/db_connection.py
@@ -1,3 +1,5 @@
+from pytest import mark
+
 import sqlite3
 import pymysql
 pymysql.install_as_MySQLdb()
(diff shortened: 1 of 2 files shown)


jmbenlloch (approved):
Until a better solution can be found. These database tests should be skipped so other PRs work properly.

jacg:
@carmenromo @halmamol @MiryamMV That's your cue!

Remember that #780 is ready to go in on top of this.

carmenromo:
@jacg , should I merge or better wait to train @halmamol and @MiryamMV ?

jacg:
Use this to train them!

carmenromo:
Ok, we are meeting this afternoon, can this PR (and the next ones) wait until then?

jacg:
Given that most of them have been ignored for weeks or months, I'm sure that another day won't make much difference. The only one which doesn't fall into this category is #781 and I'm confident that @gonzaponte and @Aretno can finish it off with #780 unmerged.

mmkekic:
Now it is me putting comments in inappropriate places, but how about @carmenromo  adds  merge instructions in docs and  @jacg reviews it? That way the merge commit message style will not be just an unwritten rule passed as oral tradition :)

carmenromo:
Yes, I think that's a good idea, I have @jacg's instructions (not only oral tradition :)) from some years ago and I can adapt them.

jacg:
Yes, but ...

1. Isn't it already written somewhere?
2. Whether it's written down or not, teaching it to newbies should be done in person, rather than by telling them to read some document.

carmenromo:
> Isn't it already written somewhere?

it isn't...

PR #785, opened 2021-04-07
Allow numpy to be used in config files

gonzaponte:
Usually, cities don't take sequences of numbers as arguments, so there was no need to have NumPy in the namespace used to read config files. Probably #781 is the first case where it is interesting in practice. For tests and demo purposes we can live without it, but in realistic cases, we might want to do something like `np.linspace(0, 100, 17)`. The pure-python equivalent would be more difficult to read, so I propose we introduce NumPy as a valid package in config files.

This is a very simple and short PR. Maybe it would be good for someone that wants to get involved in IC to review it.

### DIFF
--- a/invisible_cities/core/configure.py
+++ b/invisible_cities/core/configure.py
@@ -2,6 +2,7 @@
 JJGC August 2016
 """
 import argparse
+import numpy
 import sys
 import os
 
@@ -89,6 +90,8 @@ def make_config_file_reader():
 
     builtins = __builtins__.copy()
     builtins.update(vars(units))
+    builtins["np"] = numpy
+
     # TODO: move setting of extra 'builtins' elsewhere
     builtins['all']  = EventRange.all
     builtins['last'] = EventRange.last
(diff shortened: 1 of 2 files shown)


jacg:
Aaaah, `config.py`,  {\strong_irony one of my *favourite* parts of IC}.

I find the suggestion that this might a good place for a newcomer to start, to be comically cruel ;-)

True, it's a simple change, but the meaning of the whole module is completely impenetrable and the whole thing is so completely different from anything else in IC, that it doesn't seem appropriate as an introduction. OTOH, raising its Bus Number is probably a good idea!

And the test and implementation *are* very clear, so maybe it could be done by a novice.

I'll wait an hour or two: if there are no volunteers to have a go, I'll press the approve button on my pending review. Or I could do it sooner, if the new mergers are chomping at the bit to have something to merge.

gonzaponte:
> OTOH, raising its Bus Number is probably a good idea!

Yeah, that was also my intention. Maybe not for someone completely new, but for someone with some experience in IC, but not so much in reviewing or these obscure parts of the repo :)

jacg:
Yeah, it's a weird one this. Reviewing this PR doesn't really require understanding much about the module. (Ugh! I feel dirty saying that!)

OTOH, if someone were to raise the Bus Number, it would have to be someone interested in Python gymnastics, and would take much more effort then merely reviewing this PR.

jacg (approved):
+ Clear test demonstrates lack of the missing feature.

+ Simple implementation which
   - looks sensible
   - passes the test

LGTM

PR #787, opened 2021-04-23
Upload db

carmenromo:
This PR includes the latest sensor calibrations from run 8593. Database has been uploaded with the PMT and SiPMs gains and the pdf values. Sensor 15060 is seeing light again, so it has been unmasked.

The channel 21004 shows a suspect spectrum and will be followed closely in the next calibrations.

Plots comparing with other stable runs are attached.

sipmRunDifferencePlots_test

pmtRunDifferencePlots

Captura de pantalla 2021-04-23 a las 16 11 05

PR #788, opened 2021-04-23
Fix database github action

jmbenlloch:
For some reason the Github Action we use for calibration PRs is [not working anymore](https://github.com/next-exp/IC/pull/787/checks?check_run_id=2421133790). Apparently `pytest`was available by default before with the action `actions/setup-python@v2` but it is not anymore. This PR forces the installation of `pytest`.

I have already tested this change in my fork. I have included this commit in my fork's master and then submitted the same PR that is failing on IC (#787). In my fork, the [test passes without problem](https://github.com/jmbenlloch/IC-1/pull/7).

### DIFF
--- a/.github/workflows/check_db_calibration.yml
+++ b/.github/workflows/check_db_calibration.yml
@@ -33,7 +33,7 @@ jobs:
         git lfs pull
     - name: Install dependencies
       run: |
-        pip install pymysql numpy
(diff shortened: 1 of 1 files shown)


carmenromo (approved):
`pytest` has been added to the list of packages of the Github action. Needed for the calibration checks.

PR #791, opened 2021-05-05
Update demopp db

carmenromo:
DEMO++ database has been updated according to the last sensor calibration constants. Pdfs have been added too.


jmbenlloch (approved):
Latest calibration added with all test passing.

PR #793, opened 2021-05-17
Update NEWDB

carmenromo:
This PR includes the latest sensor calibrations from run 8634. Database has been uploaded with the PMT and SiPMs gains and the pdf values. Sensor 15060 keeps behaving suspicious, so we will leave it masked from now on.

Plots comparing with other stable runs are attached.

sipmRunDifferencePlots_test

pmtRunDifferencePlots


jmbenlloch (approved):
Latest calibration added.

PR #794, opened 2021-05-20
Check `max_time` and change `max_subevt` in Buffy and Detsim cities

### LINKED ISSUE #792: Buffy/detsim  max_time and buffer_length
Parameter `max_time` in both buffy and detsim might be confusing, it is the maximal length of the event that will be taken into account starting from the first detected signal, all signals after that are simply lost. 
As is now, it is possible to set that parameter to any value unrelated to buffer length. I would suggest that in case user given `max_time` is smaller than user given `buffer_length` we set `max_time` to `buffer_length`and throw a warning with message:
 `"max_time shorter than buffer_length, setting max_time to buffer_length"` 

Also, the `max_subevent` uses the `max_time` to set maximal number of splitting per event (this allows us to keep `event_id`unique across full nexus MC production)
https://github.com/next-exp/IC/blob/0cf3acff45925268bde860db0132bab3c51a259d/invisible_cities/cities/components.py#L960
However, the correct dependency should be 
`int(math.ceil(max_time/buffer_length))` that would keep `event_id` equal to nexus one in case `max_time == buffer_length`

MCruces-fz:
Check  `max_time` is always greater or equal than `buffer_length` and modificate `max_subevt` to match nexus `event_id`.

Fixes #792

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -20,6 +20,7 @@ import numpy  as np
 import pandas as pd
 import inspect
 import warnings
+import math
 
 from .. dataflow                  import                  dataflow as  fl
 from .. dataflow.dataflow         import                      sink
@@ -920,6 +921,27 @@ def compute_and_write_pmaps(detector_db, run_number, pmt_samp_wid, sipm_samp_wid
     return compute_pmaps, empty_indices, empty_pmaps
 
 
+def check_max_time(max_time: Union[int, float], buffer_length: float) -> Union[int, float]:
+    """
+    `max_time` must be greater than `buffer_length`. If not, raise warning 
+        and set `max_time` == `buffer_length`.
+
+    :param max_time: Maximal length of the event that will be taken into 
+        account starting from the first detected signal, all signals after 
+        that are simply lost.
+    :param buffer_length: Length of buffers.
+    :return: `max_time` if `max_time` >= `buffer_length`, else `buffer_length`.
+    """
+
+    if max_time < buffer_length:
+        warnings.warn("`max_time` shorter than `buffer_length`, "
+                      "setting `max_time` to `buffer_length`", 
+                      stacklevel=2)
+        return buffer_length
+    else:
+        return max_time
+
+
 def calculate_and_save_buffers(buffer_length    : float        ,
                                max_time         : int          ,
                                pre_trigger      : float        ,
@@ -957,7 +979,7 @@ def calculate_and_save_buffers(buffer_length    : float        ,
                                out  = "buffers"                            )
 
     saved_buffers = "buffers" if order_sensors is None else "ordered_buffers"
-    max_subevt    =  max_time // buffer_length + 1
+    max_subevt    =  math.ceil(max_time / buffer_length)
     buffer_writer_    = sink(buffer_writer( h5out
                                           , run_number = run_number
                                           , n_sens_eng = npmt
--- a/invisible_cities/cities/detsim.py
+++ b/invisible_cities/cities/detsim.py
@@ -23,6 +23,7 @@ from . components import copy_mc_info
 from . components import collect
 from . components import calculate_and_save_buffers
 from . components import MC_hits_from_files
+from . components import check_max_time
 
 from .. reco     import tbl_functions as tbl
 from .. database import load_db       as db
@@ -123,7 +124,10 @@ def detsim(*, files_in, file_out, event_range, print_mod, compression,
            detector_db, run_number, s1_lighttable, s2_lighttable, sipm_psf,
            buffer_params, physics_params, rate):
 
+    buffer_params_  = buffer_params .copy()
     physics_params_ = physics_params.copy()
+
+    buffer_params_["max_time"] = check_max_time(buffer_params_["max_time"], buffer_params_["length"])
 
     ws    = physics_params_.pop("ws")
     el_dv = physics_params_.pop("el_drift_velocity")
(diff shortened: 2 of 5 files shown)


invisible_cities/cities/components.py
```diff
@@ -920,6 +922,26 @@ def compute_and_write_pmaps(detector_db, run_number, pmt_samp_wid, sipm_samp_wid
     return compute_pmaps, empty_indices, empty_pmaps
 
 
+def check_max_time(max_time: int, buffer_length: float) -> Union[int, float]:
```
gondiaz: Couldn't `max_time` be a float?

invisible_cities/cities/components.py
```diff
@@ -920,6 +922,26 @@ def compute_and_write_pmaps(detector_db, run_number, pmt_samp_wid, sipm_samp_wid
...
+        that are simply lost.
+    :param buffer_length: Length of buffers.
+    :return: `max_time` if `max_time` >= `buffer_length`, else `buffer_length`.
+    """
+
+    if max_time < buffer_length:
+        warnings.warn("`max_time` shorter than `buffer_length`, "
+                      "setting `max_time` to `buffer_length`", 
+                      stacklevel=2)
+        max_time = copy.deepcopy(buffer_length)
```
gondiaz: This is not necessary, since `max_time` is not a mutable object. Mutable objects are lists and dictionaries. You can find more info on python mutability here [python-data-model](https://docs.python.org/3/reference/datamodel.html).
Then `max_time = buffer_length` should be correct.

invisible_cities/cities/components.py
```diff
@@ -957,7 +979,7 @@ def calculate_and_save_buffers(buffer_length    : float        ,
                                out  = "buffers"                            )
 
     saved_buffers = "buffers" if order_sensors is None else "ordered_buffers"
-    max_subevt    =  max_time // buffer_length + 1
+    max_subevt    =  math.ceil(max_time / buffer_length)
```
gondiaz, line 982: It is correct, but since `math` library is not used elsewhere, I suggest to use `numpy`, such that:
`int(np.ceil(max_time / buffer_length)`.

invisible_cities/cities/components.py
```diff
@@ -20,6 +20,8 @@
 import pandas as pd
 import inspect
 import warnings
+import math
```
gondiaz, line 23: This library is no needed if you use `numpy`

invisible_cities/cities/components.py
```diff
@@ -20,6 +20,8 @@
...
+import copy
```
gondiaz: This library is no needed

gondiaz (changes requested):

invisible_cities/cities/components.py
mmkekic: I think in this case I prefer `math` since it also does type conversion (in Python 3)

invisible_cities/cities/components.py
MCruces-fz: You're right, it could

invisible_cities/cities/components.py
MCruces-fz: Nice, it's very interesting

invisible_cities/cities/components.py
MCruces-fz: Okay, so I'll keep `math.ceil()`, right?

invisible_cities/cities/detsim.py
```diff
@@ -123,6 +124,8 @@ def detsim(*, files_in, file_out, event_range, print_mod, compression,
            detector_db, run_number, s1_lighttable, s2_lighttable, sipm_psf,
            buffer_params, physics_params, rate):
 
+    buffer_params["max_time"] = check_max_time(buffer_params["max_time"], buffer_params["length"])
```
gondiaz: We should be more cautious here: `buffer_params` is a dictionary, hence a mutable object. Now you are modifying `buffer_params` inside the function, as it is the case for `physics_params`. In order to avoid unexpected results, I would also duplicate `buffer_params`.

invisible_cities/cities/components.py
gondiaz: yes

gondiaz (changes requested):

invisible_cities/cities/detsim.py
MCruces-fz: Right, in this specific case it is better to do `deepcopy`

invisible_cities/cities/detsim.py
```diff
@@ -123,6 +125,8 @@ def detsim(*, files_in, file_out, event_range, print_mod, compression,
            detector_db, run_number, s1_lighttable, s2_lighttable, sipm_psf,
            buffer_params, physics_params, rate):
 
+    buffer_params["max_time"] = copy.deepcopy(check_max_time(buffer_params["max_time"],
```
gondiaz: This is wrong. Let me show an example of what goes on with mutable objects like this:
```
def func(a_dict):
    a_dict["b"] = "world"

d = {"a": "hello"}
func(d)
print(d)
```

As you can see, after calling func you get `d` modified. This is because dictionarys are mutable objects: inside the funcion you are not really having a new `a_dict` variable, but the passed `d`.

Then you may notice that your change does nothing different from what it was implemented before, `buffer_params` is still the same dictionary and is being modified inside the function, which we want to avoid. 
If you look at the next line (130) you can see how this is avoided for `physics_params`, which is also a dictionary. I wanted you to apply this same approach, ie creating a new dictionary inside the function itself, being a copy of the original one.

gondiaz (changes requested):

invisible_cities/cities/detsim.py
MCruces-fz: Okay, we can do the same as in the lines below
```python
buffer_params_ = buffer_params.copy()
buffer_params["max_time"] = check_max_time(buffer_params_["max_time"], buffer_params_["length"])
```
and could this be better too?
```python
def check_max_time(max_time, buffer_length):
    """ Maybe this is better to avoid 
        substitution in all cases """
    if max_time < buffer_length:
        ## WARNING ##
        return buffer_length
    else:
        return max_time
```

gondiaz: Yes, it is better.

Caution here:
```
buffer_params_ = buffer_params.copy()
buffer_params["max_time"] = check_max_time(buffer_params_["max_time"], buffer_params_["length"])
```
must be
```
buffer_params_ = buffer_params.copy()
buffer_params_["max_time"] = check_max_time(buffer_params_["max_time"], buffer_params_["length"])
```

MCruces-fz: Ah, right. But if we do
```python
buffer_params_["max_time"] = check_max_time(...)
```
then we need to update the main dictionary `buffer_params`:
```python
buffer_params = buffer_params_
```
Something like this?

gondiaz: Why?

MCruces-fz: Because I was still confused. @mmkekic helped me realize what is done with the other mutable objects in the same case. I was trying to update de ditcionary `buffer_params` with the new values in `buffer_params_`, but it's better if we use `buffer_params_` in the rest of the code.
Am I right now?

invisible_cities/cities/detsim.py
```diff
@@ -14,6 +14,7 @@
 """
 
 import os
+import copy
```
gondiaz: This library is not used

invisible_cities/cities/detsim.py
gondiaz: It is good now, it is more consistent to use `buffer_params` and `physics_params` in the same footing. Also, I would put the `.copy()` lines one after the other.

gondiaz (changes requested):
Please, rewritte commit history to something like:
- add check_max_time protection funcion and test
- add max_time protection in detsim
- add max_time protection in buffy

I will approve the PR afterwards.

gondiaz (approved):
This PR provides a protection against signal loss by an unwarned user about the meaning of `max_time` parameter, as stated in #792, and also modifies event numbering to match nexus-event to IC-event in the case of `max_time` == `buffer_length`. I approve it, good work.

PR #795, opened 2021-05-31
Add Valdrada, the trigger simulation city

Aretno:
Adds valdrada, the trigger which simulates the trigger in the FPGA.

This city takes rwf and produces a dataframe with the simulated trigger information. It takes a configurable set of parameters that imitate the detector acquisition ones and their naming convention. It provides all possible trigger candidates above a given width and flags them as valid or not based on the configuration.

Details of the output information are in the city main file but basically consist  on the values used for triggering in the FPGA. Aside from the trigger implementation itself, a modified BLR algorithm that imitates the behavior of the one in the FPGA has been added.

This notebook ilustrates the flow of Valdrada and shows how the less immediate parameters impact the trigger results.
https://github.com/Aretno/Notebooks/blob/master/TriggerSim/Valdrada.ipynb 

The implementation has been based on extensive discussion with Raúl and has been throughly tested prior to the PR. This was done jointly with Raúl, evaluating the FPGA output he obtained with the output of this implementation. Full agreement was found between the fpga simulation and the python implementation. A discrepance of 0.16% in the total number of valid triggers was found between already acquired data and data passing the trigger simulation (done for several runs). This discrepance also appeared in the FPGA simulation which is a copy-paste of the existing code. Such discrepance was traced to the high pass filter, which starts with a different value in the online acquisition (depends on previous time bins, which we do not stored) and yielded a different starting value for the coefficients. This could cause an ADC deviation which modify the overall result in edge cases.

Moreover, the tool has already been used for extensive trigger analysis with no visible deviations from the expected behavior.

### DIFF
--- a/invisible_cities/cities/valdrada.py
+++ b/invisible_cities/cities/valdrada.py
@@ -1,0 +1,172 @@
+"""
+-----------------------------------------------------------------------
+                                 valdrada
+-----------------------------------------------------------------------
+
+This city finds simulates the trigger procedure at the FPGA over PMT waveforms.
+This includes a number of tasks:
+    - Use a moving average to obtain the baseline
+    - Truncate to integers the deconvolution output
+    - Find trigger candidates and compute their characteristics, taking into
+      account the unique circumstances derived from working on a continuous scheme
+      (for example, delay in signals)
+    - Evaluate coincidences between trigger candidates.
+
+A number of general configuration parameters are needed (input as a dict, trigger_config):
+- coincidence_window : Time window (in time bins) in which valid trigger coincidences are counted
+- discard_width      : Any trigger with width less than this parameter is discarded.
+- multipeak          : Dictionary with multipeak protection parameters, otherwise None.
+     - q_min      : Integrated charge threshold of a post-trigger peak to discard
+                    a trigger due to multipeak protection. In ADC counts.
+     - time_min   : Minimum width of a post-trigger peak to discard a trigger due to
+                    multipeak protection. In time bins.
+     - time_after : For how long is the multipeak protection evaluated after a
+                    valid trigger. In mus.
+
+A individual trigger configuration can be given per channel, through a dict
+with keys equal to PMT IDs, which marks the validity range of the peak
+characteristics:
+- q_min, q_max          : Range for the integrated charge of the peak (q_min < q < q_max).
+                          In ADC counts.
+- time_min, time_max    : Range for the peak width (time_min <= width < time_max).
+                          In time mus.
+- baseline_dev, amp_max : Range for peak height (baseline_dev < height < amp_max).
+                          In ADC counts.
+- pulse_valid_ext       : Time allowed for a pulse to go below baseline_dev. In ns.
+
+The result of the city is a dataframe containing the event ID and PMT ID of each
+trigger candidate. For each trigger candidate a number of parameters is computed:
+ - trigger_time : time bin at which the trigger candidate starts.
+ - q            : integrated ADC counts within the peak.
+ - width        : Length of the peak in time bins.
+ - height       : Maximum ADC values in a given time bien within the peak.
+ - baseline     : Value of the baseline at trigger_time.
+ - max_height   : Maximum (minimum due to PMT negative signals) height in the wvf.
+
+Additionally, a set of flags are assigned depending on wether the parameters are
+within range of the trigger configuration:
+ - valid_q    : If peak is within q range.
+ - valid_w    : If peak is within width range.
+ - valid_h    : If peak is within height range.
+ - valid_peak : Only if 'multipeak' is active, True if there isn't a post-trigger
+               candidate with the configuration parametersself.
+ - valid_all  : boolean and of previous valid flags.
+
+ Finally, a series of coincidence-related values are also given:
+ - n_coinc       : Number of valid triggers within the coincidence_window,
+                   starting at the trigger trigger_time and including the trigger itself.
+                   -1 indicates no valid trigger (including the trigger itself).
+ - closest_ttime : Time difference to closest valid trigger.
+                   -1 if there are none aside from the trigger itself.
+ - closest_pmt   : PMT ID of the closest valid trigger.
+                   -1 if there are none aside from the trigger itself.
+"""
+
+import tables as tb
+import numpy  as np
+
+from .. reco                  import tbl_functions        as tbl
+from .. io  .run_and_event_io import run_and_event_writer
+from .. io  .trigger_io       import       trigger_writer, trigger_dst_writer
+
+from .. dataflow            import dataflow as fl
+from .. dataflow.dataflow   import push
+from .. dataflow.dataflow   import pipe
+from .. dataflow.dataflow   import sink
+
+from .  components import city
+from .  components import print_every
+from .  components import collect
+from .  components import copy_mc_info
+from .  components import deconv_pmt_fpga
+from .  components import WfType
+from .  components import wf_from_files
+from .  components import get_number_of_active_pmts
+
+from .. reco.trigger_functions import retrieve_trigger_information
+from .. reco.trigger_functions import get_trigger_candidates
+from .. reco.trigger_functions import check_trigger_coincidence
+
+from .. core       import system_of_units as units
+
+
+def check_empty_trigger(triggers) -> bool:
+    """
+    Filter for valdrada flow.
+    The flow stops if there are no candidate triggers/
+    """
+    return len(triggers) > 0
+
+@city
+def valdrada(files_in, file_out, compression, event_range, print_mod, detector_db, run_number,
+             trigger_config = dict(), channel_config = dict()):
+
+    # Change time units into number of waveform bins.
+    if trigger_config['multipeak'] is not None:
+        trigger_config['multipeak']['time_min'  ] /= 25*units.ns
+        trigger_config['multipeak']['time_after'] /= 25*units.ns
+    for k in channel_config.keys():
+        channel_config[k]['time_min']       /= 25*units.ns
+        channel_config[k]['time_max']       /= 25*units.ns
+        channel_config[k]['pulse_valid_ext'] = int(channel_config[k]['pulse_valid_ext']/25*units.ns)
+
+    #### Define data transformations
+    # Raw WaveForm to deconvolved waveforms
+    rwf_to_cwf           = fl.map(deconv_pmt_fpga(detector_db, run_number, list(channel_config.keys())),
+                                  args = "pmt",
+                                  out  = ("cwf", "baseline"))
+
+    # Extract all possible trigger candidates of each PMT
+    trigger_on_channels  = fl.map(retrieve_trigger_information(channel_config, trigger_config),
+                                  args = ("pmt", "cwf", "baseline", "event_number"),
+                                  out  = "triggers")
+
+    # Add coincidence between channels
+    check_coincidences   = fl.map(check_trigger_coincidence(trigger_config['coincidence_window']),
+                                  item = "triggers")
+
+    # Filter events with zero triggers
+    filter_empty_trigger = fl.map(check_empty_trigger,
+                                  args = "triggers",
+                                  out  = "empty_trigger")
+
+    event_count_in     = fl.spy_count()
+    event_count_out    = fl.spy_count()
+    events_no_triggers = fl.count_filter(bool, args = "empty_trigger")
+    evtnum_collect     = collect()
+
+    with tb.open_file(file_out, "w", filters = tbl.filters(compression)) as h5out:
+        # Define writers...
+        write_event_info_   = run_and_event_writer(h5out)
+        write_trigger_info_ = trigger_writer      (h5out, get_number_of_active_pmts(detector_db, run_number))
+        write_trigger_dst_  = trigger_dst_writer  (h5out)
+        # ... and make them sinks
+
+        write_event_info   = sink(write_event_info_  , args=( "run_number"  , "event_number"    , "timestamp"   ))
+        write_trigger_info = sink(write_trigger_info_, args=( "trigger_type", "trigger_channels"                ))
+        write_trigger_dst  = sink(write_trigger_dst_ , args=  "triggers"                                         )
+
+        result = push(source = wf_from_files(files_in, WfType.rwf),
+                      pipe   = pipe(fl.slice(*event_range, close_all=True),
+                                    print_every(print_mod),
+                                    event_count_in.spy,
+                                    rwf_to_cwf,
+                                    trigger_on_channels,
+                                    filter_empty_trigger,
+                                    events_no_triggers.filter,
+                                    check_coincidences,
+                                    event_count_out.spy,
+                                    fl.branch("event_number", evtnum_collect.sink),
+                                    fl.fork(write_trigger_dst ,
+                                            write_event_info  ,
+                                            write_trigger_info)),
+                      result = dict(events_in   = event_count_in    .future,
+                                    events_out  = event_count_out   .future,
+                                    evtnum_list = evtnum_collect    .future,
+                                    events_pass = events_no_triggers.future))
+
+        if run_number <= 0:
+            copy_mc_info(files_in, h5out, result.evtnum_list,
+                         detector_db, run_number)
+
+        return result
--- a/invisible_cities/cities/valdrada.py
+++ b/invisible_cities/cities/valdrada.py
@@ -49,7 +49,7 @@ within range of the trigger configuration:
  - valid_w    : If peak is within width range.
  - valid_h    : If peak is within height range.
  - valid_peak : Only if 'multipeak' is active, True if there isn't a post-trigger
-               candidate with the configuration parametersself.
+               candidate with the configuration parameters.
  - valid_all  : boolean and of previous valid flags.
 
  Finally, a series of coincidence-related values are also given:
--- a/invisible_cities/cities/valdrada_test.py
+++ b/invisible_cities/cities/valdrada_test.py
@@ -1,0 +1,53 @@
+import os
+import tables as tb
+
+from .  valdrada           import valdrada
+from .. core.testing_utils import assert_tables_equality
+
+
+def test_valdrada_contains_all_tables(trigger_config):
+    conf, PATH_OUT = trigger_config
+    valdrada(**conf)
+    with tb.open_file(PATH_OUT) as h5out:
+        assert "Run"             in h5out.root
+        assert "Run/events"      in h5out.root
+        assert "Run/runInfo"     in h5out.root
+        assert "Trigger"         in h5out.root
+        assert "Trigger/events"  in h5out.root
+        assert "Trigger/trigger" in h5out.root
+        assert "Trigger/DST"     in h5out.root
+
+
+def test_valdrada_exact_result_multipeak(ICDATADIR, trigger_config):
+    true_out         = os.path.join(ICDATADIR, "test_trigger_fpga_multipeak.h5")
+    conf, PATH_OUT   = trigger_config
+    valdrada(**conf)
+
+    tables = ("Trigger/events", "Trigger/trigger", "Trigger/DST",
+              "Run/events"      , "Run/runInfo")
+
+    with tb.open_file(true_out)  as true_output_file:
+        with tb.open_file(PATH_OUT) as      output_file:
+            for table in tables:
+                assert hasattr(output_file.root, table)
+                got      = getattr(     output_file.root, table)
+                expected = getattr(true_output_file.root, table)
+                assert_tables_equality(got, expected)
+
+
+def test_valdrada_exact_result(ICDATADIR, trigger_config):
+    true_out          = os.path.join(ICDATADIR, "test_trigger_fpga.h5")
+    conf, PATH_OUT    = trigger_config
+    conf['trigger_config']['multipeak'] = None
+    valdrada(**conf)
+
+    tables = ("Trigger/events", "Trigger/trigger", "Trigger/DST",
+              "Run/events"      , "Run/runInfo")
+
+    with tb.open_file(true_out)  as true_output_file:
+        with tb.open_file(PATH_OUT) as      output_file:
+            for table in tables:
+                assert hasattr(output_file.root, table)
+                got      = getattr(     output_file.root, table)
+                expected = getattr(true_output_file.root, table)
+                assert_tables_equality(got, expected)
--- a/invisible_cities/cities/valdrada_test.py
+++ b/invisible_cities/cities/valdrada_test.py
@@ -19,7 +19,7 @@ def test_valdrada_contains_all_tables(trigger_config):
 
 
 def test_valdrada_exact_result_multipeak(ICDATADIR, trigger_config):
-    true_out         = os.path.join(ICDATADIR, "test_trigger_fpga_multipeak.h5")
+    true_out          = os.path.join(ICDATADIR, "exact_result_multipeak_valdrada.h5")
     conf, PATH_OUT   = trigger_config
     valdrada(**conf)
 
@@ -36,7 +36,7 @@ def test_valdrada_exact_result_multipeak(ICDATADIR, trigger_config):
 
 
 def test_valdrada_exact_result(ICDATADIR, trigger_config):
-    true_out          = os.path.join(ICDATADIR, "test_trigger_fpga.h5")
+    true_out          = os.path.join(ICDATADIR, "exact_result_valdrada.h5")
     conf, PATH_OUT    = trigger_config
     conf['trigger_config']['multipeak'] = None
     valdrada(**conf)
--- a/invisible_cities/config/valdrada.conf
+++ b/invisible_cities/config/valdrada.conf
@@ -1,0 +1,19 @@
+files_in    = '$ICDIR/database/test_data/blr_fpga_examples.h5'
+file_out   = '$ICDIR/database/test_data/valdrada.h5'
+
+compression = 'ZLIB4'
+run_number  = 8093
+detector_db = 'new'
+event_range = all
+
+trigger_config = {'coincidence_window':64, 'discard_width':40,
+                  'multipeak':{'q_min':100000, 'time_min':2*mus, 'time_after':800*mus}}
+
+channel_config = {0: {'q_min'       :5000, 'q_max'   :50000,
+                      'time_min'    :2*mus, 'time_max': 40*mus,
+                      'baseline_dev':  10, 'amp_max' : 1000,
+                      'pulse_valid_ext':50*ns},
+                  2: {'q_min'       :5000, 'q_max'   :50000,
+                      'time_min'    :2*mus, 'time_max': 40*mus,
+                      'baseline_dev':  10, 'amp_max' : 1000,
+                      'pulse_valid_ext':50*ns}}
--- a/invisible_cities/conftest.py
+++ b/invisible_cities/conftest.py
@@ -50,6 +50,11 @@ def example_blr_wfs_filename(ICDATADIR):
     return os.path.join(ICDATADIR, "blr_examples.h5")
 
 
+@pytest.fixture(scope='session')
+def example_blr_fpga_wfs_filename(ICDATADIR):
+    return os.path.join(ICDATADIR, "blr_fpga_examples.h5")
+
+
 @pytest.fixture(scope  = 'session',
                 params = ['electrons_40keV_z250_MCRD.h5'])
 def electron_MCRD_file(request, ICDATADIR):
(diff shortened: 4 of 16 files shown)


invisible_cities/conftest.py
```diff
@@ -699,6 +704,44 @@ def deconvolution_config(ICDIR, ICDATADIR, PSFDIR, config_tmpdir):
 
     return conf, PATH_OUT
 
+<<<<<<< HEAD
+=======
```
gonzaponte: Unresolved conflict! Fixed in the last commit, but shouldn't be here anyway.

invisible_cities/io/trigger_io_test.py
```diff
@@ -0,0 +1,56 @@
...
+    size  = draw(integers(min_value=1, max_value=10))
+    elements = []
+    for _ in range(6):
+        elements.append(draw(lists(integers(min_value=0, max_value=np.iinfo(np.uint32).max),
+                                   min_size=size, max_size=size)))
+    for _ in range(4):
+        elements.append(draw(lists(booleans(), min_size=size, max_size=size)))
+    for _ in range(5):
+        elements.append(draw(lists(integers(min_value=np.iinfo(np.int32).min, max_value=np.iinfo(np.int32).max),
+                                   min_size=size, max_size=size)))
```
gonzaponte: do these numbers have specific meanings? Maybe it's better to give them names lke
`a_meaningful_name = lists(integers(min_value=0, max_value=np.iinfo(np.uint32).max)` and then `draw(a_meaninful_name)`

gonzaponte: Ah, I see that later these are assigned to each column in `col_names`. I think in terms of self-documentation it's better for the composite to draw each value individually, but this might be too slow (haven't checked), so at least you should leave a comment annotating the columns they correspond to.

I think I would also transpose the data drawing: instead of "for each column generate all rows" -> "for each row generate all columns". This would be more natural if we had a composite that generates a single row and then another one that generates a list of rows of a given size. If this doesn't affect performance, I would go for it.

invisible_cities/io/trigger_io_test.py
```diff
@@ -0,0 +1,56 @@
...
+    trigger_dst = load_dst(output_file, group='Trigger', node='DST')
+
+    col_names = ["event"     , "pmt"       , "trigger_time" , "q"
+                ,"width"     , "height"    ,  "valid_q"     , "valid_w"
+                ,"valid_h"   , "valid_peak", "valid_all"    , "baseline"
+                ,"max_height", "n_coinc"   , "closest_ttime", "closest_pmt"]
+
+    assert np.all([tname == cname for tname, cname in zip(trigger_dst.columns.values, col_names)])
+
+    # Check values stored, the if is needed because the valid_all condition is done in the writer, not on the input trigger
```
gonzaponte: The writer shouldn't manipulate the data. Why not add that column to the input instead?

invisible_cities/reco/trigger_functions_test.py
```diff
@@ -0,0 +1,330 @@
...
+    start1      = np.random.randint(2*base_window, n_baseline // 2)
+    length      = np.random.randint(n_baseline // 50, n_baseline // 4)
+    stop1       = start1 + length
+    start2      = stop1 + 10
+    stop2       = start2 + length
+
+    wf[start1:stop1] = np.full(length, np.random.randint(thr+1, 50))
+    wf[start2:stop2] = np.full(length, np.random.randint(thr+1, 50))
+    wf[1500  : 1520] = np.full(20    , np.random.randint(thr+1, 50)) # Extra pulse to check discard_width
+    return wf, thr, length, start1, start2
```
gonzaponte: I never remember how to handle this, but using random numbers is something we should avoid. At the very least we should fix the seed for the tests or fixtures that use random numbers to produce data.

invisible_cities/reco/trigger_functions_test.py
```diff
@@ -0,0 +1,330 @@
...
+
+    elements = [evt_id    , pmt_id
+               ,t_true    , q_true  , w_true, h_true
+               ,val_q     , val_w   , val_h , val_p
+               ,wf[start1-1], wf.max()
+               ,-1        , -1      , -1]
+
+    assert len(triggers) == 1
+    for i, element in enumerate(elements):
+        assert element == triggers[0][i]
```
gonzaponte: why not `elements == triggers[0]`?

invisible_cities/reco/trigger_functions_test.py
```diff
@@ -0,0 +1,330 @@
...
+    # Check that the multipeak protection discards the trigger if conditions are met.
+    channel_config = channel_conf.copy()
+    trigger_config = trigger_conf.copy()
+
+    wf     = double_square_wf[0]
+    thr    = double_square_wf[1]
+    length = double_square_wf[2]
+    start1 = double_square_wf[3]
+    start2 = double_square_wf[4]
+    channel_config  = channel_conf.copy()
```
gonzaponte: copied twice

invisible_cities/reco/trigger_functions_test.py
```diff
@@ -0,0 +1,330 @@
...
+    channel_config['pulse_valid_ext'] = 0
+    triggers = get_trigger_candidates(wf, evt_id, pmt_id
+                                     ,channel_config, trigger_config
+                                     ,np.zeros(wf.shape), wf.max())
+    for t in triggers:
+        assert t[-6] == True
```
gonzaponte: This looks like the code needs a namedtuple. Accessing element -6 is a bit cryptic.

invisible_cities/reco/trigger_functions_test.py
```diff
@@ -0,0 +1,330 @@
...
+        val_p  = True
+
+        elements = [evt_id    , pmt_id
+                   ,t_true    , q_true  , w_true, h_true
+                   ,val_q     , val_w   , val_h , val_p
+                   ,wf[start1-1], wf.max()
+                   ,-1        , -1      , -1]
+
+        for i, element in enumerate(elements):
+            assert element == triggers[j][i]
```
gonzaponte: `assert elements == triggers[j]`?

invisible_cities/reco/trigger_functions_test.py
```diff
@@ -0,0 +1,330 @@
...
+
+    for t in triggers:
+        assert t[-6] == True
+
+
+@given(evt_id=integers(0, 3))
+def test_retrieve_trigger_information(double_square_wf, evt_id
+                                     ,channel_conf    , trigger_conf):
+    channel_config = channel_conf.copy()
+    trigger_config = trigger_conf.copy()
```
gonzaponte: why these copies?

invisible_cities/sierpe/blr_test.py
```diff
@@ -144,3 +161,64 @@ def test_deconv_pmt_ad_hoc_signals_dead_sensors(ad_hoc_blr_signals):
...
     np.allclose(blr_wfs, evt_true_blr_wfs[pmt_active])
+
+
+@mark.slow
+def test_deconv_pmt_fpga_ad_hoc_signals_all(ad_hoc_blr_fpga_signals):
+    all_rwfs, all_true_blr_wfs, all_true_baseline, params = ad_hoc_blr_fpga_signals
+
+    # This test takes long, so we pick a random event.
+    # Its exhaustiveness relies on repeated test runs.
+    evt_no            = np.random.choice(all_rwfs.shape[0])
```
gonzaponte: Probably best to use hypothesis then. Take an integer as input and decorate the test with `@settings(max_examples=1)` (or, 2, 3, etc.). This will let you do (at least) two things:

1. reproduce the failure by using a hypothesis-provided hash
2. try fixes in your code quickly because hypothesis attempts to run previous failures first

Same for the next test.

invisible_cities/sierpe/blr_test.py
```diff
@@ -144,3 +161,64 @@ def test_deconv_pmt_ad_hoc_signals_dead_sensors(ad_hoc_blr_signals):
...
+    rep_base = np.repeat(params.base_window, len(pmt_active))
+    blr_wfs, baseline = zip(*map(blr.deconvolve_signal_fpga    , evt_rwfs        [pmt_active],
+                                 params.coeff_clean[pmt_active], params.coeff_blr[pmt_active],
+                                 rep_thr                       , rep_base                    ))
+
+    assert np.allclose(blr_wfs , evt_true_blr_wfs [pmt_active])
+    assert np.allclose(baseline, evt_true_baseline[pmt_active])
+
+
+def test_moving_average_baseline():
```
gonzaponte: fixing the seed would be desirable here as well.

invisible_cities/sierpe/blr.pyx
```diff
@@ -70,4 +70,138 @@ cpdef deconvolve_signal(double [:] signal_daq,
...
+                blr_val = current # When outside of BLR/discharge, flip to raw signal
+                if k>=base_window:
+                    ped       = ped + current - aux[iaux-base_window]
+                    aux[iaux] = current
+                    iaux     += 1
+
+        signal_r[k] = blr_val
+        top[k]      = ped/base_window
+
+    return np.asarray(signal_r, dtype='int16'), np.asarray(top, dtype='int16')
```
gonzaponte, line 207: from double to int16, is there a risk of over/underflow? what operation should happen in this conversion (truncate, round, ceil or floor)?

invisible_cities/reco/trigger_functions.py
```diff
@@ -0,0 +1,174 @@
...
+    ----------
+    triggers       : List with all the trigger related information.
+    """
+    # Trigger parameters
+    qrange    = [channel_conf['q_min'          ], channel_conf['q_max'   ]]
+    wrange    = [channel_conf['time_min'       ], channel_conf['time_max']]
+    hrange    = [channel_conf['baseline_dev'   ], channel_conf['amp_max' ]]
+    pulse_ext =  channel_conf['pulse_valid_ext']
+
+    fMulti        = trigger_conf['multipeak'    ] # Protection for later peaks
```
gonzaponte: rename to `multipeak` or `multipk`?

invisible_cities/reco/trigger_functions.py
```diff
@@ -0,0 +1,174 @@
...
+        # Baseline at the start of the pulse
+        baseline_t0 = baseline[start_time]
+        triggers  .append([event_id   , pmt_id
+                          ,start_time , charge        , width    , height
+                          ,trigger_q  , trigger_w     , trigger_h, True   # By default multi flag is set as true
+                          ,baseline_t0, wvf_max_height
+                          ,-1         , -1            , -1])              # Filled later in the coincidence module
+
+    # Extra-peaks protection
+    if fMulti is not None:
```
gonzaponte: shouldn't it be a boolean?

invisible_cities/reco/trigger_functions.py
```diff
@@ -0,0 +1,174 @@
...
+        # Evaluate if later peaks (must be over discard_width) are candidates for the multipeak protection
+        for i, t0 in zip(range(len(triggers[:-1])), triggers[:-1]):
+            for t1 in triggers[i+1:]:
+                multi_candidate = ((t1[4] >=                 multi_w  ) &
+                                   (t1[3] >                  multi_q  ) &
+                                   (t1[2] <= t0[2] + t0[4] + multi_t+2)) # Start of trigger + peak width + time window + FPGA delay (2)
+                if not multi_candidate:
+                    continue
+                # In case the later peak ends after the multipeak protection window ends
+                if (t1[2]+t1[4]) > (t0[2] + t0[4] + multi_t+2): # Event partially contained in the multipeak window
```
gonzaponte: I guess these magic numbers have meaning, so again, a namedtuple would be better.

invisible_cities/reco/trigger_functions.py
```diff
@@ -0,0 +1,174 @@
...
+        # Shortest time is the first value of time_diff for each trigger
+        shortest   = np.array([t[0] if len(t)>0 else -1 for t in time_diffs])
+        # Count how many triggers are within the coincidince window of each trigger
+        n_coinc    = np.array([len(t[t<coinc_window])+1 for t in time_diffs])
+        # Undo the original ordering
+        old_order  = np.argsort(order_idx)
+        # Assign the coincidence results to the trigger array
+        triggers[valid_sel, -3] = n_coinc [old_order]
+        triggers[valid_sel, -2] = shortest[old_order]
+        triggers[valid_sel, -1] = pmts    [old_order]
```
gonzaponte: Add empty lines in between blocks to make it a bit more readable.

invisible_cities/reco/trigger_functions.py
```diff
@@ -0,0 +1,174 @@
...
+        valid_sel  = np.all(triggers[:, 6:10], axis=1).astype('bool') # Use only valid triggers
+        times      = triggers[:, 2][valid_sel]         # Times of valid triggers
+        ids        = triggers[:, 1][valid_sel]         # Ids   of valid triggers
+
+        # Order by time, index so we can use this to order both times and ids
+        order_idx  = np.lexsort((ids, times))
+        order_time = times[order_idx]
+        pmts       = np.full(ids.shape, -1)
+        # Calculate all time differences between each trigger and all later ones
+        time_diffs = np.array([order_time[i+1:] - t for i, t in enumerate(order_time)])
```
gonzaponte: Converting it to an array is misleading. This would be a 1D array of type `object` each object being a 1D array of different lengths. Better leave it as a list.

invisible_cities/io/trigger_io.py
```diff
@@ -41,3 +42,41 @@ def _make_tables(hdf5_file, n_sensors, compression="ZLIB4"):
...
+        row["pmt"          ] = pmt
+        row["trigger_time" ] = trigger_time
+        row["q"            ] = charge
+        row["width"        ] = width
+        row["height"       ] = height
+        row["valid_q"      ] = valid_q
+        row["valid_w"      ] = valid_w
+        row["valid_h"      ] = valid_h
+        row["valid_peak"   ] = valid_peak
+        row["valid_all"    ] = (valid_q + valid_w + valid_h + valid_peak) == 4
```
gonzaponte: the writer shouldn't need to know anything about what it is writing. you may also want to consider creating a dataframe and saving it using the dataframe writer?

also, since these are booleans, it's more appropriate to use `all`.

invisible_cities/conftest.py
```diff
@@ -699,6 +704,44 @@ def deconvolution_config(ICDIR, ICDATADIR, PSFDIR, config_tmpdir):
...
+
+@pytest.fixture(scope='function') # Needs to be function as the config dict is modified when running
```
gonzaponte, line 708: Is the comment outdated? I've seen that you copy the config in each test, no?

gonzaponte (commented):
I haven't looked at the code in depth, but I've cached a few technical details already.

This PR is algorithmically complex and requires knowing many non-trivial details. Who else besides you is familiar with this? It would be great if this person could review the algorithmic side of things at least.

invisible_cities/conftest.py
Aretno: I think the comment is fine. This only applies to the city, the copied ones are for the `trigger_functions`.

invisible_cities/sierpe/blr.pyx
Aretno: Should be truncate (which for positive numbers behaves as floor, if I recall correctly).

invisible_cities/reco/trigger_functions_test.py
Aretno: Because I modify the dictionary on several tests and the fixture's scope is set to session. Ideally I would use 'function' but it gave an ugly warning that, while it does not worry me, adds too much noise to pytest:

` invisible_cities/reco/trigger_functions_test.py::test_retrieve_trigger_information
  /home/lior/miniconda/envs/IC-3.7-2020-06-16/lib/python3.7/site-packages/hypothesis/extra/pytestplugin.py:170: HypothesisDeprecationWarning: invisible_cities/reco/trigger_functions_test.py::test_retrieve_trigger_information uses the 'trigger_config' fixture, which is reset between function calls but not between test cases generated by `@given(...)`.  You can change it to a module- or session-scoped fixture if it is safe to reuse; if not we recommend using a context manager inside your test function.  See https://docs.pytest.org/en/latest/fixture.html#sharing-test-data for details on fixture scope.
    note_deprecation(msg % (fx.argname,), since="2020-02-29")`  

In lack of a better approach, and after discussion with @gonzaponte, we have opted to define two functions that provide the modified configuration dicts when needed.

invisible_cities/reco/trigger_functions_test.py
Aretno: Yeah, much better.

invisible_cities/reco/trigger_functions_test.py
Aretno: I created a composite now, I guess it should be ok.

invisible_cities/conftest.py
Aretno: I have no idea how to remove this from the history.

Aretno:
Some things were answered in the comments of the reviewer, anothers talked peer to peer, but I think all the issues pointed out so far in the review have been taken into account.

invisible_cities/cities/valdrada.py
```diff
@@ -0,0 +1,172 @@
...
+     - time_after : For how long is the multipeak protection evaluated after a
+                    valid trigger. In mus.
+
+A individual trigger configuration can be given per channel, through a dict
+with keys equal to PMT IDs, which marks the validity range of the peak
+characteristics:
+- q_min, q_max          : Range for the integrated charge of the peak (q_min < q < q_max).
+                          In ADC counts.
+- time_min, time_max    : Range for the peak width (time_min <= width < time_max).
+                          In time mus.
```
gonzaponte, line 32:
```suggestion
                          In mus.
```

invisible_cities/cities/valdrada.py
```diff
@@ -0,0 +1,172 @@
+"""
+-----------------------------------------------------------------------
+                                 valdrada
+-----------------------------------------------------------------------
+
+This city finds simulates the trigger procedure at the FPGA over PMT waveforms.
```
gonzaponte, line 6:
```suggestion
This city simulates the trigger procedure at the FPGA over PMT waveforms.
```

invisible_cities/cities/valdrada.py
```diff
@@ -0,0 +1,172 @@
...
+    - Use a moving average to obtain the baseline
+    - Truncate to integers the deconvolution output
+    - Find trigger candidates and compute their characteristics, taking into
+      account the unique circumstances derived from working on a continuous scheme
+      (for example, delay in signals)
+    - Evaluate coincidences between trigger candidates.
+
+A number of general configuration parameters are needed (input as a dict, trigger_config):
+- coincidence_window : Time window (in time bins) in which valid trigger coincidences are counted
+- discard_width      : Any trigger with width less than this parameter is discarded.
```
gonzaponte, line 17: units?

invisible_cities/cities/valdrada.py
```diff
@@ -0,0 +1,172 @@
...
+def valdrada(files_in, file_out, compression, event_range, print_mod, detector_db, run_number,
+             trigger_config = dict(), channel_config = dict()):
+
+    # Change time units into number of waveform bins.
+    if trigger_config['multipeak'] is not None:
+        trigger_config['multipeak']['time_min'  ] /= 25*units.ns
+        trigger_config['multipeak']['time_after'] /= 25*units.ns
+    for k in channel_config.keys():
+        channel_config[k]['time_min']       /= 25*units.ns
+        channel_config[k]['time_max']       /= 25*units.ns
```
gonzaponte, line 110: shouldn't these be forced to be ints?

invisible_cities/cities/valdrada_test.py
```diff
@@ -0,0 +1,53 @@
...
+    with tb.open_file(true_out)  as true_output_file:
+        with tb.open_file(PATH_OUT) as      output_file:
+            for table in tables:
+                assert hasattr(output_file.root, table)
+                got      = getattr(     output_file.root, table)
+                expected = getattr(true_output_file.root, table)
+                assert_tables_equality(got, expected)
+
+
+def test_valdrada_exact_result(ICDATADIR, trigger_config):
```
gonzaponte, line 38: how about
```
@mark.parametrize("multipeak", (True, False))
def test_valdrada_exact_result(ICDATADIR, trigger_config, multipeak)
    ...
    if not multipeak:
        conf['trigger_config']['multipeak'] = None
```
to combine both tests? You would need to add true_out to the parametrize as well, but it's worth it.

invisible_cities/config/valdrada.conf
```diff
@@ -0,0 +1,19 @@
...
+                  'multipeak':{'q_min':100000, 'time_min':2*mus, 'time_after':800*mus}}
+
+channel_config = {0: {'q_min'       :5000, 'q_max'   :50000,
+                      'time_min'    :2*mus, 'time_max': 40*mus,
+                      'baseline_dev':  10, 'amp_max' : 1000,
+                      'pulse_valid_ext':50*ns},
+                  2: {'q_min'       :5000, 'q_max'   :50000,
+                      'time_min'    :2*mus, 'time_max': 40*mus,
+                      'baseline_dev':  10, 'amp_max' : 1000,
+                      'pulse_valid_ext':50*ns}}
```
gonzaponte, line 19: Add `* adc` and `# time bins` as discussed offline.

invisible_cities/conftest.py
```diff
@@ -699,6 +704,41 @@ def deconvolution_config(ICDIR, ICDATADIR, PSFDIR, config_tmpdir):
...
+                                                 ,baseline_dev    = 10
+                                                 ,amp_max         = 1000
+                                                 ,pulse_valid_ext = 50*units.ns)
+                                        ,2 : dict(q_min           = 5000
+                                                 ,q_max           = 50000
+                                                 ,time_min        = 2 *units.mus
+                                                 ,time_max        = 40*units.mus
+                                                 ,baseline_dev    = 10
+                                                 ,amp_max         = 1000
+                                                 ,pulse_valid_ext = 50*units.ns)})
```
gonzaponte, line 737: Add `* adc` and `# time bins` as discussed offline.

invisible_cities/io/trigger_io.py
```diff
@@ -41,3 +42,36 @@ def _make_tables(hdf5_file, n_sensors, compression="ZLIB4"):
     trg_tables = trg_type, trg_channels
 
     return trg_tables
+
+
+def trigger_dst_writer(hdf5_file, **kwargs):#{
```
gonzaponte, line 47: Why `**kwargs`? They are not used.

gonzaponte: Also what's this `#{`? Also in line 54.

invisible_cities/io/trigger_io_test.py
```diff
@@ -0,0 +1,82 @@
...
+    col_names = ["event"     , "pmt"       , "trigger_time" , "q"
+                ,"width"     , "height"    , "valid_q"      , "valid_w"
+                ,"valid_h"   , "valid_peak", "valid_all"    , "baseline"
+                ,"max_height", "n_coinc"   , "closest_ttime", "closest_pmt"]
+
+    assert np.all([tname == cname for tname, cname in zip(trigger_dst.columns.values, col_names)])
+
+    # Check values stored
+    for i, row in trigger_dst.iterrows():
+        assert_allclose(triggers[i], list(row.values))
```
gonzaponte, line 82: I think this can be replaced by
```suggestion
    assert_allclose(np.asarray(triggers), trigger_dst.values)
```
But I'm not 100% sure it will work

invisible_cities/reco/trigger_functions.py
```diff
@@ -0,0 +1,198 @@
...
+    triggers   = []
+    indexes    = np.where(signal > hrange[0])[0]
+    if len(indexes) == 0: return triggers # Protection in case min. threshold is not surpassed
+    index_sel  = np.diff(indexes)>(pulse_ext+1)
+    candidates = np.split(indexes, np.where(index_sel)[0] + 1)
+
+    for p in candidates:
+        peak_slice = slice(p[0]-1, p[-1] + pulse_ext + 1) # DAQ gets pulse_ext counts after peaks; also, for Q it takes a bin before TODO Fix for N100 (remove -1)
+        peak       = signal[peak_slice]
+        if not (len(peak) > max(discard_width, 1)): continue
```
gonzaponte, line 64:
```suggestion
        if len(peak) <= max(discard_width, 1): continue
```

invisible_cities/reco/trigger_functions.py
```diff
@@ -0,0 +1,198 @@
...
+    pulse_ext =  channel_conf['pulse_valid_ext']
+
+    multipk       = trigger_conf['multipeak'    ] # Protection for later peaks
+    discard_width = trigger_conf['discard_width'] # Don't store events below certain width
+
+    # Multipeak protection parameters
+    if multipk:
+        multi_q =     multipk['q_min'     ]  # Minimum charge of later peaks to discard a trigger
+        multi_w = int(multipk['time_min'  ]) # Minimum time width of later peaks to discard a trigger
+        multi_t = int(multipk['time_after']) # Time window for a later peak to be considered
```
gonzaponte, line 52: Maybe this can go already inside the other if statement around line 89?

invisible_cities/reco/trigger_functions.py
```diff
@@ -0,0 +1,198 @@
...
+                                 ,start_time , charge        , width    , height
+                                 ,trigger_q  , trigger_w     , trigger_h,   True, trigger_all  # By default multiflag is set as true
+                                 ,baseline_t0, wvf_max_height
+                                 ,-1         , -1            , -1) # Filled later in the coincidence module
+        triggers  .append(trigger)
+
+    # Extra-peaks protection
+    if multipk is not None:
+        # Evaluate if later peaks (must be over discard_width) are candidates for the multipeak protection
+        for i, t0 in zip(range(len(triggers[:-1])), triggers[:-1]):
```
gonzaponte, line 91:
```suggestion
        for i, t0 in enumerate(triggers[:-1]):
```

invisible_cities/reco/trigger_functions.py
```diff
@@ -0,0 +1,198 @@
...
+            new_val = t0.valid_peak
+            for t1 in triggers[i+1:]:
+                multi_candidate = ((t1.width        >=                              multi_w  ) &
+                                   (t1.q            >                               multi_q  ) &
+                                   (t1.trigger_time <= t0.trigger_time + t0.width + multi_t+2)) # Start of trigger + peak width + time window + FPGA delay (2)
+                if not multi_candidate:
+                    continue
+                # In case the later peak ends after the multipeak protection window ends
+                if (t1.trigger_time + t1.width) > (t0.trigger_time + t0.width + multi_t + 2): # Event partially contained in the multipeak window
+                    peak_slice = slice(t1.trigger_time - 1, t0.trigger_time + t0.width + multi_t + 2 + 1)
```
gonzaponte, line 101: `t0.trigger_time + t0.width + multi_t + 2` seems to be a relevant value. It's worth to give it a name for readability.

invisible_cities/reco/trigger_functions.py
```diff
@@ -0,0 +1,198 @@
...
+                    peak       = signal[peak_slice]
+                    width      = len(peak) - 1 # TODO Remove -1 in N100
+                    charge     = peak[peak>0].sum()
+
+                    new_val    = (width >= multi_w) & (charge > multi_q)
+                else:
+                    new_val = False
+                if not new_val: break
+            triggers[i] = triggers[i]._replace(valid_peak=new_val, valid_all=all([triggers[i].valid_all, new_val]))
```
gonzaponte, line 110:
```suggestion
            triggers[i] = triggers[i]._replace(valid_peak=new_val, valid_all=triggers[i].valid_all and new_val)
```

invisible_cities/reco/trigger_functions.py
```diff
@@ -0,0 +1,198 @@
...
+
+    return triggers
+
+
+def retrieve_trigger_information(channel_config : Dict[int, int]
+                                ,trigger_config : Dict
+                                ) -> Callable:
```
gonzaponte, line 117: This looks like a city component, but maybe it just looks like one. I leave to you the decision of moving it to components.

invisible_cities/reco/trigger_functions_test.py
```diff
@@ -0,0 +1,331 @@
...
+from .. reco.trigger_functions import check_trigger_coincidence
+from .. reco.trigger_functions import TriggerInfo
+from .. core.core_functions    import in_range
+
+
+def channel_conf(**kwargs):
+    channel_dict = {'q_min'          : 5000, 'q_max'   :  50000,
+                    'time_min'       : 2000, 'time_max':  10000,
+                    'baseline_dev'   :    5, 'amp_max' :   1000,
+                    'pulse_valid_ext':   15}
```
gonzaponte, line 23: units

invisible_cities/reco/trigger_functions_test.py
```diff
@@ -0,0 +1,331 @@
...
+    wf[1500  : 1520] = np.full(20    , heights[2]) # Extra pulse to check discard_width
+
+    return wf, thr, length, start1, start2
+
+
+@composite
+def trigger_time_and_validity(draw):
+    # Set of trigger times an validity flags, to test trigger coincidence
+    size  = draw (integers(min_value=1, max_value=100))
+    flag  = lists(integers(min_value=0, max_value=1  ), min_size=4, max_size=4)
```
gonzaponte, line 59: Shouldn't it be booleans?

invisible_cities/sierpe/blr.pyx
gonzaponte: Maybe it would be good to clip the result before casting it to an int. Otherwise, 33000 would become -32536.
The limits for int16 are -32768 and +32767

gonzaponte (changes requested):
I'm ignoring cosmetics for now.

PR #796, opened 2021-06-15
Update NEWDB

MiryamMV:
This PR includes the latest sensor calibrations starting from run 8692. The database has been uploaded with the PMT/SiPMs gains and the pdf values.

Here some plots comparing the calibration runs with previous calibrations:

sipmRunDifferencePlots
pmtRunDifferencePlots


jmbenlloch (approved):

PR #799, opened 2021-07-09
Update DEMOPPDB

carmenromo:
Demo++ database has been updated with the calibration constants from R10159 for the SiPMs and pdfs. In the database, however, The `Min_run` set is 9862 because the values from the previous calibrations were incorrect after the first rotation of the dice boards 1 and 3.


jmbenlloch (approved):
Fixes last calibration in the DB.

Issue #800, opened 2021-08-30
Wrong dst loading in city sources


gondiaz:
City sources whose input is a dst-like dataframe have the following structure (see for example `cdst_from_files` or `hits_and_kdst_from_files`)

```
try:
    df = load_dst (path, group, node)
except tb.exceptions.NoSuchNodeError:
    continue
```
where load_dst is the function in `io.dst_io`.  Nevertheless, `load_dst` function outputs a warning in the case of a `NoSuchNodeError` and therefore the intended behaviour of the `try-except` clause in the city source is wrong.

mmkekic:
I'd say that whoever takes this over should also revist all 3 readers since they are all some subset of generic `dst_from_files`. If there is a need to fix the behaviour fast the 2 options are to either catch a warning or upgrade it to error for this scope

gondiaz:
I have proposed a solution in PR #801 which involves removing `NoSuchNodeError` from `load_dst` function. I found it the most natural approach since if a file does not contain the node you are requesting, you get an `Exception` instead of a `warning` which is more error prone.

It might be that your file is wrong and does not contain the node (or you are wrongly requesting a node which is not included in that file type) or you might have an spelling error. Personally, I find more natural to get an `Exception` since I would like to be aware of such a mistake.

Nevertheless, since this is a commonly used function @mmkekic suggests to get input from the likely users: @paolafer @Aretno @jjgomezcadenas @gonzaponte @ausonandres @bpalmeiro @jahernando @msorel @jmalbos @jerenner. 

The `load_dst` function was created by @jjgomezcadenas and @Aretno. The only place in IC that the `warning` is catched is in the function `safe_copy_nexus_eventmap` in `io/mc_io.py`, thus the fact of being a `warning` is not very relevant inside IC.

gonzaponte:
Solved in #801

PR #801, opened 2021-09-01
Remove NoSuchNodeError catch in dst loading

gondiaz:
This PR solves issue #800. The proposed solution removes the `NoSuchNodeError` catch in `load_dst` function.

### DIFF
--- a/invisible_cities/io/dst_io.py
+++ b/invisible_cities/io/dst_io.py
@@ -18,28 +18,34 @@ def _decode_str_columns(df):
     df        = df.apply(to_string)
     return df
 
-def load_dst(filename, group, node, evt_list=None):
+def load_dst(filename, group, node, evt_list=None, ignore_errors=False):
     """load a kdst if filename, group and node correctly found"""
-    try:
+
+    def read_dst_(filename, group, node, evt_list):
         with tb.open_file(filename) as h5in:
-            try:
-                table  = getattr(getattr(h5in.root, group), node)
-                if evt_list is None:
-                    values = table.read()
-                else:
-                    events = table.read(field='event')
-                    values = table.read_coordinates(np.where(np.isin(events, evt_list)))
-                return _decode_str_columns(pd.DataFrame.from_records(values, columns=table.colnames))
-            except NoSuchNodeError:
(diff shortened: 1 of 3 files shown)


gonzaponte:
The problem with this function is that it is used both by individual users, which may want just a warning (IIRC, this was the original reason), and the cities, which definitely want to crash if something happens to the file.
Two questions:
1. What should be done with the other exception that is turned into a warning in this very function? It looks like we need to handle that one as well.
2. Have you considered introducing an argument to the function to combine both behaviors? Something in the lines of `ignore_empty`/`unsafe`?

It's been a while since I last had an issue with this function, so other, more frequent users should comment.

gondiaz:
Agree with point 1 that the other warnings should be changed to exceptions, but probably in another dedicated PR.

For point 2: if I understand well the warnings are meant to avoid code breaking in personal analysis notebooks/scripts where the function is used and those people would also be bothered if a new argument is implemented if the default behaviour is to raise an `Exception`. Also the warning message it outputs "not of kdst type" should be changed.

gonzaponte:
> but probably in another dedicated PR.

I don't see the need to open another PR, but I don't have a problem with that.

>  if I understand well the warnings are meant to avoid code breaking in personal analysis notebooks/scripts where the function is used

Yup

> those people would also be bothered if a new argument is implemented if the default behaviour is to raise an Exception.

I'm not sure what you are saying here. The new behavior is going to be to raise an exception in any case, right? Anyone using this function will now have to deal with that anyway. The question is whether to force them to catch the exception explicitly, i.e.
```
try:
    load_dst(...)
except:
    # do something or not
```
or to add an argument to their calls, which is arguably less intrusive
```
load_dst(..., ignore_errors=True)
``` 

BTW, there is a third way: to have `load_dst` to always raise exceptions and define `load_dst_xxx` that handles the exceptions raised by `load_dst`. This is pretty much the same as the adding-a-new-parameter option.

gondiaz:
I agree that adding the extra karg is the less intrusive and probably the best option. `ignore_errrors = False` would be the default and people that uses the function should change to `ignore_errrors = True` if neccessary.

gonzaponte (approved):
Clear and concise. Approved.

PR #802, opened 2021-09-09
Add DST group in high level files

gondiaz:
This PR adds 'DST/Events' group in beersheba and subsequent isaura outputs. Until now high level analysis was performed using esmeralda files, but the two extra cities were added afterwards. These cities did not contain the DST group, which contain relevant information, for example the S1 data. Ideally, high level analysis will be performed with  isaura files, thus both cities must carry the DST group. The PR ensures backward compatibility, old files without DST group can be processed as well.

### DIFF
--- a/invisible_cities/cities/beersheba.py
+++ b/invisible_cities/cities/beersheba.py
@@ -50,6 +50,7 @@ from .. reco.deconv_functions  import InterpolationMethod
 from .. io.run_and_event_io    import run_and_event_writer
 from .. io.          dst_io    import df_writer
 from .. io.          dst_io    import load_dst
+from .. io.          kdst_io   import kdst_from_df_writer
 
 from .. evm.event_model        import HitEnergy
 
@@ -311,6 +312,7 @@ def deconv_writer(h5out, compression='ZLIB4'):
     return write_deconv
 
 
+
 @city
 def beersheba(files_in, file_out, compression, event_range, print_mod, detector_db, run_number,
               deconv_params = dict()):
@@ -411,6 +413,8 @@ def beersheba(files_in, file_out, compression, event_range, print_mod, detector_
     event_count_out       = fl.spy_count()
     events_passed_no_hits = fl.count_filter(bool, args = "cdst_passed_no_hits")
 
+    filter_out_none       = fl.filter(lambda x: x is not None, args = "kdst")
+
     evtnum_collect        = collect()
 
     with tb.open_file(file_out, "w", filters = tbl.filters(compression)) as h5out:
@@ -418,6 +422,8 @@ def beersheba(files_in, file_out, compression, event_range, print_mod, detector_
         write_event_info = fl.sink(run_and_event_writer (h5out), args = ("run_number", "event_number", "timestamp"))
         write_deconv     = fl.sink(  deconv_writer(h5out=h5out), args =  "deconv_dst")
         write_summary    = fl.sink( summary_writer(h5out=h5out), args =  "summary"   )
+        write_kdst_table = fl.sink(  kdst_from_df_writer(h5out), args =  "kdst"      )
+
         result = push(source = cdst_from_files(files_in),
                       pipe   = pipe(fl.slice(*event_range, close_all=True)    ,
                                     print_every(print_mod)                    ,
@@ -432,6 +438,7 @@ def beersheba(files_in, file_out, compression, event_range, print_mod, detector_
                                               evtnum_collect.sink)            ,
                                     fl.fork(write_deconv    ,
                                             write_summary   ,
+                                            (filter_out_none, write_kdst_table),
                                             write_event_info))                ,
                       result = dict(events_in   = event_count_in       .future,
                                     events_out  = event_count_out      .future,
--- a/invisible_cities/cities/beersheba_test.py
+++ b/invisible_cities/cities/beersheba_test.py
@@ -149,3 +149,20 @@ def test_beersheba_expandvar(deconvolution_config):
     conf['deconv_params']['psf_fname'] = '$ICDIR/database/test_data/PSF_dst_sum_collapsed.h5'
 
     beersheba(**conf)
+
+
+def test_beersheba_copy_kdst(deconvolution_config, ICDATADIR):
+    PATH_IN = os.path.join(ICDATADIR, "test_cdst_NEW_v1.2.0_bg.h5")
+    conf, PATH_OUT = deconvolution_config
+    conf["files_in"]   = PATH_IN
+    conf["run_number"] = 8515
+
+    expected_events = [3, 90, 152]
+
+    beersheba(**conf)
+
+    with tb.open_file(PATH_OUT) as output_file:
+        assert hasattr(output_file.root, "DST/Events")
+
+        got_events = output_file.root.DST.Events.read()["event"]
+        assert expected_events == got_events.tolist()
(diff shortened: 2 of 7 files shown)


invisible_cities/cities/beersheba.py
```diff
@@ -311,6 +312,14 @@ def write_deconv(df):
     return write_deconv
 
 
+def kdst_writer(h5out):
+    """ This function protects for kdst of None type"""
+    def writer(kdst):
+        if isinstance(kdst, pd.DataFrame): kdst_from_df_writer(h5out)(kdst)
```
gonzaponte: As standing now, `kdst_from_df_writer(h5out)` would be executed for each element of the source, shouldn't it happen only once?

invisible_cities/cities/components.py
```diff
@@ -503,6 +503,7 @@ def cdst_from_files(paths: List[str]) -> Iterator[Dict[str,Union[pd.DataFrame, M
         try:
             cdst_df    = load_dst (path,   'CHITS', 'lowTh')
             summary_df = load_dst (path, 'Summary', 'Events')
+            kdst_df    = load_dst (path,     'DST', 'Events', ignore_errors=True)
         except tb.exceptions.NoSuchNodeError:
             continue
```
gonzaponte: Is this try/except clause for a specific line or for both cdst and summary?
If it is for just one of them, then that line should be the only one in the try block. If it is for both of them it's fine.
The new line (kdst_df) should definitely go outside of the try block because it is not supposed to raise a `NoSuchNodeError`.

invisible_cities/cities/components.py
```diff
@@ -520,6 +521,8 @@ def cdst_from_files(paths: List[str]) -> Iterator[Dict[str,Union[pd.DataFrame, M
                 event_number, timestamp = evtinfo
                 yield dict(cdst    = cdst_df   .loc[cdst_df   .event==event_number],
                            summary = summary_df.loc[summary_df.event==event_number],
+                           kdst    = kdst_df   .loc[kdst_df   .event==event_number] if isinstance(kdst_df, pd.DataFrame) \
+                                                                                    else None,
```
gonzaponte: what do you think of this?

```suggestion
                this_event = lambda df: df.event==event_number
                yield dict(cdst    = cdst_df   .loc[this_event],
                           summary = summary_df.loc[this_event],
                           kdst    = kdst_df   .loc[this_event] if isinstance(kdst_df, pd.DataFrame) \
                                                                else None,
```

gonzaponte: I'm also thinking whether it is better to check `isinstance(kdst_df, pd.DataFrame)` or `kdst_df is not None`. The second one is perhaps more general, but I'm not convinced it's actually better. Thoughts?

invisible_cities/cities/beersheba.py
```diff
@@ -418,6 +427,8 @@ def beersheba(files_in, file_out, compression, event_range, print_mod, detector_
         write_event_info = fl.sink(run_and_event_writer (h5out), args = ("run_number", "event_number", "timestamp"))
         write_deconv     = fl.sink(  deconv_writer(h5out=h5out), args =  "deconv_dst")
         write_summary    = fl.sink( summary_writer(h5out=h5out), args =  "summary"   )
+        write_kdst_table = fl.sink(          kdst_writer(h5out), args="kdst")
```
gonzaponte: Alignment :)

invisible_cities/cities/beersheba.py
```diff
@@ -432,6 +443,7 @@ def beersheba(files_in, file_out, compression, event_range, print_mod, detector_
                                               evtnum_collect.sink)            ,
                                     fl.fork(write_deconv    ,
                                             write_summary   ,
+                                            write_kdst_table,
                                             write_event_info))                ,
```
gonzaponte: We could also replace the `kdst_writer` function with 
```
write_kdst_table = fl.sink(kdst_from_df_writer(h5out))
filter_out_none  = fl.filter(lambda x: x is not None)
[...]
fl.fork(write_deconv    ,
        write_summary   ,
        ("kdst", filter_out_none, write_kdst_table),
        write_event_info))
```
which I think is the idiomatic way to do it, but I'm afraid it might be a bit cryptic. what do you think?

invisible_cities/cities/beersheba_test.py
```diff
@@ -149,3 +149,15 @@ def test_beersheba_expandvar(deconvolution_config):
     conf['deconv_params']['psf_fname'] = '$ICDIR/database/test_data/PSF_dst_sum_collapsed.h5'
 
     beersheba(**conf)
+
+
+def test_beersheba_copy_kdst(deconvolution_config, ICDATADIR):
```
gonzaponte, line 154: Do any of the other tests check that if a file doesn't contain the kdst everything runs smoothly?

invisible_cities/cities/beersheba_test.py
```diff
@@ -149,3 +149,15 @@ def test_beersheba_expandvar(deconvolution_config):
...
+    PATH_IN = os.path.join(ICDATADIR, "test_cdst_NEW_v1.2.0_bg.h5")
+    conf, PATH_OUT = deconvolution_config
+    conf["files_in"] = PATH_IN
+    conf["run_number"] = 8515
+
+    beersheba(**conf)
+
+    with tb.open_file(PATH_OUT) as output_file:
+        assert hasattr(output_file.root, "DST/Events")
```
gonzaponte, line 165: I would also check that the event numbers are those you expect.

gonzaponte (commented):
Good work, some minor comments and we are good to go.

invisible_cities/cities/components.py
gondiaz: Both `cdst_df` and `summary_df` should be inside. Ok, it is more correct to move the `kdst_df` outside.

invisible_cities/cities/components.py
gondiaz: I like the `this_event` proposal, added. No opinion about the instance check.

invisible_cities/cities/beersheba.py
gondiaz: It looks an elegant solution to me, but I would prefer to use the argument explicitly, namely 
```
write_kdst_table = fl.sink(kdst_from_df_writer(h5out), args = "kdst")
filter_out_none  = fl.filter(lambda x: x is not None, args = "kdst")
[...]
fl.fork(write_deconv    ,
        write_summary   ,
        (filter_out_none, write_kdst_table),
        write_event_info))
```
just to follow the same structure as the other writers. Is it good or should we use the idiomatic way?

invisible_cities/cities/beersheba_test.py
gondiaz: The esmeralda used in the other tests does not contain DST group, therefore I don't think it is necessary an extra explicit test

invisible_cities/cities/beersheba.py
gondiaz: Right, I would go for a filter based solution as you suggested above

invisible_cities/cities/beersheba_test.py
gondiaz: Added

invisible_cities/cities/beersheba.py
gonzaponte: I think most users are more used to `args` and find the alternative more difficult to understand, so I'm ok with this even if it is slightly more verbose than necessary.

gonzaponte:
Please edit the description of this PR to explain briefly why we are adding this data to the output so it goes into the merge commit. I will approve the PR once this is done.

gondiaz:
I have added the same commits for Isaura city and updated the PR's description and title.

gonzaponte (approved):
Tests are good and the code is readable and simple. Approved.

Good job.

PR #803, opened 2021-09-15
Update run numbers in demoppdb

carmenromo:
There was a run number (R9859) from the Demo++ Run9 that was not included in the current gains in the DB. In order to fix that, the following changes have been added:

- Runs with `MinRun=9523` and `MaxRun=9859` --> change `MaxRun` to 9856
- Runs with `MinRun=9860` and `MaxRun=9861` --> change `MinRun` to 9857 and `MaxRun` to 9858
- Runs with `MinRun=9862` and `MaxRun=100000` --> change `MinRun` to 9859


jmbenlloch (approved):
This PR fixes the calibration run-ranges for DEMOPP.

PR #807, opened 2021-11-04
Update to Python 3.8

gondiaz:
This PR is intended to update the IC conda and libraries to allow working in Python ~~3.9~~ 3.8.

The branch has been tested on the following systems (we will update the list once more machines are tested):

✅ MacOS Monterey (12.0.1)
✅ Ubuntu 20.04 (new system for github actions)
~~❌ Ubuntu 18.04 (old system for github actions) --> failing tests~~
✅   Ubuntu 16.04 (gpu1next)
✅   Red Hat Enterprise Linux Server 7.5 (Maipo) (CESGA)
✅ CentOS Linux 7 (majorana1)
✅ CentOS Linux 7 (frontend1demo)
❌ CentOS Linux 7 (PowerPC) (argudell) --> failing tests (conda has to be installed appart, ppc64el arch)
❌ Scientific Linux 6.6 (neutrinos1) --> failing conda installation

~~Failing tests may be related with system incompatibility with newer scipy version, since they fail for tests using scipy functions producing nans, in particular in IC random-sampling and deconvolution functions.~~

The Ubuntu version of github actions has been updated to Ubuntu 20.04 from Ubuntu 18.04 to allow tests to pass.

We would appreciate some input from other machines.


invisible_cities/reco/psf_functions_test.py
```diff
@@ -55,6 +55,7 @@ def test_add_empty_sensors_and_normalize_q_file(ICDATADIR):
                                  database = load_db.DataSiPM('new', 0))
     hdst_processed.reset_index(inplace=True, drop=True)
     hdst_psf       = pd.read_hdf(PATH_TEST)
+    hdst_psf       = hdst_psf.astype(hdst_processed.dtypes) # since pandas version update
```
jacg: Minor point, but I'd mention the old and new version in the comment about the pandas version update.

And again when the same comment appears on line 72

gondiaz:
It seems that the confirmation for tests to pass expects a report from a (python=3.7, ubuntu-18.04) versions, any clue about it?

jmbenlloch:
> It seems that the confirmation for tests to pass expects a report from a (python=3.7, ubuntu-18.04) versions, any clue about it?

Don't worry about it. The `master` branch is currently setup to expect that check before any merge. It was the previous version we've been using. When this PR is approved, I'll update the configuration so github checks for `build (3.9, ubuntu-20.04)`

gondiaz:
I have downgraded to python version 3.8 and obtain the same results in the three machines tested until now (in the PR header marked with ✅ ), so I will adress the newly failing tests.

gonzaponte:
> We would appreciate some input from other machines.

We _need_ to ensure it works on every machine we use. It doesn't make sense to update to a version that doesn't work everywhere.

On that point...
> I have downgraded to python version 3.8

We cannot upgrade to python 3.9. The version of the libc library needed (>=2.14) is not available in neutrinos. We would need to install them separately and we might do so in the future, but for now, we stick to 3.8.

What is the current status of this PR?

gondiaz:
The PR should be ready to review. 

I would like to point out that some tests might fail in some machines due to the `hypothesis` default timing. That doesn't mean that the test is not passing (namely not fullfiling the assertions) but just that `hypothesis` califies them as failed. This can be addressed with a `deadline=None` setting as done in the last commit, but it might happen for other tests. This is a message example on these tests:
`Unreliable test timings! On an initial run, this test took 272.08ms, which exceeded the deadline of 200.00ms, but on a subsequent run it took 144.20 ms, which did not. If you expect this sort of variability in your test timings, consider turning deadlines off for this test by setting deadline=None.`

For example, @Aretno tested it on his machine with Ubuntu 18.02 obtaining this behaviour for the following test:
```
=========================== short test summary info ============================
FAILED invisible_cities/io/pmaps_io_test.py::test_load_pmaps - hypothesis.err...
= 1 failed, 956 passed, 13 skipped, 4 xfailed, 4985 warnings in 363.63s (0:06:03) =
```

This test in particular was not addressed in the last commit since it does not appeared at the machines tested at PR's header (should it?)

gonzaponte:
I am not able to install anything on neutrinos due to this:
`python: /lib64/libc.so.6: version GLIBC_2.14 not found (required by python)`

It is not actually because of this PR, I can't even install the main branch (how/when did this happen!?), but it's stopping us from checking that this update can be run there.

Besides the glibc problem, in that machine there are conflicts with the dependencies, so we cannot go forward with it yet.

@jmbenlloch do you have any insights on what's going on?

jacg:
> Besides the glibc problem, in that machine there are conflicts with the dependencies

Both of these are exactly the sorts of problems which are completely avoided (by design) in Nix.

(If only you could get Nix to run everywhere you need.)

gonzaponte:
> It is not actually because of this PR, I can't even install the main branch

An update on this. The environment I have right now on neutrinos works just fine, but it was installed weeks ago. Cloning it works just fine. However, after doing `conda update --all` it doesn't work because of the glibc problem. The list of updates can be found [here](https://pastebin.com/WG4KUaZJ).

More importantly, it looks like if I don't set up my conda installation or if I just try to install the environment with a different name (thus forcing conda to solve the dependencies again), it doesn't work either. Our dependencies are pinned, but the dependencies of our dependencies are not necessarily. I think the problem is coming from there somehow.

My main suspect was the `glib` package that went from `2.65.0-h3eb4bd4_0` to `2.69.1-h5202010_0`. But I haven't found anything on their changelog that could translate into our problem.

Finally, I tried to install a few different glibc versions (2.17 seems to be the absolute minimum we need) and add them to the `LD_LIBRARY_PATH`, but of course it didn't work. Updating the system libraries is not an option, apparently.

gonzaponte:
After an offline discussion with @jmbenlloch and @gondiaz, we have decided to stop supporting neutrinos until the OS is updated to some non-medieval version. The time scale for this update is O(months) so in the meantime, we will need to use other machines.

The next step for this PR is to ensure all machines can work with the proposed environment. We will contact individual users for a test run on those machines.

invisible_cities/detsim/light_tables_c_test.py
```diff
@@ -34,6 +34,7 @@ def test_LT_SiPM_optional_arguments(get_dfs):
 @given(xs=floats(min_value=-500, max_value=500),
        ys=floats(min_value=-500, max_value=500),
        sipm_indx=integers(min_value=0, max_value=1500))
+@settings(deadline=None, max_examples=100)
```
gonzaponte: can we define `few_examples = settings(deadline=None, max_examples=100)`? Also I think it makes sense for `given` to be the closest decorator to the function definition.
Also, how bad was the deadline issue?

gonzaponte (commented):
What is this thing of going from `np.type` to `type`? do you remember? If so, can you reword the commit message, so it includes the explanation?

invisible_cities/detsim/light_tables_c_test.py
gondiaz: If I remember correctly, the issue was not serious but due to a change in the default deadline of pytest in newer versions. Also, such failing tests depended on the machine, one may be faster than others in carrying the computation, I guess. I decorated all the failing tests of the machines I tested, then it may be more tests with this deadline issue in some particular machines.

gonzaponte (approved):
Important update to a more recent python version and this repository's dependencies.
Upgrade-related issues addressed clearly and warnings silenced.
Hopefully, the next upgrade won't take this long :)
Good job.

gonzaponte:
The 3.7 build is stuck and preventing this PR from being merged. Can you kill or ignore it manually @jmbenlloch ?

PR #809, opened 2021-12-01
Make paolina faster

gondiaz:
This PR makes Paolina a bit faster by avoiding duplicated function calls.
I naively estimated a 10-20% improvement.


paolafer (approved):
This PR is straightforward and makes the code more efficient. Approved!

PR #812, opened 2022-03-09
Fix attempt to download non-existing tables

gonzaponte:
Due to the mutability of lists, this code was modifying all the elements of the dictionary. This resulted in an attempt to download tables that don't exist in the database.

### DIFF
--- a/invisible_cities/database/download_test.py
+++ b/invisible_cities/database/download_test.py
@@ -21,10 +21,34 @@ def test_create_table_sqlite(dbname, output_tmpdir):
 
     connSqlite = sqlite3.connect(dbfile)
     connMySql  = pymysql.connect(host="neutrinos1.ific.uv.es",
-                                user='nextreader',passwd='readonly', db=dbname)
+                                 user='nextreader',passwd='readonly', db=dbname)
 
     cursorMySql  = connMySql .cursor()
     cursorSqlite = connSqlite.cursor()
 
     for table in db.tables:
         db.create_table_sqlite(cursorSqlite, cursorMySql, table)
+
+
+@mark.parametrize('dbname', db.dbnames)
+def test_table_assignment(dbname):
+    for name in db.common_tables:
+        assert name in db.table_dict[dbname]
+
+    for name in db.extended.get(dbname, ()):
+        assert name in db.table_dict[dbname]
+
+
+@mark.parametrize('dbname', db.dbnames)
+def test_tables_exist(dbname):
+    connMySql  = pymysql.connect(host="neutrinos1.ific.uv.es",
+                                 user='nextreader',passwd='readonly', db=dbname)
+
+    cursor    = connMySql.cursor()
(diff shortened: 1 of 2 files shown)


jacg (commented):
I believe that this does what it says on the tin. How difficult would it be to add a test?

If `tables` were a `tuple`, instead of a `list`, the original version would have worked as intended: `lhs += rhs` mutates `lhs` for mutable types (e.g. `list`) but rebinds it to a new object (`lhs + rhs`) for immutable types (e.g. `tuple`).

gondiaz (changes requested):
Good catch, this PR does what it claims. To me is ok if you add a test, but I think changing the lists to tuples would be good enough.

Also, even though the test in `download_test.py` is skipped due to random misconnections, the `dbname` and `table` are fixed inside the function, which is parametrized. Could you fix this either removing the parametrization or the harcoded `dbname`? We may decide to activate the test in the future.

gonzaponte:
> How difficult would it be to add a test?

I'm thinking of a test that checks for the existence of the corresponding tables for each database. It's not perfect, as @gondiaz mentioned we have issues with the connection to the server. I can think of another test, but it's a bit trivial, I will push it and you can comment on that.

> Could you fix this either removing the parametrization or the harcoded dbname? 

I will take a look into that, maybe it's the same as mentioned above.

jacg (commented):
Any reason why the *original* `tables` (lines 77-79) can't be a `tuple`?

gonzaponte:
No, I had just switching to that (not commited yet).

invisible_cities/database/download_test.py
```diff
@@ -28,3 +28,27 @@ def test_create_table_sqlite(dbname, output_tmpdir):
...
+def test_table_assignment(dbname):
+    for name in db.common_tables:
+        assert name in db.table_dict[dbname]
+
+    for name in db.extended.get(dbname, ()):
+        assert name in db.table_dict[dbname]
+
+
+@mark.parametrize('dbname', db.dbnames)
+def test_existing_tables(dbname):
```
gondiaz: Maybe `test_tables_exist` is a more descriptive name?

invisible_cities/database/download_test.py
```diff
@@ -28,3 +28,27 @@ def test_create_table_sqlite(dbname, output_tmpdir):
...
+    connMySql  = pymysql.connect(host="neutrinos1.ific.uv.es",
+                                user='nextreader',passwd='readonly', db=dbname)
```
gondiaz: Cosmetics
```suggestion
    connMySql  = pymysql.connect(host="neutrinos1.ific.uv.es",
                                 user='nextreader', passwd='readonly', db=dbname)
```

gondiaz (changes requested):

gondiaz (approved):
This PR fixes a bug in downloading database tables, good work, aproved!

PR #813, opened 2022-03-16
Fix incorrect calculation of peak width

gondiaz:
This PR changes the computation of signal (peak) width for signals. The current default is to sum the width of the bins with charge larger than 0. This may bias the expected/desired calculation for S1 signals, which are much shorter than S2's and it is common for them to contain zero charge bins.

### DIFF
--- a/invisible_cities/evm/pmaps_test.py
+++ b/invisible_cities/evm/pmaps_test.py
@@ -177,12 +177,7 @@ def test_Peak_height(pks):
     assert peak.height == approx(peak.pmts.sum_over_sensors.max())
 
 
-#@given(peaks())
-#def test_Peak_width(pks):
-#    _, peak = pks
-#    assert peak.width == approx(peak.times[-1] - peak.times[0])
-
-def test_Peak_width_correct():
+def test_peak_width_signal_positive():
     nsamples = 3
     times  = np.arange(nsamples)
     widths = np.full(nsamples, 1)
@@ -190,6 +185,32 @@ def test_Peak_width_correct():
 
     peak = S1(times, widths, pmts, SiPMResponses.build_empty_instance())
     assert peak.width == nsamples
+
+
+def test_peak_width_signal_not_positive():
+    nsamples = 10
+    times  = np.arange(nsamples)
+    widths = np.full(nsamples, 1)
+    signal = np.full((1, nsamples), 1)
+    signal[0, 1:nsamples-1:2] = 0
+    signal[0, 2:nsamples-1:2] = -100
+
+    pmts = PMTResponses(np.arange(1), signal)
+    peak = S1(times, widths, pmts, SiPMResponses.build_empty_instance())
+    assert peak.width == nsamples
+
+
+def test_peak_width_signal_limits():
+    nsamples = 10
+    times  = np.arange(nsamples)
+    widths = np.full(nsamples, 1)
+    signal = np.full((1, nsamples), 1)
+    signal[0, 0] = 0
+    signal[0,-1] = -100
+
+    pmts = PMTResponses(np.arange(1), signal)
+    peak = S1(times, widths, pmts, SiPMResponses.build_empty_instance())
+    assert peak.width == (nsamples-2)
 
 
 def _get_indices_above_thr(sr, thr):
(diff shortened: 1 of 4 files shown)


gonzaponte:
I would rather fix the `width_above_threshold` method, which is the one doing something fishy,

gondiaz:
what do you suggest? substitute the method by some `compute_width`?

gonzaponte:
No, I mean to fix `width_above_threshold` so it adds up all the bin widths that need to be included in the sum.

gonzaponte:
This doesn't fix the width calculation. I don't think there is anything stopping the waveform from being negative at some point, so in theory, a threshold of -1 can give you the same problem (only less frequently). I suggest that `width_above_threshold` takes the first and last (or last + 1) indices above threshold and sums everything in between. The current code is skipping indices within that range falling below the threshold, which is incorrect.
There are other options like calling the function with `-np.inf`, but it looks ugly, don't you think?

This is a typical case in which a very simple test can give you ~100% certainty that you are doing it right :). I suggest you start by coding this test using an unrealistic peak shape to cover all bases (like to not get tricked by the threshold = -1)

invisible_cities/evm/pmaps_test.py
```diff
@@ -192,6 +187,32 @@ def test_Peak_width_correct():
...
+    nsamples = 10
+    times  = np.arange(nsamples)
+    widths = np.full(nsamples, 1)
+    signal = np.full((1, nsamples), 1)
+    signal[0, 0] = 0
+    signal[0,-1] = -100
+
+    pmts = PMTResponses(np.arange(1), signal)
+    peak = S1(times, widths, pmts, SiPMResponses.build_empty_instance())
+    assert peak.width == (nsamples-2)
```
gonzaponte:
```suggestion
    assert peak.width == nsamples - 2
```

invisible_cities/evm/pmaps_test.py
```diff
@@ -192,6 +187,32 @@ def test_Peak_width_correct():
     assert peak.width == nsamples
 
 
+def test_peak_width_signal_not_positive():
```
gonzaponte:
```suggestion
def test_Peak_width_signal_not_positive():
```

invisible_cities/evm/pmaps_test.py
```diff
@@ -192,6 +187,32 @@ def test_Peak_width_correct():
...
+    signal = np.full((1, nsamples), 1)
+    signal[0, 1:nsamples-1:2] = 0
+    signal[0, 2:nsamples-1:2] = -100
+
+    pmts = PMTResponses(np.arange(1), signal)
+    peak = S1(times, widths, pmts, SiPMResponses.build_empty_instance())
+    assert peak.width == nsamples
+
+
+def test_peak_width_signal_limits():
```
gonzaponte:
```suggestion
def test_Peak_width_signal_limits():
```

invisible_cities/evm/pmaps_test.py
```diff
@@ -259,9 +280,12 @@ def test_Peak_width_above_threshold_max_zero(pks):
 def test_Peak_width_above_threshold(pks, thr):
     _, peak = pks
     i_above_thr     = _get_indices_above_thr(peak.pmts, thr)
-    expected        = (np.sum(peak.bin_widths[i_above_thr])
-                       if np.size(i_above_thr) > 0
-                       else 0)
+    if np.size(i_above_thr)>0:
```
gonzaponte:
```suggestion
    if np.size(i_above_thr) > 0:
```

invisible_cities/evm/pmaps_test.py
```diff
@@ -177,12 +177,7 @@ def test_Peak_height(pks):
     assert peak.height == approx(peak.pmts.sum_over_sensors.max())
 
 
-#@given(peaks())
-#def test_Peak_width(pks):
-#    _, peak = pks
-#    assert peak.width == approx(peak.times[-1] - peak.times[0])
-
-def test_Peak_width_correct():
+def test_peak_width_signal_positive():
```
gonzaponte:
```suggestion
def test_Peak_width_signal_positive():
```

gonzaponte (commented):
Just some cosmetics, ready to be approved.

gonzaponte (approved):
This PR fixes a long-standing bug in calculating the width of peaks. Clear and well tested. Approved!

PR #814, opened 2022-04-01
Fix isaura processing for zero-energy events

gondiaz:
Beersheba can return a deconvoluted track such all hit energies are zero. This can happen when esmeralda corrected energies are NaN values, when the event hits are outside the boundaries of the kr correction map.

When this kind of events are processed by isaura, the voxelization function returns an empty list, since no zero energy voxels are built. An empty list cannot be further processed (see `track_blob_info_creator_extractor` function).

This PR solves this issue by allowing isaura to process this events, such that paolina algorithm is not run for them.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -1094,6 +1094,7 @@ def make_event_summary(event_number  : int              ,
     DataFrame containing relevant per event information.
     """
     es = pd.DataFrame(columns=list(types_dict_summary.keys()))
+    if len(paolina_hits.hits) == 0: return es
 
     ntrks = len(topology_info.index)
     nhits = len(paolina_hits.hits)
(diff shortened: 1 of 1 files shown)


ausonandres:
As we have commented off-line, maybe it is better to replace the proposed solution: ``(hit_energies>0).all()``, by ``(hit_energies>0).any()``, in order to keep the events that also include well-reconstructed hits inside the volume.
They should be rejected anyway in the analysis posterior to the reconstruction chain, but I believe that we prefer to do it in the latest step. Do you agree?

gondiaz:
Agree

ausonandres:
And one additional question (perhaps for another PR?): do you think that there should be one filter that accounts for these events that are thrown away because all their hits have no reconstructed energy? Maybe the filter `filter_events_nohits` in `components/compute_and_write_tracks_info` could be extended to those cases, or a new one added. What do you think?

gondiaz:
Having such a filter might be fine, probably better in another PR since here just fixed the functions to be able to manage those cases.
Even though unrelated, it would also be relevant to define a configuration parameter in esmeralda to decide if you wish to run or not paolina reconstruction...

ausonandres (approved):
Although simple, this PR was necessary since it fixes a clear issue in the reconstruction.

Some changes commented in the conversation should be addressed in further PRs.

PR #815, opened 2022-04-25
Update NEXT100 activity/efficiency database

gondiaz:
Updates the activity and efficiency columns in NEXT100 database with the latest activity values from measurements and the efficiencies computed through nexus simulations. Efficiencies assume at least 2 MeV is deposited in the active volume.
Recall that the previous values were dummy, in order to test PR #806.


MiryamMV (approved):
This PR adds the correct values of activities (v3) and efficiencies to the database for NEXT100. I approve! Good job.

PR #816, opened 2022-04-27
Standardize symbols

gonzaponte:
This PR groups all symbols in a single place and tries to standardize their use. This means that the config files no longer take strings that are later converted to symbols within each city, which seems a bit silly.

I don't know who can review this, so feel free to jump in.


gonzaponte:
I need a reviewer. Any volunteers?

adamredwine:
I volunteer insofar as I am capable, though I acknowledge that I have not
yet established a reputation in the code.  Maybe I can do it in parallel?

gonzaponte:
Sure, take your time. This PR is a bit technical because affects the whole repo, but it's pretty much a refactor and reorganization of existing code.
Thanks for jumping in!

invisible_cities/types/symbols.py
```diff
@@ -0,0 +1,133 @@
...
+    individual = auto()
+
+
+class WfType(AutoNameEnumBase):
+    rwf  = auto()
+    mcrd = auto()
+
+
+ALL_SYMBOLS = {}
+for etype in dict(locals()).values():
```
gondiaz, line 126: It's correct, but might be better (less error prone in the futue) to define first the locals dictionary or list of values, such that
```suggestion
locls = dict(locals())
ALL_SYMBOLS = {}
for etype in locls.values():
```
just to avoid `ALL_SYMBOLS` to enter in the loop.

gondiaz (dismissed):

invisible_cities/reco/pmaps_functions.py
```diff
@@ -9,11 +7,7 @@
 from ..evm.pmaps           import             SiPMResponses
 from . peak_functions      import rebin_times_and_waveforms
 from ..core.core_functions import        dict_filter_by_key
-
-
-class RebinMethod(Enum):
-    stride    = 0
```
adamredwine, line 15: Why are the default values here converted to auto in the combined file?

This applies to several other classes as well so I assume it is intentional.  I won't comment in other locations.

invisible_cities/core/configure_test.py
```diff
@@ -15,9 +13,14 @@
 from . exceptions import NoInputFiles
 from . exceptions import NoOutputFile
 
+from .. types .symbols     import EventRange
 from .. cities.components  import city
 from .. cities.penthesilea import penthesilea
 
+all  = EventRange.all
```
adamredwine, line 20: Why do you have this defined as "all" here but as "all_events" in symbols.py?  Honestly, it's not clear to me why you define this alias at all rather than just calling the attribute when you need it.

invisible_cities/types/symbols.py
```diff
@@ -0,0 +1,100 @@
...
+
+class BlsMode(AutoNameEnumBase):
+    mean      = auto()
+    median    = auto()
+    scipymode = auto()
+    mode      = auto()
+
+
+class Contiguity(Enum):
+    FACE   = 1.2
```
adamredwine, line 16: Not a code comment but I don't understand why the all-caps variables get orange coloring.

adamredwine (approved):
I read every change closely and I did not see any clear problems.  The only thing that I didn't understand was why some classes had values assigned to attributes in the old classes that were replaced with auto() in the new classes.

adamredwine (approved):
This all seems fine and consistent with the first commit.  No comment.

adamredwine (approved):
Looks fine and consistent with other changes.

invisible_cities/core/configure.py
```diff
@@ -15,16 +15,16 @@
...
 
 def event_range(string):
     try:
         return int(string)
     except ValueError:
-        if string.lower() == 'all' : return EventRange.all
-        if string.lower() == 'last': return EventRange.last
-        # TODO Improve the message in the ValueError
-        raise ValueError("`--event-range` must be an int, 'all' or 'last'")
+        if string not in ALL_SYMBOLS:
```
adamredwine, line 25: I'm unfamiliar with a "not" in this usage.  I would expect to see "if not string in ALL_SYMBOLS:"

Does this provide the same functionality?

invisible_cities/types/symbols.py
```diff
@@ -98,3 +99,14 @@ class SiPMCharge(AutoNameEnumBase):
 class WfType(AutoNameEnumBase):
```
adamredwine, line 120: This was defined in the symbols.py file.  Why not import it from there?

invisible_cities/types/symbols.py
adamredwine: Probably fine.

adamredwine (commented):
I trust that the "not" statement works but I'd like to hear if there are differences in behavior because it seems quite unusual to me.

invisible_cities/cities/beersheba.py
```diff
@@ -86,14 +86,14 @@ def deconvolve_signal(det_db          : pd.DataFrame,
     iteration_tol   : Stopping threshold (difference between iterations).
     sample_width    : Sampling size of the sensors.
     bin_size        : Size of the interpolated bins.
-    energy_type     : Energy type ('E' or 'Ec', see Esmeralda) used for assignment.
-    deconv_mode     : 'joint' or 'separate', 1 or 2 step deconvolution, see description later.
```
adamredwine, line 100: These are just changes from single forward quote to single back quote? I have no issue with the change, but it seems like a pretty minor detail.

adamredwine (approved):
Now I understand that the ALL_SYMBOLS is added on to the end of the symbols.py file.  That confused me the first time around.  Okay, this looks fine.  I see how you added that as a separate commit and then used it later.  I'm starting to get the hang of the process.

adamredwine (approved):
Looks fine.

invisible_cities/core/random_sampling.py
```diff
@@ -233,7 +233,7 @@ def dark_expectation(self, sample_width : int,
 
         pad_xbins, _ = pad_pdfs(self.xbins, self.probs)
 
-        if dark_model == DarkModel.threshold:
```
adamredwine, line 241: What is the justification for switching == with "is"?  From what I see (e.g. https://realpython.com/courses/python-is-identity-vs-equality/), these are not identical comparisons.  In this case, the "threshold" attribute is probably just some number.  I don't know the process well enough to understand where it is assigned, but I can imagine it pointing to a different memory space than the dark_model variable.

adamredwine (approved):
Okay, now I've definitely, totally, for sure, gone through everything and understood it... I think.  :)  Generally it looks fine to me; I left a few comments where I didn't quite understand something.

invisible_cities/reco/pmaps_functions.py
gonzaponte: Just for consistency with everything else. There are only a couple of cases where the enum value is meaningful.

invisible_cities/core/configure_test.py
gonzaponte: I think I just did it to keep the definitions as they were before this PR so the relevant change was more obvious, but I will take a look and if there is no reason to put it like this I will replace it with `EventRange.all` instead of `all`

invisible_cities/types/symbols.py
gonzaponte: Probably interpreted as global variables by the editor, but I'm not sure.

invisible_cities/core/configure.py
gonzaponte: Yes, and actually it is the recommended syntax in python because it reads more like English.
I think the logic was that `not string in x` could be potentially read as `(not string) in x`. However this explanation clashes with this other syntax: `x is not None`.
`not None` is `True`, but `x is not None` is not the same as `x is True`. For instance for `x=3`
`x is not None` is `True`, but `x is True` is `False`,

invisible_cities/cities/beersheba.py
gonzaponte: Single quotes in python mean it is a string, which is no longer the case. Backquotes are for formatting. x in backquotes means that I am referring to some variable named x in the code (and you see that github formats it in a monospace font like `x`).
In this docstring it is important to make this distinction, because up to now the config file needed to contain strings and from this PR on they will contain symbols, i.e. instead of `energy_type = 'E'`, the config file will read `energy_type = E`

invisible_cities/core/random_sampling.py
gonzaponte: equality is an operation that takes two values and performs some logical operation on them, while identity compares memory positions. It is probably not critical in our case, but it is the right way to do it. Symbols are unique elements; thus, there is precisely one `DarkModel.threshold` (like there is only one `True` and one `False`), and you cannot make yourself another one.
I can think of two exapmles to showcase this:
- For `int`s, these two statements yield different results: `2**32 == 2**33/2`,  `2**32 is 2**33/2` (don't try it out with small integers, that's another story)
- In principle, I can write an object whose equality operator gives the wrong result. For instance
```
class A:
    def __eq__(self, other):
        return True
```
and then `A() == DarkModel.threshold` would return `True`. You cannot do the same with `is`.

regarding this:
>  the "threshold" attribute is probably just some number.

It's not. That's the associated value, which we mostly don't care about. That's the point of symbols, they are a completely separate entity. Effectively `DarkModel` is a new type and its attributes are the only possible values of this new type. Kind of like `bool` is a type and `True` and `False` are the only possible values of the `bool` type. Unlike the boolean type, `Enum`s have an associated value, but we (mostly) use them as symbols, where the associated value is meaningless and we only care about their uniqueness.

adamredwine: Awesome, thanks.

gonzaponte:
@adamredwine is this review done? Any requests? I know it's approved, but I think this was because you weren't sure about which option to use.

In the meantime, I will simply rebase on master and solve conflicts. This should require you to approve the PR again.

invisible_cities/types/symbols.py
```diff
@@ -98,3 +99,14 @@ class SiPMCharge(AutoNameEnumBase):
...
+
+ALL_SYMBOLS = {}
+for etype in dict(locals()).values():
+    if not isinstance(etype, EnumMeta): continue
+    if etype is EnumMeta              : continue
+    if etype is Enum                  : continue
+    if etype is AutoNameEnumBase      : continue
+
+    for symbol in etype:
+        ALL_SYMBOLS[symbol.name] = symbol
```
jacg, line 133: I don't like this implementation because 

+ it uses action at a distance,
+ takes everything in `locals()` and tries to exclude everything irrelevant (and the stability of this set makes me nervous in Python).

How about a decorator on the enum classes, that adds the class' symbols to `ALL_SYMBOLS`? Or it could be done with metaclasses.

But that might be more trouble than it's worth. This certainly should not further delay merging this PR!

jacg (approved):
An excellent PR that removes a flaky stringy-API and replaces it with a more robust symbol-based one. This makes it clear exactly what values are admissible. Errors will be spotted sooner, error messages will describe them more directly, it will be easier to fix them.

I thoroughly recommend that this PR be merged without further delay, especially as there are 2 more PR in the queue that are built on top of this one.

invisible_cities/types/symbols.py
gonzaponte: I like both approaches much more than mine. I will try them in a separate PR.

jacg:
Thanks for merging, @MiryamMV. You've been in the role for a couple of years now, is it time to train the next generation and pass on the baton?

MiryamMV:
Time flies and I didn’t even notice I’ve been doing this for 2 years now. I don’t want to point out anybody so.. any volunteers? In the meantime, I’ll be happy to keep on with the merges for a few more months.

PR #818, opened 2022-05-23
Simplify file compression

gonzaponte:
In this PR, I revisit the file compression in IC. The main point was that this feature relied on the same default value being used everywhere (namely `"ZLIB4"`). While addressing this, I noticed that we didn't have any check for compression and that we were doing it inconsistently. Furthermore, the compression options were redundant, as the compression library and level can be set at file opening, and the data stored inherits this configuration.

There are two significant changes implemented here:
- Writers will no longer compress something with a default value that magically happened to be the same everywhere. As a consequence, cities are now fully responsible for file compression.
- By default, the compression configuration will be set only at file opening. This, however, can be overridden for each individual writer explicitly.

These changes won't have a major impact on users or developers. Nonetheless, developers writing new cities will have to be aware of this.

### DIFF
--- a/invisible_cities/cities/beersheba.py
+++ b/invisible_cities/cities/beersheba.py
@@ -297,14 +297,13 @@ def check_nonempty_dataframe(df) -> bool:
     return len(df) > 0
 
 
-def deconv_writer(h5out, compression='ZLIB4'):
+def deconv_writer(h5out):
     """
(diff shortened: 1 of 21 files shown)


jmbenlloch (approved):
Good job! The code is more consistent now and less prone to errors. I like your `test_city_output_file_is_compressed` test. Approved!

PR #821, opened 2022-06-15
DEMO SiPMs calibration

carmenromo:
Demo++ database has been updated with the calibration constants from R11179 for the SiPMs and pdfs. SiPM 14033 seems to be dead, therefore it has been masked. Also, the sensor ids have been fixed according to the frontends change and the positioning of the dices. Two sensors are unable to calibrate but they are hidden by the light tube (plotted in pink, top left, the red one is dead):

Screenshot 2022-06-15 at 12 15 48

Screenshot 2022-06-15 at 12 12 10

Screenshot 2022-06-15 at 12 11 11


jmbenlloch (approved):
Latest calibration added and new sensor IDs included.

PR #823, opened 2022-07-15
Filter nexus delayed hits in detsim

gondiaz:
`detsim` contains a parameter `max_time` such that all hits ocurring afterwards this threshold are not included in the waveform simulation. For example, a nucleus can decay to its ground state via two gamma deexcitations, being the decay halt-life of the latter much larger than the initial. 
In such a case, if the second decay is within the limit set by `max_time`, `detsim` will be able to process both signals and split them in two different events if the second decay occurs at time larger than the `buffer_length` of the waveform.
But if the second decay time is larger than `max_time`, `detsim` is not filtering these hits and an error is raised in the waveform creation.

This PR solves this issue by adding a function that filters the hits at times larger than `max_time`. A warning is raised for these events for the user to be aware that the simulation might not provide the expected output.
This PR also fixes the fact that S1 has to be computed using the hits at the active and buffer volumes, not only in the active as was previously done.


invisible_cities/detsim/detsim_waveforms.py
```diff
@@ -1,37 +1,51 @@
...
     """
-    s1_times are histogramed in a waveform.
-    Parameters
-        :s1_times: list
-            output of generate_s1_times_from_pes, list of size equal to the
-            number of pmts. Each element is an array with the times of the S1
+    Creates the s1-waveform based on the pdf at :s1_times_pdf:
+
+    Parameters:
+        :s1_pes_at_pmts: np.ndarray
```
MiryamMV: This variable is called different in the definition of the function.

invisible_cities/detsim/detsim_waveforms.py
```diff
@@ -1,37 +1,51 @@
...
-            waveform buffer lenght in time units
-        :bin_width: float
-            waveform bin width in time units
-        :start_time:
-            waveform time of first bin edge in time units
+            waveform buffer_length, waveform will range from tmin to
+            tmin + buffer_length
+        :bin_width:
+            waveform bin width
+        :tmin: float
```
MiryamMV: Same here.

MiryamMV (approved):
This PR raises a warning instead of filtering the delayed hits after `max_time`. It also computed the `s1` using the active and buffer volumes. Well done @gondiaz!

MiryamMV:
This addition to the code fixes the error raised occasionally in low energy events. Thanks  @gonzaponte!!

gonzaponte:
My addition filters out events that do not produce any photons, which generated an error.

PR #824, opened 2022-07-18
Upgrade config

gonzaponte:
This PR forces every city to be fully-annotate its input arguments. Hopefully, this will help new users to understand what parameters they need to provide to a city and slightly improve the self-documentation of the cities.
The types of input arguments are checked before running the city. Inner functions can also be fully-annotated and type-checked by using the provided decorator, `check_annotations`.

### DIFF
--- a/invisible_cities/cities/beersheba.py
+++ b/invisible_cities/cities/beersheba.py
@@ -22,6 +22,9 @@ from numpy       import nan_to_num
 
 from typing      import Tuple
 from typing      import List
+from typing      import Union
+from typing      import Sequence
+from typing      import Optional
 
 from .  components import city
 from .  components import collect
@@ -30,6 +33,8 @@ from .  components import print_every
 from .  components import cdst_from_files
 from .  components import summary_writer
 
+from .. core.configure         import EventRangeType
+from .. core.configure         import OneOrManyFiles
 from .. core.configure         import check_annotations
 
 from .. reco                   import tbl_functions           as tbl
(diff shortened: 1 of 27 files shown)


invisible_cities/core/configure_test.py
```diff
@@ -401,3 +416,240 @@ def test_configure_numpy(config_tmpdir):
     conf_ns = configure(argv).as_namespace
     assert hasattr(conf_ns, "a_numpy_array")
     assert conf_ns.a_numpy_array.tolist() == [0, 0.5, 1]
+
+
+@mark.parametrize( "value type".split()
```
jacg, line 421: While these tests are important, let's just take a few seconds to observe that we have over 100 lines of boilerplate that we get for free in a (strong! C++ doesn't cut the mustard) static type system!

invisible_cities/cities/beersheba.py
```diff
@@ -67,13 +72,13 @@ def deconvolve_signal(det_db          : pd.DataFrame,
...
-                      cut_type        : CutType=CutType.abs,
-                      inter_method    : InterpolationMethod=InterpolationMethod.cubic,
-                      n_iterations_g  : int=0):
+                      diffusion       : Optional[Tuple[float, float, float]]=(1., 1., 0.3),
+                      energy_type     : Optional[HitEnergy]=HitEnergy.Ec,
+                      deconv_mode     : Optional[DeconvolutionMode]=DeconvolutionMode.joint,
+                      n_dim           : Optional[int]=2,
+                      cut_type        : Optional[CutType]=CutType.abs,
+                      inter_method    : Optional[InterpolationMethod]=InterpolationMethod.cubic,
+                      n_iterations_g  : Optional[int]=0):
```
jacg, line 81: Minor point: this does not conform to the IC alignment style. There are a few more instances of this in the code.

invisible_cities/cities/dorothea.py
```diff
@@ -57,15 +60,33 @@
...
+            , drift_v           : float
+            , s1_nmin           :   int, s1_nmax          :   int
+            , s1_emin           : float, s1_emax          : float
+            , s1_wmin           : float, s1_wmax          : float
+            , s1_hmin           : float, s1_hmax          : float
+            , s1_ethr           : float
+            , s2_nmin           :   int, s2_nmax          :   int
+            , s2_emin           : float, s2_emax          : float
+            , s2_wmin           : float, s2_wmax          : float
+            , s2_hmin           : float, s2_hmax          : float
```
jacg, line 83: I find the spatial proximity of the type of the earlier parameter to the name of the next parameter, a bit confusing to read. That is to say, in something like
```python
, s1_nmin          : float, s1_emax          :float
```
the natural inclination is to see the `float, s1_emax` as belonging together.

jacg (approved):
Improves automatic verification of cities' parameters, resulting in better, enforced documentation and checking of configurations. An important step forward in long-term usability and mainainability of the code.

NB: this sits on top of #816, which needs merging first. #816 has already been approved.

invisible_cities/cities/dorothea.py
gonzaponte: I'm planning a PR applying a more uniform style everywhere, all these cosmetic changes will fit in there naturally.

PR #825, opened 2022-10-17
Change `max_time` annotation in Buffy functions

halmamol:
In Buffy, `max_time` is defined as the maximum duration of the event that will be taken into account starting from the first detected signal. Despite it's a time unit it was defined as an `int`.  This PR modifies its type annotations from `int` to `float` in all the functions used by Buffy.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -992,6 +992,9 @@ def check_max_time(max_time: float, buffer_length: float) -> Union[int, float]:
     :param buffer_length: Length of buffers.
     :return: `max_time` if `max_time` >= `buffer_length`, else `buffer_length`.
     """
+    if max_time % units.mus:
+        message = "Invalid value for max_time, it has to be a multiple of 1 mus"
+        raise ValueError(message)
 
     if max_time < buffer_length:
         warnings.warn("`max_time` shorter than `buffer_length`, "
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -981,7 +981,7 @@ def compute_and_write_pmaps(detector_db, run_number, pmt_samp_wid, sipm_samp_wid
     return compute_pmaps, empty_indices, empty_pmaps
 
 
-def check_max_time(max_time: Union[int, float], buffer_length: float) -> Union[int, float]:
+def check_max_time(max_time: float, buffer_length: float) -> Union[int, float]:
     """
     `max_time` must be greater than `buffer_length`. If not, raise warning
         and set `max_time` == `buffer_length`.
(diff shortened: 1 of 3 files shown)


invisible_cities/cities/components.py
```diff
@@ -989,7 +989,7 @@ def check_max_time(max_time: Union[int, float], buffer_length: float) -> Union[i
 
 
 def calculate_and_save_buffers(buffer_length    : float        ,
-                               max_time         : int          ,
+                               max_time         : float          ,
```
gonzaponte: realign commas

invisible_cities/detsim/buffer_functions.py
```diff
@@ -18,7 +18,7 @@ def bin_sensors(sensors   : pd.DataFrame,
                 bin_width : float       ,
                 t_min     : float       ,
                 t_max     : float       ,
-                max_buffer: int         ) -> Tuple[np.ndarray, pd.Series]:
+                max_buffer: float         ) -> Tuple[np.ndarray, pd.Series]:
```
gonzaponte: realign closing parenthesis with commas

andLaing:
I'm probably very out of date with my understanding of IC but I think the int was there since the SiPMs have a binning of 1 µs so it was unlikely to be less than an integer multiple of µs that were used in the code and the dreaded float accuracy should be avoided at all costs.

gonzaponte:
I'm not sure about this. Technically they are floats, but in practice, they will only take integer values.

I have two comments. First, the argument should also apply to `buffer_length`, right? Second, regarding
>  the dreaded float accuracy should be avoided at all costs

I think we are not dealing with that, actually. Since the units are defined as floats, any value of `max_time` with the corresponding units is actually a float...

If I understand correctly, what matters (mostly) is the ratio between `max_time` and `buffer_length`, and I don't see any potential rounding problem with that because we round up the ratio.
Do you agree, @andLaing? To me, it seems safe to work with floats, with some notes on the documentation making it more clear, perhaps.

invisible_cities/cities/components.py
```diff
@@ -992,6 +992,9 @@ def check_max_time(max_time: Union[int, float], buffer_length: float) -> Union[i
     :param buffer_length: Length of buffers.
     :return: `max_time` if `max_time` >= `buffer_length`, else `buffer_length`.
     """
+    if max_time % units.mus:
+            message = "Invalid unit for max_time, it has to be a multiple of 1 mus"
+            raise ValueError(message)
```
gonzaponte: Each indentation level should be 4 spaces

gonzaponte: Also
Invalid unit -> Invalid value

gonzaponte (commented):
Just some cosmetic changes, ready for approval

gonzaponte (approved):
This PR addresses the type inconsistency of `max_time`. It is a float, but not all values of floats are allowed. This PR implements this requirement in the code while giving the correct type hint. It also adds a test to demonstrate it.
Great job, @halmamol!

PR #826, opened 2022-10-25
Add MC tables on kdsts

### LINKED ISSUE #822: Add MC info in kdsts
The title is self explanatory, MC info is currently not saved to kdsts which might be usefull for example to study the performance of position reconstruction or efficiency at large radius.

Hist: just check how MC info is added in other cities

carhc:
This is a small change in Dorothea, only to include MC info from previous files in case it’s needed and/or wanted. There’s a new boolean parameter, `include_mc`, to choose whether to copy MC tables (`True`) or not (`False`). By default, `include_mc = False` and therefore no MC info is added.

This was suggested on issue #822


gonzaponte:
Closes #822

invisible_cities/cities/dorothea.py
```diff
@@ -90,19 +97,27 @@ def dorothea(files_in, file_out, compression, event_range, print_mod, detector_d
...
+                                    event_count_in       .spy             ,
+                                    classify_peaks                        ,
+                                    pmap_passed                           ,
+                                    fl.branch(write_pmap_filter)          ,
+                                    pmap_select          .filter          ,
+                                    event_count_out      .spy             ,
+                                    build_pointlike_event                 ,
+                                    fl.branch("event_number", evtnum_collect.sink), 
+                                    fl.fork(write_pointlike_event         ,
+                                        write_event_info              )),
```
gonzaponte: branch and fork are similar but not identical. Branch creates a separate parallel pipeline, but doesn't consume the input pipeline, that is:

```
       ----
     /
 ---------
```
while fork splits the input pipeline into N pipelines consuming it, like

```
       ----
     /
 ---
     \
       ----
```
The important bit is that while branch _must_ have some other component after it (it's not a sink), fork _can't_ have anything after it because it terminates the pipeline (it is a sink).

All this crap was just to say that we should put this new branch you added into the fork, so it looks more natural. We can talk offline if you don't know how to do it.

invisible_cities/cities/dorothea.py
```diff
@@ -90,19 +97,27 @@ def dorothea(files_in, file_out, compression, event_range, print_mod, detector_d
...
+                      result = dict(events_in   = event_count_in .future,
+                                    events_out  = event_count_out.future,
+                                    evtnum_list = evtnum_collect .future, 
+                                    selection   = pmap_select    .future))
+                      
+        if run_number <= 0 and include_mc == True:
```
gonzaponte: since `include_mc` is a boolean, it is more natural to simply write `include_mc` instead of `include_mc == True`

invisible_cities/cities/dorothea.py
```diff
@@ -60,7 +63,9 @@ def dorothea(files_in, file_out, compression, event_range, print_mod, detector_d
              drift_v,
              s1_nmin, s1_nmax, s1_emin, s1_emax, s1_wmin, s1_wmax, s1_hmin, s1_hmax, s1_ethr,
              s2_nmin, s2_nmax, s2_emin, s2_emax, s2_wmin, s2_wmax, s2_hmin, s2_hmax, s2_ethr, s2_nsipmmin, s2_nsipmmax,
+             include_mc = False,
              global_reco_params=dict()):
```
gonzaponte: `include_mc` should go after `global_reco_params`

gonzaponte (changes requested):
Looks good, but some small changes are needed. Also, we need a test that checks that the MC info is included in the output. I suggest first adding a test that inspects the output file for the existence of the regular output tables and later parametrize it to look for the MC tables when `include_mc` is `True`

invisible_cities/cities/dorothea.py
jacg: Perhaps it's worth noting that `dataflow` is documented in [these highly-commented illustrative tests](https://github.com/next-exp/IC/blob/master/invisible_cities/dataflow/dataflow_test.py).

Specifically, `branch` and `fork` are compared and contrasted [here](https://github.com/next-exp/IC/blob/7b8d436ad16b94fdd04eae6b597145c498abf53c/invisible_cities/dataflow/dataflow_test.py#L345-L350).

`branch` and `fork` allow you to express the same flows, but from slightly different perspectives, so one might be more natural or convenient than the other, depending on the context.

One thing which is not evident from the documentation test linked above, is that `branch` always splits one flow into exactly 2 flows, while `fork` splits one flow into an arbitrary number of flows.

invisible_cities/cities/dorothea_test.py
```diff
@@ -101,24 +102,38 @@ def test_dorothea_filter_events(config_tmpdir, Kr_pmaps_run4628_filename):
...
+    true_output = os.path.join(ICDATADIR, "Kr83_nexus_v5_03_00_ACTIVE_7bar_3evts.KDSTwithMC.h5")
+
+    if include_mc:
+
+        file_out = os.path.join(output_tmpdir, 'dorothea_exact_result_with_MC.h5')
+        tables   = tables + ("MC/hits", "MC/particles")
+
+    else:
+
+        file_out = os.path.join(output_tmpdir, 'dorothea_exact_result_without_MC.h5')
```
gonzaponte: I think in this case it's better to gather the definitions like:
```
tables = ...
if include_mc:
    extend tables

label    = "with" if include_mc else "without"
file_out = os.path.join(output_tmpdir, f"dorothea_exact_result_{label}_MC.h5")
```

gonzaponte (commented):
I see that you are adding some unnecessary blank lines. We want to remove them in order to keep the style uniform.

More or less as a reference, I use blank lines to visually identify different parts of a procedure (like running the city and checking its output) or to group expressions/statements that are complementary in some form (like in the comment below: several expressions to properly define `tables`, a blank line, several expressions to properly define `file_out`), etc.

Also, there is a merge commit there that needs to be addressed, but we will do it at the end.

invisible_cities/cities/dorothea_test.py
```diff
@@ -101,21 +103,56 @@ def test_dorothea_filter_events(config_tmpdir, Kr_pmaps_run4628_filename):
...
+        assert "Run/events"           in h5out.root
+        assert "Run/runInfo"          in h5out.root
+        assert "Filters"              in h5out.root
+        assert "Filters/s12_selector" in h5out.root
+        assert "DST"                  in h5out.root
+        assert "DST/Events"           in h5out.root
+        if include_mc:
+            assert "MC"               in h5out.root
+            assert "MC/hits"          in h5out.root
+            assert "MC/particles"     in h5out.root
```
gonzaponte: Add also an `else` clause to check that these tables don't exist if `include_mc` is False.
Another (fancier) way would be to do
`assert not (node_name in h5out.root) ^ include_mc` 
which covers both cases without the need for a `if/else` clause, but it's perhaps a bit too cryptic :)

gonzaponte (commented):
Just a small thing to add. Then we will fix the commit history and approve.

invisible_cities/cities/dorothea_test.py
carhc: My first encounter with the caret symbol on Python, I had to look it up! The XOR makes it certainly more elegant, but I agree that maybe it can be a little bit confusing too. It might not be as straight forward as the other way in case someone is just quickly inspecting the file, even if the reader is familiar with the symbol. But maybe it's worth changing it anyway, I'm OK with that!

gonzaponte:
@jmbenlloch can you add @carhc to the organization?

gonzaponte (approved):
This PR includes the option of adding MC info to kdst files. The functionality is appropriately tested and well-implemented.

Good job!

gonzaponte:
Ready to merge @MiryamMV

MiryamMV:
Just merged! Congrats @carhc on your first contribution to IC!!

PR #827, opened 2022-11-21
Rename argument to avoid confussion

gonzaponte:
The first argument to the gauss function was named `amp` which might mislead the reader to believe that it's the function amplitude at `x=mu`. It has been renamed to `area`


carmenromo (approved):
Gaussian function argument has been renamed to avoid misleading. Approved!

PR #830, opened 2023-03-20
Remove event model

martinperezmaneiro:
This PR removes the use of the event model from the cities and replaces it for pandas DataFrames. This will probably make the processing faster.


gonzaponte:
Superseded by #912 and another one to come after that.

PR #831, opened 2023-03-21
Update database host

jmbenlloch:
The MySQL database server we were using (neutrinos1) is too old and very difficult to update in that machine. I have migrated all the databases to a new server (next). This PR updates the hostname in the source code, so it can still download all the tables.

The phpmyadmin web interface is working with the new server already. In the future, new calibrations or any other changes must be performed either using phpmyadmin or connecting to the new server. The MySQL server at neutrinos1 will be stopped as soon as this PR is approved.


paolafer (approved):
This PR changes the name of the database server. The database is correctly downloaded and the tests pass: approved!

PR #832, opened 2023-04-06
Update NEXT100 SiPM channel mapping

jmbenlloch:
This PR updates the channel mapping of SiPM in the database, according the assembly plans for NEXT-100.

The different dice boards (DB) are placed like this:

image

and the internal mapping of each DB goes like this:

image

The `SensorID` is the one used on the simulation and `ChannelID` is the one we are getting from the DAQ.

Using the DB number and the sensor number a new ID is built. Basically one has to add `(DB number + 1)*1000 + sensor_number`. For instance, the sensor 0 of DB 0 would be sensor number 1000. DB 0 goes from sensor 1000 to 1063. The same pattern applies until DB 55 with sensors 56000 to 56063.

Looking at the previous table, the sensor with `SensorID` 1000 would have `ChannelID` 1013. The same applies for all other sensors. Every DB is mounted with the same angle, so there is no relative rotation between one and another.

That is the pattern included in the database `NEXT100DB` for all the 3584 SiPMs.


paolafer (approved):
This PR updates the mapping of NEXT100 SiPMs according to the current situation. Approved!

PR #272, opened 2017-07-17
Cities dev

jjgomezcadenas:
This PR simplifies the structure of the cities, making then one step closer to the goal of being truly invisibles. The most important features are:

1) Most cities have the same run() method, that now goes to the base (only Zaira needs to re-write the run() method).

2) Cities reading RWF (Isidora and Irene) have the same file_loop() method, that now goes to their common base.

3) Cities reading PMAPS (Dorothea and Penthiselea) have the same file_loop() method that now goes to their common base (Dorothea and Penthiselea both inherit now from HitCity).

4) A counter class allows easy bookeping and common go() method. 

Overall, the concrete cities are much thiner and simpler, with considerable reduction of noise and sizeable repayment of TD).

PR #833, opened 2023-04-13
Refactor xy algorithms

### LINKED ISSUE #587: Parametrized algorithm for xy reconstruction
I'd like to bring up again the matter of being able to choose a particular algorithm for xy reconstruction, via configuration parameter. Right now, I'm testing variations of corona and also different algorithms and I think that it would be better to have a code with several options, at least for a while.
In the ancient times, we had this possibility, but it was removed at some point.
If you agree, I volunteer to implement the code to do this.

gonzaponte:
This PR provides a refactor of the basic reconstruction algorithms. It includes, among other things:
- the decoupling of `corona`  and `barycenter`
- the addition (and check) of annotations to the xy reconstruction functions
- a generalization of penthesilea and dorothea to allow different reconstruction algorithms
- a refactor of the functions related to hit-building used in penthesilea

Closes #587.


gonzaponte:
Any volunteers to review this?

adamredwine:
I’d be willing to try if my poor effort wasn’t too off putting last time.

invisible_cities/reco/xy_algorithms.py
```diff
@@ -58,13 +59,15 @@ def count_masked(cs, d, datasipm, is_masked):
...
 
 
-def corona(pos, qs, all_sipms,
-           Qthr            =  0 * units.pes,
-           Qlm             =  5 * units.pes,
-           lm_radius       =  0 * units.mm,
-           new_lm_radius   = 15 * units.mm,
-           msipm           =  3,
-           consider_masked = False):
+def corona( pos             : np.ndarray
```
adamredwine: For what it's worth, you didn't change the Usage Example below to match the new input expectations (i.e. no default values) reflected here.

invisible_cities/reco/xy_algorithms.py
```diff
@@ -25,8 +25,7 @@ def barycenter(pos : np.ndarray, qs : np.ndarray):
         ...
         [xs, ys])
        qs = vector (q1, q2...qs) --> (1xn)
-
-        """
```
adamredwine, line 31: When Steve Jobs' father told him as a kid to paint the fence of their yard, he painted only the outside. When his dad called him on it, he said no one would see the inside and his dad responded "but we will."

Nice fence painting.  :)

invisible_cities/cities/penthesilea.py
```diff
@@ -54,26 +55,28 @@
...
-               , s1_wmin           : float, s1_wmax     : float
-               , s1_hmin           : float, s1_hmax     : float
-               , s1_ethr           : float
-               , s2_nmin           :   int, s2_nmax     :   int
-               , s2_emin           : float, s2_emax     : float
-               , s2_wmin           : float, s2_wmax     : float
-               , s2_hmin           : float, s2_hmax     : float
-               , s2_ethr           : float
-               , s2_nsipmmin       :   int, s2_nsipmmax :   int
+def penthesilea( files_in           : OneOrManyFiles
```
adamredwine, line 58: No real comment here I just don't understand what has been changed other than the two additional arguments (slice_reco_algo and global_reco_algo) at the bottom.  It looks like the deleted and added lines are the same... right?

invisible_cities/conftest.py
```diff
@@ -378,43 +378,43 @@ def KrMC_hdst(ICDATADIR):
              7, 9, 2, 7, 2, 4]
 
     X     = [137.02335083, 136.05361214, 155, 137.50739563,
-             139.44340629, 135, 0.,
+             139.44340629, 135, NN,
```
adamredwine, line 381: I don't doubt that this is appropriate, but I am curious what the intention is here.  It seems odd to me to mix numbers and NN in lists like this.  I can understand if there is a concern about divide by zeroes or something I suppose.  What is the purpose of this change?

adamredwine (approved):
On the whole, this seemed pretty straightforward.  I had no big technical comments, just a couple points of curiosity.

invisible_cities/reco/xy_algorithms.py
gonzaponte: Good catch. I will update it.

invisible_cities/cities/penthesilea.py
gonzaponte: I believe the whole thing is marked because I realigned all the colons one place to the right.

invisible_cities/conftest.py
gonzaponte: `NN` is not actually a `nan` as you might expect. It's defined as `-99999` or something like that. I missed this part of the development but I think the reason to use `NN` and not `nan` was related to I/O. I plan to revisit this in the near future, but for now, it will stay like this.
As for your actual question, I replaced 0 with NN because 0 is a realistic value that you might not notice when analyzing data. For instance, if you plot a bunch of events and only a few of them have `0` you will not see anything weird. If they have `-99999` you surely will.

jacg: I remember discussing this many moons ago, and being unconvinced, but I don't recall the details at all. Here is the [relevant commit](https://github.com/next-exp/IC/commit/1050194986d6ac1d2a897b9a3bc1d3c9d5a166a9) with a message which offers a little explanation, but not enough for me understand exactly what problem it is supposed to solve and how.

The commit was part of #272, but a quick glance at the discussion doesn't reveal any deep insights.

gonzaponte: Thanks, this is what I vaguely remembered. I will try to dig deeper when I finally tackle this.

invisible_cities/reco/xy_algorithms.py
gonzaponte: Wait, the usage example does specify all arguments. Unless you mean the K1, K2, and K3 values, which are left unspecified because the description of the functionality does not depend on those.

gonzaponte: @adamredwine ?

adamredwine:
Odd... I thought I replied to this comment but my response is not showing
up on github; surely a user error.  I was referring to the K values in my
comment but also the default on the consider_masked argument appears to
have changed from True to False.

gonzaponte:
Mmm, I am seeing that it was already `False` before. Am I missing something?

adamredwine:
Ah, no, my bad.  I was misreading line 124.  I'm accustomed to seeing the
defaults given in this way.  I think you're good.

invisible_cities/reco/xy_algorithms.py
```diff
@@ -95,17 +110,13 @@ def corona(pos, qs, all_sipms,
                 ***In general lm_radius should typically be set to 0, or some value slightly
                 larger than pitch or pitch*sqrt(2).***
 
-                ***If lm_radius is set to a negative number, the algorithm will simply return
-                the overall barycenter all the SiPms above threshold.***
-
-
                 ---------
                     This kwarg has some physical motivation. It exists to try to partially
                 compensate problem that the NEW tracking plane is not continuous even though light
```
paolafer: problem --> the problem.

invisible_cities/reco/xy_algorithms.py
```diff
@@ -119,7 +130,7 @@ def corona(pos, qs, all_sipms,
                     lm_radius can always be set to 0 mm, but setting it to 15 mm (slightly larger
```
paolafer: This is a repetition of line 110, isn't it?

invisible_cities/reco/xy_algorithms.py
```diff
@@ -95,17 +110,13 @@ def corona(pos, qs, all_sipms,
...
-                the overall barycenter all the SiPms above threshold.***
-
-
                 ---------
                     This kwarg has some physical motivation. It exists to try to partially
                 compensate problem that the NEW tracking plane is not continuous even though light
                 can be emitted by the EL at any (x,y). When lm_radius < pitch, the search for SiPMs
                 that might contribute pos and charge to a new Cluster is always centered about
                 the position of hottest_sipm. That is, SiPMs within new_lm_radius of
-                hottest_sipm are taken into account by barycenter(). In contrast, when
```
paolafer, line 108: I suggest that the paragraph be modified to improve it.
Current version: "Consider the case where at a local maximum there are four nearly equally 'hot' SiPMs. new_local_maximum would yield a pos, pos1, between these hot SiPMs. Searching for SiPMs that contribute to this cluster within new_lm_radius of pos1 might be better than searching searching for SiPMs  within new_lm_radius of hottest_sipm."
Proposal: "Consider the case where there are four nearly equally 'hot' SiPMs around a local maximum. new_local_maximum would yield a position, pos1, somewhere among these hot SiPMs. Searching for SiPMs that contribute to this cluster within new_lm_radius from pos1 might be better than searching for SiPMs  within new_lm_radius from hottest_sipm."

invisible_cities/reco/xy_algorithms.py
paolafer: Also, `15 mm` makes sense in NEW, but in NEXT-100 there's a different pitch. It would be better to use `pitch * sqrt(2)`.

paolafer (changes requested):

invisible_cities/reco/xy_algorithms.py
```diff
@@ -8,38 +9,49 @@
...
-
-        """
-
-    if not len(pos)   : raise SipmEmptyList
-    if np.sum(qs) == 0: raise SipmZeroCharge
+    """
+    pos, qs = threshold_check(pos, qs, Qthr)
     mu, var = weighted_mean_and_var(pos, qs, axis=0)
     # For uniformity of interface, all xy algorithms should return a
     # list of clusters. barycenter always returns a single clusters,
```
paolafer: single clusters --> single cluster.

invisible_cities/reco/xy_algorithms_test.py
```diff
@@ -173,13 +173,19 @@ def test_barycenter_single_cluster_generic(p_q):
...
     with raises(SipmEmptyList):
-        algorithm(np.array([]), None)
+        algorithm(np.array([]), np.array([]))
 
 
-@parametrize("algorithm", (barycenter, partial(corona, all_sipms=datasipm)))
+@parametrize("algorithm", (barycenter, corona_default))
 def test_raises_sipm_zero_charge(algorithm):
     with raises(SipmZeroCharge):
         algorithm(np.array([[1, 2]]), np.array([0, 0]))
```
paolafer: It seems like there's one position, but two charges. Is this possible? The test doesn't fail, but is that sensible?

invisible_cities/reco/xy_algorithms_test.py
```diff
@@ -188,6 +194,7 @@ def test_raises_sipm_zero_charge(algorithm):
 def test_corona_barycenter_can_be_the_same_with_one_cluster(toy_sipm_signal, datasipm):
```
paolafer: Why `can be` the same? Shouldn't they give the same result always?

invisible_cities/reco/xy_algorithms_test.py
```diff
@@ -205,7 +212,9 @@ def test_corona_multiple_clusters(toy_sipm_signal, datasipm):
     """
     pos, qs = toy_sipm_signal
     clusters = corona(pos, qs, datasipm,
-                      msipm=1, new_lm_radius=15*units.mm, Qlm=4.9*units.pes)
+                      Qthr = 0, Qlm=4.9*units.pes,
```
paolafer, line 219: In line 209: cluster.xy --> cluster.XY.

invisible_cities/reco/xy_algorithms_test.py
```diff
@@ -237,7 +246,10 @@ def test_corona_min_threshold_Qthr(datasipm):
...
     pos, qs = toy_sipm_signal
     with raises(ClusterEmptyList):
-        corona(pos, qs, datasipm, msipm=2)
+        corona( pos, qs, datasipm
+              , Qthr=0, Qlm=5 * units.pes
+              , lm_radius=0, new_lm_radius=15 * units.mm
+              , msipm=2)
 
 
 @mark.skip
```
paolafer: As discussed offline, we could maybe resuscitate this test, adding the missing arguments, if you think it's worth it.

paolafer (changes requested):

invisible_cities/reco/xy_algorithms.py
```diff
@@ -58,100 +70,119 @@ def count_masked(cs, d, datasipm, is_masked):
...
+        Charge threshold to apply to all SiPMs. SiPMs with lower
+        charge are ignored
+
+    Qlm : float
+        Charge threshold to find a local maximum
+
+    msipm : int
+        Minimum number of SiPMs in a cluster (see `consider_masked`)
+
+    pitch : float
```
paolafer: I don't see the pitch among the arguments. Also, you could move `msipm` below, in the order it appears in the function. Finally, `all_sipms` is missing.

invisible_cities/reco/xy_algorithms.py
```diff
@@ -58,100 +70,119 @@ def count_masked(cs, d, datasipm, is_masked):
...
+      charge distribution
+    - obtaining the barycenter from the SiPMs within `new_lm_radius`
+      of the new local maximum
+    - creating a cluster from the SiPMs neighbouring the barycenter of
+      the new local maximum, provided it contains at least `msipm`
+    - discarding (non-destructively) the sipms that formed the cluster
+    - repeating the previous steps until there are no more SiPMs with
+      charge higher than `Qlm`
+
+    Parameters
```
paolafer, line 144: Why did you change `arguments` for `Parameters`? Aren't they the arguments of the function?

gonzaponte: Ugh, I always get confused with this, but:
A parameter is a variable in a function definition. An argument is the actual value.

Besides, in general, I tried to replicate the style that I started to implement for this repo (see for example the core submodule), which is pretty much like numpy's.

jacg: Yup: parameters are names, arguments are values.

Because parameters are bound to arguments when a function is called, and therefore represent and give access to arguments within a function's definition, it is *often* OK to use the terms interchangeably. But *sometimes* the difference is crucial. For example when discussing a situation in which the type of the argument does not match the type of the parameter it is essential to have the linguistic means to talk about them as distinct entities.

The difference between parameter and argument is essentially the difference between variable and value: parameters are a specific kind of variable: those defined/declared[^1] in the function's parameter list. Most of the time it's OK to say "`t` is the time" rather than "`t` is the variable which is bound to the time", but if you literally believe that `t` *is* the time, and do not understand that they are different things and you do not have the vocabulary to talk about them as distinct things, then you will run into trouble sooner or later.

[^1]: Another pair of words with related-yet-different meaning.

invisible_cities/reco/xy_algorithms_test.py
gonzaponte: Good catch, it should be the same number of entries in both

invisible_cities/reco/xy_algorithms_test.py
gonzaponte: I guess it depends on the arguments given to both `corona` and `barycenter`

paolafer: Ok, but in the test you are asserting that they are actually the same, no?

gonzaponte: Yes, I'm asserting that, with the appropriate values for both algorithms, we obtain the same result.

paolafer (approved):
All my comments have been addressed. This is a very nice, long-due PR cleaning up the xy reconstruction algorithms. Approved!

gonzaponte:
Cleaned up the history and rebased. Ready to be merged if the CI doesn't complain.

PR #841, opened 2023-10-04
Update NEXT-100 SiPM DB mapping

jmbenlloch:
This PR updates the channel mapping according to the hardware connections made during the assembly at LSC.

The MC positions of the DBs are these:
image

The corresponding mapping of the "hardware ID" (FEB ID) for each DB goes like this:
image

Inside each DB, the corresponding mappings are these (**ChannelID is the same as ElecID. The former is used in IC, while the latter is the name in the decoder**):
image

To compute the ID of a sensor the following formula is used: `(DB_ID + 1) * 1000 + sensor`. For instance, the lower-left sensor of the first DB (lower-left as well) would be:

MC sensor ID: `(0 + 1) * 1000 + 0 = 1000`
Channel ID: `(35 + 1) * 1000 + 13 = 36013`

In the same way, SensorID 1001 corresponds to ChannelID 36012, etc.

This PR includes the new NEXT100 database with those relations for all 3584 SiPMs.


jmbenlloch:
I have updated this PR with a new mapping that includes the changes detected during the first tests at Canfranc. 

The MC map is the same:

image

The new FEB ID map is this:

image

The internal mapping is the same as before.

halmamol (approved):
This PR updates the NEXT-100 database to match the channel mapping with the connections of the dices. These have been reviewed at LSC and with calibration runs.

PR #845, opened 2024-01-09
Update NEXT100 PMT mapping

jmbenlloch:
This PR fixes the PMT mapping of NEXT100 according to the installation data provided by the engineers.


gonzaponte:
Superseded by #849.

PR #848, opened 2024-01-23
Update PMT positions in NEXT100 database

jmbenlloch:
This PR updates NEXT100's PMT positions according to the latest geometry implemented in nexus.


paolafer (approved):
The database has been changed according to the new reference system used in the simulations: approved!

PR #854, opened 2024-02-15
Allow null event range

### LINKED ISSUE #573: `event_range = 0` doesn't work
I was trying to run `diomira` with `event_range = 0` for a test and it crashes.

- Is this behaviour expected?
- Should it be fixed?

gonzaponte:
This PR adds the option to run a city with `event_range = 0`, which can be useful for testing.

Closes #573

### DIFF
--- a/invisible_cities/cities/cities_test.py
+++ b/invisible_cities/cities/cities_test.py
@@ -10,11 +10,11 @@ from pytest import raises
 
 from .. core.configure      import configure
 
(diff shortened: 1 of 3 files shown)


martinperezmaneiro (approved):
Good PR. Adds protection to `slice` function for `event_range = 0`, and creates the corresponding test. Also changes a file structure test to run over no events. Everything works as expected :)

PR #856, opened 2024-02-16
Add fix to be able to create the current environment on MacOS 13.3.1

paolafer:
I had problems creating the `3.8-2022-04-13` environment  and this fix by Corey Adams solves it.

### DIFF
--- a/manage.sh
+++ b/manage.sh
@@ -83,6 +83,9 @@ function make_environment {
 
     cat <<EOF > ${YML_FILENAME}
 name: ${CONDA_ENV_NAME}
+channels:
+- conda-forge
+- defaults
 dependencies:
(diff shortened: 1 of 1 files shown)


gonzaponte:
I am not convinced this won't have (undesired) side effects in other systems. The GHA passed, so I guess it doesn't, but I don't trust it either. I have to dwell on it.

gonzaponte:
After dealing with this myself, I realized that we must also change the architecture depending on the system. Right now it's hardcoded to `x86_64`, but we need `arm64`, especially for the new Macs with an M1 chip. It will be painful to test this properly.

gonzaponte (approved):
I believe this is ok and does not disrupt the installation process in other machines.

PR #858, opened 2024-02-22
Propagate contiguity variable across the code

### LINKED ISSUE #805: `make_tracks` does not use `contiguity` argument
Its value should be propagated to the relevant function calls.

paolafer:
This PR propagates the `contiguity` variable across several paolina functions, where the default value was used.
Closes #805.

### DIFF
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -289,8 +289,10 @@ def make_tracks(evt_number       : float,
     return tc
 
 
-def drop_end_point_voxels(voxels: Sequence[Voxel], energy_threshold: float, min_vxls: int = 3,
-                          contiguity: Contiguity = Contiguity.CORNER) -> Sequence[Voxel]:
+def drop_end_point_voxels(voxels           : Sequence[Voxel],
+                          energy_threshold : float,
+                          min_vxls         : int = 3,
+                          contiguity       : Contiguity = Contiguity.CORNER) -> Sequence[Voxel]:
     """Eliminate voxels at the end-points of a track, recursively,
        if their energy is lower than a threshold. Returns 1 if the voxel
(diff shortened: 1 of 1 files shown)


invisible_cities/reco/paolina_functions.py
```diff
@@ -289,18 +289,19 @@ def make_tracks(evt_number       : float,
     return tc
 
 
-def drop_end_point_voxels(voxels: Sequence[Voxel], energy_threshold: float, min_vxls: int = 3) -> Sequence[Voxel]:
+def drop_end_point_voxels(voxels: Sequence[Voxel], energy_threshold: float, min_vxls: int = 3,
+                          contiguity: Contiguity = Contiguity.CORNER) -> Sequence[Voxel]:
```
gonzaponte: If the list of parameters must be split into several lines, the convention is to put one per line.

invisible_cities/reco/paolina_functions.py
```diff
@@ -289,18 +289,21 @@ def make_tracks(evt_number       : float,
     return tc
 
 
-def drop_end_point_voxels(voxels: Sequence[Voxel], energy_threshold: float, min_vxls: int = 3) -> Sequence[Voxel]:
+def drop_end_point_voxels(voxels: Sequence[Voxel],
+                          energy_threshold: float,
+                          min_vxls: int = 3,
+                          contiguity: Contiguity = Contiguity.CORNER) -> Sequence[Voxel]:
```
gonzaponte: We also tend to align things like this:
```suggestion
def drop_end_point_voxels(voxels           : Sequence[Voxel],
                          energy_threshold : float,
                          min_vxls         : int = 3,
                          contiguity       : Contiguity = Contiguity.CORNER) -> Sequence[Voxel]:
```

gonzaponte (approved):
Fixed a long-standing issue. Excellent!

PR #865, opened 2024-03-07
Add data preparation functions for the Kr map computation

### LINKED ISSUE #862: ICAROS: data preparation
ICAROS data preparation.

Input:
- dst (Kr dst from Dorothea)

Select and validate the Kr input data
- [x] Select 1S1, monitor and check efficiency
- [x] Select 1S2, monitor and check efficiency
- [x] Select S2 vs drift-time band, monitor and check efficiency

Ouput: 
- filtered dst (a deep-copy), and selection masks on the initial dst

Write Out:
- information of the selection/filter

bpalmeiro:
This PR links to Issue #862 and presents a proposal for the set of functions to perform the data selection and checking for the preparation for the Kr map creation. However, there are a couple of things pending up for discussion:

- The band selection relies on the lt functions that we considered adding in the map_creation PR (since we don't think they belong in this PR  workflow), but this is convoluted. It may be worth adding a previous PR with those functions.
- The previous point makes the band selection function only a scheme of the final implementation, being the only function not tested for the present PR.
- The workflow related to implementing this function in the city was left for the general PR —or even a posterior PR—related to the city itself. We considered that this makes this one much clearer since it relates only to adding the functions. Thoughts?

### DIFF
--- a/invisible_cities/core/core_functions.py
+++ b/invisible_cities/core/core_functions.py
@@ -9,7 +9,11 @@ import numpy as np
 from typing import Sequence
 from typing import Tuple
 
+from . exceptions import ValueOutOfRange
+
 from ..types.symbols import NormMode
+from ..types.symbols import Strictness
+
 
 def timefunc(f):
     """
@@ -76,6 +80,43 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
     upper_bound = data <= maxval if right_closed else data < maxval
     return lower_bound & upper_bound
 
+
+def check_if_values_in_interval(data         : np.array,
+                                minval       : float   ,
+                                maxval       : float   ,
+                                display_name : str = '',
+                                strictness   : Strictness = Strictness.stop_proccess,
+                                **kwargs)->bool:
+    """
+    Checks whether input values are all inside the interval (minval, maxval).
+
+    Parameters
+    ----------
+    data : np.array
+        Input array to check.
+    minval: float
+        Lower limit of the interval.
+    maxval: float
+        Upper limit of the interval.
+    display_name: string
+        Name for the message to print if exception raises.
+    strictness: Strictness
+        If 'warning', function returns a False if the criteria
+        is not match. If 'stop_proccess' it raises an exception.
+    Returns
+    ----------
+        True if values are in the interval. False if not and strictness
+        is set to 'warning'. Otherwise, it raises an exception.
+    """
+    if in_range(data, minval, maxval, **kwargs).all():
+        return True;
+
+    elif strictness is Strictness.warning:
+        return False
+    else:
+        raise ValueOutOfRange(f' Variable {display_name} is out of bounds ({minval}, {maxval})')
+
+    return
 
 def weighted_mean_and_var(data       : Sequence,
                           weights    : Sequence,
--- a/invisible_cities/core/core_functions.py
+++ b/invisible_cities/core/core_functions.py
@@ -83,11 +83,11 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
     return lower_bound & upper_bound
 
 
-def all_in_range(data         : np.ndarray                                  ,
-                 minval       : float                                       ,
-                 maxval       : float                                       ,
-                 display_name : Optional[str] = ''                          ,
-                 strictness   :Optional[Strictness] = Strictness.raise_error,
+def all_in_range(data         : np.ndarray                                   ,
+                 minval       : float                                        ,
+                 maxval       : float                                        ,
+                 display_name : Optional[str] = ''                           ,
+                 strictness   : Optional[Strictness] = Strictness.raise_error,
                  **kwargs)->bool:
     """
     Checks whether input values are all inside the interval (minval, maxval).
@@ -103,17 +103,20 @@ def all_in_range(data         : np.ndarray                                  ,
     display_name: string
         Label to be displayed in case of warning or exception.
     strictness: Strictness
-        It `silent`, it returns a False if the criteria
-        is not match.
-        If `warning`, it returns a False and raises a warning.
-        If `raise_error`, it raises an exception.
+        Describes the behaviour when the output is `False`. If `strictness` is:
+        - `Strictness.raise_error`: an exception is raised.
+        - `Strictness.warning`: a warning is raised before returning.
+        - `Strictness.silent`: the function returns quietly.
 
     **kwargs:
         Optional arguments being passed to `in_range`.
     Returns
     -------
-        True if values are in the interval. False if not and strictness
-        is set to `warning` or `silent`. Otherwise, it raises an exception.
+    True if values are in the interval. False otherwise.
+
+    Raises
+    ------
+    ValueOutOfRange: if the output is `False` and `strictness` is `Strictness.raise_error`
     """
 
     values_in_interval = in_range(data, minval, maxval, **kwargs)
--- a/invisible_cities/core/core_functions.py
+++ b/invisible_cities/core/core_functions.py
@@ -9,6 +9,8 @@ import numpy as np
 
 from typing import Sequence
 from typing import Tuple
+from typing import Optional
+
 
 from . exceptions import ValueOutOfRange
 from . configure  import check_annotations
@@ -82,18 +84,18 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
     return lower_bound & upper_bound
 
 
-def all_in_range(data         : np.ndarray                           ,
-                 minval       : float                                ,
-                 maxval       : float                                ,
-                 display_name : str = ''                             ,
-                 strictness   : Strictness = Strictness.stop_proccess,
+def all_in_range(data         : np.ndarray                                    ,
+                 minval       : float                                         ,
+                 maxval       : float                                         ,
+                 display_name : Optional[str] = ''                            ,
+                 strictness   :Optional[Strictness] = Strictness.stop_proccess,
                  **kwargs)->bool:
     """
     Checks whether input values are all inside the interval (minval, maxval).
 
     Parameters
     ----------
-    data : np.array
+    data : np.ndarray
         Input array to check.
     minval: float
         Lower limit of the interval.
--- a/invisible_cities/core/core_functions.py
+++ b/invisible_cities/core/core_functions.py
@@ -84,11 +84,11 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
     return lower_bound & upper_bound
 
 
-def all_in_range(data         : np.ndarray                                    ,
-                 minval       : float                                         ,
-                 maxval       : float                                         ,
-                 display_name : Optional[str] = ''                            ,
-                 strictness   :Optional[Strictness] = Strictness.stop_proccess,
+def all_in_range(data         : np.ndarray                                  ,
+                 minval       : float                                       ,
+                 maxval       : float                                       ,
+                 display_name : Optional[str] = ''                          ,
+                 strictness   :Optional[Strictness] = Strictness.raise_error,
                  **kwargs)->bool:
     """
     Checks whether input values are all inside the interval (minval, maxval).
--- a/invisible_cities/core/core_functions.py
+++ b/invisible_cities/core/core_functions.py
@@ -13,7 +13,6 @@ from typing import Optional
 
 
 from . exceptions import ValueOutOfRange
-from . configure  import check_annotations
 from ..types.symbols import NormMode
 from ..types.symbols import Strictness
 
--- a/invisible_cities/core/core_functions.py
+++ b/invisible_cities/core/core_functions.py
@@ -82,12 +82,12 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
     return lower_bound & upper_bound
 
 
-def all_in_range(data         : np.ndarray,
-                 minval       : float   ,
-                 maxval       : float   ,
-                 display_name : str = '',
+def all_in_range(data         : np.ndarray                           ,
+                 minval       : float                                ,
+                 maxval       : float                                ,
+                 display_name : str = ''                             ,
                  strictness   : Strictness = Strictness.stop_proccess,
-                **kwargs)->bool:
+                 **kwargs)->bool:
     """
     Checks whether input values are all inside the interval (minval, maxval).
 
@@ -120,7 +120,7 @@ def all_in_range(data         : np.ndarray,
         return False
 
     text  = f'Variable {display_name} has {n_outliers} values out of bounds ({minval}, {maxval}\n'
-    text += f'Outsiders: {outliers}'
+    text += f'Outliers: {outliers}'
 
     if strictness is Strictness.warning:
         warnings.warn(text, UserWarning)
--- a/invisible_cities/core/core_functions.py
+++ b/invisible_cities/core/core_functions.py
@@ -111,17 +111,16 @@ def all_in_range(data         : np.ndarray,
     """
 
     values_in_interval = in_range(data, minval, maxval, **kwargs)
-    outsiders_mask     = np.logical_not(values_in_interval)
-    outsiders_list     = data[outsiders_mask]
-    n_outsiders        = np.sum(outsiders_mask)
+    outliers   = data[~values_in_interval]
+    n_outliers = len(outliers)
 
     if values_in_interval.all():
         return True
     elif strictness is Strictness.silent:
         return False
 
-    text  = f'Variable {display_name} has {n_outsiders} values out of bounds ({minval}, {maxval}\n'
-    text += f'Outsiders: {outsiders_list}'
+    text  = f'Variable {display_name} has {n_outliers} values out of bounds ({minval}, {maxval}\n'
+    text += f'Outsiders: {outliers}'
 
     if strictness is Strictness.warning:
         warnings.warn(text, UserWarning)
--- a/invisible_cities/core/core_functions.py
+++ b/invisible_cities/core/core_functions.py
@@ -102,14 +102,19 @@ def all_in_range(data         : np.ndarray                                  ,
     maxval: float
         Upper limit of the interval.
     display_name: string
-        Name for the message to print if exception raises.
+        Label to be displayed in case of warning or exception.
     strictness: Strictness
-        If 'warning', function returns a False if the criteria
-        is not match. If 'stop_proccess' it raises an exception.
+        It `silent`, it returns a False if the criteria
+        is not match.
+        If `warning`, it returns a False and raises a warning.
+        If `raise_error`, it raises an exception.
+
+    **kwargs:
+        Optional arguments being passed to `in_range`.
     Returns
-    ----------
+    -------
         True if values are in the interval. False if not and strictness
-        is set to 'warning'. Otherwise, it raises an exception.
+        is set to `warning` or `silent`. Otherwise, it raises an exception.
     """
 
     values_in_interval = in_range(data, minval, maxval, **kwargs)
--- a/invisible_cities/core/core_functions.py
+++ b/invisible_cities/core/core_functions.py
@@ -82,12 +82,12 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
     return lower_bound & upper_bound
 
 
-def check_if_values_in_interval(data         : np.ndarray,
-                                minval       : float   ,
-                                maxval       : float   ,
-                                display_name : str = '',
-                                strictness   : Strictness = Strictness.stop_proccess,
-                                **kwargs)->bool:
+def all_in_range(data         : np.ndarray,
+                 minval       : float   ,
+                 maxval       : float   ,
+                 display_name : str = '',
+                 strictness   : Strictness = Strictness.stop_proccess,
+                **kwargs)->bool:
     """
     Checks whether input values are all inside the interval (minval, maxval).
 
--- a/invisible_cities/core/core_functions.py
+++ b/invisible_cities/core/core_functions.py
@@ -3,6 +3,7 @@ Core functions
 This module includes utility functions.
 """
 import time
+import warnings
 
 import numpy as np
 
@@ -10,7 +11,7 @@ from typing import Sequence
 from typing import Tuple
 
 from . exceptions import ValueOutOfRange
-
+from . configure  import check_annotations
 from ..types.symbols import NormMode
 from ..types.symbols import Strictness
 
@@ -81,7 +82,7 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
     return lower_bound & upper_bound
 
 
-def check_if_values_in_interval(data         : np.array,
+def check_if_values_in_interval(data         : np.ndarray,
                                 minval       : float   ,
                                 maxval       : float   ,
                                 display_name : str = '',
@@ -108,15 +109,26 @@ def check_if_values_in_interval(data         : np.array,
         True if values are in the interval. False if not and strictness
         is set to 'warning'. Otherwise, it raises an exception.
     """
-    if in_range(data, minval, maxval, **kwargs).all():
-        return True;
 
-    elif strictness is Strictness.warning:
+    values_in_interval = in_range(data, minval, maxval, **kwargs)
+    outsiders_mask     = np.logical_not(values_in_interval)
+    outsiders_list     = data[outsiders_mask]
+    n_outsiders        = np.sum(outsiders_mask)
+
+    if values_in_interval.all():
+        return True
+    elif strictness is Strictness.silent:
+        return False
+
+    text  = f'Variable {display_name} has {n_outsiders} values out of bounds ({minval}, {maxval}\n'
+    text += f'Outsiders: {outsiders_list}'
+
+    if strictness is Strictness.warning:
+        warnings.warn(text, UserWarning)
         return False
     else:
-        raise ValueOutOfRange(f' Variable {display_name} is out of bounds ({minval}, {maxval})')
+        raise ValueOutOfRange(text)
 
-    return
 
 def weighted_mean_and_var(data       : Sequence,
                           weights    : Sequence,
--- a/invisible_cities/core/core_functions_test.py
+++ b/invisible_cities/core/core_functions_test.py
@@ -117,7 +117,8 @@ def test_in_range_right_close_interval(data):
                                   min_value  = 0,
                                   max_value  = 100))
 def test_all_in_range_when_all_fall_inside(data):
-    assert core.all_in_range(data, 0, 100, left_closed=True, right_closed=True)
+    assert core.all_in_range(data, 0, 100, left_closed=True,right_closed=True,
+                             strictness=core.Strictness.raise_error)
 
 
 @given(random_length_float_arrays(min_length = 1,
--- a/invisible_cities/core/core_functions_test.py
+++ b/invisible_cities/core/core_functions_test.py
@@ -133,7 +133,7 @@ def test_all_in_range_when_some_fall_outside_silent(data):
                                   max_value  = 100))
 def test_all_in_range_when_some_fall_outside_warns(data):
     minvalue = np.min(data)
-    with warns(UserWarning):
+    with warns(UserWarning, match="values out of bounds"):
         core.all_in_range(data, minvalue+1, 100, strictness=core.Strictness.warning)
 
 @given(random_length_float_arrays(min_length = 1,
--- a/invisible_cities/core/core_functions_test.py
+++ b/invisible_cities/core/core_functions_test.py
@@ -141,7 +141,7 @@ def test_all_in_range_when_some_fall_outside_warns(data):
 def test_all_in_range_when_some_fall_outside_exception(data):
     minvalue =np.min(data)
     with raises(core.ValueOutOfRange, match="values out of bounds"):
-        core.all_in_range(data, minvalue+1, 100, strictness=core.Strictness.stop_proccess)
+        core.all_in_range(data, minvalue+1, 100, strictness=core.Strictness.raise_error)
 
 
 @mark.parametrize(" first  second       norm_mode        expected".split(),
--- a/invisible_cities/core/core_functions_test.py
+++ b/invisible_cities/core/core_functions_test.py
@@ -6,8 +6,6 @@ import pandas as pd
 import numpy  as np
 import numpy.testing as npt
 
-import warnings
-
 from pytest import approx
 from pytest import mark
 from pytest import raises
--- a/invisible_cities/core/core_functions_test.py
+++ b/invisible_cities/core/core_functions_test.py
@@ -125,14 +125,14 @@ def test_all_in_range_when_all_fall_inside(data):
                                   min_value  = 0,
                                   max_value  = 100))
 def test_all_in_range_when_some_fall_outside_silent(data):
-    minvalue =np.min(data)
+    minvalue = np.min(data)
     assert not core.all_in_range(data, minvalue+1, 100, strictness=core.Strictness.silent)
 
 @given(random_length_float_arrays(min_length = 1,
                                   min_value  = 0,
                                   max_value  = 100))
 def test_all_in_range_when_some_fall_outside_warns(data):
-    minvalue =np.min(data)
+    minvalue = np.min(data)
     with warns(UserWarning):
         core.all_in_range(data, minvalue+1, 100, strictness=core.Strictness.warning)
 
@@ -140,7 +140,7 @@ def test_all_in_range_when_some_fall_outside_warns(data):
                                   min_value  = 0,
                                   max_value  = 100))
 def test_all_in_range_when_some_fall_outside_exception(data):
-    minvalue =np.min(data)
+    minvalue = np.min(data)
     with raises(core.ValueOutOfRange, match="values out of bounds"):
         core.all_in_range(data, minvalue+1, 100, strictness=core.Strictness.raise_error)
 
--- a/invisible_cities/core/core_functions_test.py
+++ b/invisible_cities/core/core_functions_test.py
@@ -6,9 +6,13 @@ import pandas as pd
 import numpy  as np
 import numpy.testing as npt
 
+import warnings
+
 from pytest import approx
 from pytest import mark
 from pytest import raises
+from pytest import warns
+
 
 from flaky                  import flaky
 from hypothesis             import given
@@ -121,10 +125,17 @@ def test_check_if_values_in_interval_when_all_fall_inside(data):
 @given(random_length_float_arrays(min_length = 1,
                                   min_value  = 0,
                                   max_value  = 100))
-def test_check_if_values_in_interval_when_some_fall_outside_warning(data):
+def test_check_if_values_in_interval_when_some_fall_outside_silent(data):
     minvalue =np.min(data)
-    assert not core.check_if_values_in_interval(data, minvalue+1, 100, strictness=core.Strictness.warning)
+    assert not core.check_if_values_in_interval(data, minvalue+1, 100, strictness=core.Strictness.silent)
 
+@given(random_length_float_arrays(min_length = 1,
+                                  min_value  = 0,
+                                  max_value  = 100))
+def test_check_if_values_in_interval_when_some_fall_outside_warns(data):
+    minvalue =np.min(data)
+    with warns(UserWarning):
+        core.check_if_values_in_interval(data, minvalue+1, 100, strictness=core.Strictness.warning)
 
 @given(random_length_float_arrays(min_length = 1,
                                   min_value  = 0,
--- a/invisible_cities/core/core_functions_test.py
+++ b/invisible_cities/core/core_functions_test.py
@@ -131,9 +131,9 @@ def test_check_if_values_in_interval_when_some_fall_outside_warning(data):
                                   max_value  = 100))
 def test_check_if_values_in_interval_when_some_fall_outside_exception(data):
     minvalue =np.min(data)
-    npt.assert_raises(core.ValueOutOfRange,
-                      core.check_if_values_in_interval,
-                      data, minvalue+1, 100, strictness=core.Strictness.stop_proccess)
+    raises(core.ValueOutOfRange,
+           core.check_if_values_in_interval,
+           data, minvalue+1, 100, strictness=core.Strictness.stop_proccess)
 
 
 @mark.parametrize(" first  second       norm_mode        expected".split(),
--- a/invisible_cities/core/core_functions_test.py
+++ b/invisible_cities/core/core_functions_test.py
@@ -140,9 +140,8 @@ def test_all_in_range_when_some_fall_outside_warns(data):
                                   max_value  = 100))
 def test_all_in_range_when_some_fall_outside_exception(data):
     minvalue =np.min(data)
-    raises(core.ValueOutOfRange,
-           core.all_in_range,
-           data, minvalue+1, 100, strictness=core.Strictness.stop_proccess)
+    with raises(core.ValueOutOfRange, match="values out of bounds"):
+        core.all_in_range(data, minvalue+1, 100, strictness=core.Strictness.stop_proccess)
 
 
 @mark.parametrize(" first  second       norm_mode        expected".split(),
--- a/invisible_cities/core/core_functions_test.py
+++ b/invisible_cities/core/core_functions_test.py
@@ -111,6 +111,31 @@ def test_in_range_right_close_interval(data):
     assert all(output)
 
 
+@given(random_length_float_arrays(min_length = 1,
+                                  min_value  = 0,
+                                  max_value  = 100))
+def test_check_if_values_in_interval_when_all_fall_inside(data):
+    assert core.check_if_values_in_interval(data, 0, 100, left_closed=True, right_closed=True)
+
+
+@given(random_length_float_arrays(min_length = 1,
+                                  min_value  = 0,
+                                  max_value  = 100))
+def test_check_if_values_in_interval_when_some_fall_outside_warning(data):
+    minvalue =np.min(data)
+    assert not core.check_if_values_in_interval(data, minvalue+1, 100, strictness=core.Strictness.warning)
+
+
+@given(random_length_float_arrays(min_length = 1,
+                                  min_value  = 0,
+                                  max_value  = 100))
+def test_check_if_values_in_interval_when_some_fall_outside_exception(data):
+    minvalue =np.min(data)
+    npt.assert_raises(core.ValueOutOfRange,
+                      core.check_if_values_in_interval,
+                      data, minvalue+1, 100, strictness=core.Strictness.stop_proccess)
+
+
 @mark.parametrize(" first  second       norm_mode        expected".split(),
                   ((  1  ,    2  , core.NormMode.first ,   -1    ),
                    (  4  ,    2  , core.NormMode.second,    1    ),
--- a/invisible_cities/core/core_functions_test.py
+++ b/invisible_cities/core/core_functions_test.py
@@ -118,32 +118,32 @@ def test_in_range_right_close_interval(data):
 @given(random_length_float_arrays(min_length = 1,
                                   min_value  = 0,
                                   max_value  = 100))
-def test_check_if_values_in_interval_when_all_fall_inside(data):
-    assert core.check_if_values_in_interval(data, 0, 100, left_closed=True, right_closed=True)
+def test_all_in_range_when_all_fall_inside(data):
+    assert core.all_in_range(data, 0, 100, left_closed=True, right_closed=True)
 
 
 @given(random_length_float_arrays(min_length = 1,
                                   min_value  = 0,
                                   max_value  = 100))
-def test_check_if_values_in_interval_when_some_fall_outside_silent(data):
+def test_all_in_range_when_some_fall_outside_silent(data):
     minvalue =np.min(data)
-    assert not core.check_if_values_in_interval(data, minvalue+1, 100, strictness=core.Strictness.silent)
+    assert not core.all_in_range(data, minvalue+1, 100, strictness=core.Strictness.silent)
 
 @given(random_length_float_arrays(min_length = 1,
                                   min_value  = 0,
                                   max_value  = 100))
-def test_check_if_values_in_interval_when_some_fall_outside_warns(data):
+def test_all_in_range_when_some_fall_outside_warns(data):
     minvalue =np.min(data)
     with warns(UserWarning):
-        core.check_if_values_in_interval(data, minvalue+1, 100, strictness=core.Strictness.warning)
+        core.all_in_range(data, minvalue+1, 100, strictness=core.Strictness.warning)
 
 @given(random_length_float_arrays(min_length = 1,
                                   min_value  = 0,
                                   max_value  = 100))
-def test_check_if_values_in_interval_when_some_fall_outside_exception(data):
+def test_all_in_range_when_some_fall_outside_exception(data):
     minvalue =np.min(data)
     raises(core.ValueOutOfRange,
-           core.check_if_values_in_interval,
+           core.all_in_range,
            data, minvalue+1, 100, strictness=core.Strictness.stop_proccess)
 
 
--- a/invisible_cities/reco/corrections.py
+++ b/invisible_cities/reco/corrections.py
@@ -243,9 +243,9 @@ def get_normalization_factor(map_e0    : ASectorMap,
 
     return norm_value
 
-def apply_geo_correction(map_e0         : ASectorMap,
-                         norm_strat     : NormStrategy         = NormStrategy.max,
-                         norm_value     : Optional[float]      = None
+def apply_geo_correction(map_e0     : ASectorMap,
+                         norm_strat : NormStrategy    = NormStrategy.max,
+                         norm_value : Optional[float] = None
                         ) -> Callable:
     """
     For a map for each correction, it returns a function
@@ -265,12 +265,11 @@ def apply_geo_correction(map_e0         : ASectorMap,
     -------
         A function that returns geometric correction factor without passing a map.
     """
-
     normalization   = get_normalization_factor(map_e0, norm_strat, norm_value)
     get_xy_corr_fun = maps_coefficient_getter(map_e0.mapinfo, map_e0.e0)
 
-    def geo_correction_factor(x : np.array,
-                                y : np.array)-> np.array:
+    def geo_correction_factor(x : np.ndarray,
+                              y : np.ndarray)-> np.ndarray:
         geo_factor = correct_geometry_(get_xy_corr_fun(x,y))
         factor     = geo_factor * normalization
         return factor
--- a/invisible_cities/reco/corrections.py
+++ b/invisible_cities/reco/corrections.py
@@ -248,14 +248,14 @@ def apply_geo_correction(map_e0     : ASectorMap,
                          norm_value : Optional[float] = None
                         ) -> Callable:
     """
-    For a map for each correction, it returns a function
-    that provides a geometric only correction factor for a
-    given hit collection when (x,y,z,time) is provided.
+    Given a map and a correction strategy, it returns a
+    function that provides a geometric only correction factor
+    for a given hit collection when (x,y) is provided.
 
     Parameters
     ----------
     map_e0 : AsectorMap
-        Correction map for geometric corrections.
+        Correction map for geometric effects.
     norm_strat : NormStrategy
         Provides the desired normalization to be used.
     norm_value : Float(optional)
--- a/invisible_cities/reco/corrections.py
+++ b/invisible_cities/reco/corrections.py
@@ -243,6 +243,41 @@ def get_normalization_factor(map_e0    : ASectorMap,
 
     return norm_value
 
+def apply_geo_correction(map_e0         : ASectorMap,
+                         norm_strat     : NormStrategy         = NormStrategy.max,
+                         norm_value     : Optional[float]      = None
+                        ) -> Callable:
+    """
+    For a map for each correction, it returns a function
+    that provides a geometric only correction factor for a
+    given hit collection when (x,y,z,time) is provided.
+
+    Parameters
+    ----------
+    map_e0 : AsectorMap
+        Correction map for geometric corrections.
+    norm_strat : NormStrategy
+        Provides the desired normalization to be used.
+    norm_value : Float(optional)
+        If norm_strat is selected to be custom, user must provide the
+        desired scale.
+    Returns
+    -------
+        A function that returns geometric correction factor without passing a map.
+    """
+
+    normalization   = get_normalization_factor(map_e0, norm_strat, norm_value)
+    get_xy_corr_fun = maps_coefficient_getter(map_e0.mapinfo, map_e0.e0)
+
+    def geo_correction_factor(x : np.array,
+                                y : np.array)-> np.array:
+        geo_factor = correct_geometry_(get_xy_corr_fun(x,y))
+        factor     = geo_factor * normalization
+        return factor
+
+    return geo_correction_factor
+
+
 
 def apply_all_correction_single_maps(map_e0         : ASectorMap,
                                      map_lt         : ASectorMap,
--- a/invisible_cities/reco/corrections_test.py
+++ b/invisible_cities/reco/corrections_test.py
@@ -269,34 +269,34 @@ def test_apply_all_correction_single_maps_raises_exception_when_invalid_map(map_
 
 @few_examples
 @given(float_arrays(size      = 1,
-                   min_value = -198,
-                   max_value = +198),
+                    min_value = -198,
+                    max_value = +198),
        float_arrays(size      = 1,
-                   min_value = -198,
-                   max_value = +198))
+                    min_value = -198,
+                    max_value = +198))
 def test_apply_geo_correction_properly(map_filename, x, y):
     """
     Due the map taken as input, the geometric correction
     factor must be 1.
     """
     maps      = read_maps(map_filename)
-    load_corr = apply_geo_correction(map_e0     = maps, norm_strat = NormStrategy.max)
-    corr   = load_corr(x, y)
+    load_corr = apply_geo_correction(map_e0 = maps, norm_strat = NormStrategy.max)
+    corr      = load_corr(x, y)
     assert_allclose (corr, np.ones_like(corr))
 
 @few_examples
 @given(float_arrays(size      = 1,
-                   min_value = -198,
-                   max_value = +198),
+                    min_value = -198,
+                    max_value = +198),
        float_arrays(size      = 1,
-                   min_value = -198,
-                   max_value = +198),
+                    min_value = -198,
+                    max_value = +198),
        float_arrays(size      = 1,
-                   min_value = 0,
-                   max_value = 5e2),
+                    min_value = 0,
+                    max_value = 5e2),
        float_arrays(size      = 1,
-                   min_value = 0,
-                   max_value = 1e5))
+                    min_value = 0,
+                    max_value = 1e5))
 def test_apply_all_correction_single_maps_properly(map_filename, x, y, z, t):
     """
     Due the map taken as input, the geometric correction
--- a/invisible_cities/reco/corrections_test.py
+++ b/invisible_cities/reco/corrections_test.py
@@ -280,7 +280,7 @@ def test_apply_geo_correction_properly(map_filename, x, y):
     factor must be 1.
     """
     maps      = read_maps(map_filename)
-    load_corr = apply_geo_correction(map_e0 = maps, norm_strat = NormStrategy.max)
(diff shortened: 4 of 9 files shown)


jahernando:
Some comments:

1.  *check_if_values_in_interval* seems a decorator of *in_range*, can be done in a more elegant way?
2.  *apply_geo_correction* is not a good name: something more as *get_xy_geo_correction_function*
3.  *apply_geo_correction* is a useful wrapper, but what is a *ASectorMap* object? Do we need it? Do you keep it for backward compatibility?
4.  *selection_nS_mask_and_checking* is not a good name, something as *check_nS_efficiency_is_in_range*, add comment that what are the values, I guess the range interval of the allowed efficiencies on S1 and S2.
5.  *band_selector_and_check*, replace name for *select_and_check_S2signal_vs_dtime_band*
6.  *selection_in_band* this is a generic function, maybe it is better in a more general scope, it is in fact an operation over a 2d profile.


Tests:
- ok *check_if_values_in_interval* 
- ok *apply_geo_correction*
- ok *selection_nS_mask_and_checking*

invisible_cities/core/core_functions.py
```diff
@@ -77,6 +81,43 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
...
+    Returns
+    ----------
+        True if values are in the interval. False if not and strictness
+        is set to 'warning'. Otherwise, it raises an exception.
+    """
+    if in_range(data, minval, maxval, **kwargs).all():
+        return True;
+
+    elif strictness is Strictness.warning:
+        return False
```
gonzaponte: This should be `Strictness.silent` since you don't actually produce a warning.

invisible_cities/core/core_functions.py
```diff
@@ -77,6 +81,43 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
...
+    else:
+        raise ValueOutOfRange(f' Variable {display_name} is out of bounds ({minval}, {maxval})')
```
gonzaponte: Can it be useful to add how many values are out of range and by how much?

invisible_cities/core/core_functions_test.py
```diff
@@ -111,6 +111,31 @@ def test_in_range_right_close_interval(data):
...
+
+
+@given(random_length_float_arrays(min_length = 1,
+                                  min_value  = 0,
+                                  max_value  = 100))
+def test_check_if_values_in_interval_when_some_fall_outside_exception(data):
+    minvalue =np.min(data)
+    npt.assert_raises(core.ValueOutOfRange,
+                      core.check_if_values_in_interval,
+                      data, minvalue+1, 100, strictness=core.Strictness.stop_proccess)
```
gonzaponte: Use `pytest.raises` to follow the same structure as other tests in the repo. No need for hypothesis here.

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,185 @@
...
+    eff_interval
+        Limits of the range where selection efficiency
+        is considered correct.
+    Returns
+    ----------
+        A  mask corresponding to the selection made.
+    """
+    if input_mask is None:
+        input_mask = [True] * len(dst)
+    else: pass;
```
gonzaponte: This line is not needed

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,185 @@
...
+
+    effsel   = dst[sel_krband].event.nunique()/dst[input_mask].event.nunique()
+
+    check_if_values_in_interval(data         = np.array(effsel)  ,
+                                minval       = eff_interval[0]   ,
+                                maxval       = eff_interval[1]   ,
+                                display_name = "Z-band selection",
+                                strictness   = strictness        ,
+                                right_closed = True)
+
```
gonzaponte: `dst[input_mask]` is repeated 3 times. It's worth to define a variable and perform this selection only once. Also maybe worth considering something like
`X, Y, Z = dst.loc[input_mask, list("XYZ")].values.T`
and use these variables directly

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,185 @@
...
+    e_sel = e[in_range(e, *range_e)]
+
+    res_fit      = RANSACRegressor().fit(z_sel, np.log(e_sel).reshape(-1, 1))
+
+    in_mask      = res_fit.inlier_mask_
+    residuals_ln = e_sel[in_mask] - np.exp(res_fit.predict(z_sel[in_mask].reshape(-1, 1))).flatten()
+    resy, resx   = np.histogram(residuals_ln, 100)
+    resx         = resx[:-1] + np.diff(resx)
+    fitres       = fit(gauss, resx, resy, seed=[4e3,0,10])
+    fitsigma     = fitres.val[2]
```
gonzaponte: It might be worth hiding all this logic (sigma estimation) in a separate function to make this function more clear

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+from .. core.testing_utils import assert_dataframes_equal
+from .. core               import core_functions   as core
+
+from .                     import icaro_components as icarcomp
+
+@given(integers(min_value = 1,
+                max_value = 1e4))
+def test_selection_nS_mask_and_checking_right_output(nsignals):
```
gonzaponte: Parametrize on S1/S2.

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,86 @@
...
+    data = pd.DataFrame({'nS1': data, 'event': range(nevt)})
+
+    mask = icarcomp.selection_nS_mask_and_checking(data, icarcomp.type_of_signal.nS1)
+
+    assert np.sum(mask) == nsignals
+
+
+@given(integers(min_value = 1,
+                max_value = 1e4))
+def test_selection_nS_mask_and_checking_self_concatenating(nsignals):
```
gonzaponte:
```suggestion
def test_selection_nS_mask_and_checking_consistency(nsignals):
```

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,86 @@
...
+    dataS2 = np.concatenate([np.zeros(nevt- ns2),
+                             np.ones (ns2)])
+    np.random.shuffle(dataS1)
+    np.random.shuffle(dataS2)
+
+    data   = pd.DataFrame({'nS1': dataS1, 'nS2': dataS2, 'event': range(nevt)})
+    maskS1 = icarcomp.selection_nS_mask_and_checking(data, icarcomp.type_of_signal.nS1)
+    maskS2 = icarcomp.selection_nS_mask_and_checking(data, icarcomp.type_of_signal.nS2, maskS1)
+
+    assert_dataframes_equal(data[maskS2], data[maskS1][data[maskS1].nS2 ==1])
```
gonzaponte: This feels like it's repeating the source code. Maybe there is something simpler to check like the count on maskS2 <= maskS1 and that every False in maskS1 is still False in maskS2

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,86 @@
...
+    data   = pd.DataFrame({'nS1': dataS1, 'event': range(nevt)})
+
+    eff = ns1 / 1e4
+    min_eff = min(lim1, lim2)
+    max_eff = max(lim1, lim2)
+    if (eff < min_eff) or (eff > max_eff):
+        npt.assert_raises(core.ValueOutOfRange,
+                          icarcomp.selection_nS_mask_and_checking,
+                          data,  icarcomp.type_of_signal.nS1, None,
+                          [min_eff, max_eff], icarcomp.Strictness.stop_proccess)
```
gonzaponte: Simplify this test removing hypothesis and setting a known case by hand. Also use pytest.raises.

gonzaponte (commented):
First round. This is being done with peras, more or less.

invisible_cities/core/core_functions.py
bpalmeiro: Silent option added and implemented

invisible_cities/core/core_functions.py
bpalmeiro: Added

invisible_cities/core/core_functions.py
```diff
@@ -77,6 +82,54 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
     return lower_bound & upper_bound
 
 
+def check_if_values_in_interval(data         : np.ndarray,
```
gonzaponte: suggestion to keep it closer to `in_range`: `all_in_range`

invisible_cities/core/core_functions.py
```diff
@@ -77,6 +82,54 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
...
+    Returns
+    ----------
+        True if values are in the interval. False if not and strictness
+        is set to 'warning'. Otherwise, it raises an exception.
+    """
+
+    values_in_interval = in_range(data, minval, maxval, **kwargs)
+    outsiders_mask     = np.logical_not(values_in_interval)
+    outsiders_list     = data[outsiders_mask]
+    n_outsiders        = np.sum(outsiders_mask)
```
gonzaponte:
```suggestion
    outliers   = data[~values_in_interval]
    n_outliers = len(outliers)
```

and change names below.

invisible_cities/core/core_functions_test.py
```diff
@@ -6,9 +6,13 @@
 import numpy  as np
 import numpy.testing as npt
 
+import warnings
+
```
gonzaponte: not needed

invisible_cities/core/core_functions_test.py
```diff
@@ -111,6 +115,38 @@ def test_in_range_right_close_interval(data):
...
+        core.check_if_values_in_interval(data, minvalue+1, 100, strictness=core.Strictness.warning)
+
+@given(random_length_float_arrays(min_length = 1,
+                                  min_value  = 0,
+                                  max_value  = 100))
+def test_check_if_values_in_interval_when_some_fall_outside_exception(data):
+    minvalue =np.min(data)
+    raises(core.ValueOutOfRange,
+           core.check_if_values_in_interval,
+           data, minvalue+1, 100, strictness=core.Strictness.stop_proccess)
```
gonzaponte: Use the context manager:
```suggestion
    with raises(core.ValueOutOfRange, match="a_partial_match_of_the_msg"):
        core.check_if_values_in_interval(data, minvalue+1, 100, strictness=core.Strictness.stop_proccess)
```

invisible_cities/reco/corrections.py
```diff
@@ -243,6 +243,41 @@ def get_normalization_factor(map_e0    : ASectorMap,
 
     return norm_value
 
+def apply_geo_correction(map_e0         : ASectorMap,
+                         norm_strat     : NormStrategy         = NormStrategy.max,
+                         norm_value     : Optional[float]      = None
```
gonzaponte: Cosmetics:
- reduce spaces before colon
- reduce spaces before equal signs

invisible_cities/reco/corrections.py
```diff
@@ -243,6 +243,41 @@ def get_normalization_factor(map_e0    : ASectorMap,
...
+    norm_strat : NormStrategy
+        Provides the desired normalization to be used.
+    norm_value : Float(optional)
+        If norm_strat is selected to be custom, user must provide the
+        desired scale.
+    Returns
+    -------
+        A function that returns geometric correction factor without passing a map.
+    """
+
```
gonzaponte: remove blank line

invisible_cities/reco/corrections.py
```diff
@@ -243,6 +243,41 @@ def get_normalization_factor(map_e0    : ASectorMap,
...
+    normalization   = get_normalization_factor(map_e0, norm_strat, norm_value)
+    get_xy_corr_fun = maps_coefficient_getter(map_e0.mapinfo, map_e0.e0)
+
+    def geo_correction_factor(x : np.array,
+                                y : np.array)-> np.array:
```
gonzaponte: alignement. and use `np.ndarray` as type hint

invisible_cities/reco/corrections_test.py
```diff
@@ -266,6 +267,23 @@ def test_apply_all_correction_single_maps_raises_exception_when_invalid_map(map_
                   map_e, map_e, map_e,
                   apply_temp = True)
 
+@few_examples
+@given(float_arrays(size      = 1,
+                   min_value = -198,
+                   max_value = +198),
+       float_arrays(size      = 1,
+                   min_value = -198,
+                   max_value = +198))
```
gonzaponte: alignment

invisible_cities/reco/corrections_test.py
```diff
@@ -266,6 +267,23 @@ def test_apply_all_correction_single_maps_raises_exception_when_invalid_map(map_
...
+def test_apply_geo_correction_properly(map_filename, x, y):
+    """
+    Due the map taken as input, the geometric correction
+    factor must be 1.
+    """
+    maps      = read_maps(map_filename)
+    load_corr = apply_geo_correction(map_e0     = maps, norm_strat = NormStrategy.max)
+    corr   = load_corr(x, y)
```
gonzaponte: alignment and spaces before =

invisible_cities/reco/corrections_test.py
```diff
@@ -266,6 +267,23 @@ def test_apply_all_correction_single_maps_raises_exception_when_invalid_map(map_
...
+    assert_allclose (corr, np.ones_like(corr))
+
 @few_examples
 @given(float_arrays(size      = 1,
                    min_value = -198,
```
gonzaponte: alignment

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,203 @@
...
+
+    dst_sel = dst[input_mask]
+
+    emaps = apply_geo_correction(boot_map, norm_strat  = norm_strat)
+    E0    = dst_sel.S2e.values * emaps(dst_sel.X.values,
+                                       dst_sel.Y.values)
+
+    sel_krband = np.zeros_like(input_mask)
+    sel_krband[input_mask] = selection_in_band(dst_sel.DT,
+                                               E0,
```
gonzaponte:
```suggestion
    sel_krband[input_mask] = selection_in_band(dst_sel.DT, E0,
```

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,203 @@
...
+                                               range_dt = range_DT,
+                                               range_e = range_E,
+                                               nsigma  = nsigma_sel)
```
gonzaponte: alignment

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,203 @@
...
+                                maxval       = eff_interval[1]   ,
+                                display_name = "DT-band selection",
+                                strictness   = strictness        ,
+                                right_closed = True)
+
+    return sel_krband
+
+
+def selection_in_band(dt        : np.array,
+                      e         : np.array,
```
gonzaponte: type hints np.ndarray

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,203 @@
...
+    nsigma: float
+        Number of sigmas to set the band width
+    Returns
+    ----------
+        A  mask corresponding to the selection made.
+    """
+    # Reshapes and flattens are needed for RANSAC function
+
+    dt_sel = dt[in_range(dt, *range_dt)]
+    e_sel = e[in_range(e, *range_e)]
```
gonzaponte: alignment

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,203 @@
...
+
+    res_fit      = RANSACRegressor().fit(dt_sel, np.log(e_sel).reshape(-1, 1))
+    sigma        = sigma_estimation(dt_sel, e_sel, res_fit)
+
+    prefict_fun  = lambda dt: res_fit.predict(dt.reshape(-1, 1)).flatten()
+    upper_band   = lambda dt: prefict_fun(dt) + nsigma * sigma
+    lower_band   = lambda dt: prefict_fun(dt) - nsigma * sigma
+
```
gonzaponte: remove blank line

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,85 @@
...
+def test_selection_nS_mask_and_checking_right_output(nsignals, signal):
+    nevt = int(1e4)
+    data = np.concatenate([np.zeros(nevt- nsignals),
+                          np.ones  (nsignals)])
+    np.random.shuffle(data)
+
+    data = pd.DataFrame({'nS1': data, 'nS2': data,'event': range(nevt)})
+
+    mask = icarcomp.selection_nS_mask_and_checking(data, signal)
+
```
gonzaponte: try to make it a bit more compact

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,101 @@
...
+                             np.ones (ns2)])
+    np.random.shuffle(dataS1)
+    np.random.shuffle(dataS2)
+
+    data   = pd.DataFrame({'nS1': dataS1, 'nS2': dataS2, 'event': range(nevt)})
+    maskS1 = icarcomp.selection_nS_mask_and_checking(data, icarcomp.type_of_signal.nS1)
+    maskS2 = icarcomp.selection_nS_mask_and_checking(data, icarcomp.type_of_signal.nS2, maskS1)
+
+    assert( np.sum(maskS1) >=  np.sum(maskS2))
+    assert( np.logical_not(maskS2[np.logical_not(maskS1)].all()))
```
gonzaponte: remove extra parentheses

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,101 @@
...
+    np.random.shuffle(dataS1)
+
+    data   = pd.DataFrame({'nS1': dataS1, 'event': range(nevt)})
+
+    eff = ns1 / 1e4
+
+    raises(core.ValueOutOfRange,
+           icarcomp.selection_nS_mask_and_checking,
+           data,  icarcomp.type_of_signal.nS1, None,
+           [min_eff, max_eff], icarcomp.Strictness.stop_proccess)
```
gonzaponte: context manager

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,101 @@
...
+
+@given(floats(min_value = 0.01,
+              max_value = 20))
```
gonzaponte: Can't use hypothesis with random numbers

invisible_cities/core/core_functions.py
```diff
@@ -77,6 +82,53 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
     return lower_bound & upper_bound
 
 
+def all_in_range(data         : np.ndarray,
+                 minval       : float   ,
+                 maxval       : float   ,
+                 display_name : str = '',
+                 strictness   : Strictness = Strictness.stop_proccess,
```
gonzaponte: `stop_process` is too specific to the context this function was developed in. I suggest `raise_` or `error`

invisible_cities/core/core_functions.py
```diff
@@ -77,6 +82,53 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
...
+        Input array to check.
+    minval: float
+        Lower limit of the interval.
+    maxval: float
+        Upper limit of the interval.
+    display_name: string
+        Name for the message to print if exception raises.
+    strictness: Strictness
+        If 'warning', function returns a False if the criteria
+        is not match. If 'stop_proccess' it raises an exception.
```
gonzaponte: Describe the behavior for the three values. Use backticks (`) to refer to specific values.

Document kwargs.

invisible_cities/core/core_functions.py
```diff
@@ -77,6 +82,53 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
...
+    Parameters
+    ----------
+    data : np.array
+        Input array to check.
+    minval: float
+        Lower limit of the interval.
+    maxval: float
+        Upper limit of the interval.
+    display_name: string
+        Name for the message to print if exception raises.
```
gonzaponte:
```suggestion
        Label to be displayed in case of warning or exception.
```

invisible_cities/core/core_functions.py
```diff
@@ -77,6 +82,53 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
...
+    strictness: Strictness
+        If 'warning', function returns a False if the criteria
+        is not match. If 'stop_proccess' it raises an exception.
+    Returns
+    ----------
+        True if values are in the interval. False if not and strictness
+        is set to 'warning'. Otherwise, it raises an exception.
+    """
+
```
gonzaponte, line 121: Remove blank line.

invisible_cities/core/core_functions.py
```diff
@@ -77,6 +82,53 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
...
+        Upper limit of the interval.
+    display_name: string
+        Name for the message to print if exception raises.
+    strictness: Strictness
+        If 'warning', function returns a False if the criteria
+        is not match. If 'stop_proccess' it raises an exception.
+    Returns
+    ----------
+        True if values are in the interval. False if not and strictness
+        is set to 'warning'. Otherwise, it raises an exception.
```
gonzaponte:
```suggestion
        is set to `warning` or `silent`. Otherwise, it raises an exception.
```

invisible_cities/core/core_functions.py
```diff
@@ -77,6 +82,53 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
...
+        Lower limit of the interval.
+    maxval: float
+        Upper limit of the interval.
+    display_name: string
+        Name for the message to print if exception raises.
+    strictness: Strictness
+        If 'warning', function returns a False if the criteria
+        is not match. If 'stop_proccess' it raises an exception.
+    Returns
+    ----------
```
gonzaponte:
```suggestion
    Returns
    -------
```

(matching in length)

invisible_cities/core/core_functions_test.py
```diff
@@ -111,6 +113,37 @@ def test_in_range_right_close_interval(data):
     assert all(output)
 
 
+@given(random_length_float_arrays(min_length = 1,
+                                  min_value  = 0,
+                                  max_value  = 100))
+def test_all_in_range_when_all_fall_inside(data):
+    assert core.all_in_range(data, 0, 100, left_closed=True, right_closed=True)
```
gonzaponte: set strictness explicitly here

invisible_cities/core/core_functions_test.py
```diff
@@ -111,6 +113,37 @@ def test_in_range_right_close_interval(data):
...
+
+
+@given(random_length_float_arrays(min_length = 1,
+                                  min_value  = 0,
+                                  max_value  = 100))
+def test_all_in_range_when_some_fall_outside_silent(data):
+    minvalue =np.min(data)
```
gonzaponte:
```suggestion
    minvalue = np.min(data)
```
here and in the next couple of tests.

invisible_cities/core/core_functions_test.py
```diff
@@ -111,6 +113,37 @@ def test_in_range_right_close_interval(data):
...
+    assert not core.all_in_range(data, minvalue+1, 100, strictness=core.Strictness.silent)
+
+@given(random_length_float_arrays(min_length = 1,
+                                  min_value  = 0,
+                                  max_value  = 100))
+def test_all_in_range_when_some_fall_outside_warns(data):
+    minvalue =np.min(data)
+    with warns(UserWarning):
```
gonzaponte: use `match = ` within `warns` as well

invisible_cities/core/core_functions.py
```diff
@@ -3,13 +3,18 @@
...
 import time
+import warnings
 
 import numpy as np
 
 from typing import Sequence
 from typing import Tuple
 
+from . exceptions import ValueOutOfRange
+from . configure  import check_annotations
```
gonzaponte: not used

invisible_cities/reco/corrections.py
```diff
@@ -243,6 +243,40 @@ def get_normalization_factor(map_e0    : ASectorMap,
 
     return norm_value
 
+def apply_geo_correction(map_e0     : ASectorMap,
+                         norm_strat : NormStrategy    = NormStrategy.max,
+                         norm_value : Optional[float] = None
+                        ) -> Callable:
+    """
+    For a map for each correction, it returns a function
```
gonzaponte: Did you mean 
```suggestion
    Given a map and a correction strategy, it returns a function
```
?

invisible_cities/reco/corrections.py
```diff
@@ -243,6 +243,40 @@ def get_normalization_factor(map_e0    : ASectorMap,
...
+    that provides a geometric only correction factor for a
+    given hit collection when (x,y,z,time) is provided.
```
gonzaponte: The inner function only uses x,y

invisible_cities/reco/corrections.py
```diff
@@ -243,6 +243,40 @@ def get_normalization_factor(map_e0    : ASectorMap,
...
+
+    Parameters
+    ----------
+    map_e0 : AsectorMap
+        Correction map for geometric corrections.
```
gonzaponte:
```suggestion
        Map for geometric corrections.
```
or
```suggestion
        Correction map for geometric effects.
```

invisible_cities/reco/corrections_test.py
```diff
@@ -268,17 +269,34 @@ def test_apply_all_correction_single_maps_raises_exception_when_invalid_map(map_
...
+                    max_value = +198),
        float_arrays(size      = 1,
-                   min_value = -198,
-                   max_value = +198),
+                    min_value = -198,
+                    max_value = +198))
+def test_apply_geo_correction_properly(map_filename, x, y):
+    """
+    Due the map taken as input, the geometric correction
+    factor must be 1.
```
gonzaponte:
```suggestion
    The input map is homogeneous, therefore the geometric correction
    factor must be 1.
```

invisible_cities/reco/corrections_test.py
```diff
@@ -268,17 +269,34 @@ def test_apply_all_correction_single_maps_raises_exception_when_invalid_map(map_
...
+    """
+    maps      = read_maps(map_filename)
+    load_corr = apply_geo_correction(map_e0 = maps, norm_strat = NormStrategy.max)
```
gonzaponte:
```suggestion
    load_corr = apply_geo_correction(map_e0=maps, norm_strat=NormStrategy.max)
```
When keyword arguments are given in the same line, it's usually preferred to remove the whitespaces around the `=`

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,201 @@
...
+from .. types.symbols       import type_of_signal
+from .. types.symbols       import Strictness
+from .. types.symbols       import NormStrategy
+from .. core.core_functions import all_in_range
+from .. core.core_functions import in_range
+from .. core.core_functions import shift_to_bin_centers
+from .. core.fit_functions  import fit
+from .. core.fit_functions  import gauss
+
+def selection_nS_mask_and_checking(dst        : pd.DataFrame                         ,
```
gonzaponte:
```suggestion
def select_nS_mask_and_check(dst        : pd.DataFrame                         ,
```

function names in imperative

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,201 @@
...
+                 minval       = interval[0]  ,
+                 maxval       = interval[1]  ,
+                 display_name = column.value ,
+                 strictness   = strictness   ,
+                 right_closed = True)
+
+    return mask
+
+
+def band_selector_and_check(dst         : pd.DataFrame                                 ,
```
gonzaponte:
```suggestion
def select_band_and_check(dst         : pd.DataFrame                                 ,
```
imperative

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,201 @@
...
+    This function returns a selection of the events that
+    are inside the Kr E vz Z band, and checks
+    if the selection efficiency is correct.
+
+    Parameters
+    ----------
+    dst : pd.DataFrame
+        Krypton dataframe.
+    boot_map: str
+        Name of bootstrap map file.
```
gonzaponte, line 88: When documentation is written, we should link to it here to explain what the bootstrap map is.

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,201 @@
...
+    norm_strt: norm_strategy
+        Provides the desired normalization to be used.
```
gonzaponte:
```suggestion
    norm_strat: norm_strategy
        Provides the desired normalization to be used.
```

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,201 @@
...
+        Range in Z-axis
+    range_E: Tuple[np.array, np.array]
+        Range in Energy-axis
+    nsigma_sel: float
+        Number of sigmas to set the band width
+    eff_interval
+        Limits of the range where selection efficiency
+        is considered correct.
+    Returns
+    ----------
```
gonzaponte:
```suggestion
    Returns
    -------
```

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,201 @@
...
+        A  mask corresponding to the selection made.
+    """
+    if input_mask is None:
+        input_mask = [True] * len(dst)
+
+    dst_sel = dst[input_mask]
+
+    emaps = apply_geo_correction(boot_map, norm_strat  = norm_strat)
```
gonzaponte:
```suggestion
    emaps = apply_geo_correction(boot_map, norm_strat=norm_strat)
```

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,201 @@
...
+    E0    = dst_sel.S2e.values * emaps(dst_sel.X.values,
+                                       dst_sel.Y.values)
```
gonzaponte:
```suggestion
    E0    = dst_sel.S2e.values * emaps(dst_sel.X.values, dst_sel.Y.values)
```

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,201 @@
...
+
+    sel_krband = np.zeros_like(input_mask)
+    sel_krband[input_mask] = selection_in_band(dst_sel.DT, E0     ,
+                                               range_dt = range_DT,
+                                               range_e  = range_E ,
+                                               nsigma   = nsigma_sel)
+
+    effsel   = dst[sel_krband].event.nunique()/dst[input_mask].event.nunique()
```
gonzaponte:
```suggestion
    effsel = dst[sel_krband].event.nunique()/dst[input_mask].event.nunique()
```

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,201 @@
...
+from .. core.core_functions import all_in_range
+from .. core.core_functions import in_range
+from .. core.core_functions import shift_to_bin_centers
+from .. core.fit_functions  import fit
+from .. core.fit_functions  import gauss
+
+def selection_nS_mask_and_checking(dst        : pd.DataFrame                         ,
+                                   column     : type_of_signal                       ,
+                                   input_mask : Optional[np.array]  = None           ,
+                                   interval   : Tuple[float, float] = [0,1]          ,
```
gonzaponte: The function below uses `eff_interval` which I think is a clearer name

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,201 @@
...
+                 minval       = eff_interval[0]    ,
+                 maxval       = eff_interval[1]    ,
+                 display_name = "DT-band selection",
+                 strictness   = strictness         ,
+                 right_closed = True)
+
+    return sel_krband
+
+
+def selection_in_band(dt        : np.ndarray         ,
```
gonzaponte: `select_band`, `find_values_within_band` or similar

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,201 @@
...
+    Parameters
+    ----------
+    dt: np.array
+        axial (dt/z) values
+    e: np.array
+        energy values
+    range_dt: Tuple[np.array, np.array]
+        Range in DT-axis
+    range_e: Tuple[np.array, np.array]
+        Range in Energy-axis
```
gonzaponte: wrong inner types

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,201 @@
...
+    nsigma: float
+        Number of sigmas to set the band width
+    Returns
+    ----------
+        A  mask corresponding to the selection made.
```
gonzaponte, line 166:
```suggestion
    Returns
    -------
        A mask corresponding to the selection made.
```

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,201 @@
...
+    sigma        = sigma_estimation(dt_sel, np.log(e_sel), res_fit)
+
+    prefict_fun  = lambda dt: res_fit.predict(dt.reshape(-1, 1)).flatten()
+    upper_band   = lambda dt: prefict_fun(dt) + nsigma * sigma
+    lower_band   = lambda dt: prefict_fun(dt) - nsigma * sigma
+    sel_inband   = in_range(np.log(e), lower_band(dt), upper_band(dt))
+
+    return  sel_inband
+
+def sigma_estimation(dt     : np.ndarray     ,
```
gonzaponte: `estimate_sigma`

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,201 @@
...
+    ----------
+    dt: np.array
+        axial (dt/z) values
+    e: np.array
+        energy values
+    res_fit: RANSACRegressor
+        RANSAC object fitted to the data
+
+    Returns
+    ----------
```
gonzaponte:
```suggestion
    Returns
    -------
```

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,201 @@
...
+        The sigma of the residuals as a float.
+    """
+    # Reshapes and flattens are needed for RANSAC function
+
+    in_mask      = res_fit.inlier_mask_
+    e_predict    = res_fit.predict(dt[in_mask].reshape(-1, 1)).flatten()
+    residuals_ln = e[in_mask] - e_predict
+    resy, resx   = np.histogram(residuals_ln, 100)
+    resx         = shift_to_bin_centers(resx)
+    fitres       = fit(gauss, resx, resy, seed=[4e3,0,10])
```
gonzaponte: is this seed robust in general?

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,98 @@
...
+
+from   sklearn.linear_model import RANSACRegressor
+
+from .. core.testing_utils  import assert_dataframes_equal
+from .. core                import core_functions   as core
+
+from .                      import icaro_components as icarcomp
+
+
+@mark.parametrize("signal", (icarcomp.type_of_signal.nS1, icarcomp.type_of_signal.nS2))
```
gonzaponte:
```suggestion
@mark.parametrize("signal", icarcomp.type_of_signal)
```
this is enough, I think

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,98 @@
...
+@given(nsignals= integers(min_value = 1,
+                          max_value = 1e4))
+def test_selection_nS_mask_and_checking_right_output(nsignals, signal):
+    nevt = int(1e4)
+    data = np.concatenate([np.zeros(nevt- nsignals), np.ones(nsignals)])
+    np.random.shuffle(data)
```
gonzaponte: we cannot use hypothesis with random. Parametrize it with nsignals using 3 or 4 values. You can take the extremes (0, 10**4) and a couple of values in between. Also, you should test that this function filters not only 0 S1/S2 but any other value that's not 1.

If you want to use hypothesis you would have to do something along the lines of:
`@given(nsignals = arrays(int, array_shapes(max_dims=1), elements=integers(0, some_max_value)))`

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,98 @@
...
+    data = pd.DataFrame({'nS1': data, 'nS2': data, 'event': range(nevt)})
+
```
gonzaponte: remove blank line

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,98 @@
...
+    mask = icarcomp.selection_nS_mask_and_checking(data, signal)
+
+    assert np.sum(mask) == nsignals
```
gonzaponte: `np.count_nonzero` is more appropriate than `np.sum`, even though they provide the same result

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+@given(integers(min_value = 1,
+                max_value = 1e4))
```
gonzaponte:
```suggestion
@given(integers(1, 10**4))
```
I think the arguments are self-explanatory in this case and we can reduce the noise a bit.
In any case, same comment as above, can't use hypothesis and random. I suggest parametrizing it with a few cases.

Side note:
`1e4` is a `float`, `10**4` is an `int`

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,98 @@
...
+       integers(min_value = 1,
+                max_value = 1e4))
+def test_selection_nS_mask_and_checking_concatenating(ns1, ns2):
+    nevt = int(1e4)
+    dataS1 = np.concatenate([np.zeros(nevt- ns1),
+                             np.ones (ns1)])
+    dataS2 = np.concatenate([np.zeros(nevt- ns2),
+                             np.ones (ns2)])
+    np.random.shuffle(dataS1)
+    np.random.shuffle(dataS2)
```
gonzaponte: hypothesis & random

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,98 @@
...
+
+    data   = pd.DataFrame({'nS1': dataS1, 'nS2': dataS2, 'event': range(nevt)})
+    maskS1 = icarcomp.selection_nS_mask_and_checking(data, icarcomp.type_of_signal.nS1)
+    maskS2 = icarcomp.selection_nS_mask_and_checking(data, icarcomp.type_of_signal.nS2, maskS1)
+
+    assert np.sum(maskS1) >=  np.sum(maskS2)
+    assert np.logical_not(maskS2[np.logical_not(maskS1)].all())
```
gonzaponte:
```suggestion
    assert np.sum(maskS1) >= np.sum(maskS2)
    assert not maskS2[~maskS1].all()
```

Simplify boolean logic.
The second assertion is ok because you don't allow ns1/ns2 to be zero, but
- you need to test ns1/ns2 = 0
- you will need to adapt that line when you test for those cases

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,98 @@
...
+    ns1  = int(1e3)
+    min_eff = 0.5
+    max_eff = 1
+    dataS1 = np.concatenate([np.zeros(nevt- ns1),
+                             np.ones (ns1)])
+    np.random.shuffle(dataS1)
+
+    data   = pd.DataFrame({'nS1': dataS1, 'event': range(nevt)})
+
+    eff = ns1 / 1e4
```
gonzaponte:
```suggestion
    eff = ns1 / nevt
```

try to reduce a bit the number of blank lines

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,98 @@
...
+
+    with raises(core.ValueOutOfRange, match="values out of bounds"):
+           icarcomp.selection_nS_mask_and_checking(data,  icarcomp.type_of_signal.nS1,
+                                                   None,[min_eff, max_eff],
```
gonzaponte:
```suggestion
                                                   None, [min_eff, max_eff],
```

gonzaponte (commented):
A more in-depth review now. Mostly cosmetics though.

invisible_cities/reco/icaro_components_test.py
bpalmeiro: I don't necessarily agree with this one. Although it's true random are used, it is still deterministic, it's only for ordering. Am I missing something?

gonzaponte: As discussed with peras, while in this particular case it doesn't matter because the function doesn't care about the order, it is not a good practice. Also, hypothesis can be heavy on runtime, so it's better reserved for cases where it is more useful.

invisible_cities/reco/icaro_components.py
```diff
@@ -194,7 +194,12 @@ def estimate_sigma(dt     : np.ndarray     ,
     residuals_ln = e[in_mask] - e_predict
     resy, resx   = np.histogram(residuals_ln, 100)
     resx         = shift_to_bin_centers(resx)
-    fitres       = fit(gauss, resx, resy, seed=[4e3,0,10])
+    seed         = [np.max(resy) * (2 * np.pi) **.5 * np.std(residuals_ln),
+                    0., np.std(residuals_ln)]
+    # Ideally, a np.mean for the mean should be better that a plain 0 but the
+    # fit does not cope well with that input value, so 0 (fair guess anyway) is
+    # used instead
+    fitres       = fit(gauss, resx+3, resy, seed=seed)
```
gonzaponte: Ýou forgot to remove the +3

invisible_cities/reco/icaro_components_test.py
```diff
@@ -41,10 +37,8 @@ def test_select_nS_mask_and_check_consistency(nsignals):
     npt.assert_equal(mask, mask_re)
 
 
-@given(integers(min_value = 1,
-                max_value = 10*1000),
-       integers(min_value = 1,
-                max_value = 10*1000))
+@mark.parametrize("ns1", [   1, 100, 9999, 10*1000])
```
gonzaponte, line 41: why not nS1=0?

invisible_cities/core/core_functions.py
```diff
@@ -77,6 +83,58 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
     return lower_bound & upper_bound
 
 
+def all_in_range(data         : np.ndarray                                  ,
+                 minval       : float                                       ,
+                 maxval       : float                                       ,
+                 display_name : Optional[str] = ''                          ,
+                 strictness   :Optional[Strictness] = Strictness.raise_error,
```
gonzaponte:
```suggestion
                 strictness   : Optional[Strictness] = Strictness.raise_error,
```

invisible_cities/core/core_functions.py
```diff
@@ -77,6 +83,58 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
...
+        Lower limit of the interval.
+    maxval: float
+        Upper limit of the interval.
+    display_name: string
+        Label to be displayed in case of warning or exception.
+    strictness: Strictness
+        It `silent`, it returns a False if the criteria
+        is not match.
+        If `warning`, it returns a False and raises a warning.
+        If `raise_error`, it raises an exception.
```
gonzaponte:
```suggestion
        Describes the behaviour when the output is `False`. If `strictness` is:
        - `Strictness.raise_error`: an exception is raised.
        - `Strictness.warning`: a warning is raised before returning.
        - `Strictness.silent`: the function returns quietly.
```

invisible_cities/core/core_functions.py
```diff
@@ -77,6 +83,58 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
...
+
+    **kwargs:
+        Optional arguments being passed to `in_range`.
+    Returns
+    -------
+        True if values are in the interval. False if not and strictness
+        is set to `warning` or `silent`. Otherwise, it raises an exception.
```
gonzaponte:
```suggestion
    Returns
    -------
        True if values are in the interval. False otherwise.
    
    Raises
    ------
    ValueOutOfRange: if the output is `False` and `strictness` is `Strictness.raise_error`
```

invisible_cities/reco/corrections_test.py
```diff
@@ -268,21 +269,38 @@ def test_apply_all_correction_single_maps_raises_exception_when_invalid_map(map_
...
        float_arrays(size      = 1,
-                   min_value = 0,
-                   max_value = 1e5))
+                    min_value = 0,
+                    max_value = 1e5))
 def test_apply_all_correction_single_maps_properly(map_filename, x, y, z, t):
     """
-    Due the map taken as input, the geometric correction
-    factor must be 1, the temporal correction 1 and the
+The input map is homogeneous, therefore the geometric
```
gonzaponte: Indentation

invisible_cities/reco/corrections_test.py
```diff
@@ -268,21 +269,38 @@ def test_apply_all_correction_single_maps_raises_exception_when_invalid_map(map_
...
+    """
+    maps      = read_maps(map_filename)
+    load_corr = apply_geo_correction(map_e0=maps, norm_strat=NormStrategy.max)
+    corr      = load_corr(x, y)
+    assert_allclose (corr, np.ones_like(corr))
+
+@few_examples
+@given(float_arrays(size      = 1,
+                    min_value = -198,
+                    max_value = +198),
```
gonzaponte:
```suggestion
@given(float_arrays(size=1, min_value=-198, max_value=+198),
```
I think this is short enough to be done in one line, no? If you agree, do it for all 4 lines.

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,205 @@
...
+from .. core.core_functions import in_range
+from .. core.core_functions import shift_to_bin_centers
+from .. core.fit_functions  import fit
+from .. core.fit_functions  import gauss
+
+def select_nS_mask_and_check(dst          : pd.DataFrame                       ,
+                             column       : type_of_signal                     ,
+                             input_mask   : Optional[np.ndarray]  = None       ,
+                             eff_interval : Tuple[float, float] = [0,1]        ,
+                             strictness   : Strictness = Strictness.raise_error
```
gonzaponte: These should be marked as optional. In general I would like to reduce the number of optional arguments, so it would be even better to leave only strictness as optional (and remove their default values)

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,205 @@
...
+def select_band_and_check(dst         : pd.DataFrame                                     ,
+                          boot_map    : ASectorMap                                       ,
+                          norm_strat  : NormStrategy                   = NormStrategy.max,
+                          input_mask  : np.ndarray                     = None            ,
+                          range_DT    : Tuple[np.ndarray, np.ndarray]  = (10, 1300)      ,
+                          range_E     : Tuple[np.ndarray, np.ndarray]  = (10.0e+3,14e+3) ,
+                          nsigma_sel  : float                          = 3.5             ,
+                          eff_interval: Tuple[float, float]            = [0,1]           ,
+                          strictness  : Strictness = Strictness.raise_error
+                          )->np.array:
```
gonzaponte: Same comment about optionality. In this case, perhaps `nsigma_sel` can also be optional

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,205 @@
...
+                             column       : type_of_signal                     ,
+                             input_mask   : Optional[np.ndarray]  = None       ,
+                             eff_interval : Tuple[float, float] = [0,1]        ,
+                             strictness   : Strictness = Strictness.raise_error
+                             )->np.ndarray:
+    """
+    Selects nS1(or nS2) == 1 for a given kr dst and
+    returns the mask. It also computes selection efficiency,
+    checking if the value is within a given interval, and
+    saves histogram parameters.
```
gonzaponte: and  saves histogram parameters.?

invisible_cities/reco/icaro_components.py
gonzaponte: This optionality should also be described in the docstring, stating the default value.

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,205 @@
...
+                 right_closed = True)
+
+    return sel_krband
+
+
+def select_band(dt        : np.ndarray         ,
+                e         : np.ndarray         ,
+                range_dt  : Tuple[float, float],
+                range_e   : Tuple[float, float],
+                nsigma    : float   = 3.5) ->np.ndarray:
```
gonzaponte: optional

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,205 @@
...
+    res_fit      = RANSACRegressor().fit(dt_sel.reshape(-1,1),
+                                         np.log(e_sel).reshape(-1, 1))
+    sigma        = estimate_sigma(dt_sel, np.log(e_sel), res_fit)
+
+    prefict_fun  = lambda dt: res_fit.predict(dt.reshape(-1, 1)).flatten()
+    upper_band   = lambda dt: prefict_fun(dt) + nsigma * sigma
+    lower_band   = lambda dt: prefict_fun(dt) - nsigma * sigma
+    sel_inband   = in_range(np.log(e), lower_band(dt), upper_band(dt))
+
+    return  sel_inband
```
gonzaponte:
```suggestion
    return sel_inband
```

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,85 @@
...
+    nevt      = 10*1000
+    xrange    = [0, 1000]
+    slope     = 100
+    n0        = 100
+    x         = np.random.uniform(*xrange, nevt)
+    y         = slope * x + n0
+    y         = np.random.normal(y, sigma)
+    res_fit   = RANSACRegressor().fit(x.reshape(-1,1), y.reshape(-1, 1))
+    sigma_est = icarcomp.estimate_sigma(x, y, res_fit)
+    np.testing.assert_allclose(sigma, sigma_est, rtol=0.1)
```
gonzaponte:
```suggestion
    npt.assert_allclose(sigma_est, sigma, rtol=0.1)
```

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,85 @@
...
+
+@mark.parametrize("nsignals", [1, 100, 9999, 10*1000])
+def test_select_nS_mask_and_check_consistency(nsignals):
+    nevt = 10*1000
+    data = np.concatenate([np.zeros(nevt - nsignals), np.ones(nsignals)])
+    np.random.shuffle(data)
+    data    = pd.DataFrame({'nS1': data, 'event': range(nevt)})
+    mask    = icarcomp.select_nS_mask_and_check(data, icarcomp.type_of_signal.nS1)
+    mask_re = icarcomp.select_nS_mask_and_check(data, icarcomp.type_of_signal.nS1, input_mask=mask)
+    npt.assert_equal(mask, mask_re)
```
gonzaponte:
```suggestion
    npt.assert_equal(mask_re, mask)
```
Although it's commutative, we tend to (with exceptions) write the target as the first argument and the expected value as the second. This applies also to `==`  and other comparisons.

invisible_cities/reco/icaro_components_test.py
```diff
@@ -0,0 +1,85 @@
...
+    dataS1  = np.concatenate([np.zeros(nevt- ns1),
+                              np.ones (ns1)])
+    np.random.shuffle(dataS1)
+    data   = pd.DataFrame({'nS1': dataS1, 'event': range(nevt)})
+    eff    = ns1 / nevt
+
+    with raises(core.ValueOutOfRange, match="values out of bounds"):
+                icarcomp.select_nS_mask_and_check(data,  icarcomp.type_of_signal.nS1,
+                                                  None,[min_eff, max_eff],
+                                                  icarcomp.Strictness.raise_error)
```
gonzaponte:
```suggestion
    with raises(core.ValueOutOfRange, match="values out of bounds"):
        icarcomp.select_nS_mask_and_check(data,  icarcomp.type_of_signal.nS1,
                                          None,[min_eff, max_eff],
                                          icarcomp.Strictness.raise_error)
```

gonzaponte (commented):
Mostly cosmetics

invisible_cities/reco/icaro_components_test.py
bpalmeiro: It is not expected to have no input events. Thus, in the concatenated use, if the first selection removes all events, the efficiency computation has a zero division. Should we contemplate this case in the function?

gonzaponte: It makes sense for this test. Leave a comment clarifying it, just in case.

invisible_cities/core/core_functions.py
gonzaponte: Still there

invisible_cities/core/core_functions.py
```diff
@@ -77,6 +83,61 @@ def in_range(data, minval=-np.inf, maxval=np.inf, left_closed=True, right_closed
...
+        Label to be displayed in case of warning or exception.
+    strictness: Strictness
+        Describes the behaviour when the output is `False`. If `strictness` is:
+        - `Strictness.raise_error`: an exception is raised.
+        - `Strictness.warning`: a warning is raised before returning.
+        - `Strictness.silent`: the function returns quietly.
+
+    **kwargs:
+        Optional arguments being passed to `in_range`.
+    Returns
```
gonzaponte, line 113:
```suggestion
    
    Returns
```
add blank line before Returns

invisible_cities/reco/corrections.py
```diff
@@ -243,6 +243,40 @@ def get_normalization_factor(map_e0    : ASectorMap,
...
+    Parameters
+    ----------
+    map_e0 : AsectorMap
+        Correction map for geometric effects.
+    norm_strat : NormStrategy
+        Provides the desired normalization to be used.
+    norm_value : Float(optional)
+        If norm_strat is selected to be custom, user must provide the
+        desired scale.
+    Returns
```
gonzaponte, line 264: Add blank line before Returns

invisible_cities/reco/corrections.py
```diff
@@ -243,6 +243,40 @@ def get_normalization_factor(map_e0    : ASectorMap,
...
+
+    Parameters
+    ----------
+    map_e0 : AsectorMap
+        Correction map for geometric effects.
+    norm_strat : NormStrategy
+        Provides the desired normalization to be used.
+    norm_value : Float(optional)
+        If norm_strat is selected to be custom, user must provide the
+        desired scale.
```
gonzaponte, line 263:
```suggestion
    norm_strat : NormStrategy, optional
        Provides the desired normalization to be used. Default: NormStrategy.max.
    norm_value : float, optional
        If norm_strat is selected to be custom, user must provide the
        desired scale.
```

invisible_cities/reco/corrections_test.py
```diff
@@ -268,21 +269,30 @@ def test_apply_all_correction_single_maps_raises_exception_when_invalid_map(map_
...
+    maps      = read_maps(map_filename)
+    load_corr = apply_geo_correction(map_e0=maps, norm_strat=NormStrategy.max)
+    corr      = load_corr(x, y)
+    assert_allclose (corr, np.ones_like(corr))
+
+@few_examples
+@given(float_arrays(size      = 1, min_value = -198, max_value =      198),
+       float_arrays(size      = 1, min_value = -198, max_value =      198),
+       float_arrays(size      = 1, min_value =    0, max_value =      500),
+       float_arrays(size      = 1, min_value =    0, max_value = 100*1000))
```
gonzaponte, line 291: Remove unnecessary spaces after "size"

invisible_cities/reco/corrections_test.py
```diff
@@ -268,21 +269,30 @@ def test_apply_all_correction_single_maps_raises_exception_when_invalid_map(map_
...
-                   min_value = 0,
-                   max_value = 5e2),
+                    min_value = -198,
+                    max_value = +198),
        float_arrays(size      = 1,
-                   min_value = 0,
-                   max_value = 1e5))
+                    min_value = -198,
+                    max_value = +198))
+def test_apply_geo_correction_properly(map_filename, x, y):
```
gonzaponte, line 277: Display each `float_arrays` call as a single line like the test below

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,219 @@
...
+from .. core.core_functions import all_in_range
+from .. core.core_functions import in_range
+from .. core.core_functions import shift_to_bin_centers
+from .. core.fit_functions  import fit
+from .. core.fit_functions  import gauss
+
+def select_nS_mask_and_check(dst          : pd.DataFrame                                 ,
+                             column       : type_of_signal                               ,
+                             input_mask   : Optional[np.ndarray]  = None                 ,
+                             eff_interval : Optional[Tuple[float, float]] = [0,1]        ,
```
gonzaponte, line 22: I would remove the default value for these two arguments (and thus their Optional type hint)

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,219 @@
...
+
+def select_band_and_check(dst         : pd.DataFrame                                              ,
+                          boot_map    : ASectorMap                                                ,
+                          norm_strat  : Optional[NormStrategy]                  = NormStrategy.max,
+                          input_mask  : Optional[np.ndarray]                    = None            ,
+                          range_DT    : Optional[Tuple[np.ndarray, np.ndarray]] = (10, 1300)      ,
+                          range_E     : Optional[Tuple[np.ndarray, np.ndarray]] = (10.0e+3,14e+3) ,
+                          nsigma_sel  : Optional[float]                         = 3.5             ,
+                          eff_interval: Optional[Tuple[float, float]]           = [0,1]           ,
+                          strictness  : Optional[Strictness]                    = Strictness.raise_error
```
gonzaponte, line 76: Of these, I think `nsigma_sel` and `strictness` could/should be optional, the rest I prefer that they are not. I'm not sure about `norm_strat`, because there might be some assumptions in the code that are optimized for that choice. I think you will have a better feeling for that one.

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,219 @@
...
+    range_DT: Tuple[np.array, np.array] (Optional)
+        Range in DT-axis
+        Defaults to (10, 300).
+    range_E: Tuple[np.array, np.array] (Optional)
+        Range in Energy-axis.
+        Defaults to (10e3, 14e3).
+    nsigma_sel: float (Optional)
+        Number of sigmas to set the band width.
+        Defaults to 3.5.
+    eff_interval (Optional)
```
gonzaponte, line 104:
```suggestion
    eff_interval: (float, float), optional
```

```suggestion
    eff_interval: Tuple[float, float], optional
```
Please keep on of these formats everywhere to make the style of the documentation consistent. We are heavily inspired by numpy's documentation, so in case of doubt, that is the reference.

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,219 @@
...
+                 right_closed = True)
+
+    return sel_krband
+
+
+def select_band(dt        : np.ndarray         ,
+                e         : np.ndarray         ,
+                range_dt  : Tuple[float, float],
+                range_e   : Tuple[float, float],
+                nsigma    : Optional[float] = 3.5) ->np.ndarray:
```
gonzaponte, line 146:
```suggestion
def select_band(dt       : np.ndarray         ,
                e        : np.ndarray         ,
                range_dt : Tuple[float, float],
                range_e  : Tuple[float, float],
                nsigma   : Optional[float] = 3.5) ->np.ndarray:
```

invisible_cities/reco/icaro_components.py
```diff
@@ -0,0 +1,219 @@
...
+    e: np.array
+        energy values
+    range_dt: Tuple[float, float]
+        Range in DT-axis
+    range_e: Tuple[float, float]
+        Range in Energy-axis
+    nsigma: float (Optional)
+        Number of sigmas to set the band width
+        Defaults to 3.5
+    Returns
```
gonzaponte, line 164: Add blank line before Returns

gonzaponte:
Superseded by #948.

PR #866, opened 2024-03-08
Add main functions for the computation of Kr maps

### LINKED ISSUE #863: ICAROS: calibration map creator
Creates the calibration map

Input:
- filtered dst (selected Kr events)

tasks:
- [x] bin Kr data in (x,y), monitor and check, Indicate fraction of active bins which have not enough Kr events
- [ ] tests for the binning
- [x] create map, selecting a fit method, fit data, set parameters, error an p-value per bin. monitor, and check
- [ ] tests for the map creation
- [ ] regularize - fill bins with external information (default: use the mean)
- [ ] tests for the regularization

returns
- a *dataframe* with the calibrations per (x, y) bin

carhc:
This PR addresses issue #863 (ICAROS: Calibration map creation) and lies on top of PR #865 (Data preparation functions). 

It contains the main and auxiliary functions needed for the core of Icaro city. Many different functions have been defined in order to prepare, fit and produce the Kr calibration maps. 

WIP: 

- Complete tests
- Define regularization.
- Define how to select map bins that fall within the detector volume (maybe using DB).
- Fix the prescription for the p-value calculation. 
- Cosmetics and unify style. Many of the functions here have been changing almost daily for the last week (even today!), so there may still be inconsistencies, especially regarding the functions message for parameters and returns.

### DIFF
--- a/invisible_cities/reco/krmap_functions.py
+++ b/invisible_cities/reco/krmap_functions.py
@@ -380,11 +380,10 @@ def valid_bin_counter(map_df             : pd.DataFrame,
     inside = np.count_nonzero(map_df.in_active)
     valid  = np.count_nonzero(map_df.valid)
 
-    valid_per = valid / inside
+    valid_per = valid / inside if valid <= inside else 0
 
     if valid_per <= validity_parameter:
         warnings.warn(f"{inside-valid} inner bins are not valid. {100*valid_per:.1f}% are successful.", UserWarning)
-    else:       print(f"{inside-valid} inner bins are not valid. {100*valid_per:.1f}% are successful.")
 
     return valid_per
 
--- a/invisible_cities/reco/krmap_functions.py
+++ b/invisible_cities/reco/krmap_functions.py
@@ -248,28 +248,6 @@ def get_number_of_bins(nevents : Optional[int] = None,
     else: return  np.array([100, 100]);
 
 
-def get_XY_bins(n_bins   : np.array,
-                XYrange  : Tuple[float, float]):
-    '''
-    Returns the bins that will be used to make the map.
-
-    Parameters
-    ---------
-    b_nins: np.array
-        array of len = 2 containing the number of bins in X and Y
-    XYrange: Tuple[float, float]
-        Limits (mm) of X and Y for the map computation
-
-    Returns
-    ---------
-    bins: Tuple[np.array, np.array]
-        Bins in each direction (X,Y) (square map).
-    '''
-    bins_x = np.linspace(*XYrange, n_bins[0]+1)
-    bins_y = np.linspace(*XYrange, n_bins[1]+1)
-    return bins_x, bins_y
-
-
 def get_bin_counts_and_event_bin_id(dst  : pd.DataFrame,
                                     bins : Tuple[np.array, np.array]):
     '''
@@ -361,7 +339,6 @@ def calculate_residuals(x          : np.array,
 
 
 def calculate_pval(residuals : np.array):
-
     '''
     Calculate the p-value for the Shapiro-Wilk normality test of residuals.
 
--- a/invisible_cities/reco/krmap_functions.py
+++ b/invisible_cities/reco/krmap_functions.py
@@ -401,8 +401,8 @@ def fit_and_fill_map(map_bin : pd.DataFrame,
     function specified in the fittype parameter, calculate all the different map values
     and fill the dataframe with them.
 
-      Parameters
-    --------------
+    Parameters
+    ----------
     map_bin : pd.DataFrame
          KrMap dataframe
     dst     : pd.DataFrame
@@ -410,31 +410,30 @@ def fit_and_fill_map(map_bin : pd.DataFrame,
     fittype : str
          Type of fit to perform.
 
-       Returns
-    -------------
+    Returns
+    -------
     kr_map : pd.DataFrame
          DataFrame containing map parameters.
     '''
     if not map_bin.in_active or not map_bin.has_min_counts: return map_bin
 
-    dst_bin  = dst.query(f'bin_index == {map_bin.bin}')
+    dst_bin  = dst.loc[dst.bin_index == map_bin.bin]
 
     fit_func, seed = get_function_and_seed_lt(fittype = fittype)
     x, y           = select_fit_variables    (fittype = fittype,
                                               dst     = dst_bin)
 
-    fit_output, _, _, ier = fit(func        = fit_func,
-                                x           = x,
-                                y           = y,
-                                seed        = seed(x, y),
-                                full_output = True)
+    fit_output = fit(func        = fit_func,
+                     x           = x,
+                     y           = y,
+                     seed        = seed(x, y))
 
-    par, err, cov = transform_parameters(fittype    = fittype,
-                                         fit_output = fit_output)
+    par, err, cov = transform_parameters(fit_output = fit_output)
 
-    res, std = calculate_residuals(x, y, fit_output) # Still considering this
-    chi2, _  = get_chi2_and_pvalue(y, fit_output.fn(x), len(x)-len(par), std)
-    pval     = calculate_pval(res)
+    res  = y - fit_output.fn(x)
+    std  = res.std()
+    chi2 = get_chi2_and_pvalue(y, fit_output.fn(x), len(x)-len(par), std)[0]
+    pval = stats.shapiro(res)[1] if (len(res) > 3) else 0.
 
     map_bin['e0']          = par[0]
     map_bin['ue0']         = err[0]
@@ -444,7 +443,7 @@ def fit_and_fill_map(map_bin : pd.DataFrame,
     map_bin['res_std']     = std
     map_bin['chi2']        = chi2
     map_bin['pval']        = pval
-    map_bin['fit_success'] = ier in range(1, 5)
+    map_bin['fit_success'] = fit_output.ier in range(1, 5)
     map_bin['valid']       = map_bin['fit_success'] and map_bin['has_min_counts'] and map_bin['in_active']
 
     return map_bin
--- a/invisible_cities/reco/krmap_functions.py
+++ b/invisible_cities/reco/krmap_functions.py
@@ -221,6 +221,7 @@ def get_number_of_bins(nevents : Optional[int] = None,
     '''
     Computes the number of XY bins to be used in the creation
     of correction map regarding the number of selected events.
+
     Parameters
     ---------
     nevents: int (optional)
@@ -233,8 +234,9 @@ def get_number_of_bins(nevents : Optional[int] = None,
         However, if no number of bins is given in advance, this will
         automatically select a value depending on the amount of events
         contained in the dst and the threshold.
+
     Returns
-    ---------
+    -------
     n_bins: int
         Number of bins in each direction (X,Y) (square map).
     '''
--- a/invisible_cities/reco/krmap_functions.py
+++ b/invisible_cities/reco/krmap_functions.py
@@ -378,9 +378,9 @@ def valid_bin_counter(map_df             : pd.DataFrame,
     '''
 
     inside = np.count_nonzero(map_df.in_active)
-    valid  = np.count_nonzero(map_df.valid)
+    valid  = np.count_nonzero(map_df.in_active & map_df.valid)
 
-    valid_per = valid / inside if valid <= inside else 0
+    valid_per = valid / inside if inside else 0
 
     if valid_per <= validity_parameter:
         warnings.warn(f"{inside-valid} inner bins are not valid. {100*valid_per:.1f}% are successful.", UserWarning)
--- a/invisible_cities/reco/krmap_functions.py
+++ b/invisible_cities/reco/krmap_functions.py
@@ -332,18 +332,17 @@ def get_bin_counts_and_event_bin_id(dst  : pd.DataFrame,
     return counts, bin_labels
 
 
-def calculate_residuals(x          : pd.Series,
-                        y          : pd.Series,
+def calculate_residuals(x          : np.array,
+                        y          : np.array,
                         fit_output : FitFunction):
-
     '''
     Calculate residuals and their standard deviation for the fitted data.
 
     Parameters
     ----------
-    x : pd.Series
+    x : pd.array
         Independent variable
-    y : pd.Series
+    y : pd.array
         Dependent variable
     fit_output : FitFunction
         Container for IC's fit function result
@@ -356,10 +355,7 @@ def calculate_residuals(x          : pd.Series,
         Standard deviation of residuals.
     '''
 
-    function = fit_output.fn
-
-    res = y - function(x)
-    std = res.std()
+    res = y - fit_output.fn(x); std = res.std()
 
     return res, std
 
--- a/invisible_cities/reco/krmap_functions.py
+++ b/invisible_cities/reco/krmap_functions.py
@@ -474,28 +474,6 @@ def fit_and_fill_map(map_bin : pd.DataFrame,
     return map_bin
 
 
-def find_outliers(maps    : pd.DataFrame,
-                  x2range : Tuple[float, float] = (0, 2)):
-    '''
-    For a given maps and deserved range, it returns a mask where values are
-    within the interval.
-
-    Parameters
-    ---------
-    maps: pd.DataFrame
-        Map to check the outliers
-    x2range : Tuple[float, float]
-        Range for chi2
-
-    Returns
-    ---------
-    mask: pd.Series
-        Mask.
-    '''
-    mask = in_range(maps.chi2, *x2range)
-    return mask
-
-
 def regularize_map(maps    : pd.DataFrame,
                    x2range : Tuple[float, float]):
     '''
@@ -517,8 +495,7 @@ def regularize_map(maps    : pd.DataFrame,
 
     new_map   = copy.deepcopy(maps)
 
-    outliers  = maps.in_active
-    outliers &= np.logical_not(find_outliers(new_map, x2range))
+    outliers  = maps.in_active & ~in_range(new_map.chi2, *x2range)
 
     new_map['e0'] [outliers] = np.nanmean(maps['e0'])
     new_map['lt'] [outliers] = np.nanmean(maps['lt'])
--- a/invisible_cities/reco/krmap_functions.py
+++ b/invisible_cities/reco/krmap_functions.py
@@ -398,17 +398,14 @@ def valid_bin_counter(map_df             : pd.DataFrame,
         Percentage of valid bins.
     '''
 
-    inside = len(map_df[map_df['in_active'] == True])
-    valid  = len(map_df[map_df['valid']     == True])
+    inside = np.count_nonzero(map_df.in_active)
+    valid  = np.count_nonzero(map_df.valid)
 
-    valid_per = valid / inside * 100
+    valid_per = valid / inside
 
     if valid_per <= validity_parameter:
-        # If the percentage of valid bins is below the threshold, issue a warning
-        warnings.warn(f"{inside-valid} inner bins are not valid. {valid_per:.1f}% are successful.", UserWarning)
-    else:
-        # If the percentage of valid bins meets the threshold, print a message
-        print(f"{inside-valid} inner bins are not valid. {valid_per:.1f}% are successful.")
+        warnings.warn(f"{inside-valid} inner bins are not valid. {100*valid_per:.1f}% are successful.", UserWarning)
+    else:       print(f"{inside-valid} inner bins are not valid. {100*valid_per:.1f}% are successful.")
 
     return valid_per
 
--- a/invisible_cities/reco/krmap_functions.py
+++ b/invisible_cities/reco/krmap_functions.py
@@ -1,9 +1,18 @@
-import numpy  as np
-import pandas as pd
+import copy
+import warnings
+import itertools
 
+import numpy        as np
+import pandas       as pd
+import scipy.stats  as stats
+
+
+from   typing               import Tuple, Optional
 from .. types.symbols       import KrFitFunction
 from .. evm.ic_containers   import FitFunction
 from .. core.fit_functions  import polynom, expo
+from ..core.core_functions  import in_range, shift_to_bin_centers
+from ..core.fit_functions   import fit, get_chi2_and_pvalue
 
 
 def lin_seed(x : np.array, y : np.array):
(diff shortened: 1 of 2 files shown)


jahernando:
Some comments [1]:

-  *check_if_values_in_interval* seems a decorator of *in_range*, can be done in a more elegant way?
- *apply_geo_correction* is not a good name: something more as *get_xy_geo_correction_function*
- *apply_geo_correction* is a useful wrapper, but what is a *ASectorMap* object? Do we need it? Do you keep it for backward compatibility?

Tests:
- ok *check_if_values_in_interval* 
- ok *apply_geo_corection*

jahernando:
Comments & Corrections:

1. *get_number_of_bins* ok
2. *get_binned_data* what is does is: *get_bin_counts_and_event_bin_id*, change name!
3. *update_dst* extend: *update_dst_with_event_bin_id*
4. in *calculate_residuals* line 611, it is missing *KrFitFunction* in *lin_function*?
5. *fit_and_fill_map*, I think the *try* is not necessary. 
6. *fit_and_fill_map*, I will remove the *else* associated with *if not map_bin['in_active'] or not map_bin['has_min_counts']:*, and put the rest of the code (most of the function body) in the first identation level. 
7.  *fit_and_fill_map*, L736, *name  = get_par_name_from_fittype(fittype = fittype)*, you are transforming the fit paramters into (E0, LT), so this is not necessary [it is already solved in a posterior commit!].
8. *icaro* city is not commented, you need to indicate the meaning of the parameters, in particular what is *nStimeprofile*
9. *icaro* city, I understand it is still a draft, but nevertheless, we should add a *check* function after the *map_builder* and *map_evol* main functions.
10. *icaro* city, L80, L81, about the *sinks*, what are the arguments: *"pointlike_event", "pmap_passed"*
11. *compute_map*, has some hardwired values, this is not acceptable, these parameters should be in a configuration file, even if they are almost never modified [Some modifications done in a later commit].
12. What is the decision about computing the chi2?
13. Most of the functions have not an associated test!


Tests:
 -  ok *get_number_of_bins*

jahernando:
This is a large commit, better do small commits with specific changes. Otherwise is difficult to follow the flow of changes.

I understant that:
- some (empty) functions are introduced in the main flow to check the map and the map evolution internally
- some functions have been renamed to gain in clarity
- some arguments are not passed by default to gain robustness

Do I miss other changes?

There is a comment of many lines related with time-series, I understant this is the time-evolution part, that will be re-adapted in a new PR.

@carhc Are you preparing to commit more tests of the functions used in the code?

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,374 @@ def transform_parameters(fit_output : FitFunction):
...
+
+    kr_map  = pd.DataFrame(columns = columns)
+
+    n_xbins   = len(bins[0])-1
+    n_ybins   = len(bins[1])-1
+    b_center  = [shift_to_bin_centers(axis) for axis in bins]
+
+    bin_index = range(n_xbins*n_ybins)
+    geom_comb = itertools.product(b_center[1], b_center[0])
+    r_values  = np.array([np.sqrt(x**2+y**2)for x, y in itertools.product(b_center[1], b_center[0])])
```
gonzaponte: We usually take x as the outer look and y as the inner one. Here `geom_comb` will loop over x first, I suggest to change it, if it doesn't have any implications.

Also, you have two instances of this cartesian product. I suggest making a list and use it in the next line and below in the assignment to the columns

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,374 @@ def transform_parameters(fit_output : FitFunction):
...
+
+    kr_map['bin']            = bin_index
+    kr_map['counts']         = counts
+    kr_map['R']              = r_values
+    kr_map[['Y', 'X']]       = pd.DataFrame(geom_comb)
+    kr_map['in_active']      = kr_map['R']      <= r_max
+    kr_map['has_min_counts'] = kr_map['counts'] >= n_min
+    kr_map['fit_success']    = False
+    kr_map['valid']          = False
```
gonzaponte: Since you are going to enumerate most columns anyway, why not create the dataframe using the syntax:
`pd.DataFrame(dict(col0 = values0, ...)`?

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,374 @@ def transform_parameters(fit_output : FitFunction):
...
+    Returns
+    ---------
+    bins: Tuple[np.array, np.array]
+        Bins in each direction (X,Y) (square map).
+    '''
+    bins_x = np.linspace(*XYrange, n_bins[0]+1)
+    bins_y = np.linspace(*XYrange, n_bins[1]+1)
+    return bins_x, bins_y
+
+
```
gonzaponte: Unless this function is used in many many places, I don't think it's worth it.

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,374 @@ def transform_parameters(fit_output : FitFunction):
...
+    std : float
+        Standard deviation of residuals.
+    '''
+
+    function = fit_output.fn
+
+    res = y - function(x)
+    std = res.std()
+
+    return res, std
```
gonzaponte, line 338: make this more compact: it can fit in one line.

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,374 @@ def transform_parameters(fit_output : FitFunction):
...
+    residuals : np.array
+        Residuals from the fitted model.
+
+    Returns
+    -------
+    pval : float
+        p-value for the Shapiro-Wilk normality test.
+    '''
+
+    pval = stats.shapiro(residuals)[1] if (len(residuals) > 10) else 0.
```
gonzaponte: is 10 a magic number?

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,374 @@ def transform_parameters(fit_output : FitFunction):
...
+        Threshold for the ratio of valid bins (default set to 0.9).
+
+    Returns
+    -------
+    valid_per : float
+        Percentage of valid bins.
+    '''
+
+    inside = len(map_df[map_df['in_active'] == True])
+    valid  = len(map_df[map_df['valid']     == True])
```
gonzaponte:
```suggestion
    inside = np.count_nonzero(map_df.in_active)
    valid  = np.count_nonzero(map_df.valid)
```

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,374 @@ def transform_parameters(fit_output : FitFunction):
...
+        p-value for the Shapiro-Wilk normality test.
+    '''
+
+    pval = stats.shapiro(residuals)[1] if (len(residuals) > 10) else 0.
+
+    return pval
+
+
+def valid_bin_counter(map_df             : pd.DataFrame,
+                      validity_parameter : Optional[float] = 0.9):
```
gonzaponte, line 362: Is this per one or percent? The docsstring suggests per one, the usage suggests percent. I suggest renaming it to make it more clear (`validity_ratio` or `validity_percent`)

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,374 @@ def transform_parameters(fit_output : FitFunction):
...
+    valid  = len(map_df[map_df['valid']     == True])
+
+    valid_per = valid / inside * 100
+
+    if valid_per <= validity_parameter:
+        # If the percentage of valid bins is below the threshold, issue a warning
+        warnings.warn(f"{inside-valid} inner bins are not valid. {valid_per:.1f}% are successful.", UserWarning)
+    else:
+        # If the percentage of valid bins meets the threshold, print a message
+        print(f"{inside-valid} inner bins are not valid. {valid_per:.1f}% are successful.")
```
gonzaponte: remove superfluous comments

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,374 @@ def transform_parameters(fit_output : FitFunction):
...
+         kdst dataframe.
+    fittype : str
+         Type of fit to perform.
+
+       Returns
+    -------------
+    kr_map : pd.DataFrame
+         DataFrame containing map parameters.
+    '''
+    if not map_bin['in_active'] or not map_bin['has_min_counts']: return map_bin
```
gonzaponte: Use `.column` wherever possible

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,374 @@ def transform_parameters(fit_output : FitFunction):
...
+
+    k        = map_bin.bin
+
```
gonzaponte: superfluous variable, please remove

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,374 @@ def transform_parameters(fit_output : FitFunction):
...
+    map_bin['e0']          = par[0]
+    map_bin['ue0']         = err[0]
+    map_bin['lt']          = par[1]
+    map_bin['ult']         = err[1]
+    map_bin['covariance']  = cov
+    map_bin['res_std']     = std
+    map_bin['chi2']        = chi2
+    map_bin['pval']        = pval
+    map_bin['fit_success'] = True if ier in [1, 2, 3, 4] else False
```
gonzaponte:
```suggestion
    map_bin['fit_success'] = ier in range(1, 5)

File: invisible_cities/reco/krmap_functions.py [hunk 12]
```diff
@@ -149,3 +158,374 @@ def transform_parameters(fit_output : FitFunction):
...
+    map_bin['valid']       = map_bin['fit_success'] & map_bin['has_min_counts'] & map_bin['in_active']
```
gonzaponte: for booleans it's more idiomatic to use `and` instead of `&`

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,374 @@ def transform_parameters(fit_output : FitFunction):
...
+
+    Returns
+    ---------
+    mask: pd.Series
+        Mask.
+    '''
+    mask = in_range(maps.chi2, *x2range)
+    return mask
+
+
```
gonzaponte: This function is also superfluous

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,374 @@ def transform_parameters(fit_output : FitFunction):
...
+    Returns
+    ---------
+    new_map: pd.DataFrame
+        Regularized map.'''
+
+    new_map   = copy.deepcopy(maps)
+
+    outliers  = maps.in_active
+    outliers &= np.logical_not(find_outliers(new_map, x2range))
```
gonzaponte:
```suggestion
    outliers  = maps.in_active & ~in_range(new_map.chi2, *x2range))
```

gonzaponte (commented):
This is being done with peras

invisible_cities/reco/krmap_functions.py
carhc: Kind of, `stats.shapiro` requires an input of at least len 3, but I initially thought of 10 because "it made sense" for me in the context of where the function is used. However, where this bit of code appears, this should not be a problem, because for low event count bins there would be no lifetime fit performed. I have updated the number from 10 to 3.

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,325 @@ def transform_parameters(fit_output : FitFunction):
...
+                     n_min  : int,
+                     r_max  : float)->pd.DataFrame:
+    '''
+    This function creates the dataframe in which the map parameters are stored.
+
+    Parameters
+    ----------
+
+    fittype : KrFitFunction
+        Chosen fit function for map computation
```
gonzaponte: This is no longer a parameter. Remove blank line before.

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,325 @@ def transform_parameters(fit_output : FitFunction):
...
+    bins : Tuple[np.array, np.array]
+        Tuple containing bins in both axis
```
gonzaponte:
```suggestion
        Tuple containing bins in both axes
```

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,325 @@ def transform_parameters(fit_output : FitFunction):
...
+    counts : np.array
+        Number of events falling into each bin
+    n_min : int
+        Min number of events per bin required to perform fits
+    r_max : float
+        Radius defining the active area of the detector
+
+    Returns
+    -------
+
```
gonzaponte: blank line

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,325 @@ def transform_parameters(fit_output : FitFunction):
...
+        events per bin, bin in/outside the active volume, bin position
+        (X, Y, R), etc.
+    '''
+    n_xbins   = len(bins[0])-1
+    n_ybins   = len(bins[1])-1
+    b_center  = [shift_to_bin_centers(axis) for axis in bins]
+
+    bin_index = range(n_xbins*n_ybins)
+    geom_comb = list(itertools.product(b_center[0], b_center[1]))
+    r_values  = np.array([np.sqrt(x**2+y**2)for x, y in geom_comb])
```
gonzaponte:
```suggestion
    geom_comb = np.array(list(itertools.product(*b_center))
    r_values  = np.sum(geom_comb**2, axis=1)**0.5
```

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,325 @@ def transform_parameters(fit_output : FitFunction):
...
+
+    kr_map    = pd.DataFrame(dict(bin            = bin_index,
+                                  counts         = counts,
+                                  X              = list(zip(*geom_comb))[0],
+                                  Y              = list(zip(*geom_comb))[1],
```
gonzaponte: Following the suggestion above...
```suggestion
                                  X              = geom_comb.T[0],
                                  Y              = geom_comb.T[1],
```

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,325 @@ def transform_parameters(fit_output : FitFunction):
...
+                                  ue0            = np.nan,
+                                  lt             = np.nan,
+                                  ult            = np.nan,
+                                  cov            = np.nan,
+                                  res_std        = np.nan,
+                                  chi2           = np.nan,
+                                  pval           = np.nan))
+
+    kr_map.in_active      = kr_map.R      <= r_max
+    kr_map.has_min_counts = kr_map.counts >= n_min
```
gonzaponte: you can add these to the dictionary above since you defined `r_values` and `counts`

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,325 @@ def transform_parameters(fit_output : FitFunction):
...
+    n_bins: int
+        Number of bins in each direction (X,Y) (square map).
+    '''
+    if    n_bins is not None: return n_bins;
+    elif  nevents < thr: return np.array([50, 50]);
+    else: return  np.array([100, 100]);
+
+
+def get_bin_counts_and_event_bin_id(dst  : pd.DataFrame,
+                                    bins : Tuple[np.array, np.array]):
```
gonzaponte: why not accept two arrays instead of a tuple of arrays? here and in `create_df_kr_map`.

A couple of tests for this function could be:
- Check the size of the output arrays
- Check that the sum of counts corresponds to the number of events in the dst

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,325 @@ def transform_parameters(fit_output : FitFunction):
...
+
+      Returns
+    -------------
+    counts     : np.array
+         Total number of events falling into each bin
+    bin_labels : np.array
+         1D bin label for each event
+
+      Further info:
+    -----------------
```
gonzaponte:
```suggestion
    Parameters
    ----------
    dst  : pd.DataFrame
         Krypton dataframe.
    bins : Tuple[np.array, np.array]
         Bins used to compute the map.

    Returns
    -------
    counts     : np.array
         Total number of events falling into each bin
    bin_labels : np.array
         1D bin label for each event

    Further info:
    -------------
```

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,325 @@ def transform_parameters(fit_output : FitFunction):
...
+
+    Returns
+    -------
+    res : np.array
+        Residuals.
+    std : float
+        Standard deviation of residuals.
+    '''
+
+    res = y - fit_output.fn(x); std = res.std()
```
gonzaponte, line 336:
```suggestion
    res = y - fit_output.fn(x)
    std = res.std()
```
I think I told you to make it a single line, but I didn't realize you needed the two variables. In any case, this function is so trivial, that if it is only used in `fit_and_fill_map`, you can remove it.

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,325 @@ def transform_parameters(fit_output : FitFunction):
...
+
+    Returns
+    -------
+    pval : float
+        p-value for the Shapiro-Wilk normality test.
+    '''
+
+    pval = stats.shapiro(residuals)[1] if (len(residuals) > 3) else 0.
+
+    return pval
```
gonzaponte, line 358: I think you can also remove this function.

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,325 @@ def transform_parameters(fit_output : FitFunction):
...
+         Type of fit to perform.
+
+       Returns
+    -------------
+    kr_map : pd.DataFrame
+         DataFrame containing map parameters.
+    '''
+    if not map_bin.in_active or not map_bin.has_min_counts: return map_bin
+
+    dst_bin  = dst.query(f'bin_index == {map_bin.bin}')
```
gonzaponte:
```suggestion
    dst_bin  = dst.loc[dst_bin.bin_index == map_bin.bin]
```

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,325 @@ def transform_parameters(fit_output : FitFunction):
...
+    if valid_per <= validity_parameter:
+        warnings.warn(f"{inside-valid} inner bins are not valid. {100*valid_per:.1f}% are successful.", UserWarning)
+    else:       print(f"{inside-valid} inner bins are not valid. {100*valid_per:.1f}% are successful.")
+
+    return valid_per
+
+
+def fit_and_fill_map(map_bin : pd.DataFrame,
+                     dst     : pd.DataFrame,
+                     fittype : KrFitFunction):
```
gonzaponte, line 393: This function is not tested

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,325 @@ def transform_parameters(fit_output : FitFunction):
...
+
+    fit_func, seed = get_function_and_seed_lt(fittype = fittype)
+    x, y           = select_fit_variables    (fittype = fittype,
+                                              dst     = dst_bin)
+
+    fit_output, _, _, ier = fit(func        = fit_func,
+                                x           = x,
+                                y           = y,
+                                seed        = seed(x, y),
+                                full_output = True)
```
gonzaponte: There is no longer a `full_output` parameter and the output is only `fit_output`

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,325 @@ def transform_parameters(fit_output : FitFunction):
...
+    x2range : Tuple[float, float]
+        Range for chi2
+
+    Returns
+    ---------
+    new_map: pd.DataFrame
+        Regularized map.'''
+
+    new_map   = copy.deepcopy(maps)
+
```
gonzaponte: blank line

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,325 @@ def transform_parameters(fit_output : FitFunction):
...
+        Map to check the outliers
+    x2range : Tuple[float, float]
+        Range for chi2
+
+    Returns
+    ---------
+    new_map: pd.DataFrame
+        Regularized map.'''
+
+    new_map   = copy.deepcopy(maps)
```
gonzaponte:
```suggestion
    new_map   = maps.copy()

File: invisible_cities/reco/krmap_functions_test.py [hunk 30]
```diff
@@ -104,3 +105,62 @@ def test_transform_parameters(a, b):
...
+
+@given(integers(min_value = 0, max_value = 1e10),
+       integers(min_value = 0, max_value = 1e10),
+       arrays  (dtype     = np.int64, shape = (2,),
+                elements  = integers(min_value = 1,
+                                     max_value = 1e4)))
+def test_get_number_of_bins_performance_default_value(nevents, thr, n_bins):
+    bins = krf.get_number_of_bins(nevents, thr, n_bins)
+
+    npt.assert_array_equal(bins, n_bins)
```
gonzaponte: Simplify this test: fix a value for `nevents` and `thr` and parametrize `n_bins` with a few values, instead of letting hypothesis generate a zillion values.

invisible_cities/reco/krmap_functions_test.py
```diff
@@ -104,3 +105,62 @@ def test_transform_parameters(a, b):
...
+
+@given(integers(min_value = 0, max_value = 1e10),
+       integers(min_value = 0, max_value = 1e10),
+       arrays  (dtype = np.int64,  shape = (2,),
+                elements = integers(min_value = 1,
+                                    max_value = 1e4)))
+def test_get_number_of_bins_returns_type(nevents, thr, n_bins):
+
+    assert type(krf.get_number_of_bins(nevents, thr) ) == np.ndarray
+    assert type(krf.get_number_of_bins(n_bins=n_bins)) == np.ndarray
```
gonzaponte: Remove hypothesis and do a simple fixed cases. Why do we want to test the output type?

invisible_cities/reco/krmap_functions_test.py
```diff
@@ -104,3 +105,62 @@ def test_transform_parameters(a, b):
...
+
+
+@given(n_bins=arrays(dtype = int,      shape = (2,),
+                     elements = integers(min_value = 2,
+                                         max_value = 100)),
+       n_min=integers(min_value=1,  max_value=100),
+       r_max=floats  (min_value=50, max_value=450))
```
gonzaponte: again, a single case is enough.

invisible_cities/reco/krmap_functions_test.py
```diff
@@ -104,3 +105,62 @@ def test_transform_parameters(a, b):
...
+
+    bins   = [np.linspace(*XYrange, n_bins_x+1), np.linspace(*XYrange, n_bins_y+1)]
+    counts =  np.full(shape=(n_bins_x, n_bins_y), fill_value = 100, dtype=int).flatten()
+    df = krf.create_df_kr_map(bins, counts, n_min, r_max)
+
+    columns  = ['bin', 'counts', 'e0', 'ue0', 'lt', 'ult', 'cov', 'res_std', 'chi2',
+                'pval', 'in_active', 'has_min_counts', 'fit_success', 'valid', 'R', 'X', 'Y']
+
+    assert all(element in columns for element in df.columns.values)
+
```
gonzaponte: Check also the number of rows in the output and rename the test to something like `test_create_df_kr_map_shape`.

Also, add another test checking that the non-dummy values set in the function fall in the expected range.

invisible_cities/reco/krmap_functions.py
gonzaponte: Also, need tests for this function

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +158,325 @@ def transform_parameters(fit_output : FitFunction):
...
+    map_bin['chi2']        = chi2
+    map_bin['pval']        = pval
+    map_bin['fit_success'] = ier in range(1, 5)
+    map_bin['valid']       = map_bin['fit_success'] and map_bin['has_min_counts'] and map_bin['in_active']
+
+    return map_bin
+
+
+def regularize_map(maps    : pd.DataFrame,
+                   x2range : Tuple[float, float]):
```
gonzaponte: Tests. At least:
- check it doesn't modify the input
- create a silly map with ones and a few "invalid entries", then check the invalid entries are restored to 1.

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +157,326 @@ def transform_parameters(fit_output : FitFunction):
...
+    counts, _, _, bin_indexes = stats.binned_statistic_2d(x=dst.X, y=dst.Y, values=None,
+                                                          bins=[bins_x, bins_y], statistic='count',
+                                                          expand_binnumbers=True)
+
+    counts       = counts.flatten()
+    bin_indexes -= 1
+    bin_labels   = np.ravel_multi_index(bin_indexes, dims=(n_xbins, n_ybins),
+                                        mode='clip', order = 'F')
+
+    return counts, bin_labels
```
gonzaponte, line 310:
```suggestion
    # indexes 0 and len+1 represent underflow and overflow, thus we subtract 1
    bin_labels   = np.ravel_multi_index(bin_indexes - 1, dims=(n_xbins, n_ybins),
                                        mode='clip', order = 'F')

    return counts.flatten(), bin_labels
```

gonzaponte: if I understand correctly `mode="clip"` means that bins out of range will be assigned the closest bin. Is this what we used to do @bpalmeiro ?

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +157,326 @@ def transform_parameters(fit_output : FitFunction):
...
+
+    valid_per = valid / inside if inside else 0
+
+    if valid_per <= validity_parameter:
+        warnings.warn(f"{inside-valid} inner bins are not valid. {100*valid_per:.1f}% are successful.", UserWarning)
+
+    return valid_per
+
+
+def fit_and_fill_map(map_bin : pd.DataFrame,
```
gonzaponte, line 391: If I understand correctly, this is a `pd.Series`, not a `pd.DataFrame`

invisible_cities/reco/krmap_functions.py
```diff
@@ -149,3 +157,326 @@ def transform_parameters(fit_output : FitFunction):
...
+
+    Returns
+    ---------
+    new_map: pd.DataFrame
+        Regularized map.'''
+    new_map   = maps.copy()
+    outliers  = new_map.in_active & ~new_map.valid
+
+    if isinstance(x2range, Tuple):
+          outliers &= ~in_range(new_map.chi2, *x2range)
```
gonzaponte, line 474: I think the logic is wrong here. The outliers should be those that are in the active and (not valid OR not in range). Truth table:
```
in_active  valid   x2_in_range  outlier  expected
--------------------------------------------------
True       True    True         False    False   <--- satisfied
True       True    False        False    True    <--- not satisfied
True       False   True         False    True    <--- not satisfied
True       False   False        True     True    <--- satisfied
False      x       x            False    False   <--- satisfied
```

invisible_cities/reco/krmap_functions_test.py
```diff
@@ -104,3 +106,133 @@ def test_transform_parameters(a, b):
...
+    bins_y   = np.linspace(*XYrange, n_bins_y+1)
+    counts   = np.full(shape=(n_bins_x, n_bins_y), fill_value = 100, dtype=int).flatten()
+
+    df = krf.create_df_kr_map(bins_x, bins_y, counts, n_min, r_max)
+
+    columns  = ['bin', 'counts', 'e0', 'ue0', 'lt', 'ult', 'cov', 'res_std', 'chi2',
+                'pval', 'in_active', 'has_min_counts', 'fit_success', 'valid', 'R', 'X', 'Y']
+
+    assert all(element in columns for element in df.columns.values)
+    assert df.bin.nunique() == n_bins_x*n_bins_y
```
gonzaponte, line 151: should the length od the dataframe be also `n_binx_x*n_bins_y`?

invisible_cities/reco/krmap_functions_test.py
```diff
@@ -104,3 +106,133 @@ def test_transform_parameters(a, b):
...
+                   (( 3, 3), 100,  0.2)))
+def test_valid_bin_counter_warning(n_bins, rmax, validity_parameter):
+    counts = np.array(range(n_bins[0]*n_bins[1]))
+    krmap  = krf.create_df_kr_map(bins_x = np.linspace(-rmax, +rmax, n_bins[0]+1),
+                                  bins_y = np.linspace(-rmax, +rmax, n_bins[1]+1),
+                                  counts = counts,
+                                  n_min  = 0,
+                                  r_max  = np.nextafter(np.sqrt(2)*rmax, np.inf))
+
+    krmap.valid.iloc[0 : 9] = True
```
gonzaponte, line 215: as discussed offline, simplify the data in this test

invisible_cities/reco/krmap_functions_test.py
```diff
@@ -104,3 +106,133 @@ def test_transform_parameters(a, b):
...
+
+    if validity_parameter == 1:
+        with warns(UserWarning, match = "inner bins are not valid."):
+            krf.valid_bin_counter(map_df = krmap, validity_parameter = validity_parameter)
```
gonzaponte, line 219: for `validity_parameter = 0.2`, you are not checking anything. Check this: https://stackoverflow.com/a/45671804

gonzaponte (commented):
a few more comments.

invisible_cities/reco/krmap_functions.py
bpalmeiro: In our case, we had three values for "fiducial":

- The R max for the selection (just a quality one to ensure everything was inside)

- The R max for the map, that defines the maximum extension of it. In this case, if the event didn't happen to fall inside the bins, it was disregarded.

- Also (not sure here it's the most appropriate place to mention, tho), we had a posterior step where the peripheral bins (those further than a fiducial r -but within said Rmax for map production-) were set to nan.

The latter was used to ensure that, in the posterior analysis, all the hits reconstructed outside the volume were automatically flagged as nan.

bpalmeiro: That said, the best approach would be to disregard events outside boundaries instead of fakely merging them in the border bins, even if we rely on a previous selection. It's not a very strong opinion, but I think it adds redundancy (and, therefore, a bit of robustness) to the process.

carhc: Yes, my bad... I assumed that there would not be any events outside the boundaries here since I was expecting a "clean" `dst` in this part of the code, so when I defined `bin_labels` like that I didn't think of the consequences in case some event was actually out of the range.

I am thinking in a solution like this one: 

```suggestion
    counts       = counts.flatten()
    bin_indexes -= 1
    valid_mask = ( in_range(bin_indexes[0], 0, n_xbins, right_closed=True) &
                   in_range(bin_indexes[1], 0, n_ybins, right_closed=True) )

    bin_labels             = np.full(bin_indexes.shape[1], fill_value=np.nan)
    bin_labels[valid_mask] = np.ravel_multi_index((bin_indexes[0, valid_mask],
                                                   bin_indexes[1, valid_mask]),
                                                   dims=(n_xbins, n_ybins),
                                                   order='F')
```

In case some events are outside the desired range, their label would be a `NaN` instead of a numerical index, so when grouping the events bin by bin, none would be assigned to those events...

gonzaponte: ok

gonzaponte:
Superseded by #948.

PR #870, opened 2024-04-24
Modify detsim to read nexus files with int instead of strings

### LINKED ISSUE #869: Modify detsim to deal with string-free nexus files
In the latest nexus releases, we have added the possibility of replacing strings with ints, to reduce the size of the output file. This is done optionally, and in practice it is used only in cosmogenic simulations.
We have to modify detsim to take that into account, in the `hits_selector` function, allowing for the possibility of selecting on ints (using the string map of the nexus file) instead of on strings.

paolafer:
This PR modifies a few functions all over the code, to deal with nexus files with `int` instead of `string` in some columns of their dataframes.
The main modification is in the `detsim` city, in the selection of hits in active or buffer, but a few other functions needed to be updated, too.
A test has been added to prove that the code works.

Closes #869.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -617,14 +617,8 @@ def MC_hits_from_files(files_in : List[str], rate: float) -> Generator:
         except tb.exceptions.NoSuchNodeError:
             continue
 
-        strings = False
-
         l_type = hits_df.dtypes['label']
-        if l_type == np.int32:
-            map_df = load_mcstringmap(filename)
-        else:
-            strings = True
-
+        map_df = load_mcstringmap(filename) if l_type == np.int32 else None
 
         for evt, hits in hits_df.groupby(level=0):
             yield dict(event_number = evt,
@@ -635,8 +629,8 @@ def MC_hits_from_files(files_in : List[str], rate: float) -> Generator:
                        time         = hits.time  .values,
                        label        = hits.label .values,
                        timestamp    = timestamp(evt),
-                       name         = map_df.name.values if not strings else "",
-                       name_id      = map_df.index.values if not strings else 0)
+                       name         = map_df.name .values if map_df is not None else "",
+                       name_id      = map_df.index.values if map_df is not None else  0)
 
 
 @check_annotations
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -618,35 +618,23 @@ def MC_hits_from_files(files_in : List[str], rate: float) -> Generator:
             continue
 
         strings = False
-        try:
+        if type(hits_df.iloc[0].label) is not str:
             map_df = load_mcstringmap(filename)
-        except:
+        else:
             strings = True
 
-        if strings:
-            for evt, hits in hits_df.groupby(level=0):
-                yield dict(event_number = evt,
-                           x            = hits.x     .values,
-                           y            = hits.y     .values,
-                           z            = hits.z     .values,
-                           energy       = hits.energy.values,
-                           time         = hits.time  .values,
-                           label        = hits.label .values,
-                           timestamp    = timestamp(evt),
-                           name         = "",
-                           name_id      = 0)
-        else:
-            for evt, hits in hits_df.groupby(level=0):
-                yield dict(event_number = evt,
-                           x            = hits.x     .values,
-                           y            = hits.y     .values,
-                           z            = hits.z     .values,
-                           energy       = hits.energy.values,
-                           time         = hits.time  .values,
-                           label        = hits.label .values,
-                           timestamp    = timestamp(evt),
-                           name         = map_df.name.values,
-                           name_id      = map_df.name_id.values)
+
+        for evt, hits in hits_df.groupby(level=0):
+            yield dict(event_number = evt,
+                       x            = hits.x     .values,
+                       y            = hits.y     .values,
+                       z            = hits.z     .values,
+                       energy       = hits.energy.values,
+                       time         = hits.time  .values,
+                       label        = hits.label .values,
+                       timestamp    = timestamp(evt),
+                       name         = map_df.name.values if not strings else "",
+                       name_id      = map_df.name_id.values if not strings else 0)
 
 
 @check_annotations
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -64,6 +64,7 @@ from .. io     .event_filter_io   import       event_filter_writer
 from .. io     .pmaps_io          import               pmap_writer
 from .. io     .rwf_io            import             buffer_writer
 from .. io     .mcinfo_io         import            load_mchits_df
+from .. io     .mcinfo_io         import          load_mcstringmap
 from .. io     .dst_io            import                 df_writer
 from .. types  .ic_types          import                  NoneType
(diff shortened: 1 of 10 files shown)


invisible_cities/cities/components.py
```diff
@@ -615,15 +616,37 @@ def MC_hits_from_files(files_in : List[str], rate: float) -> Generator:
...
+                yield dict(event_number = evt,
+                           x            = hits.x     .values,
+                           y            = hits.y     .values,
+                           z            = hits.z     .values,
+                           energy       = hits.energy.values,
+                           time         = hits.time  .values,
+                           label        = hits.label .values,
+                           timestamp    = timestamp(evt),
+                           name         = map_df.name.values,
+                           name_id      = map_df.name_id.values)
```
gonzaponte: 1. Why not look at the type of the label column to determine whether we need to map between ints and strings? instead of try/except?
2. I don't like the repetition. I suggest unifying both loops and inserting the ifs only for the relevant fields. e.g.
```
dict(...
     name = map_df.name.values if strings else ""
     ...)
```

invisible_cities/cities/detsim.py
```diff
@@ -73,10 +73,20 @@ def hits_selector(active_only: bool=True):
...
+            sel = (label == "ACTIVE")
+            if not active_only:
+                sel =  sel | (label == "BUFFER")
+        else:
+            active_id = name_id[name == "ACTIVE"][0]
+            sel = (label == active_id)
+            if not active_only:
+                buffer_id = name_id[name == "BUFFER"][0]
+                sel =  sel | (label == buffer_id)
+
```
gonzaponte: Again I suggest unifying both logical paths. The only difference between both is whether you select values equal to `"ACTIVE"/"BUFFER"` or `name_id[name == "ACTIVE"/"BUFFER"]`.
Also, numpy arrays contain type information in `an_array.dtype`. I would perhaps test the opposite, that is, that the array contains ints instead of strings (easier to check with `dtype`). I believe this would allow you to remove the check for zero length.

invisible_cities/io/mcinfo_io.py
```diff
@@ -948,3 +951,19 @@ def _read_mchit_info(h5f, event_range=(0, int(1e9))) -> Mapping[int, Sequence[MC
...
+    parameters
+    ----------
+    file_name : str
+                Name of the file containing MC info.
+
+    returns
+    -------
+    pd.DataFrame with the int-string pairs within the file.
+    """
+    return pd.read_hdf(file_name, 'MC/string_map')
```
gonzaponte: It's probably worth it setting the `name_id` field as the index of the dataframe, so the search you perform in `select_hits`, is a bit simpler.

invisible_cities/cities/detsim_test.py
```diff
@@ -261,3 +261,15 @@ def test_detsim_buffer_times(ICDATADIR, output_tmpdir):
...
+
+def test_detsim_hits_without_strings(ICDATADIR, output_tmpdir):
+    PATH_IN  = os.path.join(ICDATADIR    , "nexus_next100_nostrings.h5")
+    PATH_OUT = os.path.join(output_tmpdir, "detsim_nostrings.h5")
+    conf = configure('detsim $ICTDIR/invisible_cities/config/detsim_next100.conf'.split())
+    conf.update(dict(files_in    = PATH_IN,
+                     file_out    = PATH_OUT,
+                     run_number  = 0))
+    result = detsim(**conf)
+
```
gonzaponte: It would be great if you checked that the string map is also present in the output file.

gonzaponte (changes requested):

invisible_cities/cities/components.py
paolafer: Regarding 1., I'm having problems checking the type of the column. What don't you like about try/except?
Regarding 2, done.

invisible_cities/cities/detsim.py
paolafer: I think that the type is `int32`; how do you check that exactly? `int` does not work.

invisible_cities/cities/components.py
paolafer: I think now it works. However, why is that better than try/except?

invisible_cities/cities/detsim.py
paolafer: I did change the code as suggested. However, I'm not sure I'm using `name_id` as an index as in another comment.

invisible_cities/io/mcinfo_io.py
paolafer: See above.

invisible_cities/cities/detsim_test.py
paolafer: Done.

invisible_cities/cities/detsim_test.py
```diff
@@ -261,3 +261,17 @@ def test_detsim_buffer_times(ICDATADIR, output_tmpdir):
...
+    pd.read_hdf(PATH_OUT, 'MC/string_map')
```
gonzaponte: Maybe check that it's not empty?

invisible_cities/cities/components.py
```diff
@@ -615,6 +616,16 @@ def MC_hits_from_files(files_in : List[str], rate: float) -> Generator:
...
+
+        strings = False
+
+        l_type = hits_df.dtypes['label']
+        if l_type == np.int32:
+            map_df = load_mcstringmap(filename)
+        else:
+            strings = True
+
+
```
gonzaponte: Perhaps we can simplify this a bit more. Now you have 1 extra variable (`strings`) to keep track of whether `map_df` has a value. How about
```
map_df = load_mcstringmap(filename) if l_type == np.int32 else None
````
and later on checking if `map_df` is `None`?

File: invisible_cities/cities/components.py [hunk 7]
```diff
@@ -623,7 +634,9 @@ def MC_hits_from_files(files_in : List[str], rate: float) -> Generator:
                        energy       = hits.energy.values,
                        time         = hits.time  .values,
                        label        = hits.label .values,
-                       timestamp    = timestamp(evt))
+                       timestamp    = timestamp(evt),
+                       name         = map_df.name.values if not strings else "",
+                       name_id      = map_df.index.values if not strings else 0)
```
```suggestion
                       name         = map_df.name .values if not strings else "",
                       name_id      = map_df.index.values if not strings else 0)
```

invisible_cities/cities/components.py
gonzaponte: try/except is too generic. You could narrow it down by looking for the specific exception thrown if the table is not found. However, in this case, there are only two possibilities: label contains strings or ints. A generic code can help misbehavior to slip through or create confusing error messages.

invisible_cities/cities/detsim.py
gonzaponte: I think I got confused with this. I thought `name_id` was a dataframe, not an array. If it were a dataframe, the best would be to do  `name_id.loc[some_label]`. Is there a good reason to split the dataframe in two arrays in `MC_hits_from_files`?

gonzaponte: Discussed offline. Will stay like this for now.

gonzaponte (approved):
A necessary update to deal with the change in the file structure produced upstream. Good job!

PR #871, opened 2024-05-07
Add a protection for possible mismatch in sensor IDs

### LINKED ISSUE #779: Avoid shape mismatch in buffy with inconsistent sensor ids
Buffy source function
https://github.com/next-exp/IC/blob/a747d537c3ac0e38505b256ec85b60740f114e60/invisible_cities/cities/components.py#L349
reads sensors responses from nexus files. These sensor responses are not divided in PMTs or SIPMs, but they are labeled by the Sensor ID number. Ideally, this number from nexus should correspond to SensorID from IC database. If it is not the case, a buggy behaviour arises.

To see it (as it is actually the case for `demopp`), assume that PMT sensor ids from IC are `pmt_ids = [2, 3, 4]`, but PMT sensor ids from nexus are [0, 1, 2]. Then, the PMT sensor response selected in buffy source correspond to sensor ids [2, 3, 4]. The sensor ids [0, 1] doesn't exist in nexus `demopp` database, thus the just the response for id [2] is taken as pmt response. Nevertheless the response for SIPMs is taken as `~pmt_indx`, thus sensors [0, 1] are considered as SIPMs. This produces a shape mismatch latter in the city since sipm ids are explicity selected.

Since there is also a `sns_positions` table in nexus with sensor ids and names, I think buffy source should break at this level to avoid the difficult to track shape mismatch.

```
    pmt_ids  = load_db.DataPMT(db_file, run_number).SensorID

    for file_name in paths:
        sns_resp = mcinfo_io.load_mcsensor_response_df(file_name              ,
                                                       return_raw = False     ,
                                                       db_file    = db_file   ,
                                                       run_no     = run_number)

        for evt in mcinfo_io.get_event_numbers_in_file(file_name):

            timestamp = create_timestamp(evt, rate) / units.ms

            try:
                ## Assumes two types of sensor, all non pmt
                ## assumed to be sipms. NEW, NEXT100 and DEMOPP safe
                ## Flex with this structure too.
                pmt_indx  = sns_resp.loc[evt].index.isin(pmt_ids)
                pmt_resp  = sns_resp.loc[evt][ pmt_indx]
                sipm_resp = sns_resp.loc[evt][~pmt_indx]
```

paolafer:
This PR adds a protection that raises an error if the nexus file has PMT IDs which are not in the MySQL database.
Closes #779.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -465,13 +465,11 @@ def mcsensors_from_file(paths     : List[str],
                                                        run_no     = run_number)
 
         if not is_oldformat_file(file_name):
-
             nexus_sns_pos = pd.read_hdf(file_name, 'MC/sns_positions')
-            sns_labels    = np.array([name.casefold() for name in nexus_sns_pos.sensor_name.values])
-            pmt_condition = np.argwhere(['Pmt'.casefold() in lbl for lbl in sns_labels]).flatten()
-            nexus_pmt_ids = nexus_sns_pos.sensor_id.values[pmt_condition]
+            pmt_condition = nexus_sns_pos.sensor_name.str.casefold().str.contains('pmt')
+            nexus_pmt_ids = nexus_sns_pos[pmt_condition].sensor_id
 
-            if not all(x in pmt_ids for x in nexus_pmt_ids):
+            if not nexus_pmt_ids.isin(pmt_ids).all():
                 raise SensorIDMismatch('Some PMT IDs in nexus file do not appear in database')
 
 
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -464,15 +464,15 @@ def mcsensors_from_file(paths     : List[str],
                                                        db_file    = db_file   ,
                                                        run_no     = run_number)
 
-        sns_pos_tbl = 'MC/sensor_positions' if is_oldformat_file(file_name) else 'MC/sns_positions'
+        if not is_oldformat_file(file_name):
 
-        nexus_sns_pos = pd.read_hdf(file_name, sns_pos_tbl)
-        sns_labels    = np.array([name.casefold() for name in nexus_sns_pos.sensor_name.values])
-        pmt_condition = np.argwhere(['Pmt'.casefold() in lbl for lbl in sns_labels]).flatten()
-        nexus_pmt_ids = nexus_sns_pos.sensor_id.values[pmt_condition]
+            nexus_sns_pos = pd.read_hdf(file_name, 'MC/sns_positions')
+            sns_labels    = np.array([name.casefold() for name in nexus_sns_pos.sensor_name.values])
+            pmt_condition = np.argwhere(['Pmt'.casefold() in lbl for lbl in sns_labels]).flatten()
+            nexus_pmt_ids = nexus_sns_pos.sensor_id.values[pmt_condition]
 
-        if not all(x in pmt_ids for x in nexus_pmt_ids):
-            raise SensorIDMismatch('Some PMT IDs in nexus file do not appear in database')
+            if not all(x in pmt_ids for x in nexus_pmt_ids):
+                raise SensorIDMismatch('Some PMT IDs in nexus file do not appear in database')
 
 
         for evt in mcinfo_io.get_event_numbers_in_file(file_name):
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -38,6 +38,7 @@ from .. core   .exceptions        import           MCEventNotFound
 from .. core   .exceptions        import              NoInputFiles
 from .. core   .exceptions        import              NoOutputFile
 from .. core   .exceptions        import InvalidInputFileStructure
+from .. core   .exceptions        import          SensorIDMismatch
 from .. core   .configure         import          event_range_help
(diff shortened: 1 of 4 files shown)


invisible_cities/cities/components.py
```diff
@@ -451,6 +453,16 @@ def mcsensors_from_file(paths     : List[str],
                                                        return_raw = False     ,
                                                        db_file    = db_file   ,
                                                        run_no     = run_number)
+        sns_pos_tbl = 'MC/sns_positions'
+        if is_oldformat_file(file_name): sns_pos_tbl = 'MC/sensor_positions'
```
```suggestion
        sns_pos_tbl = 'MC/sensor_positions' if is_oldformat_file(file_name) else 'MC/sns_positions'
```
more idiomatic

invisible_cities/cities/components.py
```diff
@@ -451,6 +453,16 @@ def mcsensors_from_file(paths     : List[str],
...
+
+        nexus_sns_pos       = pd.read_hdf(file_name, sns_pos_tbl)
+        nexus_pmt_ids       = nexus_sns_pos[nexus_sns_pos.sensor_id < 100].sensor_id.values
```
gonzaponte: This is dangerous and I believe there is another instance in the code where we rely on a magic number. We should write the sensor type in this table in nexus. For now, we can add a comment that clarifies this number.

invisible_cities/cities/components.py
```diff
@@ -451,6 +453,16 @@ def mcsensors_from_file(paths     : List[str],
...
+        nexus_pmt_arg_in_db = np.argwhere([sid in pmt_ids for sid in nexus_pmt_ids])
+
+        if len(nexus_pmt_ids) != len(nexus_pmt_arg_in_db):
```
gonzaponte: A more concise way to express that every number in a sequence is present in another sequence, using a `pd.Series`, is:
```
nexus_pmt_ids = nexus_sns_pos[nexus_sns_pos.sensor_id < 100].sensor_id
if nexus_pmt_ids.isin(pmt_ids).all():
```

gonzaponte (changes requested):
Some minor comments.

paolafer:
I've rebased my branch on master; now it should be ready to merge, if approved.

invisible_cities/cities/components.py
```diff
@@ -453,6 +455,15 @@ def mcsensors_from_file(paths     : List[str],
                                                        db_file    = db_file   ,
                                                        run_no     = run_number)
 
+        sns_pos_tbl = 'MC/sensor_positions' if is_oldformat_file(file_name) else 'MC/sns_positions'
+
+        nexus_sns_pos       = pd.read_hdf(file_name, sns_pos_tbl)
+        nexus_pmt_ids       = nexus_sns_pos[nexus_sns_pos.sensor_id < 100].sensor_id ## sensors with IDs below 100 are PMTs, the rest are SiPMs
```
```suggestion
        nexus_sns_pos = pd.read_hdf(file_name, sns_pos_tbl)
        nexus_pmt_ids = nexus_sns_pos[nexus_sns_pos.sensor_id < 100].sensor_id ## sensors with IDs below 100 are PMTs, the rest are SiPMs
```
just cosmetics

gonzaponte (commented):
I would like to add a test for this.
Perhaps creating a file on the fly with wrong sensor ids and verify that it raises the expected exception. what do you think?

invisible_cities/cities/components_test.py
```diff
@@ -318,3 +319,14 @@ def test_check_max_time_units():
...
+    """
+    Check if an error is raised when wrong PMT IDs are read from nexus file
+    """
+    file_in    = os.path.join(ICDATADIR, "nexus_next100_full_wrong_PMT_IDs.h5")
+    run_number = 0
+    rate       = 0.5
+
+    with raises(SensorIDMismatch):
+        sns_gen = mcsensors_from_file([file_in], 'next100', run_number, rate)
+        _ = next(sns_gen)
```
gonzaponte:
```suggestion
    sns_gen = mcsensors_from_file([file_in], 'next100', run_number, rate)
    with raises(SensorIDMismatch):
        next(sns_gen)
```
to make it more clear where the exception is raised.

invisible_cities/cities/components_test.py
```diff
@@ -318,3 +319,14 @@ def test_check_max_time_units():
     with raises(ValueError):
         check_max_time(max_time, buffer_length)
 
+def test_read_wrong_pmt_ids(ICDATADIR):
+    """
+    Check if an error is raised when wrong PMT IDs are read from nexus file
```
gonzaponte: Maybe explain that the nexus file contains one or more sensor IDs that are not present in the database, that we expect this to produce an error, and that therefore we check for that.

gonzaponte (commented):
I'm happy with this. Just two minor comments and I'll approve it.

invisible_cities/cities/components_test.py
paolafer: Done.

paolafer:
The code is rebased on master and some commits have been squashed to clear up the history.

gonzaponte (approved):
Fixes a long-standing issue. Great job!

paolafer:
I found some conflicts rebasing on master, so I have applied the commits of this PR on the master branch one by one by hand and pushed again to the same branch. There should not be any more `out-of-date` message now, let me know if you see anything wrong.

Issue #705, opened 2020-03-10
event_count repeated in Phyllis result


jacg:
Is Phyllis being used?

Its result contains 4 values, two of which are the same value repeated under different names: `events_in` and `event_count` are both the result of `event_count.future`:

            result = dict(events_in   = event_count     .future,
                          spe         = accumulate_light.future,
                          dark        = accumulate_dark .future,
                          event_count =      event_count.future)

(Additionally, the second `event_count` is misaligned.)

jacg:
... and exactly the same problem appears it Trude.

In fact, the amount of code duplication between Phyllis and Trude looks quite scary! (It's not immediately obvious what the best way of avoiding this would be, but we can definitely do better.)

carmenromo:
Yes, both of them are being used! They compute the spectra for SiPMs (Trude) and PMTs (Phyllis)

jacg:
Oh dear. Is the `event_count` being used?

gonzaponte:
Done in #873 and #880.

PR #873, opened 2024-05-27
Remove repeated and not-used variable from result

paolafer:
This PR removes an unnecessary variable from two cities, which was repeated.
Closes issue #705.


gonzaponte (approved):
Good!

PR #874, opened 2024-05-29
Remove default binning

### LINKED ISSUE #857: Remove default binning in `Buffy`
This line is extremely dangerous as it uses a default value that the user surely is not aware of and probably doesn't understand. This has caused troubles in the past. I believe it doesn't cause them anymore, but it's better to be safe than sorry.

https://github.com/next-exp/IC/blob/86e39c483455e07e644931824294674dec78a0d9/invisible_cities/cities/buffy.py#L165

paolafer:
This PR removes the assignment of a default time binning to sensors, when no sensor response is found in the bunch of files being processed. The city will crash, instead.
Also, a test is modified to account for this change and another one is added.

Closes #857

### DIFF
--- a/invisible_cities/cities/buffy.py
+++ b/invisible_cities/cities/buffy.py
@@ -23,6 +23,8 @@ import warnings
 from typing import Callable
 from typing import     List
 
+from argparse import Namespace
+
 from .. core                   import        system_of_units as units
 from .. core  .configure       import         EventRangeType
 from .. core  .configure       import         OneOrManyFiles
@@ -65,9 +67,9 @@ def buffy( files_in          : OneOrManyFiles
     try:
         pmt_wid, sipm_wid = pmt_and_sipm_bin_width_safe_(files_in)
     except tb.exceptions.NoSuchNodeError:
-        return dict(events_in   = 0,
-                    events_resp = 0,
-                    evtnum_list = [])
+        return Namespace(events_in   = 0,
+                         events_resp = 0,
+                         evtnum_list = [])
 
     nsamp_pmt         = int(buffer_length /  pmt_wid)
     nsamp_sipm        = int(buffer_length / sipm_wid)
--- a/invisible_cities/cities/buffy.py
+++ b/invisible_cities/cities/buffy.py
@@ -62,7 +62,13 @@ def buffy( files_in          : OneOrManyFiles
 
     max_time          = check_max_time(max_time, buffer_length)
     npmt, nsipm       = get_n_sensors(detector_db, run_number)
-    pmt_wid, sipm_wid = pmt_and_sipm_bin_width_safe_(files_in)
+    try:
+        pmt_wid, sipm_wid = pmt_and_sipm_bin_width_safe_(files_in)
+    except tb.exceptions.NoSuchNodeError:
+        return dict(events_in   = 0,
+                    events_resp = 0,
+                    evtnum_list = [])
+
     nsamp_pmt         = int(buffer_length /  pmt_wid)
     nsamp_sipm        = int(buffer_length / sipm_wid)
 
@@ -162,4 +168,5 @@ def pmt_and_sipm_bin_width_safe_(files_in: List[str]):
         except (tb.HDF5ExtError, tb.NoSuchNodeError) as e:
             warnings.warn(f' no useful bin widths: {0}'.format(e), UserWarning)
             continue
-    return 25 * units.ns, 1 * units.mus
+
+    raise tb.NoSuchNodeError
--- a/invisible_cities/cities/buffy_test.py
+++ b/invisible_cities/cities/buffy_test.py
@@ -125,7 +125,7 @@ def test_buffy_filters_empty(config_tmpdir, ICDATADIR):
 
         evt_filter = h5out.root.Filters.detected_events.read()
         assert len(evt_filter) == nevt
-        assert len(evt_filter[evt_filter['passed'] == True]) == 2
+        assert np.count_nonzero(evt_filter['passed']) == 2
 
 
 @mark.parametrize("fn_first fn_second".split(),
--- a/invisible_cities/cities/buffy_test.py
+++ b/invisible_cities/cities/buffy_test.py
@@ -109,11 +109,14 @@ def test_buffy_filters_empty(config_tmpdir, ICDATADIR):
     assert buffy_result.events_resp.n_passed == n_passed
     with tb.open_file(file_out) as h5out:
 
+        assert h5out.root.Run.events.shape == (2,)
+
         assert hasattr(h5out.root        ,         'Filters')
         assert hasattr(h5out.root.Filters, 'detected_events')
 
         evt_filter = h5out.root.Filters.detected_events.read()
         assert len(evt_filter) == nevt
+        assert len(evt_filter[evt_filter['passed'] == True]) == 2
 
 
 def test_buffy_exact_result(config_tmpdir, ICDATADIR):
--- a/invisible_cities/cities/buffy_test.py
+++ b/invisible_cities/cities/buffy_test.py
@@ -105,7 +105,6 @@ def test_buffy_filters_empty(config_tmpdir, ICDATADIR):
         warnings.simplefilter("ignore", category=UserWarning)
         buffy_result = buffy(**conf)
 
-    print(buffy_result.events_in, buffy_result.events_resp.n_passed)
     assert buffy_result.events_in            == nevt
     assert buffy_result.events_resp.n_passed == n_passed
     with tb.open_file(file_out) as h5out:
--- a/invisible_cities/cities/buffy_test.py
+++ b/invisible_cities/cities/buffy_test.py
@@ -86,13 +86,10 @@ def test_buffy_no_file_without_sns_response(config_tmpdir, ICDATADIR):
     conf = configure('buffy invisible_cities/config/buffy.conf'.split())
     conf.update(dict(files_in=file_in, file_out=file_out, event_range=nevt))
 
-    try:
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=UserWarning)
         buffy(**conf)
-    except:
-        if not os.path.exists(file_out):
-            assert True
-        else:
-            assert False
+    assert not os.path.exists(file_out)
 
 
 def test_buffy_filters_empty(config_tmpdir, ICDATADIR):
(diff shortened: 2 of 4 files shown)


invisible_cities/cities/buffy.py
```diff
@@ -62,7 +62,13 @@ def buffy( files_in          : OneOrManyFiles
 
     max_time          = check_max_time(max_time, buffer_length)
     npmt, nsipm       = get_n_sensors(detector_db, run_number)
-    pmt_wid, sipm_wid = pmt_and_sipm_bin_width_safe_(files_in)
+    try:
+        pmt_wid, sipm_wid = pmt_and_sipm_bin_width_safe_(files_in)
+    except tb.exceptions.NoSuchNodeError:
+        return dict(events_in   = 0,
+                    events_resp = 0,
+                    evtnum_list = [])
```
gonzaponte: In order to match the output type replace `dict` with `Namespace` (`from argparse import Namespace`)

invisible_cities/cities/buffy_test.py
```diff
@@ -72,31 +72,48 @@ def test_buffy_kr(config_tmpdir, full_sim_file):
         assert np.all(sipm_integ[sipm_nonzero] == sns_sums[sns_sums.index > 12])
 
 
-def test_buffy_filters_empty(config_tmpdir, ICDATADIR):
+def test_buffy_no_file_without_sns_response(config_tmpdir, ICDATADIR):
     file_in  = os.path.join(ICDATADIR, 'nexus_new_kr83m_fast.oldformat.sim.h5')
-    file_out = os.path.join(config_tmpdir, 'filtered_empty.buffers.h5')
+    file_out = os.path.join(config_tmpdir, 'no_file.h5')
```
gonzaponte: I'm trying to set the rule that the output file of a test should be named like the test for practical reasons, so it would be great if you rename the output file accordingly :)

invisible_cities/cities/buffy_test.py
```diff
@@ -72,31 +72,48 @@ def test_buffy_kr(config_tmpdir, full_sim_file):
...
     conf.update(dict(files_in=file_in, file_out=file_out, event_range=nevt))
 
-    # Warning expected since no MC tables present.
+    try:
+        buffy(**conf)
+    except:
+        if not os.path.exists(file_out):
+            assert True
+        else:
+            assert False
```
gonzaponte: The correct way to test that some code raises an exception is with pytest's `raises`:
```
with raises(SomeException, match="a regex that matches the exception message"):
    the code that raises an exception
```

You can then check that the output file does not exist simply with `assert not os.path.exists(file_out)`

gonzaponte:
Just so we don't forget, there is still a problem if the first file does not contain sensor responses. If we don't fix it here, we will open an issue.

invisible_cities/cities/buffy_test.py
```diff
@@ -72,31 +72,48 @@ def test_buffy_kr(config_tmpdir, full_sim_file):
...
+
+
+def test_buffy_filters_empty(config_tmpdir, ICDATADIR):
+    file_in  = os.path.join(ICDATADIR, 'nexus_new_kr83m_*.oldformat.sim.h5')
```
gonzaponte: when #877 is merged you should be able to specify the input files without a wildcard, which will make this more clear.

Regarding the treatment of files with missing sensor response, I suggest we add a parameterized test with both file orders (the offending file first or last) and mark the order that doesn't work as `xfail` until we fix it.

Remember to add a docstring documenting what the test does and to comment that we use the fast file to mimic a full file without sensor responses.

Remove the commented line below.

The variable should be renames to `files_in` (with an s)

invisible_cities/cities/buffy_test.py
```diff
@@ -72,31 +72,48 @@ def test_buffy_kr(config_tmpdir, full_sim_file):
...
+    assert buffy_result.events_resp.n_passed == n_passed
     with tb.open_file(file_out) as h5out:
-        assert h5out.root.Run.events.shape == (0,)
 
         assert hasattr(h5out.root        ,         'Filters')
         assert hasattr(h5out.root.Filters, 'detected_events')
 
         evt_filter = h5out.root.Filters.detected_events.read()
         assert len(evt_filter) == nevt
-        assert not np.any(evt_filter['passed'])
```
gonzaponte: Instead of removing these lines, can we put the expected values?

invisible_cities/cities/buffy_test.py
```diff
@@ -72,31 +72,48 @@ def test_buffy_kr(config_tmpdir, full_sim_file):
...
+    file_in  = os.path.join(ICDATADIR, 'nexus_new_kr83m_*.oldformat.sim.h5')
+    #file_in  = os.path.join(ICDATADIR, 'nexus_new_kr83m_full.oldformat.sim.h5')
+    file_out = os.path.join(config_tmpdir, 'filtered_empty.buffers.h5')
+
+    nevt     = 4
+    n_passed = 2
+    conf = configure('buffy invisible_cities/config/buffy.conf'.split())
+    conf.update(dict(files_in=file_in, file_out=file_out, event_range=nevt))
+
+    # Exception expected since no MC sensor response is present.
```
gonzaponte: no MC sensor response is present ... in one of the files?

gonzaponte (changes requested):
Some small comments

invisible_cities/cities/buffy.py
paolafer: Ok, what does that mean?

invisible_cities/cities/buffy_test.py
paolafer: Ok, it would be useful to have all these style rules written down somewhere (maybe they are already and I'm not aware).

invisible_cities/cities/buffy_test.py
paolafer: For some reason, the test is failing claiming that no such file exist... it's like it's not catching the exception with `raises`.

invisible_cities/cities/buffy_test.py
paolafer: Done.

invisible_cities/cities/buffy_test.py
paolafer: Done.

invisible_cities/cities/buffy.py
gonzaponte: All cities return a `Namespace` (which you can import from `argparse`). Here you are returning a `dict`, which breaks that pattern. I suggest returning a `Namespace`, for which you only need to replace `dict` -> `Namespace`, because they have a similar constructor.

invisible_cities/cities/buffy_test.py
gonzaponte: Yes, indeed, but we have never managed to do it. I will open an issue.

invisible_cities/cities/buffy_test.py
gonzaponte: Wait, I didn't stop to think before. Buffy should not raise an error, right? The error comes from a different place.

All cities have automatic indexing (sort of) of the tables in the output file. In this case, the output file is not generated, so the function that attempts to index the tables complains that there is no file. Please fix the `city` function in components.py to call `index_tables` only if the output file exists.

paolafer: Right, we return an empty `Namespace`. Where exactly is this `index_tables` function called?

gonzaponte: In the function `city` defined in components.py.

invisible_cities/cities/buffy_test.py
```diff
@@ -72,31 +74,55 @@ def test_buffy_kr(config_tmpdir, full_sim_file):
...
+
+        assert h5out.root.Run.events.shape == (2,)
 
         assert hasattr(h5out.root        ,         'Filters')
         assert hasattr(h5out.root.Filters, 'detected_events')
 
         evt_filter = h5out.root.Filters.detected_events.read()
         assert len(evt_filter) == nevt
-        assert not np.any(evt_filter['passed'])
+        assert len(evt_filter[evt_filter['passed'] == True]) == 2
```
gonzaponte:
```suggestion
        assert np.count_nonzero(evt_filter['passed']) == 2
```

invisible_cities/cities/components.py
```diff
@@ -179,6 +179,8 @@ def index_tables(file_out):
     -checks if any columns in the tables have been marked to be indexed by writers
     -indexes those columns
     """
+    if not os.path.exists(file_out):
+        return
```
gonzaponte: Please move this check to the function `city`. This function shouldn't exit silently if the input is missing. It's in the cities where we want to handle that case.

invisible_cities/cities/buffy_test.py
```diff
@@ -72,31 +74,55 @@ def test_buffy_kr(config_tmpdir, full_sim_file):
...
     nevt = 2
     conf = configure('buffy invisible_cities/config/buffy.conf'.split())
     conf.update(dict(files_in=file_in, file_out=file_out, event_range=nevt))
 
-    # Warning expected since no MC tables present.
-    # Suppress since irrelevant in test.
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=UserWarning)
+        buffy(**conf)
+    assert not os.path.exists(file_out)
```
gonzaponte: I guess you can also check that the output of the city is correct. `events_in == 0`, etc.

paolafer:
I've rebased on the new master. The PR is now ready to be reviewed one last time.

invisible_cities/cities/components.py
paolafer: How do we want to handle this case? Do we want an exception to be raised?

invisible_cities/cities/buffy_test.py
paolafer: Mmm, in fact it seems that the output of the city is not correct: it says it's of `NoneType`.

invisible_cities/cities/components.py
gonzaponte: Here we don't. If the file doens't exist, the function should raise an exception. In `city` we know that there might not be an output file and so we should call `index_tables` only if the file exists.

invisible_cities/cities/buffy_test.py
paolafer: I've checked that the code enters in the exception of line 69 in `buffy`, so I don't know why a `None` type is returned.

gonzaponte: that is weird. Can you try to debug it and see why this is happening?

paolafer: Solved.

gonzaponte (approved):
Excellent. We remove one of the many instances of dangerous default values in our code. We now handle the case with missing values explicitly, but we still have to address the case where the first file does not contain data, for which we will open an issue. The code looks good and is properly tested. Good job!

PR #877, opened 2024-06-01
Generalize city input files

gonzaponte:
So far we could only deal with either a single file or a wildcard. This PR extends it to accept lists of files and lists of wildcards.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -117,7 +117,12 @@ def city(city_function):
             conf.files_in = [conf.files_in]
 
         globbed_files = map(glob, map(expandvars, conf.files_in))
-        conf.files_in = sorted(f for fs in globbed_files for f in fs)
+        globbed_files = sorted(f for fs in globbed_files for f in fs)
+        if len(set(globbed_files)) != len(globbed_files):
+            warnings.warn("files_in contains repeated values. Ignoring duplicate files.", UserWarning)
+            globbed_files = sorted(set(globbed_files))
+
+        conf.files_in = globbed_files
         conf.file_out = expandvars(conf.file_out)
 
         conf.event_range  = event_range(conf)
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -112,9 +112,13 @@ def city(city_function):
         if 'files_in' not in kwds: raise NoInputFiles
         if 'file_out' not in kwds: raise NoOutputFile
 
+        # always a sequence, so we can generalize the code below
+        if isinstance(conf.files_in, str):
+            conf.files_in = [conf.files_in]
 
-        conf.files_in  = sorted(glob(expandvars(conf.files_in)))
-        conf.file_out  =             expandvars(conf.file_out)
+        globbed_files = map(glob, map(expandvars, conf.files_in))
+        conf.files_in = sorted(f for fs in globbed_files for f in fs)
+        conf.file_out = expandvars(conf.file_out)
 
         conf.event_range  = event_range(conf)
         # TODO There were deamons! self.daemons = tuple(map(summon_daemon, kwds.get('daemons', [])))
--- a/invisible_cities/cities/components_test.py
+++ b/invisible_cities/cities/components_test.py
@@ -87,17 +87,17 @@ def write_config_file(filename, **kwargs):
 
 
 @mark.parametrize( "case_ files_in expected".split()
-                 , ( ( 0
-                     , "electrons_511keV_z250_RWF.h5"  # a single file without wildcard
+                 , ( ( 0 # a single file without wildcard
+                     , "electrons_511keV_z250_RWF.h5"
                      , "electrons_511keV_z250_RWF.h5")
-                   , ( 1
-                     , "electrons_*keV_z250_RWF.h5"   # a single wildcard leading to three files
+                   , ( 1 # a single wildcard leading to three files
+                     , "electrons_*keV_z250_RWF.h5"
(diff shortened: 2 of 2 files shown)


jwaiton:
Looks good! The code allows for multiple input files to be introduced as is demonstrated within the test, which passes on my system.

I didn't find it intuitive to alter a config file such that it allowed for a list of files to be processed at once, perhaps I am expecting behaviour that isn't intended in this PR. Is the intended behaviour that multiple h5 input files can be piped into a singular output file? If this is the intended behaviour, it may be helpful to include some sort of formatting help in the template config files to explain how to implement this.

I also tested an edge case in which someone inserts a list including both a wildcard and a specific file (which should result in an overlap of the input files), and added it [here](https://github.com/next-exp/IC/pull/877/files#diff-79c16ea78db6a7806308d8f3865a58cbbcd296763e68133b6445ee71bc4ffba9R89) as such:

```
@mark.parametrize( "case_ files_in expected".split()
                 , ( ( 0
                     , "electrons_511keV_z250_RWF.h5"  # a single file without wildcard
                     , "electrons_511keV_z250_RWF.h5")
                   , ( 1
                     , "electrons_*keV_z250_RWF.h5"   # a single wildcard leading to three files
                     , ["electrons_511keV_z250_RWF.h5" , "electrons_1250keV_z250_RWF.h5", "electrons_2500keV_z250_RWF.h5"])
                   , ( 2
                     , [ "electrons_511keV_z250_RWF.h5", "electrons_1250keV_z250_RWF.h5"] # a list of two files without wildcards
                     , [ "electrons_511keV_z250_RWF.h5", "electrons_1250keV_z250_RWF.h5"] )
                   , ( 3
                     , [ "electrons_5*keV_z250_RWF.h5", "electrons_1*keV_z250_RWF.h5"] # a list of wildcards leading to n files
                     , [ "electrons_511keV_z250_RWF.h5", "electrons_1250keV_z250_RWF.h5"] )
                    , (4
                     , [ "electrons_*keV_z250_RWF.h5", "electrons_1250keV_z250_RWF.h5"] # a list with an overlapping wildcard leading to n files
                     , ["electrons_511keV_z250_RWF.h5" , "electrons_1250keV_z250_RWF.h5", "electrons_2500keV_z250_RWF.h5"] )
                   )
                 )
```

This returned the error:

```
E       AssertionError: assert ['/home/e7836..._z250_RWF.h5'] == ['/home/e7836..._z250_RWF.h5']
E         At index 1 diff: '/home/e78368jw/Documents/NEXT_CODE/IC/invisible_cities/database/test_data/electrons_1250keV_z250_RWF.h5' != '/home/e78368jw/Documents/NEXT_CODE/IC/invisible_cities/database/test_data/electrons_2500keV_z250_RWF.h5'
E         Left contains one more item: '/home/e78368jw/Documents/NEXT_CODE/IC/invisible_cities/database/test_data/electrons_511keV_z250_RWF.h5'
E         Use -v to get the full diff

invisible_cities/cities/components_test.py:144: AssertionError

```
This test fails I believe due to the output of the test having two instances of `electrons_1250keV_z250_RWF.h5`. This is an unlikely edge case that I would consider only occurring due to user error, and the consequence I can see is that perhaps a file is processed twice within a single output, which may have unforeseen issues in certain cities. 

Would it be worth adding a check to ensure that if duplicate input files are found, the duplicates are removed?

gonzaponte:
> I didn't find it intuitive to alter a config file such that it allowed for a list of files to be processed at once, perhaps I am expecting behaviour that isn't intended in this PR

It is very much intended, it should be as easy as `files_in = [file_1, file_2, ...]` and I'm surprised that it didn't work. A bit silly of me not to have tried that myself. This calls for another test, which I will implement.

> This test fails I believe due to the output of the test having two instances of `electrons_1250keV_z250_RWF.h5`.

Indeed. Adding this case is an excellent idea.

> perhaps a file is processed twice within a single output, which may have unforeseen issues in certain cities.
> Would it be worth adding a check to ensure that if duplicate input files are found, the duplicates are removed?

It is a weird case and I'm unsure how much effort to put into it. I'm tempted to raise a warning and carry on with the set of unique files. It's hard to imagine a sensible behaviour that doesn't involve ignoring the duplicate file.

jwaiton:
> It is very much intended, it should be as easy as `files_in = [file_1, file_2, ...]` and I'm surprised that it didn't work.

I believe this was just a silly issue with formatting on my part. The correct format is `files_in = ['file_1', 'file_2', ...]` which works as expected. This didn't work for me initially (due to a typo in retrospect I think) so I tried all other formatting possibilities to no avail. 

It may be useful to have the format for `files_in` in an example config file somewhere, but I think that may be overkill.

> It is a weird case and I'm unsure how much effort to put into it. I'm tempted to raise a warning and carry on with the set of unique files. It's hard to imagine a sensible behaviour that doesn't involve ignoring the duplicate file.

I think that would be perfect :)

gonzaponte:
>  The correct format is `files_in = ['file_1', 'file_2', ...]` which works as expected

Indeed. Config files are just python files with some magic applied behind the scenes. Filenames must be quoted to be interpreted as strings. I've added a comment that hopefully clarifies the syntax. Let me know what you think.
We will have to add this behaviour to the documentation and clarify the syntax there as well.

> I'm surprised that it didn't work. A bit silly of me not to have tried that myself. This calls for another test, which I will implement.

My memory failed me. I did not remember that the test writes config files already, so this is covered.

jwaiton:
> I've added a comment that hopefully clarifies the syntax. Let me know what you think.

The comment is perfect. :)

I've rerun the tests with the new case, everything passed. This change allows for lists of input files and lists of wildcards to be used rather than singular files and wildcards. Good work!

jwaiton (approved):
This change allows for lists of input files and lists of wildcards to be used rather than singular files and wildcards. Good work!

PR #872, opened 2024-05-14
Add fit functions for map computation

carhc:
This PR provides the fundamental elements of the map creation, which are the fit functions. It contains the modification of IC fit function, the basic fit functions (and seeds) needed for the map computation and it also defines the current available fits. Other functions to unify input/output are included. 

This is done in a different PR to allow for modularity. This functions are useful for #860, #861, #862...

PR #878, opened 2024-06-01
Update scipy to 1.9.3

gonzaponte:
This is needed for #872. Tests pass, so we assume the version change doesn't change anything significant.


carhc (approved):
This new environment, making use of helpful updates in both numpy and scipy will allow for some new functionalities in IC. Approved!

PR #880, opened 2024-06-05
Remove unused variable

paolafer:
This PR removes an unused variable from the result of Berenice city, as PR #873 did for Trude and Phyllis.


gonzaponte (approved):
:partying_face:

PR #881, opened 2024-06-11
Retain order of input files

gonzaponte:
An amend to #877. We don't want to reorder the input as it might be relevant to keep the order. For sure we want to have the ability to specify the order in testing to catch edge cases.

### DIFF
--- a/invisible_cities/cities/components_test.py
+++ b/invisible_cities/cities/components_test.py
@@ -152,6 +152,36 @@ def test_city_files_in(case_, files_in, expected, config_tmpdir, ICDATADIR):
     assert sorted(result) == sorted(expected)
 
 
+@mark.parametrize("order", ((0,1), (1,0)))
+def test_city_keeps_input_file_ordering(ICDATADIR, config_tmpdir, order):
+    files_in = [ os.path.join(ICDATADIR, "electrons_511keV_z250_RWF.h5")
+               , os.path.join(ICDATADIR, "electrons_1250keV_z250_RWF.h5") ]
+    files_in = [files_in[i] for i in order]
+
+    @city
+    def dummy_city( files_in    : Union[str, list]
+                  , file_out    : str
+                  , event_range : tuple):
+        with tb.open_file(file_out, "w"): pass
+        return files_in
+
+    config_file = os.path.join(config_tmpdir, f"test_city_keeps_input_file_ordering.conf")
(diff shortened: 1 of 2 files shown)


jwaiton:
Good catch! I hadn't considered that to be an issue. I can't think of any particular conditions that changing the ordering would cause problems, but it's better to be safe rather than sorry. Is there a reason they were sorted ([L123](https://github.com/next-exp/IC/pull/881/files#diff-4241aad994b566cd332688ff48689b507388a77a911e9c9c3a87d73f4b91d99dL123)) in the initial format?

The tests pass locally, and reading through the changes it makes sense. So as long as they pass the tests on github I'm happy to approve them.

gonzaponte:
> Is there a reason they were sorted ([L123](https://github.com/next-exp/IC/pull/881/files#diff-4241aad994b566cd332688ff48689b507388a77a911e9c9c3a87d73f4b91d99dL123)) in the initial format?

Not really. It's just a habit.

invisible_cities/cities/components_test.py
```diff
@@ -152,6 +152,36 @@ def dummy_city( files_in    : Union[str, list]
     assert sorted(result) == sorted(expected)
 
 
+@mark.parametrize("order", ((0,1), (1,0)))
+def test_city_keeps_input_file_ordering(ICDATADIR, config_tmpdir, order):
+    files_in = [ os.path.join(ICDATADIR, "electrons_511keV_z250_RWF.h5")
+               , os.path.join(ICDATADIR, "electrons_1250keV_z250_RWF.h5") ]
+    files_in = [files_in[i] for i in order]
```
jwaiton, line 159: Just so I understand exactly how this works, `order` sets it such that `files_in` converted into a list of two lists, with the order of the files flipped. Is that correct?

gonzaponte: Not a list of two lists, but a list of strings. This produces two tests. In one of them, `order == (0,1)`, and keep the original order. In the other one `order == (1,0)` and the order is reversed. In both cases, `files_in` is a list with two strings.

jwaiton: Sorry yes, I was thinking of each string as a list (of characters). Okay that makes sense, thanks!

jwaiton (approved):
This PR introduces a change and a new test that ensures input files retain their order.

PR #882, opened 2024-06-11
Update NEXT100 DB

gonzaponte:
Contains the updated blr coefficients and calibration constants.


paolafer (approved):
The new database has been downloaded and read, giving the expected non-zero values for the `coeff_blr`, `coeff_c` and `adc_to_pes` variables. Approved.

PR #820, opened 2022-06-03
Update event ID

paolafer:
This PR changes the `event_ID` field of the IC table from `int32` to `int64` to avoid overflow in the case of large background productions.

PR #884, opened 2024-06-27
Change the last remaining event types to int64

### LINKED ISSUE #838: `hits_from_df` forces event numbers to be int32
The implementation of this function forces the event number to be a 32-bit signed integer. This creates a problem when processing backgrounds for NEXT100, where event numbers need to be represented as 64-bit integers.
I do not remember the reason for this choice and the comment in the code doesn't explain it fully.
We need to assess whether there is a real need for this restriction or if it's just a ghost from the past.
Initial checks by @martinperezmaneiro indicate that it is no longer necessary.

paolafer:
This PR changes the last items of `int32` types assigned to the event ID. Since PR #820, all event IDs should be `int64`, to deal with large background productions. These items probably escaped our attention at the time of that PR.
This PR closes #838.


gonzaponte (approved):
Good job!

PR #887, opened 2024-07-23
Reorganize IC

### LINKED ISSUE #501: Reorganization of folders

### LINKED ISSUE #885: Reorganization of the reco folder
I would like to complete the job done in PR #707 but never finished, meaning, reorganizing the `reco` folder.
At the moment this folder contains very different functions and some of them are not related to reconstruction, from my point of view.

My proposal is the following: 

1. `sensor_functions` and `spe_response` are part of the MC simulation of the events, rather than the reconstruction, therefore they should go to a dedicated folder (maybe a new `mcsim`?). 
2. `tbl_functions` are very generic functions to deal with tables, so they could go to `core`.
3. I'm not sure about `wvf_functions`; they deal with waveforms, so they are part of the reconstruction in a way.
4. Same about the two calib files. Calibration can be considered part of the reconstruction, so they can stay here.

Opinions?

paolafer:
This PR completes PR #707, reorganizing the `reco` folder.

Closes #501
Closes #885


gonzaponte (approved):
Looks good to me, good job!

PR #891, opened 2024-09-20
Store configuration parameters

### LINKED ISSUE #837: Store cities' arguments in the output file
We want to keep the parameters used in each city and propagate them as we go through the chain. This way we always know what configuration was used without doubts or having to rely on other files.
This is done in nexus already, which stores the macros under `/MC/configuration`. I propose doing the same and keeping them under `/some_descriptive_group_name/city_name`.

gonzaponte:
This PR stores the city configuration in the output file for bookkeeping purposes. It also propagates the configuration from previous cities further down the chain.

Closes #837

### DIFF
--- a/invisible_cities/cities/cities_test.py
+++ b/invisible_cities/cities/cities_test.py
@@ -107,6 +107,36 @@ def test_city_output_contains_configuration(config_tmpdir, city):
         assert len(table[:]) > 0
 
 
+def test_cities_carry_on_configuration_from_previous_ones(ICDATADIR, config_tmpdir):
+    rwf_file   = os.path.join(    ICDATADIR, "electrons_40keV_ACTIVE_10evts_RWF.h5")
+    pmaps_file = os.path.join(config_tmpdir, f"test_cities_carry_on_configuration_pmaps.h5")
+    kdst_file  = os.path.join(config_tmpdir, f"test_cities_carry_on_configuration_kdst.h5")
+
+    conf_irene = configure("dummy invisible_cities/config/irene.conf".split())
+    conf_irene.update(dict( files_in    = rwf_file
+                          , file_out    = pmaps_file
+                          , event_range = 0))
+
+    conf_dorothea = configure("dummy invisible_cities/config/dorothea.conf".split())
+    conf_dorothea.update(dict( files_in = pmaps_file
+                             , file_out = kdst_file
+                             , event_range = 0))
+
+    for city, conf in (("irene", conf_irene), ("dorothea", conf_dorothea)):
+        module_name   = f'invisible_cities.cities.{city}'
+        city_function = getattr(import_module(module_name), city)
+        city_function(**conf)
+
+    # now check that the second output file contains the configuration from both cities
+    with tb.open_file(kdst_file) as file:
+        assert "config" in file.root
+
+        for city in "irene dorothea".split():
+            assert city in file.root.config
+            table = getattr(file.root.config, city)
+            assert len(table[:]) > 0
+
+
 @mark.filterwarnings("ignore::UserWarning")
 @mark.parametrize("city", all_cities)
 def test_city_missing_detector_db(city):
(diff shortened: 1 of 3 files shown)


gonzaponte:
This PR needs some others merged first.

invisible_cities/cities/components.py
```diff
@@ -198,6 +201,30 @@ def index_tables(file_out):
...
+                            , city_name: str
+                            , args     : dict):
+    args = {k: str(v) for k,v in args.items()}
+
+    with tb.open_file(filename, "a") as file:
+        df = (pd.Series(args)
+                .to_frame()
+                .reset_index()
+                .rename(columns={"index": "variable", 0: "value"}))
+        df_writer(file, df, "config", city_name, f"configuration for {city_name}", str_col_length=300)
```
bpalmeiro, line 219: Maybe just out of curiosity, why the 300 and why hardcoded?

invisible_cities/cities/components.py
```diff
@@ -198,6 +201,30 @@ def index_tables(file_out):
...
+
+
+def copy_cities_configuration( file_in : str, file_out : str):
+    with tb.open_file(file_in, "r") as fin:
+        if "config" not in fin.root: return
```
bpalmeiro: Why a silent return? Isn't it better any sort of warning?

invisible_cities/cities/cities_test.py
```diff
@@ -60,6 +60,28 @@ def test_city_output_file_is_compressed(config_tmpdir, ICDATADIR, city):
...
+    module_name   = f'invisible_cities.cities.{city}'
+    city_function = getattr(import_module(module_name), city)
+
+    city_function(**conf)
+
+    with tb.open_file(file_out) as file:
+        assert "config" in file.root
+        assert city     in file.root.config
+        table = getattr(file.root.config, city)
+        assert len(table[:]) > 0
```
bpalmeiro, line 107: I guess you don't compare the actual values for the specific test regarding the function, right?

invisible_cities/cities/cities_test.py
```diff
@@ -60,6 +60,28 @@ def test_city_output_file_is_compressed(config_tmpdir, ICDATADIR, city):
...
+
```
bpalmeiro, line 108: I really like this test and just wondering if it is easy to expand it for testing the copy. Maybe too messy to be worth it.

invisible_cities/cities/components.py
bpalmeiro: This behavior's worth testing anyway.

bpalmeiro (commented):
Long-awaited enhancement finally coming to place! Nice and simple changes. Just a few comments/curiosities

invisible_cities/cities/components.py
gonzaponte: It is just long enough for our use case. There is no better reason than that, and there is no "right" value that can guarantee all potential values will fit. It's also the same value used in nexus to store the configuration, so it seemed appropriate.

invisible_cities/cities/components.py
gonzaponte: If there is no config group, there is nothing to copy. Mostly useful for backward-compatibility reasons.

Is your proposal adding a test that checks that if no config is present nothing is copied?

invisible_cities/cities/cities_test.py
gonzaponte: I don't understand what you mean here

invisible_cities/cities/cities_test.py
gonzaponte: Hmm, expanding it would be tricky because the input does not contain the config group, but I can make a separate test for that, at least for a single case (two consecutive cities).

gonzaponte:
> I really like this test and just wondering if it is easy to expand it for testing the copy

I've added a test that ensures that the configuration is propagated for two consecutive cities. Doing it in general is a bit 

> Why a silent return? Isn't it better any sort of warning?
> This behavior's worth testing anyway.

I've added a warning and the corresponding test

bpalmeiro (approved):
Just to repeat myself, long-awaited enhancement finally coming to place! All the suggestions were taken care of and the code is neat and clear. Great job!

PR #892, opened 2024-09-21
Fix type hint in esmeralda and beersheba

gonzaponte:
The corrections in esmeralda and beersheba are optional. The code assumed that `corrections_file = None` meant no corrections, but the type hints in the function signature did not allow that.

Note that we don't want to use default values in the cities' signatures to force the user to understand what they are doing. In this case, the optional status of the arguments related to energy corrections is reflected by the possibility of choosing `None` as a value.

### DIFF
--- a/invisible_cities/cities/beersheba.py
+++ b/invisible_cities/cities/beersheba.py
@@ -45,10 +45,6 @@ from os   .path  import expandvars
 from scipy.stats import multivariate_normal
 from numpy       import nan_to_num
 
-from typing import Tuple
-from typing import List
-from typing import Optional
-
 from .  components import city
 from .  components import collect
 from .  components import copy_mc_info
@@ -91,6 +87,9 @@ from .. types.symbols          import DeconvolutionMode
 
 from .. core                   import system_of_units as units
 
+from typing import Tuple
+from typing import List
+from typing import Optional
 from typing import Union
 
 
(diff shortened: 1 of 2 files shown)


jwaiton:
if `corrections_file` is optional, is it expected that the user can leave them out of the config and expect it to run with the assumption that `corrections_file = None`? If so, this currently doesn't work.

gonzaponte:
No, this is what I tried to explain above. The process of applying corrections is optional, but the arguments related to that are not. We want to force the user to think about what is happening with their config file. If they wish to apply corrections, they should specify a valid path for the map file, if they don't want them, they should provide `None`. But they *must* explicitly set it to something, encouraging them to understand what they are doing.

jwaiton:
Apologies, you explained it well I just misread that section.

invisible_cities/cities/beersheba.py
```diff
@@ -83,13 +83,16 @@
...
 
+from .. types.ic_types         import NoneType
 from .. types.symbols          import HitEnergy
 from .. types.symbols          import InterpolationMethod
 from .. types.symbols          import CutType
 from .. types.symbols          import DeconvolutionMode
 
 from .. core                   import system_of_units as units
 
+from typing import Union
```
jwaiton, line 93: Can this line be moved up to be with the other `typing` imports?

gonzaponte: Oops, yes, I didn't see them. They are usually at the end of the import block, I will move the other ones down as it is our convention.

jwaiton (approved):
PR allows for `None` to be given for `corrections_file` and `apply_temp`, but requires input in config files to ensure users understand what they're doing. Works as intended and passes all tests. Good job! :smile:

PR #893, opened 2024-09-21
Add protection for no input files

gonzaponte:
Currently, there is a protection against not specifying `files_in`. However, if we do set this variable, but it does not correspond to a valid file or files through globbing, the city will silently finish causing confusion. This PR fixes that.


jwaiton:
A good PR that passes all tests and works as intended :)

One oddity I noticed is that if you provide two input files, one non-existent and one good, it will not throw up any complaints about the first (or second, ordering doesnt matter) missing file. Should this be filed under a new/separate issue?

gonzaponte:
Good catch. No, it should at least let you know. I will add a test and fix it.

jwaiton (approved):
This change adds protections for the case in which malformed input files are provided to the city, which previously caused the city to stop silently with no information provided. Good job!

Issue #876, opened 2024-06-01
Revisit skipped and xfailed tests


gonzaponte:
They have been there for a while. It's time to either fix them or get rid of them.

PR #894, opened 2024-09-27
Add kr evolution functions 

### LINKED ISSUE #864: ICAROS: parameter evolution
compute the evolution in time of the main parameters: E0, electron life-time, drift-velocity, S1, S2, and others

Input
- filtered Kr dst

Tasks:
- [ ] Split dst in time intervals
- [ ] Compute S1, S2, and other average values in time intervals
- [ ] Compute E0, LT in a selected central region in time intervals
- [ ] Compute drift-velocity in time intervals

Output:
- a *DataFrame* with the parameters information in time intervals

carhc:
This PR provides the functions for the parameter evolution calculation. It should be the last one of the related PR (https://github.com/next-exp/IC/pull/872, https://github.com/next-exp/IC/pull/876...) to be approved since it sits on top of them. Test Suit will fail at the beggining. Function tests to be commited in the following days.

### DIFF
--- a/invisible_cities/reco/krmap_evolution.py
+++ b/invisible_cities/reco/krmap_evolution.py
@@ -175,9 +175,9 @@ def get_time_series_df(ntimebins  : int,
     return shift_to_bin_centers(time_bins), masks
 
 
-def compute_drift_v(dtdata    : np.array,
+def compute_drift_v(dtdata   : np.array,
                     nbins    : int,
-                    dtrange   : Tuple[float, float],
+                    dtrange  : Tuple[float, float],
                     seed     : Tuple[float, float, float, float],
                     detector : str)->Tuple[float, float]:
 
@@ -263,7 +263,7 @@ def computing_kr_parameters(data       : DataFrame,
                             emaps      : pd.DataFrame,
                             fittype    : KrFitFunction,
                             nbins_dv   : int,
-                            dtrange_dv  : List[float],
+                            dtrange_dv : List[float],
                             detector   : str)->DataFrame:
 
     '''
@@ -325,7 +325,7 @@ def computing_kr_parameters(data       : DataFrame,
 
         par, err, cov = transform_parameters(fit_output)
 
-        e0, lt = par
+        e0, lt   = par
         e0u, ltu = err
 
     else:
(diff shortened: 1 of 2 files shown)


jahernando:
The sigmoid function maybe should be in the module with other functions, or import from scipy.

jahernando:
*gaus_sig_function* Not clear what are x and y, and how yoy compute the seed. Please add comments

jahernando:
*resolution* add comments that you return the relative resolution and error

jahernando:
*get_number_of_time_bins* this seems to me overcoding, when this function, if only used one in the code, can be resumed in just one line of code easy to read!

jahernando:
*get_time_series_df* the comment is not clear, you say that you devide the *event* but in reality you divide the *dst* in masks with consecutive time intervals. Please clarify the comment on the definition of the function.

jahernando:
In *compute_drift_v*, please x-check is there is still no access to the DB!

jahernando:
*e0_xy_correction* function, if you use only one time or in one place, this function is maybe over-coding, as it is just very fre lines, but anyhow, it is acceptable.

jahernando:
In * computing_kr_parameters* as we discussed this morning, there is no need for the *norm_strat* and *norm_value*  arguments.

jahernando:
*cut_effs_evolution* as we discussed in the meeting, it should a function, than from the dst and a list of masks returns a list of efficiencies. The *computing_kr_parameter* functions should call this function to obtain the S1, S2 and Band efficiencies and then store then in the dst and not update them.

jahernando:
* add_krevol* should take as input onloy the configuration parameters, from those you obtain the parameters that configures the internal function *add_krevol* that does not need to be passed to the arguments of the function.

The internal function *add_krevol* should take as arguments, if any, *kdst, map,* and *masks* (a list of masks). Then the function should:

1)  apply the masks to the dst
2) get the time range (without the function *get_number_of_time_bins*
3) to the function *get_time_series_df* you can pass the filtered dst *dstf * insteas of *dst* and then you can remove line 580
4) remove *cut_effs_evolution* that now it should be internal to *compute_kr_parameters*

The comments of what does this function are not right. Please indicate that the function, applies the mask to the dst, split the dst in time intervals, and for each one computes a set of variables, for example life-time, drift-velocity, S1, S2 averaget, etc and returns them in a DataFrame.

gonzaponte:
Superseded by #948.

PR #895, opened 2024-10-14
Fix types in Eutropia's output

gonzaponte:
This PRs freezes the type of the PSF produced by Eutropia. A mismatch of the datatypes can cause problems when merging PSFs from different files.


martinperezmaneiro (approved):
Clear PR, good job!

PR #896, opened 2024-10-14
Improve utilities for structure comparison in tests

### LINKED ISSUE #840: Improve error messages in assertions
The utilities in testing_utils provide a handy way of comparing tables, dataframes, etc. But when an assertion fails they don't give enough information to understand what happened on the fly. We should try to improve that.

gonzaponte:
When a table or a dataframe comparion fails in a test, we don't get much information. This PR adds a bit of information to the failure to help identify their causes.

This branch is currently on top of #895, the corresponding commits will be removed before merging.

Closes #840

### DIFF
--- a/invisible_cities/core/testing_utils_test.py
+++ b/invisible_cities/core/testing_utils_test.py
@@ -1,6 +1,7 @@
 import numpy as np
 
 from pytest                       import mark
+from pytest                       import raises
 from flaky                        import flaky
 from hypothesis                   import given
 from hypothesis.strategies        import floats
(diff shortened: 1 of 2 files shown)


jwaiton:
The code is self explanatory, adds beneficial structure to the comparison of two tables, and runs with no issues. 

There seems to be no testing of the conditions in which the function `assert_tables_equality()` provides a False return in `testing_utils_test.py`. Would this be beneficial to add to make sure the function returns False when we desire it to?

gonzaponte:
Added 3 tests.

invisible_cities/core/testing_utils_test.py
```diff
@@ -41,3 +42,98 @@ def test_assert_tables_equality_withNaN():
...
+                      , ('two', 2, 2.0)],
+                      dtype=[('text', 'U10'), ('integer', 'i4'), ('float', 'f4')])
+
+    # different lengths
+    with raises(AssertionError):
+        assert_tables_equality(table1, table2)
+
+    # different widths
+    with raises(AssertionError):
+        assert_tables_equality(table1, table3)
```
jwaiton, line 139: Does this need to be broken up into two tests still, now that there is only one condition for checking the shape?

gonzaponte: Do you mean broken into two different _assertions_? It does. Because execution is interrupted when the exception is raised and it resumes outside the `with raises` statement. i.e.
```
    with raises(AssertionError):
        assert_tables_equality(table1, table2)
        print("this line will not be executed if the one above raises an exception")
        
    print("but this one will")
```

jwaiton: My fear is that if the lengths are the same (so the test fails) but the widths are different, the widths AssertionError won't be tested. Perhaps I'm misunderstanding the way the code works here?

gonzaponte: Let me see if I understand what you are saying. You are concerned that in the case where the test fails in the first scenario (different lengths), you don't know if the second scenario would also fail. Is this right?
If that is the case, it's not a big deal. Maybe having it in separate tests would give you a bit more information. In this case I feel like we don't gain enough by splitting the test in two, but if you feel otherwise I'll be happy to do so.

jwaiton: No I think it's fine. In testing if the first case fails, fixing it and testing again would demonstrate whether the second case fails. I just wanted to make sure I wasn't misunderstanding the intent.

jwaiton (approved):
The code is self explanatory, adds beneficial structure to the comparison of two tables (or arrays), and runs with no issues. Good job!

PR #899, opened 2024-10-17
Update list of cities in `command_line_execution_test`

gonzaponte:
This test has gone under the radar for too long. This PR updates the list of cities that we run the test for.


carhc (approved):
Good catch! Now we are testing our whole city set.

As discussed offline, maybe there is no need to keep cities outside the current reconstruction flow in `all_cities` (i.e., `penthesilea`). However, this is no concern for me, especially given everything seems to be working just fine. Approved!

Issue #900, opened 2024-10-20
What to do about type checking


gonzaponte:
There is a potential problem with the type checking introduced in #824. The implementation checks the input types only the first time the function is called to avoid overhead.

In production mode, there should be no issue as we only want to do this at the city level to warn the user of wrong input values. However, when we run the test suite, we probably want to check those values every time as the various tests may call the same function with different inputs.

- Is there really a *significant* overhead in checking the inputs at every call?
- Should we have different global modes and check annotations differently in each case? I don't like it.
- Maybe do nothing and assume that `check_annotations` is only meant to be used in production mode?

PR #906, opened 2024-10-23
Fix non-interpolated deconvolution

### LINKED ISSUE #905: Beersheba: no interpolation creates an image of different size
See this test for example

```
@fixture(scope="session")
def new_grid():
    det_db   = DataSiPM('new', 0)
    det_grid = [ np.arange( det_db[var].min() + bs/2
                          , det_db[var].max() - bs/2 + np.finfo(np.float32).eps, bs)
                 for var, bs in zip("X Y".split(), [1., 1.])]
    return det_grid


@mark.parametrize("interp_method", InterpolationMethod)
def test_deconvolution_input_interpolation_method(data_hdst, new_grid, interp_method):
    """
    Check that it runs with all interpolation methods
    """
    hdst = load_dst(data_hdst, 'RECO', 'Events')
    hdst = hdst[(hdst.event == 3021916) & (hdst.npeak == 0) & (hdst.Q>40)]

    deconvolve = deconvolution_input([10., 10.], new_grid, interp_method)
    output     = deconvolve((hdst.X, hdst.Y), hdst.Q)

    assert output[0].shape == (70, 100)
    assert len(output[1])  == 2
    assert output[1][0].shape == (7000,)
    assert output[1][1].shape == (7000,)

```

gonzaponte:
The output of the deconvolution without interpolation produces an image with a different size than with interpolation.

Fixes #905.

### DIFF
--- a/invisible_cities/reco/deconv_functions.py
+++ b/invisible_cities/reco/deconv_functions.py
@@ -14,6 +14,7 @@ from scipy.spatial.distance import cdist
 from scipy                  import ndimage as ndi
 
 from ..core .core_functions import shift_to_bin_centers
+from .. core.core_functions import binedges_from_bincenters
 from ..core .core_functions import in_range
 from ..core .configure      import check_annotations
 
@@ -210,15 +211,17 @@ def deconvolution_input(sample_width : List[float     ],
                             weight      : np.ndarray
                            ) -> Tuple[np.ndarray, Tuple[np.ndarray, ...]]:
 
-        ranges = [[coord.min() - 1.5 * sw, coord.max() + 1.5 * sw] for coord, sw in zip(data, sample_width)]
-        if inter_method in (InterpolationMethod.linear, InterpolationMethod.cubic, InterpolationMethod.nearest):
-            allbins   = [np.arange(rang[0], rang[1] + np.finfo(np.float32).eps, sw) for rang, sw in zip(ranges, sample_width)]
-            Hs, edges = np.histogramdd(data, bins=allbins, normed=False, weights=weight)
-        elif inter_method is InterpolationMethod.nointerpolation:
+        eps    = np.finfo(np.float32).eps
+        ranges = [ [ coord.min() - 1.5 * sw
+                   , coord.max() + 1.5 * sw + eps]
+                   for coord, sw in zip(data, sample_width) ]
+        if inter_method is InterpolationMethod.nointerpolation:
             allbins   = [grid[in_range(grid, *rang)] for rang, grid in zip(ranges, det_grid)]
+            allbins   = [binedges_from_bincenters(bins) for bins in allbins]
             Hs, edges = np.histogramdd(data, bins=allbins, normed=False, weights=weight)
         else:
-            raise ValueError(f'inter_method {inter_method} is not a valid interpolatin mode.')
+            allbins   = [np.arange(*rang, sw) for rang, sw in zip(ranges, sample_width)]
+            Hs, edges = np.histogramdd(data, bins=allbins, normed=False, weights=weight)
 
         inter_points = np.meshgrid(*(shift_to_bin_centers(edge) for edge in edges), indexing='ij')
         inter_points = tuple      (inter_p.flatten() for inter_p in inter_points)
--- a/invisible_cities/reco/deconv_functions_test.py
+++ b/invisible_cities/reco/deconv_functions_test.py
@@ -6,6 +6,7 @@ import pytest
 
 from pytest                       import mark
 from pytest                       import raises
+from pytest                       import fixture
 
 from hypothesis                   import given
 from hypothesis.strategies        import floats
@@ -140,7 +141,17 @@ def test_interpolate_signal():
     assert np.allclose(grid             , sorted(set(inter_position[1])))
 
 
-def test_deconvolution_input(data_hdst, data_hdst_deconvolved):
+@fixture(scope="session")
+def new_grid_1mm():
+    bin_sizes = (1., 1.)
+    det_db    = DataSiPM('new', 0)
+    det_grid  = [ np.arange( det_db[var].min() - bs/2
+                           , det_db[var].max() + bs/2 + np.finfo(np.float32).eps, bs)
+                  for var, bs in zip("X Y".split(), bin_sizes)]
+    return det_grid
+
+
+def test_deconvolution_input(data_hdst, data_hdst_deconvolved, new_grid_1mm):
     ref_interpolation = np.load(data_hdst_deconvolved)
 
     hdst   = load_dst(data_hdst, 'RECO', 'Events')
@@ -148,10 +159,7 @@ def test_deconvolution_input(data_hdst, data_hdst_deconvolved):
     h      = h.groupby(['X', 'Y']).Q.sum().reset_index()
     h      = h[h.Q > 40]
 
-    det_db   = DataSiPM('new', 0)
-    det_grid = [np.arange(det_db[var].min() + bs/2, det_db[var].max() - bs/2 + np.finfo(np.float32).eps, bs)
-               for var, bs in zip(['X', 'Y'], [1., 1.])]
-    interpolator = deconvolution_input([10., 10.], det_grid, InterpolationMethod.cubic)
+    interpolator = deconvolution_input([10., 10.], new_grid_1mm, InterpolationMethod.cubic)
     inter        = interpolator((h.X, h.Y), h.Q)
 
     assert np.allclose(ref_interpolation['e_inter'], inter[0])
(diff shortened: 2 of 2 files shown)


invisible_cities/reco/deconv_functions_test.py
```diff
@@ -140,27 +141,51 @@ def test_interpolate_signal():
...
-    det_db   = DataSiPM('new', 0)
-    det_grid = [np.arange(det_db[var].min() + bs/2, det_db[var].max() - bs/2 + np.finfo(np.float32).eps, bs)
-               for var, bs in zip(['X', 'Y'], [1., 1.])]
-    interpolator = deconvolution_input([10., 10.], det_grid, InterpolationMethod.cubic)
+    interpolator = deconvolution_input([10., 10.], new_grid_1mm, InterpolationMethod.cubic)
     inter        = interpolator((h.X, h.Y), h.Q)
 
     assert np.allclose(ref_interpolation['e_inter'], inter[0])
     assert np.allclose(ref_interpolation['x_inter'], inter[1][0])
     assert np.allclose(ref_interpolation['y_inter'], inter[1][1])
```
jwaiton: It would perhaps be nice to have a docstring explaining exactly what this test does. As far as I can tell, it ensures that interpolation produces results similar to those expected from `data_hdst_deconvolved`'s interpolation data in terms of x, y and e, is that correct?

invisible_cities/reco/deconv_functions_test.py
```diff
@@ -140,27 +141,51 @@ def test_interpolate_signal():
...
 
 
+@mark.parametrize("interp_method", InterpolationMethod)
+def test_deconvolution_input_interpolation_method(data_hdst, new_grid_1mm, interp_method):
+    """
+    Check that it runs with all interpolation methods
+    """
+    hdst = load_dst(data_hdst, 'RECO', 'Events')
+    hdst = hdst[(hdst.event == 3021916) & (hdst.npeak == 0) & (hdst.Q>40)]
```
jwaiton: Is there a reason you apply a cut to `Q` > 40 in particular?

invisible_cities/reco/deconv_functions_test.py
```diff
@@ -140,27 +141,51 @@ def test_interpolate_signal():
...
+
+    deconvolve = deconvolution_input([10., 10.], new_grid_1mm, interp_method)
+    output     = deconvolve((hdst.X, hdst.Y), hdst.Q)
```
jwaiton: This is really nitpicky, but `deconvolve()` is the name of a function within `deconv_functions.py`, and differs from your naming in `test_deconvolution_input` of `interpolator()`, a function that achieves the same task.

invisible_cities/reco/deconv_functions_test.py
```diff
@@ -140,27 +141,51 @@ def test_interpolate_signal():
...
+
+    assert output[0].shape == (70, 100)
+    assert len(output[1])  == 2
+    assert output[1][0].shape == (7000,)
+    assert output[1][1].shape == (7000,)
```
jwaiton: Is it possible to alter this such that the expected shapes are ascertained from the hdst file and provided variables (`sample_width`, `det_grid`).

jwaiton: Nevermind, the method `deconvolution_input` is intended such that it ascertains these values so they have to be hardcoded I believe?

invisible_cities/reco/deconv_functions.py
```diff
@@ -210,15 +211,17 @@ def deconvolution_input(data        : Tuple[np.ndarray, ...],
...
                            ) -> Tuple[np.ndarray, Tuple[np.ndarray, ...]]:
 
-        ranges = [[coord.min() - 1.5 * sw, coord.max() + 1.5 * sw] for coord, sw in zip(data, sample_width)]
-        if inter_method in (InterpolationMethod.linear, InterpolationMethod.cubic, InterpolationMethod.nearest):
-            allbins   = [np.arange(rang[0], rang[1] + np.finfo(np.float32).eps, sw) for rang, sw in zip(ranges, sample_width)]
-            Hs, edges = np.histogramdd(data, bins=allbins, normed=False, weights=weight)
-        elif inter_method is InterpolationMethod.nointerpolation:
+        eps    = np.finfo(np.float32).eps
+        ranges = [ [ coord.min() - 1.5 * sw
+                   , coord.max() + 1.5 * sw + eps]
```
jwaiton, line 216: Perhaps this is a silly question, but why do we include this epsilon component in the binning?

invisible_cities/reco/deconv_functions.py
```diff
@@ -210,15 +211,17 @@ def deconvolution_input(data        : Tuple[np.ndarray, ...],
...
+                   for coord, sw in zip(data, sample_width) ]
+        if inter_method is InterpolationMethod.nointerpolation:
             allbins   = [grid[in_range(grid, *rang)] for rang, grid in zip(ranges, det_grid)]
+            allbins   = [binedges_from_bincenters(bins) for bins in allbins]
             Hs, edges = np.histogramdd(data, bins=allbins, normed=False, weights=weight)
         else:
-            raise ValueError(f'inter_method {inter_method} is not a valid interpolatin mode.')
+            allbins   = [np.arange(*rang, sw) for rang, sw in zip(ranges, sample_width)]
+            Hs, edges = np.histogramdd(data, bins=allbins, normed=False, weights=weight)
```
jwaiton, line 224: Does this not imply that someone can pass anything in other than `InterpolationMethod.nointerpolation` and get the behaviour from the else statement?

Just in testing:
```
>>> inter_method = 10
>>> if inter_method is InterpolationMethod.nointerpolation:
...     print("No interpolation please :)")
... else:
...     print("interp it up")
...
interp it up
```

This seems like an 'optional' default that will not be intuitive, as it wont throw an error if the user has input an invalid interpolation method.

jwaiton, line 224: **Edit: Sorry I for some reason appeared to reference the previous commit so this comment may have been posted twice.**

Does this not imply that someone can pass anything in other than `InterpolationMethod.nointerpolation` and get the behaviour from the else statement?

Just in testing:
```
>>> inter_method = 10
>>> if inter_method is InterpolationMethod.nointerpolation:
...     print("No interpolation please :)")
... else:
...     print("interp it up")
...
interp it up
```

This seems like an 'optional' default that will not be intuitive, as it wont throw an error if the user has input an invalid interpolation method.

invisible_cities/reco/deconv_functions.py
gonzaponte: in order to include the upper limit. Like this:

```
>>> eps = np.finfo(np.float32).eps
>>> np.arange(0.5, 3.5)
array([0.5, 1.5, 2.5])
>>> np.arange(0.5, 3.5+eps)
array([0.5, 1.5, 2.5, 3.5])
```

I guess technically, the correct implementation would be
```
>>> np.arange(0.5, np.nextafter(3.5, np.inf))
array([0.5, 1.5, 2.5, 3.5])
```
but it might be even more obscure.

invisible_cities/reco/deconv_functions.py
gonzaponte: True. The idea is that type checking is done at the city level (using the `check_annotations` decorator) so we don't need to do it everywhere. I'm undecided about whether we need to do it everywhere. For now, I will also add the decorator here and recover the corresponding test.

invisible_cities/reco/deconv_functions_test.py
gonzaponte: I haven't actually tried to understand this test, I've just refactored the `det_grid`/`new_grid_1mm` bit. But I think it does exactly that. I will add a docstring.

invisible_cities/reco/deconv_functions_test.py
gonzaponte: I just copied the idea from a similar test. I checked and it just creates a larger image, but the numbers with this cut are a bit rounder. My guess is that this comes from a specific case in which things look nice. I will add a comment to clarify it.

invisible_cities/reco/deconv_functions_test.py
gonzaponte: I think so.

gonzaponte (commented):
I've tried to reply as a review to reduce the amount of notifications. Let's see how it works.

invisible_cities/reco/deconv_functions.py
gonzaponte: Ah, I didn't see it when I replied to your comment, but in the outer function, we check that `inter_method` has the correct type, so it can only have one of the values of `InterpolationMethod`:
```
    if not isinstance(inter_method, InterpolationMethod):
        raise ValueError(f'inter_method {inter_method} is not a valid interpolation method.')
```

I will recover the corresponding test. I would also like to replace this explicit check with the `check_annotations` decorator, but it's not possible at the moment because of #900.

jwaiton (approved):
This PR resolves a difficult to spot issue within the interpolation and improves the clarity of the interpolation process in beersheba. Good work! 👍

PR #910, opened 2024-10-28
Handle hitless events post Sophronia

### LINKED ISSUE #897: What to do with hitless events in Sophronia
Currently, we are filtering out hitless events **only in the `/RECO/Events` output**. This means there is a mismatch between the `/Run/events` and the `/RECO/Events` outputs. The former contains all events[^1] while the latter does not. While this is not an issue _per se_, it fails a data sanity check implemented in `Esmeralda`. The solution in this case would be to remove the data sanity check.

The alternative is to remove the hitless event from the `/Run/events` output, which would satisfy the data sanity check in `Esmeralda` at the cost of filtering events, which should not be done in the processing chain.

This issue will be discussed in a dedicated meeting and the solution will be posted here later.

[^1]: for this discussion I'm ignoring the other filter in the city (the pmap filter)

gonzaponte:
As discussed in #897, we implement a solution to hitless events in Esmeralda and Beersheba. Hitless events are preserved in the `/Run/events` table but are (naturally) missing in `/RECO/Events`.

Closes #897.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -602,6 +602,7 @@ def hits_and_kdst_from_files( paths : List[str]
                 if len(hits):
                     hits = hits[event_number]
                 else:
+                    warnings.warn(f"Event {event_number} does not contain hits", UserWarning)
                     hits = HitCollection(event_number, timestamp, [])
                 yield dict(hits = hits,
                            kdst = kdst_df.loc[kdst_df.event==event_number],
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -594,12 +594,16 @@ def hits_and_kdst_from_files( paths : List[str]
             except (tb.exceptions.NoSuchNodeError, IndexError):
                 continue
 
-            check_lengths(event_info, hits_df.event.unique())
+            check_lengths(event_info, kdst_df.event.unique())
 
             for evtinfo in event_info:
                 event_number, timestamp = evtinfo.fetch_all_fields()
                 hits = hits_from_df(hits_df.loc[hits_df.event == event_number])
-                yield dict(hits = hits[event_number],
+                if len(hits):
+                    hits = hits[event_number]
+                else:
+                    hits = HitCollection(event_number, timestamp, [])
+                yield dict(hits = hits,
                            kdst = kdst_df.loc[kdst_df.event==event_number],
                            run_number = run_number,
                            event_number = event_number,
--- a/invisible_cities/cities/esmeralda_test.py
+++ b/invisible_cities/cities/esmeralda_test.py
@@ -3,6 +3,8 @@ import shutil
 import numpy  as np
 import tables as tb
 
+from pytest import mark
+
 from .. core                 import system_of_units as units
 from .. io                   import dst_io      as dio
 from .  esmeralda            import esmeralda
@@ -202,6 +204,8 @@ def test_esmeralda_filters_events_with_too_many_hits(esmeralda_config, Th228_tra
     assert            filter_output.passed.tolist() == evt_pass
 
 
+@mark.filterwarnings("ignore:Event .* does not contain hits")
+@mark.filterwarnings("ignore:Input file does not contain /config group")
 def test_esmeralda_does_not_crash_with_no_hits(esmeralda_config, Th228_hits_missing, config_tmpdir):
     path_out  = os.path.join(config_tmpdir, "esmeralda_does_not_crash_with_no_hits.h5")
     esmeralda_config.update(dict( files_in    = Th228_hits_missing
--- a/invisible_cities/cities/esmeralda_test.py
+++ b/invisible_cities/cities/esmeralda_test.py
@@ -1,4 +1,5 @@
 import os
+import shutil
 import numpy  as np
 import tables as tb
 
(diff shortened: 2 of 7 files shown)


invisible_cities/cities/components.py
```diff
@@ -594,12 +594,16 @@ def hits_and_kdst_from_files( paths : List[str]
...
 
             for evtinfo in event_info:
                 event_number, timestamp = evtinfo.fetch_all_fields()
                 hits = hits_from_df(hits_df.loc[hits_df.event == event_number])
-                yield dict(hits = hits[event_number],
+                if len(hits):
+                    hits = hits[event_number]
+                else:
+                    hits = HitCollection(event_number, timestamp, [])
+                yield dict(hits = hits,
```
jwaiton: For `if len(hits)`, the `True` condition occurs under any number of rows within `hits` as long as they aren't zero, which returns `False`.  This feels a bit obtuse, perhaps a comment to explain that the code is checking for hitless events would be useful?

invisible_cities/cities/components.py
```diff
@@ -594,12 +594,17 @@ def hits_and_kdst_from_files( paths : List[str]
...
             for evtinfo in event_info:
                 event_number, timestamp = evtinfo.fetch_all_fields()
                 hits = hits_from_df(hits_df.loc[hits_df.event == event_number])
-                yield dict(hits = hits[event_number],
+                if len(hits):
+                    hits = hits[event_number]
+                else:
+                    warnings.warn(f"Event {event_number} does not contain hits", UserWarning)
+                    hits = HitCollection(event_number, timestamp, [])
+                yield dict(hits = hits,
```
jwaiton, line 607: `if len(hits):` returns `True` for any number of rows in the `hits` dataframe, and `False` only for 0 rows, is that correct? This usage of `if len(hits):` interpreting the length of a dataframe like this feels a bit obtuse to me, but perhaps that is due to my lack of familiarity with boolean behaviour in python.

gonzaponte: Yes, this is idiomatic in Python. When interpreted as booleans, all numbers except 0 are `True` (even `np.nan`), and 0 is `False`.
This pattern is particularly common in sequences. `if len(something)` means "if it's not empty". You may also see `if not len(something)`, which means "if it's empty".
In this particular case, `hits` is a dictionary that maps event numbers to `HitCollection`s. The `len` of a dictionary corresponds to the number of entries and therefore the meaning above is preserved.

invisible_cities/cities/sophronia_test.py
```diff
@@ -105,3 +109,7 @@ def test_sophronia_filters_events_with_only_nn_hits(config_tmpdir, sophronia_con
         # (event_number, passed_flag)
         assert     output_file.root.Filters.s12_selector[0][1]
         assert not output_file.root.Filters.valid_hit   [0][1]
+
+        # Check that it's present in /Run/events, but not in /RECO/Events
+        assert event_number == output_file.root.Run.events[0][0]
+        assert event_number not in output_file.root.RECO.Events.col("event")
```
jwaiton: This addition to `test_sophronia_filters_events_with_only_nn_hits()` covers the case we're considering (hitless events retaining `Run/events` information but not `RECO/Events` information), but the test itself isn't descriptive of this condition we're checking for outside of the comment L113. 

Is it worth separating this case we're checking for (hitless events showing appropriate behaviour) into a separate test to make clear that this is a condition we're checking for? Or perhaps improving the docstring/name of the test to ensure that this is clear?

invisible_cities/cities/esmeralda_test.py
```diff
@@ -199,3 +200,22 @@ def test_esmeralda_filters_events_with_too_many_hits(esmeralda_config, Th228_tra
...
+    # copy file and remove the hits from the first event
+    shutil.copy(Th228_hits, path_in)
+    with tb.open_file(path_in, "r+") as file:
+        first_evt = file.root.Run.events[0][0]
+        evt_rows  = [row[0] == first_evt for row in file.root.RECO.Events]
+        n_delete  = sum(evt_rows)
+        file.root.RECO.Events.remove_rows(0, n_delete)
+
+    # just test that it doesn't crash
+    esmeralda(**esmeralda_config)
```
jwaiton, line 216: I'm assuming that Beersheba works identically to this, and so it doesn't need a test to verify it doesn't crash?

gonzaponte: It works in a similar fashion, but a test wouldn't hurt. This is also true for the component itself. I will add both.

jwaiton (approved):
Introduces a solution for hitless events post-Sophronia that stops Esmeralda and Beersheba from crashing, and provides a warning to the user if they occur. Good work!

invisible_cities/cities/components.py
jacg: Idiomatic Python would omit the `len`. Empty containers (be they sequences, mappings, whatever) are false in boolean contexts in Python and you are expected to take advantage of this ... except that numpy arrays throw a spanner in the works by being sequences in most respects, with the notable exception of their boolean interpretation. Consequently, in environments where numpy is used, it becomes less clear that the standard Python idioms are a good idea:

```python
import numpy as np

samples = (
                 ''  ,              'a'  ,
                 []  ,              [0]  ,
                 ()  ,              (0,) ,
                 {}  ,              {0:0},
              set()  ,              {0}  ,
    #   np.array([]) ,     np.array([0]) , # Oh dear, numpy is a git
    len(np.array([])), len(np.array([0])), # `len` provides a workaround
)

def bbb(x):
	print(x, bool(x))
    
for x in samples:
	bbb(x)
```

So, while the superfluous-for-all-containers-except-numpy-arrays `len`, is bad style in Python in general, it might be justified in projects that use numpy, for the sake of consistency and avoiding numpy-caused surprises.

PR #911, opened 2024-10-28
Generalize correction application in cities

gonzaponte:
This PR modifies the cities to apply corrections in a more general way. Currently, the correction strategy is hard-wired to `NormStrategy.kr` which limits our capabilities. This PR extends the method to allow different correction strategies.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -15,6 +15,7 @@ from typing          import Dict
 from typing          import Tuple
 from typing          import Union
 from typing          import Any
+from typing          import Optional
 
 import tables as tb
 import numpy  as np
@@ -1514,7 +1515,11 @@ def hits_thresholder(threshold_charge : float, same_peak : bool ) -> Callable:
 
 
 @check_annotations
-def hits_corrector(map_fname : str, apply_temp : bool) -> Callable:
+def hits_corrector( filename   : str
+                  , apply_temp : bool
+                  , norm_strat : NormStrategy
+                  , norm_value : Optional[float] = 0
+                  ) -> Callable:
     """
     Applies energy correction map and converts drift time to z.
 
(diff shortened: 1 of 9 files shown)


invisible_cities/cities/components.py
```diff
@@ -1531,11 +1536,14 @@ def hits_corrector(map_fname : str, apply_temp : bool) -> Callable:
     A function that takes a HitCollection as input and returns
     the same object with modified Ec and Z fields.
     """
-    map_fname = os.path.expandvars(map_fname)
-    maps      = read_maps(map_fname)
-    get_coef  = apply_all_correction(maps, apply_temp = apply_temp, norm_strat = NormStrategy.kr)
-    time_to_Z = (get_df_to_z_converter(maps) if maps.t_evol is not None else
-                 lambda x: x)
+    assert (norm_strat is not NormStrategy.custom) or (norm_value > 0)
```
jwaiton: While it doesn't matter as `get_normalization_factor()` doesn't use the `norm_value` provided unless `norm_strat` is custom, should this assert be more stringent (perhaps a XOR rather than OR)? While unlikely, someone may change `get_normalization_factor()` in the future in a manner that makes this logic untenable, such as if a normalisation strategy that isn't custom requires a normalisation value.

invisible_cities/cities/components.py
```diff
@@ -1514,7 +1515,11 @@ def threshold_hits(hitc : HitCollection) -> HitCollection:
 
 
 @check_annotations
-def hits_corrector(map_fname : str, apply_temp : bool) -> Callable:
+def hits_corrector( filename   : str
+                  , apply_temp : bool
+                  , norm_strat : NormStrategy
+                  , norm_value : Optional[float] = 0
+                  ) -> Callable:
```
jwaiton: Is doesn't appear that there's any testing for this function in `components_test.py`, is it worth doing so?

gonzaponte: Always!

invisible_cities/cities/components.py
gonzaponte: Agreed. I will also change the assertion to a different exception, this was me being lazy.

invisible_cities/cities/components_test.py
```diff
@@ -465,6 +471,54 @@ def test_read_wrong_pmt_ids(ICDATADIR):
...
+                  ( (NormStrategy.kr    , None) # None marks the default value
+                  , (NormStrategy.max   , None)
+                  , (NormStrategy.mean  , None)
+                  , (NormStrategy.custom,  1e3)
+                  ))
+@mark.parametrize("apply_temp", (False, True))
+def test_hits_corrector_valid_normalization_options( correction_map_filename
+                                                   , norm_strat
+                                                   , norm_value
+                                                   , apply_temp ):
```
jwaiton, line 484: A docstring to explain this test may be useful, although I think its self explanatory. I think `test_hits_corrector_invalid_normalization_options_raises` doesn't require one as the comments describing the fixtures + the name of the test explain (at least to me) what the test is looking to do.

jwaiton (approved):
This PR introduces methods for correction beyond the standard `NormStrategy.kr` including custom correction methods. The code is well documented and works as intended. Good job!

PR #912, opened 2024-10-28
Remove event model - phase1

gonzaponte:
This PR is the first step towards removing the superfluous parts of the event model. It replaces the `Hit` class and its subsidiaries with DataFrames, which are much more convenient.

### DIFF
--- a/invisible_cities/cities/beersheba.py
+++ b/invisible_cities/cities/beersheba.py
@@ -94,7 +94,6 @@ from typing import Union
 
 
 def event_info_adder(timestamp : float, dst : pd.DataFrame):
-    dst = dst.astype(dict(npeak=np.float64))
     return dst.assign(time=timestamp/1e3, nsipm=0, Xrms=0, Yrms=0)
 
 
--- a/invisible_cities/cities/beersheba.py
+++ b/invisible_cities/cities/beersheba.py
@@ -93,32 +93,8 @@ from typing import Optional
 from typing import Union
 
 
-# Temporary. The removal of the event model will fix this.
-def hitc_to_df_(hitc):
-    columns = "event time npeak Xpeak Ypeak nsipm X Y Xrms Yrms Z Q E Qc Ec track_id Ep".split()
-    columns = {col:[] for col in columns}
-
-    for hit in hitc.hits:
-        columns["event"   ].append(hitc.event)
-        columns["time"    ].append(hitc.time)
-        columns["npeak"   ].append(hit .npeak)
-        columns["Xpeak"   ].append(hit .Xpeak)
-        columns["Ypeak"   ].append(hit .Ypeak)
-        columns["nsipm"   ].append(hit .nsipm)
-        columns["X"       ].append(hit .X)
-        columns["Y"       ].append(hit .Y)
-        columns["Xrms"    ].append(hit .Xrms)
-        columns["Yrms"    ].append(hit .Yrms)
-        columns["Z"       ].append(hit .Z)
-        columns["Q"       ].append(hit .Q)
-        columns["E"       ].append(hit .E)
-        columns["Qc"      ].append(hit .Qc)
-        columns["Ec"      ].append(hit .Ec)
-        columns["track_id"].append(hit .track_id)
-        columns["Ep"      ].append(hit .Ep)
-    return pd.DataFrame(columns)
-
 def event_info_adder(timestamp : float, dst : pd.DataFrame):
+    dst = dst.astype(dict(npeak=np.float64))
     return dst.assign(time=timestamp/1e3, nsipm=0, Xrms=0, Yrms=0)
 
 
@@ -490,7 +466,6 @@ def beersheba( files_in         : OneOrManyFiles
     """
     correct_hits   = fl.map(hits_corrector(**corrections), item="hits")
     threshold_hits = fl.map(hits_thresholder(threshold, same_peak), item="hits")
-    hitc_to_df     = fl.map(hitc_to_df_, item="hits")
 
     deconv_params['psf_fname'       ] = expandvars(deconv_params['psf_fname'])
     deconv_params['satellite_params'] = satellite_params
@@ -537,7 +512,6 @@ def beersheba( files_in         : OneOrManyFiles
                                     correct_hits                              ,
                                     threshold_hits                            ,
                                     fl.branch(write_thr_hits)                 ,
-                                    hitc_to_df                                ,
                                     cut_sensors                               ,
                                     drop_sensors                              ,
                                     filter_events_no_hits                     ,
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -1498,21 +1498,10 @@ def hits_thresholder(threshold_charge : float, same_peak : bool ) -> Callable:
     The energy of NN hits is redistributed among neighbors.
     """
 
-    def threshold_hits(hitc : HitCollection) -> HitCollection:
-        t = hitc.time
-        thr_hits = hif.threshold_hits(hitc.hits, threshold_charge     )
-        mrg_hits = hif.merge_NN_hits ( thr_hits, same_peak = same_peak)
-
-        cor_hits = []
-        for hit in mrg_hits:
-            cluster = Cluster(hit.Q, xy(hit.X, hit.Y), hit.var, hit.nsipm)
-            xypos   = xy(hit.Xpeak, hit.Ypeak)
-            hit     = Hit(hit.npeak, cluster, hit.Z, hit.E, xypos, hit.Ec)
-            cor_hits.append(hit)
-
-        new_hitc      = HitCollection(hitc.event, t)
-        new_hitc.hits = cor_hits
-        return new_hitc
+    def threshold_hits(hits: pd.DataFrame) -> pd.DataFrame:
+        thr_hits = hif.threshold_hits(    hits, threshold_charge     )
+        mrg_hits = hif.merge_NN_hits (thr_hits, same_peak = same_peak)
+        return mrg_hits
 
     return threshold_hits
 
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -980,44 +980,6 @@ def hit_builder(dbfile, run_number, drift_v,
     return build_hits
 
 
-def sipms_above_threshold(xys: np.ndarray, qs: np.ndarray, thr:float, energy: float):
-    """
-    Finds SiPMs with charge above threshold and returns their position, charge
-    and associated energy.
-
-    Inputs
-    ------
-    xys: np.ndarray, shape (n,2)
-        SiPM positions
-    qs: np.ndarray, shape (n,)
-        Charge of each SiPM.
-    thr: float
-        Threshold on SiPM charge.
-    energy: float
-        Energy to be shared among the hits.
-
-    Returns
-    -------
-    xs: np.ndarray, shape (m,)
-        x positions of the SiPMs above threshold
-    ys: np.ndarray, shape (m,)
-        y positions of the SiPMs above threshold
-    qs: np.ndarray, shape (m,)
-        Charge of the SiPMs above threshold
-    es: np.ndarray, shape (m,)
-        Associated energy of each hit
-    """
-    over_thr = qs >= thr
-    nonempty = np.any(over_thr)
-
-    xy =            xys[over_thr]
-    qs =             qs[over_thr] if nonempty else [NN]
-    xs =                 xy[:, 0] if nonempty else [NN]
-    ys =                 xy[:, 1] if nonempty else [NN]
-    es = hif.e_from_q(qs, energy) if nonempty else [energy]
-    return xs, ys, qs, es
-
-
 def sipms_as_hits(dbfile, run_number, drift_v,
                   rebin_slices, rebin_method,
                   q_thr,
@@ -1049,7 +1011,7 @@ def sipms_as_hits(dbfile, run_number, drift_v,
             xys      = sipm_xys[peak.sipms.ids]
 
             for (slice_z, slice_e, sipm_qs) in zip(slice_zs, slice_es, sipm_charge):
-                sipm_xs, sipm_ys, sipm_qs, sipm_es = sipms_above_threshold(xys, sipm_qs, q_thr, slice_e)
+                sipm_xs, sipm_ys, sipm_qs, sipm_es = hif.sipms_above_threshold(xys, sipm_qs, q_thr, slice_e)
                 sipm_hs  = dict( event    = event_number
                                , time     = timestamp * 1e-3
                                , npeak    = peak_no
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -989,19 +989,17 @@ def sipms_as_hits(dbfile, run_number, drift_v,
     sipm_noise = NoiseSampler(dbfile, run_number).signal_to_noise
     epsilon    = np.finfo(np.float64).eps
 
-    def make_cluster(q, x, y):
-        return Cluster(q, xy(x, y), xy(0,0), 1)
+    def e_from_q(qs, e_slice):
+        return qs * e_slice / (qs.sum() + epsilon)
 
     def build_hits(pmap, selector_output, event_number, timestamp):
-        hitc = HitCollection(event_number, timestamp * 1e-3)
         s1_t = get_s1_time(pmap, selector_output)
-
+        hits = []
         for peak_no, (passed, peak) in enumerate(zip(selector_output.s2_peaks,
                                                      pmap.s2s)):
             if not passed: continue
 
             peak = pmf.rebin_peak(peak, rebin_slices, rebin_method)
-
             xys  = sipm_xys[peak.sipms.ids]
             qs   = peak.sipm_charge_array(sipm_noise, charge_type,
                                           single_point = True)
@@ -1018,26 +1016,35 @@ def sipms_as_hits(dbfile, run_number, drift_v,
 
             for (slice_z, slice_e, sipm_qs) in zip(slice_zs, slice_es, sipm_charge):
                 over_thr = sipm_qs >= q_thr
-                if np.any(over_thr):
-                    sipm_qs  = sipm_qs[over_thr]
-                    sipm_xy  = xys[over_thr]
-                    sipm_es  = sipm_qs * slice_e / (np.sum(sipm_qs) + epsilon)
+                nonempty = np.any(over_thr)
+                sipm_xy  =              xys[over_thr]
+                sipm_qs  =          sipm_qs[over_thr] if nonempty else [NN]
+                sipm_xs  =              sipm_xy[:, 0] if nonempty else [NN]
+                sipm_ys  =              sipm_xy[:, 1] if nonempty else [NN]
+                sipm_es  = e_from_q(sipm_qs, slice_e) if nonempty else [slice_e]
+                sipm_hs  = dict( event    = event_number
+                               , time     = timestamp * 1e-3
+                               , npeak    = peak_no
+                               , Xpeak    = xy_peak[0]
+                               , Ypeak    = xy_peak[1]
+                               , nsipm    = 1
+                               , X        = sipm_xs
+                               , Y        = sipm_ys
+                               , Xrms     = 0.
+                               , Yrms     = 0.
+                               , Z        = slice_z
+                               , Q        = sipm_qs
+                               , E        = sipm_es
+                               , Qc       = -1.
+                               , Ec       = -1.
+                               , track_id = -1
+                               , Ep       = -1.)
+                hits.append(pd.DataFrame(sipm_hs))
 
-                    for q, e, (x, y) in zip(sipm_qs, sipm_es, sipm_xy):
-                        hitc.hits.append( Hit( peak_no
-                                             , make_cluster(q, x, y)
-                                             , slice_z
-                                             , e
-                                             , xy_peak))
+        hits = pd.concat(hits, ignore_index=True)
+        hits = hits.astype(dict(npeak=np.uint16, nsipm=np.uint16))
+        return hits
 
-                else:
-                    hitc.hits.append( Hit( peak_no
-                                         , Cluster.empty()
-                                         , slice_z
-                                         , slice_e
-                                         , xy_peak))
-
-        return hitc
     return build_hits
 
 
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -1441,8 +1441,8 @@ def hitc_to_df(hitc: HitCollection):
 
 
 def compute_and_write_tracks_info(paolina_params, h5out,
-                                  hit_type, filter_hits_table_name='hits_select',
-                                  write_paolina_hits=None):
+                                  hit_type, filter_hits_table_name,
+                                  hits_writer):
 
     filter_events_nohits = fl.map(lambda x : len(x.hits) > 0,
                                       args = 'hits',
@@ -1472,8 +1472,6 @@ def compute_and_write_tracks_info(paolina_params, h5out,
                                             args = ('event_number', 'topology_info', 'paolina_hits', 'out_of_map'),
                                             out  = 'event_info')
 
-    to_hits_df = fl.map(hitc_to_df, item="paolina_hits")
-
     # Define writers and make them sinks
     write_tracks          = fl.sink(   track_writer     (h5out=h5out)             , args="topology_info"      )
     write_summary         = fl.sink( summary_writer     (h5out=h5out)             , args="event_info"         )
@@ -1484,7 +1482,9 @@ def compute_and_write_tracks_info(paolina_params, h5out,
 
     make_and_write_summary  = make_final_summary, write_summary
     select_and_write_tracks = events_passed_topology.filter, write_tracks
-    write_hits              = (to_hits_df, write_paolina_hits) if write_paolina_hits else None
+
+    to_hits_df = fl.map(hitc_to_df)
+    write_hits = ("paolina_hits", to_hits_df, fl.sink(hits_writer))
 
     fork_pipes = filter(None, ( make_and_write_summary
                               , write_topology_filter
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -1476,11 +1476,7 @@ def compute_and_write_tracks_info(paolina_params, h5out,
 
 @check_annotations
 def hits_merger(same_peak : bool) -> Callable:
-    def merge_hits(hc : HitCollection) -> HitCollection:
-        merged_hits = hif.merge_NN_hits(hc.hits, same_peak)
-        return HitCollection(hc.event, hc.time, merged_hits)
-
-    return merge_hits
+    return partial(hif.merge_NN_hits, same_peak=same_peak)
 
 
 @check_annotations
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -930,11 +930,6 @@ def sipm_positions(dbfile, run_number):
     return sipm_xys
 
 
-def e_from_q(qs, e_slice):
-    epsilon = np.finfo(np.float64).eps
-    return qs * e_slice / (qs.sum() + epsilon)
-
-
 def hit_builder(dbfile, run_number, drift_v,
                 rebin_slices, rebin_method,
                 global_reco, slice_reco,
@@ -972,7 +967,7 @@ def hit_builder(dbfile, run_number, drift_v,
                 try:
                     clusters = slice_reco(xys, sipm_qs)
                     qs       = np.array([c.Q for c in clusters])
-                    es       = e_from_q(qs, e_slice)
+                    es       = hif.e_from_q(qs, e_slice)
                     for c, e in zip(clusters, es):
                         hit  = Hit(peak_no, c, z_slice, e, xy_peak)
                         hitc.hits.append(hit)
@@ -1018,11 +1013,11 @@ def sipms_as_hits(dbfile, run_number, drift_v,
             for (slice_z, slice_e, sipm_qs) in zip(slice_zs, slice_es, sipm_charge):
                 over_thr = sipm_qs >= q_thr
                 nonempty = np.any(over_thr)
-                sipm_xy  =              xys[over_thr]
-                sipm_qs  =          sipm_qs[over_thr] if nonempty else [NN]
-                sipm_xs  =              sipm_xy[:, 0] if nonempty else [NN]
-                sipm_ys  =              sipm_xy[:, 1] if nonempty else [NN]
-                sipm_es  = e_from_q(sipm_qs, slice_e) if nonempty else [slice_e]
+                sipm_xy  =                  xys[over_thr]
+                sipm_qs  =              sipm_qs[over_thr] if nonempty else [NN]
+                sipm_xs  =                  sipm_xy[:, 0] if nonempty else [NN]
+                sipm_ys  =                  sipm_xy[:, 1] if nonempty else [NN]
+                sipm_es  = hif.e_from_q(sipm_qs, slice_e) if nonempty else [slice_e]
                 sipm_hs  = dict( event    = event_number
                                , time     = timestamp * 1e-3
                                , npeak    = peak_no
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -1547,12 +1547,9 @@ def hits_corrector( filename     : str
                                     , **norm_options)
     time_to_Z = get_df_to_z_converter(maps) if maps.t_evol is not None and apply_z else identity
 
-    def correct(hitc : HitCollection) -> HitCollection:
-        for hit in hitc.hits:
-            corr    = get_coef([hit.X], [hit.Y], [hit.Z], hitc.time)[0]
-            hit.Ec  = hit.E * corr
-            hit.xyz = (hit.X, hit.Y, time_to_Z(hit.Z)) # ugly, but temporary
-        return hitc
+    def correct(hits : pd.DataFrame) -> pd.DataFrame:
+        corr_factors = get_coef(hits.X, hits.Y, hits.Z, hits.time)
+        return hits.assign(Ec = hits.E * corr_factors)
 
     return correct
 
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -599,17 +599,17 @@ def hits_and_kdst_from_files( paths : List[str]
 
             for evtinfo in event_info:
                 event_number, timestamp = evtinfo.fetch_all_fields()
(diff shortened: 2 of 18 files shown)


invisible_cities/cities/components.py
```diff
@@ -598,17 +598,17 @@ def hits_and_kdst_from_files( paths : List[str]
 
             for evtinfo in event_info:
                 event_number, timestamp = evtinfo.fetch_all_fields()
-                hits = hits_from_df(hits_df.loc[hits_df.event == event_number])
-                if len(hits):
-                    hits = hits[event_number]
-                else:
+                this_event = lambda df: df.event == event_number
+                hits = hits_df.loc[this_event]
+                kdst = kdst_df.loc[this_event]
```
jwaiton, line 614: Could this not be written in a manner that doesn't require multiple iterative lookups? Perhaps instead with `kdst.groupby('event')`? Unless this isn't possible due to events that have no kdst and no hits.

Of course, then you need to fetch the `evtinfo` anyway, so perhaps theres no neat way around this outside of sorting the hits, kdst and evtinfo then grouping that way.

invisible_cities/cities/components.py
```diff
@@ -949,24 +954,24 @@ def build_hits(pmap, selector_output, event_number, timestamp):
...
             sipm_charge = peak.sipm_charge_array(sipm_noise        ,
                                                  charge_type       ,
                                                  single_point=False)
 
-            for slice_no, (t_slice, qs) in enumerate(zip(peak.times ,
-                                                         sipm_charge)):
-                z_slice = (t_slice - s1_t) * units.ns * drift_v
-                e_slice = peak.pmts.sum_over_sensors[slice_no]
+            slice_zs = (peak.times - s1_t) * units.ns * drift_v
+            slice_es = peak.pmts.sum_over_sensors
```
jwaiton, line 1039: Perhaps this is my ignorance, but this area of the code (functions related to `SensorResponses()` such as `sum_over_sensors` as an example in this case) is quite dark.

I believe I understand how it works, but it may be useful to improve the docstrings in tandem while the functions are being improved.

invisible_cities/cities/components.py
```diff
@@ -929,6 +929,11 @@ def sipm_positions(dbfile, run_number):
     return sipm_xys
 
 
+def e_from_q(qs, e_slice):
+    epsilon = np.finfo(np.float64).eps
+    return qs * e_slice / (qs.sum() + epsilon)
```
jwaiton: This replaces [split_energy()](https://github.com/gonzaponte/IC/blob/378b6425438f8dbd5acef1a0ad0b8385405b6190/invisible_cities/reco/hits_functions.py#L9), but perhaps it shouldn't sit like this in `components.py`. Would there be a suitable file for it to sit in?

invisible_cities/cities/components.py
```diff
@@ -949,24 +954,24 @@ def build_hits(pmap, selector_output, event_number, timestamp):
...
                 try:
-                    xys      = sipm_xys[peak.sipms.ids]
-                    clusters = slice_reco(xys, qs)
-                    es       = hif.split_energy(e_slice, clusters)
+                    clusters = slice_reco(xys, sipm_qs)
+                    qs       = np.array([c.Q for c in clusters])
+                    es       = e_from_q(qs, e_slice)
                     for c, e in zip(clusters, es):
                         hit  = Hit(peak_no, c, z_slice, e, xy_peak)
                         hitc.hits.append(hit)
```
jwaiton, line 1049: The `hit` class is retained here, and used for `penthesilea`. Is this intended?

jwaiton: I have a feeling that the `hit_builder()` and `sipm_as_hits()` function very similarly (returns hits from similar inputs). Is there a reason these two functions aren't merged?

invisible_cities/cities/components.py
```diff
@@ -1017,26 +1016,35 @@ def build_hits(pmap, selector_output, event_number, timestamp):
...
-                    hitc.hits.append( Hit( peak_no
-                                         , Cluster.empty()
-                                         , slice_z
-                                         , slice_e
-                                         , xy_peak))
+                nonempty = np.any(over_thr)
+                sipm_xy  =              xys[over_thr]
+                sipm_qs  =          sipm_qs[over_thr] if nonempty else [NN]
+                sipm_xs  =              sipm_xy[:, 0] if nonempty else [NN]
+                sipm_ys  =              sipm_xy[:, 1] if nonempty else [NN]
```
jwaiton: The behaviour for `!nonempty` providing a `NN` result (`-9999`), is a novel component of this function. Should it be tested that this works as expected/is accounted for appropriately?

invisible_cities/cities/components.py
jwaiton: Upon further reading, it appears that `penthesilea` is no longer used in the official production flow. Is there any need to update these functions rather than deprecate/remove them?

invisible_cities/cities/sophronia.py
```diff
@@ -70,8 +71,8 @@
 
 
 @check_annotations
-def count_valid_hits(hitc : HitCollection):
-    return sum(1 for hit in hitc.hits if hit.Q != NN)
+def count_valid_hits(hits: pd.DataFrame):
+    return np.count_nonzero(hits.Q.values != NN)
```
jwaiton, line 74: May be a bit derivative, but could there be a simple test for this? Or is it already encapsulated well enough by other tests. It's implemented in [enough_valid_hits](https://github.com/gonzaponte/IC/blob/378b6425438f8dbd5acef1a0ad0b8385405b6190/invisible_cities/cities/sophronia.py#L167), which also doesn't seem to be tested explicitly.

invisible_cities/reco/hits_functions_test.py
```diff
@@ -55,54 +77,44 @@ def thresholds(draw, min_value=1, max_value=1):
...
+@mark.parametrize( "on_corrected  col".split()
+                 , ( (     False, "Q" )
+                   , (      True, "Qc")))
 @given(hits=list_of_hits(), th=floats())
-def test_threshold_hits_all_larger_than_th(hits, th, on_corrected):
-    hits_thresh  = threshold_hits(hits, th, on_corrected = on_corrected)
-    assert (h.Q  > th or h.Q  == NN for h in hits_thresh)
-    assert (h.Qc > th or h.Qc == NN for h in hits_thresh)
+def test_threshold_hits_all_larger_than_th(hits, th, col, on_corrected):
+    hits_thresh = threshold_hits(hits, th, on_corrected = on_corrected)
```
jwaiton, line 178: Unrelated; I may open an issue, but it may be worth consolidating the different 'threshold' functions (`cut_over_Q()` in Beersheba, `threshold_hits()` in hit_functions, `hits_thresholder()` in components, `sipm_as_hits()` in Sophronia technically does so as well).

invisible_cities/reco/hits_functions.py
```diff
@@ -1,81 +1,87 @@
...
-    if len(clusters) == 1:
-        return [total_e]
-    qs = np.array([c.Q for c in clusters])
-    return total_e * qs / np.sum(qs)
+EPSILON = np.finfo(np.float64).eps
+
 
-def merge_NN_hits(hits : List[evm.Hit], same_peak : bool = True) -> List[evm.Hit]:
+def merge_NN_hits(hits : pd.DataFrame, same_peak : bool = True) -> pd.DataFrame:
     """ Returns a list of the hits where the  energies of NN hits are distributed to the closest hits such that the added energy is proportional to
```
jwaiton: Better docstrings would be nice :smile_cat:

invisible_cities/reco/hits_functions.py
```diff
@@ -1,81 +1,87 @@
...
+        candidates = hits.loc[hits.npeak == nn_hit.npeak] if same_peak else hits
+        if len(candidates) == 0: continue # drop hit !!! dangerous
+
+        dz      = np.abs(candidates.Z - nn_hit.Z)
+        closest = candidates.loc[np.isclose(dz, dz.min())]
+        index   = closest.index
+
+        corr_e  = nn_hit.E  * closest.E  / closest.E .sum()
+        corr_ec = nn_hit.Ec * closest.Ec / closest.Ec.sum()
+        corrections.loc[index, "E Ec".split()] += np.stack([corr_e, corr_ec], axis=1)
```
jwaiton: Comments are a bit sparse which in conjunction with the docstring make this a bit harder to comprehend.

For my own sanity (to double check):

The function iterates through all hits with `Q = NN`, finds those within the `same_peak` (line 26) and then for each `nn_hit` row, finds the closest hits that aren't `Q=NN` (lines 29-30) and distributes the `nn_hit` energy across them (lines 33-35). This iterates over each row and is then added to the returned hits (line 39).

invisible_cities/reco/hits_functions.py
```diff
@@ -1,81 +1,87 @@
...
+                            , Y        = NN
+                            , Xrms     = 0
+                            , Yrms     = 0
+                            , Z        = z
+                            , Q        = NN
+                            , E        = e
+                            , Qc       = -1
+                            , Ec       = ec
+                            , track_id = -1
+                            , Ep       = -1), index=[0])
```
jwaiton: I can't think of a better place for this to go, but it seems more like a 'type', if that makes sense?

invisible_cities/reco/hits_functions.py
```diff
@@ -1,81 +1,87 @@
...
+
+
+def apply_threshold(hits : pd.DataFrame, th : float, on_corrected : bool=False) -> pd.DataFrame:
```
jwaiton: No docstrings or comments :)

invisible_cities/reco/hits_functions.py
```diff
@@ -1,81 +1,87 @@
...
+    raw_e_slice = hits.E.sum()
+    cor_e_slice = np.nansum(hits.Ec) + np.finfo(np.float64).eps
```
jwaiton: To aid my understanding, the addition of the epsilon here is to ensure that L78-79 doesn't result in a division by zero, right? Can this issue not also occur with `raw_e_slice`?

invisible_cities/reco/hits_functions.py
```diff
@@ -1,81 +1,87 @@
...
+
+    col         = "Qc" if on_corrected else "Q"
+    mask_thresh = hits[col] >= th
+
+    if not mask_thresh.any():
+        first = hits.iloc[0]
+        return empty_hit( first.event, first.time
+                        , first.npeak, first.Xpeak, first.Ypeak
+                        , first.Z, raw_e_slice, cor_e_slice)
```
jwaiton, line 209: As far as I understand, this occurs only if there are no hits that pass the threshold. Is this(/will this be) standard behaviour for when a filter is applied to hits/dataframes? It may be beneficial to provide some sort of warning to the user that their cut has removed all hits.

invisible_cities/reco/hits_functions.py
```diff
@@ -1,81 +1,87 @@
...
+
+    hits = hits.loc[mask_thresh].copy()
+    qsum = np.nansum(hits.Q) + EPSILON
```
jwaiton, line 212: Since `EPSILON` isn't used elsewhere, should it be defined within the function?

EDIT: If it gets used in `apply_threshold()`, this comment can be ignored :)

invisible_cities/reco/hits_functions.py
```diff
@@ -1,81 +1,87 @@
...
+    hits.loc[:, "E" ] = hits.Q / qsum * raw_e_slice
+    hits.loc[:, "Ec"] = hits.Q / qsum * cor_e_slice
```
jwaiton: Similarly in L33-34, I find expressing these sorts of operations without explicit brackets to be quite confusing, but this is just personal preference.

invisible_cities/cities/components.py
```diff
@@ -1540,12 +1540,9 @@ def hits_corrector(map_fname : str, apply_temp : bool) -> Callable:
...
 
-    def correct(hitc : HitCollection) -> HitCollection:
-        for hit in hitc.hits:
-            corr    = get_coef([hit.X], [hit.Y], [hit.Z], hitc.time)[0]
-            hit.Ec  = hit.E * corr
-            hit.xyz = (hit.X, hit.Y, time_to_Z(hit.Z)) # ugly, but temporary
-        return hitc
+    def correct(hits : pd.DataFrame) -> pd.DataFrame:
+        corr_factors = get_coef(hits.X, hits.Y, hits.Z, hits.time)
+        return hits.assign(Ec = hits.E * corr_factors)
```
jwaiton: Removing the Z correction brings up conflicts with newer PRs right? Specifically #933, will there be an alternative for correcting z position for hits that haven't had z corrections applied?

invisible_cities/cities/components.py
```diff
@@ -1497,21 +1497,10 @@ def hits_thresholder(threshold_charge : float, same_peak : bool ) -> Callable:
...
-            hit     = Hit(hit.npeak, cluster, hit.Z, hit.E, xypos, hit.Ec)
-            cor_hits.append(hit)
-
-        new_hitc      = HitCollection(hitc.event, t)
-        new_hitc.hits = cor_hits
-        return new_hitc
+    def threshold_hits(hits: pd.DataFrame) -> pd.DataFrame:
+        thr_hits = hif.threshold_hits(    hits, threshold_charge     )
+        mrg_hits = hif.merge_NN_hits (thr_hits, same_peak = same_peak)
+        return mrg_hits
```
jwaiton: Very clean resulting function! I don't see any testing of this component within `components_test.py`, should there be?

invisible_cities/io/hits_io.py
```diff
@@ -130,16 +133,10 @@ def load_hits_skipping_NN(DST_file_name : str, group_name : str = 'RECO', table_
...
-                             fformat     = HitsTable,
-                             description = 'Hits',
-                             compression = compression)
-    # Mark column to index after populating table
-    hits_table.set_attr('columns_to_index', ['event'])
-
-    def write_hits(hits_event):
-        hits_event.store(hits_table)
-    return write_hits
+def hits_writer(hdf5_file, group_name, table_name, *, compression=None):
```
jwaiton, line 136: Removing the default `group_name` and `table_name` may cause some confusion when building new cities/modifying the pipe, right? An example where this could be an issue is [here](https://github.com/next-exp/IC/pull/912/files#diff-4241aad994b566cd332688ff48689b507388a77a911e9c9c3a87d73f4b91d99dR1486), although perhaps I'm misunderstanding its implementation.

invisible_cities/io/hits_io_test.py
```diff
@@ -75,32 +79,28 @@ def test_load_hits_double_ratio_e_q_equals_one_skipping_NN(TlMC_hits_skipping_NN
...
-    assert_allclose(np.sqrt(xstd), dst.Xrms .values)
-    assert_allclose(np.sqrt(ystd), dst.Yrms .values)
-    assert_allclose(z    , dst.Z    .values)
-    assert_allclose(q    , dst.Q    .values)
-    assert_allclose(e    , dst.E    .values)
+        write = hits_writer(h5out, "RECO", "Events")
+        write(original_hits)
+
+    read_hits = load_dst(output_file, group = "RECO", node = "Events")
+    assert_dataframes_close(read_hits, original_hits)
```
jwaiton, line 103: This is a much cleaner test now! :)

invisible_cities/cities/beersheba.py
```diff
@@ -94,32 +94,8 @@
...
-        columns["Q"       ].append(hit .Q)
-        columns["E"       ].append(hit .E)
-        columns["Qc"      ].append(hit .Qc)
-        columns["Ec"      ].append(hit .Ec)
-        columns["track_id"].append(hit .track_id)
-        columns["Ep"      ].append(hit .Ep)
-    return pd.DataFrame(columns)
-
 def event_info_adder(timestamp : float, dst : pd.DataFrame):
+    dst = dst.astype(dict(npeak=np.float64))
```
jwaiton: Is there a reason that npeak must be a `np.float64` here? Perhaps I'm misunderstanding the purpose of this line.

invisible_cities/cities/components.py
```diff
@@ -1414,6 +1414,31 @@ def sort_hits(hitc):
...
+                                     , Z        = hit .Z
+                                     , Q        = hit .Q
+                                     , E        = hit .E
+                                     , Qc       = hit .Qc
+                                     , Ec       = hit .Ec
+                                     , track_id = hit .track_id
+                                     , Ep       = hit .Ep), index=[0]))
+    df = pd.concat(hits, ignore_index=True)
+    df = df.astype(dict(event=np.int64, npeak=np.uint16, nsipm=np.uint16, Qc=np.float64, Ec=np.float64, Ep=np.float64))
+    return df
```
jwaiton, line 1568: I'm assuming this will be removed in a later phase, as well as its usage in the topological functions (`compute_and_write_tracks_info()` for example).

invisible_cities/cities/penthesilea.py
```diff
@@ -101,6 +102,8 @@ def penthesilea( files_in           : OneOrManyFiles
                                                sipm_charge_type),
                                    args = ("pmap", "selector_output", "event_number", "timestamp"),
                                    out  = "hits"                                                 )
+
+    to_hits_df            = df.map(hitc_to_df, item="hits")
```
jwaiton, line 106: Likewise I assume this is temporary, or are we assuming that cities that rely on hits will retain these conversions functions? If so, It would make sense that both `hitc_to_df()` and `hitc_from_df()` are moved to `hit_functions.py` right? And adequately tested also.

invisible_cities/cities/components.py
jwaiton: Both hit functions `threshold_hits()` and `merge_NN_hits()` are tested, so this isn't an issue.

Would it be possible to try and think of differing names for this function, due to its repetition (`threshold_hits()` calling `hif.threshold_hits()` is a bit ugly).

invisible_cities/cities/isaura.py
```diff
@@ -96,11 +97,17 @@ def isaura( files_in       : OneOrManyFiles
...
+        write_hits         = ("paolina_hits", to_hits_df, write_paolina_hits)
 
         evtnum_collect = collect()
 
-        compute_tracks = compute_and_write_tracks_info(paolina_params, h5out, hit_type=HitEnergy.E)
+        compute_tracks = compute_and_write_tracks_info( paolina_params
+                                                      , h5out
+                                                      , HitEnergy.E
+                                                      , "hits_select"
+                                                      , write_hits)
```
jwaiton: Moving `write_hits()` into `compute_and_write_tracks_info()` seems to be a bit much for a single function does it not? It's now `compute_and_write_tracks_info_and_hits()`, which I feel could be split up as done previously, unless there is some reason that makes this unfeasible/less elegant.

invisible_cities/cities/components.py
```diff
@@ -1483,7 +1481,9 @@ def compute_and_write_tracks_info(paolina_params, h5out,
 
     make_and_write_summary  = make_final_summary, write_summary
     select_and_write_tracks = events_passed_topology.filter, write_tracks
-    write_hits              = (to_hits_df, write_paolina_hits) if write_paolina_hits else None
+
+    to_hits_df = fl.map(hitc_to_df)
+    write_hits = ("paolina_hits", to_hits_df, fl.sink(hits_writer))
```
jwaiton, line 1615: likewise as mentioned previously. This function now does a lot (extracts and writes topology info, summary, and hits all in one).

invisible_cities/cities/isaura.py
jwaiton: The obvious issue here is that multiple cities extract topological information, so breaking this function up would inevitably introduce wasted repetition across isaura/esmeralda.

I think a nicer solution would be to implement a topology city that acts on the results regardless of the previous processing (agnostic to beersheba/sophronia/anything else thats given). I may open an issue on this if you think its pertinent.

invisible_cities/io/hits_io_test.py
```diff
@@ -74,34 +78,32 @@ def test_load_hits_double_ratio_e_q_equals_one_skipping_NN(TlMC_hits_skipping_NN
...
 
 
+# TODO: this test does not test make any sense
```
jwaiton, line 106: Assuming this is resolved in phase 2 :smile_cat:

invisible_cities/reco/hits_functions.py
```diff
@@ -1,81 +1,87 @@
...
-
-            raw_es_new = split_energy(raw_e_slice, hits_pass_th)
-            cor_es_new = split_energy(cor_e_slice, hits_pass_th)
-            for hit, raw_e, cor_e in zip(hits_pass_th, raw_es_new, cor_es_new):
-                hit.E  = raw_e
-                hit.Ec = cor_e
-                new_hits.append(hit)
-        return new_hits
+    if th <= 0: return hits
+    return (hits.groupby("Z", as_index=False)
```
jwaiton, line 254: Very nice refactor :)

jwaiton (changes requested):
Apologies for the wait! I have done a first read of all the changes. I have undoubtedly missed things, most of which I'll hopefully catch on my next read-through.
Github's review page often hides docstrings if they've not been modified, so I may have commented asking for docstrings where they already exist! Apologies, I tried to fix this but some may have slipped through.

Regardless, docstrings/comments for the new/modified functions are a bit lacking. I'd appreciate it if they could be included. :)

Other than that, good job! I'll have another read-through to ensure that I haven't missed anything soon (in particular related to testing).

invisible_cities/reco/hits_functions.py
jwaiton: Change `np.finfo(np.float64).eps` to `EPSILON`, since its been defined :)

invisible_cities/cities/components.py
gonzaponte: I don't recall the details but I think I had a problem with groupby. Indeed `groupby` combined with `get_group` is faster. But your own explanation sounds like something I could have run into. If either `hits_df` or `kdst_df` doesn't have an event it will raise an error. The speedup is not significant in this context so I'm tempted to leave it as is, since it is very self-explanatory.

invisible_cities/cities/components.py
gonzaponte: Which docstrings are you referring to?
Can you ellaborate on what you find dark so I can focus the docstring on that?

invisible_cities/cities/components.py
gonzaponte: Agreed. Probably in `hits_functions`.

invisible_cities/cities/components.py
gonzaponte: Yes, `sipms_as_hits` is essentially a rewrite of `hit_builder` to make it do what it was already doing with a less convoluted approach. The latter is more general, and could reproduce the former under a specific (magical) configuration.

The idea behind updating these functions is to keep `penthesilea` alive as we didn't (and still don't) know what reconstruction strategy will be necessary in this new detector.

invisible_cities/cities/components.py
jwaiton: The function `hits_and_kdst_from_file()` itself doesn't have a great docstring.

I'll follow through my process to hopefully clarify what I find dark about this section of the code. Sorry if this is a bit tedious:
- `peak` uses class components like `times` and `pmts`. I wasn't sure where these came from or how they were defined, and my IDE couldn't pick up their origin.
- Tracing back through the function, I find that peak comes from `selector_output` as seen [here](https://github.com/gonzaponte/IC/blob/378b6425438f8dbd5acef1a0ad0b8385405b6190/invisible_cities/cities/components.py#L948) in the `build_hits()` function. 
- Neither `build_hits()` nor `hit_builder()` have type formatting or a docstring explaining what the `selector_output` is, so I have to find it being used in [`penthesilea`](https://github.com/gonzaponte/IC/blob/master/invisible_cities/cities/penthesilea.py#L95).
- From here, I can see that the `selector_output` is from [`peak_classifier()`](https://github.com/gonzaponte/IC/blob/master/invisible_cities/cities/components.py#L811), but you have to understand how the dataflow system works to see this. `peak_classifier()` is a `partial` of the [`S12Selector` class](https://github.com/gonzaponte/IC/blob/master/invisible_cities/filters/s1s2_filter.py#L66).
- The `S12Selector` class itself has no overarching docstring explaining what it does, although I feel it's relatively straightforward. It uses the [`_Peak` class](https://github.com/gonzaponte/IC/blob/master/invisible_cities/evm/pmaps.py#L27) which itself holds the `times` and `pmts` information, although it also has no docstrings.

I feel like the majority of these steps could be avoided if `hit_builder()` had type formatting and a docstring, but if that isn't the case, the path back to the `_Peak` class is not that simple/is relatively undocumented. I don't think altering this is within the scope of this PR though :)

invisible_cities/cities/components.py
gonzaponte: Yes.

invisible_cities/reco/hits_functions_test.py
gonzaponte: Agreed. But that's well beyond the scope of this PR. I'll be happy to have an issue os we don't forget.
Of those, `hits_thresholder` is meant to be a component that simply composes existing functions and should not fall in the same category as the other "inner functions" or whatever we want to call them.

invisible_cities/reco/hits_functions.py
gonzaponte: Correct. NN-hits are "merged" with their closest non-NN neighbours.

invisible_cities/reco/hits_functions.py
gonzaponte: I understand what you mean. A dataframe with these columns constitutes a sort of hit-collection type. And this function could be an associated method that constructs an empty hit. Sadly, it's not a type just yet. This could be part of a follow-up PR where we find a way of translating specific dataframes into types (I haven't explored if this is in any way possible)

invisible_cities/reco/hits_functions.py
gonzaponte: > To aid my understanding, the addition of the epsilon here is to ensure that L78-79 doesn't result in a division by zero, right?

Yup. Will document.

> Can this issue not also occur with raw_e_slice?

In theory, because of the way we process data, no. But it wouldn't hurt adding a protection, I guess.

> Change np.finfo(np.float64).eps to EPSILON, since its been defined :)

Thanks! I forgot.

invisible_cities/reco/hits_functions.py
gonzaponte: > Is this(/will this be) standard behaviour for when a filter is applied to hits/dataframes?

Not sure what you mean here. Do we (will we) always produce an empty hit if no sipms survive the threshold? We are doing it now, we could decide to do it differently in the future (slice concatenation could be an option).

As for the warning, it can be added. Not sure it's useful in practice, but we can always produce a wrapper that silences that warning.

invisible_cities/cities/components.py
gonzaponte: Oops, shouldn't have removed the Z correction, actually.

invisible_cities/cities/components.py
gonzaponte: The output of a function that simply composes two other functions should be quite predictable. But that's no excuse to avoid testing, so yes, there should always be at least one simple test. I wouldn't stop this PR for that, though, it would be a good exercise for a newcomer.

gonzaponte: Ah and regarding the name, we can rename it. Not because there could be a name collision (its name is only visible within the outer function), but because it can avoid confusion for users.

invisible_cities/io/hits_io.py
gonzaponte: Not sure where the confusion comes from. If anything it should make it more clear which table is being written. If I recall correctly, the idea was to use this function in different cities, where each city might write data to a different node.

invisible_cities/cities/beersheba.py
gonzaponte: Yes, there is... sadly I forgot what the reason was. Hopefully, commenting that line triggers some test failures, which I will use to remember the reason and document this line.

invisible_cities/cities/components.py
gonzaponte: Yes.

invisible_cities/cities/penthesilea.py
gonzaponte: They will be removed in phase 2.

invisible_cities/cities/isaura.py
gonzaponte: > Moving write_hits() into compute_and_write_tracks_info() seems to be a bit much for a single function does it not?

right. That might need a big refactoring, though (haven't delved enough into it). Depending on that I might push it to a later PR, if you are happy with that.

> I think a nicer solution would be to implement a topology city

You probably know better than me. What you say makes sense, but I don't understand if there are any technical limitations. I'm happy to have a topology city like you say.

invisible_cities/io/hits_io_test.py
gonzaponte: I think I mean it literally. The test takes data from a file and ensures it belongs to a year range. But the data is static, it doesn't change at any point. It could be modified in `load_hits`, but it doesn't seem to be the case.

invisible_cities/reco/hits_functions_test.py
jwaiton: Created an issue (#936)

invisible_cities/reco/hits_functions.py
jwaiton: Sorry I didn't express myself clearly.

The reason for my concern is due to the experience with `max_num_hits` in `isaura`, where you may end up with isaura not processing any information due to each event exceeding the `max_num_hits`. This process includes no warning (as can be seen [here](https://github.com/next-exp/IC/blob/315ed2e0fd8254db9af44f0815cf1e4da966327b/invisible_cities/cities/components.py#L1329)). 

I see the same pattern of behaviour happening here, as both are 'filters' so to speak. I think it would be generally beneficial for the user to know when particular conditions are met that result in empty outputs (topological, empty hits, or anything in between). 

I also believe this is also beyond the scope of this PR, but it may be nice to consider some sort of generalised solution for this :sweat_smile:.

invisible_cities/reco/hits_functions.py
```diff
@@ -1,81 +1,255 @@
...
+                        , first.npeak, first.Xpeak, first.Ypeak
+                        , first.Z, raw_e_slice, cor_e_slice)
+
+    hits = hits.loc[mask_thresh].copy()
+    qsum = np.nansum(hits.Q) + EPSILON
+
+    # EPSILON added to avoid division by zero
+    # This should be only necessary for Ec, but we apply it to E to be safe
+    hits.loc[:, "E" ] = hits.Q / qsum * (raw_e_slice + EPSILON)
+    hits.loc[:, "Ec"] = hits.Q / qsum * (cor_e_slice + EPSILON)
```
jwaiton: This has been altered, but the underlying issue still remains:
> I find expressing these sorts of operations without explicit brackets to be quite confusing, but this is just personal preference.

I guess what I'm referencing is instead possible replace it with:
`hits.Q / (qsum * (raw_e_slice + EPSILON))`,
although this now looks a bit busy.

invisible_cities/cities/beersheba.py
jwaiton: Any update on this? :)

jwaiton (changes requested):
New commits resolve most of the issues, although a few still remain to be addressed (including two from just after the original review, in case they're missed) :)

invisible_cities/reco/hits_functions.py
gonzaponte: `*` and `/` have the same precedence, therefore a combination of them is evaluated from left to right like:

`(hits.Q / qsum) * (raw_e_slice + EPSILON)`

I'll add the parentheses if you think it's worth it

jwaiton: I think this is just a bias of mine, so feel free to ignore this. 

I like to remove all doubt because I often get it wrong (as I did in my first comment :sweat_smile:).

jacg: FWIW, I'm in favour of the parentheses: they reduce the chance of a human parsing it incorrectly.

Real Programmers (TM) know their precedence tables by heart, never make a mistake in parsing complex expressions and frown upon superfluous (for the compiler) hints. But a fundamental philosophy of IC's design is to cater for real people, not Real Programmers (TM).

invisible_cities/cities/beersheba.py
gonzaponte: I think this had something to do with `pandas` doing something wrong. There is some operation (probably `groupby`) that changes the type of the variable. The end result is the same, just of type `float` instead of `int`. After inspection, I've updated the test files to contain ints instead of floats and seems to work, so I'll push it.

invisible_cities/cities/components.py
gonzaponte: > `peak` uses class components like times and pmts. I wasn't sure where these came from or how they were defined, and my IDE couldn't pick up their origin.

These are in the constructor. Perhaps it doesn't peak them up because it derives from a base class?

I am adding some documentation for the other bits.

invisible_cities/cities/components.py
```diff
@@ -934,10 +947,30 @@ def hit_builder(dbfile, run_number, drift_v,
...
+    - For each S2:
+      - Compute the overall position of the signal according to `global_reco`
+        (typically barycenter in XYZ)
+      - For each (rebinned) slice of the S2:
+        - Clusterize the SiPM responses according to `slice_reco`
+          - Failing XY reconstructions (e.g. not enough SiPMs with signal)
+            generate "empty" (a.k.a. NN) clusters
+        - Assign each cluster the corresponding fraction of the energy in the
+          slice
+    """
```
jwaiton: The documentation is good, but the outer function is missing `typing` type hints (I'm not sure if this matters here). 

The documentation added to `components.py` differs from the 'standard' structure used in most of the city/city related functions. 

This is a consistent pattern I've noticed across IC (the documentation often differs in style and verbosity from file to file). Perhaps I should add an issue addressing this?

invisible_cities/reco/hits_functions.py
jwaiton: As discussed offline, the `Filters` group does what I've suggested. There is an issue where the `Filters` tables are empty in isaura files when they shouldn't be (hence my confusion), I've added an issue for this (#939).

On the other hand, with some further reading I've noticed that the `max_num_hits` cut is applied in such a manner within isaura that it would be considered within the `topology_select` table, rather than the `hits_select` table, which I believe should be adjusted. I'll look into this.

jwaiton (commented):
I have just one relevant comment, and another related to more underlying issues that are highlighted in this PR. Other than these, I'm happy with the PR and will approve once they've been noted :)

jwaiton (approved):
This PR begins the much desired process of removing the `Hit` class from IC, including changes across many cities and functions to accommodate this removal while retaining the same functionality.

It also adds documentation to some darker areas of the code related to these changes, and has highlighted some new issues (#939, #936).

Approved! Great work, I'm excited for phase 2 :smile_cat:

PR #917, opened 2024-11-06
Add lazy `PMap` reader

gonzaponte:
Cities that read `PMap`s consume a large amount of RAM because all events are kept in memory simultaneously. This PR adds new functions to read the `PMap`s on an event-by-event basis, in a lazy manner. The old readers have been renamed using the suffix `_eager` for clarity, while the new ones contain the suffix `_lazy`. The names of the old readers have been reused in the interfacing functions that allow the user to switch between the lazy and the eager readers. The readers' default to the eager mode for backward compatibility. However, the cities will read them lazily to avoid large memory allocations.

Moreover, the lazy versions contain two arguments to read a subset of all pmaps. These are `skip`, and `n`, which will skip the first `skip` events and read the following `n` events.

This PR has been tested in production mode.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -570,7 +570,7 @@ def pmap_from_files(paths):
             for evtinfo, (evt, pmap) in zip(event_info, pmaps):
                 event_number, timestamp = evtinfo.fetch_all_fields()
                 if event_number != evt:
-                    raise RuntimeError("Inconsistent data: event number mismatch")
+                    raise InvalidInputFileStructure("Inconsistent data: event number mismatch")
                 yield dict(pmap=pmap, run_number=run_number,
                            event_number=event_number, timestamp=timestamp)
 
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -554,7 +554,7 @@ def wf_from_files(paths, wf_type):
 def pmap_from_files(paths):
     for path in paths:
         try:
-            pmaps = load_pmaps(path)
+            pmaps = load_pmaps(path, lazy=True)
         except tb.exceptions.NoSuchNodeError:
             continue
 
@@ -567,11 +567,11 @@ def pmap_from_files(paths):
             except IndexError:
                 continue
 
-            check_lengths(event_info, pmaps)
-
-            for evtinfo in event_info:
+            for evtinfo, (evt, pmap) in zip(event_info, pmaps):
                 event_number, timestamp = evtinfo.fetch_all_fields()
-                yield dict(pmap=pmaps[event_number], run_number=run_number,
+                if event_number != evt:
+                    raise RuntimeError("Inconsistent data: event number mismatch")
+                yield dict(pmap=pmap, run_number=run_number,
                            event_number=event_number, timestamp=timestamp)
 
 
--- a/invisible_cities/io/pmaps_io.py
+++ b/invisible_cities/io/pmaps_io.py
@@ -78,8 +78,9 @@ def _make_tables(hdf5_file, compression):
     return pmp_tables
 
 
-def load_pmaps_as_df(filename):
-    return load_pmaps_as_df_eager(filename)
+def load_pmaps_as_df(filename, lazy=False):
+    loader = load_pmaps_as_df_lazy if lazy else load_pmaps_as_df_eager
+    return loader(filename)
 
 
 def load_pmaps_as_df_eager(filename):
@@ -115,8 +116,9 @@ def _build_ipmtdf_from_sumdf(sumdf):
     return ipmtdf
 
 
-def load_pmaps(filename):
-    return load_pmaps_eager(filename)
+def load_pmaps(filename, lazy=False):
+    loader = load_pmaps_lazy if lazy else load_pmaps_eager
+    return loader(filename)
 
 
 def load_pmaps_eager(filename):
--- a/invisible_cities/io/pmaps_io.py
+++ b/invisible_cities/io/pmaps_io.py
@@ -78,7 +78,7 @@ def _make_tables(hdf5_file, compression):
     return pmp_tables
 
 
-def load_pmaps_as_df(filename):
+def load_pmaps_as_df_eager(filename):
     with tb.open_file(filename, 'r') as h5f:
         pmap  = h5f.root.PMAPS
         to_df = pd.DataFrame.from_records
@@ -96,9 +96,9 @@ def _build_ipmtdf_from_sumdf(sumdf):
     ipmtdf['npmt'] = -1
     return ipmtdf
 
-def load_pmaps(filename):
+def load_pmaps_eager(filename):
     pmap_dict = {}
-    s1df, s2df, sidf, s1pmtdf, s2pmtdf = load_pmaps_as_df(filename)
+    s1df, s2df, sidf, s1pmtdf, s2pmtdf = load_pmaps_as_df_eager(filename)
     # Hack fix to allow loading pmaps without individual pmts
     if s1pmtdf is None: s1pmtdf = _build_ipmtdf_from_sumdf(s1df)
     if s2pmtdf is None: s2pmtdf = _build_ipmtdf_from_sumdf(s2df)
--- a/invisible_cities/io/pmaps_io.py
+++ b/invisible_cities/io/pmaps_io.py
@@ -106,7 +106,7 @@ def load_pmaps_as_df_eager(filename):
                 to_df(pmap.S2Pmt.read()) if 'S2Pmt' in pmap else None)
 
 
-def load_pmaps_as_df_lazy(filename):
+def load_pmaps_as_df_lazy(filename, skip=0, n=None):
     def read_event(table, event):
         if table is None: return None
         records = table.read_where(f"event=={event}")
@@ -118,6 +118,8 @@ def load_pmaps_as_df_lazy(filename):
 
         events = h5f.root.Run.events.read(field="evt_number")
         tables = [getattr(h5f.root.PMAPS, table, None) for table in tables]
+        n = events.size if n is None else n
+        events = events[skip : skip+n]
         for event in events:
             yield tuple(read_event(table, event) for table in tables)
 
@@ -130,9 +132,9 @@ def _build_ipmtdf_from_sumdf(sumdf):
     return ipmtdf
 
 
-def load_pmaps(filename, lazy=False):
+def load_pmaps(filename, lazy=False, **kwargs):
     loader = load_pmaps_lazy if lazy else load_pmaps_eager
-    return loader(filename)
+    return loader(filename, **kwargs)
 
 
 def load_pmaps_eager(filename):
@@ -180,8 +182,8 @@ def load_pmaps_eager(filename):
     return pmap_dict
 
 
-def load_pmaps_lazy(filename):
-    for (s1df, s2df, sidf, s1pmtdf, s2pmtdf) in load_pmaps_as_df_lazy(filename):
+def load_pmaps_lazy(filename, skip=0, n=None):
+    for (s1df, s2df, sidf, s1pmtdf, s2pmtdf) in load_pmaps_as_df_lazy(filename, skip, n):
         # Hack fix to allow loading pmaps without individual pmts
         if s1pmtdf is None: s1pmtdf = _build_ipmtdf_from_sumdf(s1df)
         if s2pmtdf is None: s2pmtdf = _build_ipmtdf_from_sumdf(s2df)
--- a/invisible_cities/io/pmaps_io.py
+++ b/invisible_cities/io/pmaps_io.py
@@ -78,6 +78,10 @@ def _make_tables(hdf5_file, compression):
     return pmp_tables
 
 
+def load_pmaps_as_df(filename):
+    return load_pmaps_as_df_eager(filename)
+
+
 def load_pmaps_as_df_eager(filename):
     with tb.open_file(filename, 'r') as h5f:
         pmap  = h5f.root.PMAPS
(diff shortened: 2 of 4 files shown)


invisible_cities/io/pmaps_io_test.py
```diff
@@ -201,6 +205,18 @@ def test_load_pmaps_as_df_lazy(KrMC_pmaps_filename, KrMC_pmaps_dfs):
...
 
+def test_load_pmaps_as_df_eager_lazy(KrMC_pmaps_filename):
+    eager = pmpio.load_pmaps_as_df(KrMC_pmaps_filename, lazy=False)
+    lazy  = pmpio.load_pmaps_as_df(KrMC_pmaps_filename, lazy=True )
+    assert len(eager) == 5
+    assert all(isinstance(item, pd.DataFrame) for item in eager)
+
+    assert isinstance(lazy, Generator)
+    element = next(lazy)
+    assert len(element   ) == 5
```
jwaiton: Is there a reason this spacing after `element` is the way it is? It doesn't appear to align with anything

invisible_cities/io/pmaps_io_test.py
```diff
@@ -214,14 +244,46 @@ def test_load_pmaps(output_tmpdir, data):
...
     assert len(read_pmaps) == len(true_pmaps)
     assert np.all(list(read_pmaps.keys()) == event_numbers)
     for true_pmap, read_pmap in zip(true_pmaps, read_pmaps.values()):
         assert_PMap_equality(read_pmap, true_pmap)
 
 
+def test_load_pmaps_lazy(KrMC_pmaps_filename):
+    """Ensure the lazy and non-lazy versions provide the same result"""
+    pmaps_eager = pmpio.load_pmaps_eager(KrMC_pmaps_filename)
+    pmaps_lazy  = pmpio.load_pmaps_lazy (KrMC_pmaps_filename)
```
jwaiton, line 375: Should these not follow the new format shown in the test below: `load_pmaps(..., lazy=False/True)`

invisible_cities/io/pmaps_io_test.py
```diff
@@ -214,14 +244,46 @@ def test_load_pmaps(output_tmpdir, data):
...
+    # combine all events in the same dictionary
+    pmaps_lazy = dict(pmaps_lazy)
+
+    assert len(pmaps_lazy) == len(pmaps_eager)
+    for evt, pmap_lazy in pmaps_lazy.items():
+        assert evt in pmaps_eager
+        assert_PMap_equality(pmap_lazy, pmaps_eager[evt])
+
+
+def test_load_pmaps_eager_lazy(KrMC_pmaps_filename):
```
jwaiton: This test appears to be ensuring that the content from `load_pmaps` is as expected (type and length-wise), but the test name is a big vague in that sense. Especially since the previous test (`test_load_pmaps_lazy`) is almost the same name, but accomplishes a different task. Maybe this should be changes?

invisible_cities/io/pmaps_io_test.py
```diff
@@ -203,7 +233,7 @@ def test_load_pmaps_as_df_without_ipmt(KrMC_pmaps_without_ipmt_filename, KrMC_pm
 
 
 @given(stg.data())
-def test_load_pmaps(output_tmpdir, data):
+def test_load_pmaps_eager(output_tmpdir, data):
```
jwaiton: Likewise in naming convention for this test, its a bit vague since it matches the other related tests in naming convention but differs in content. 

It appears to be writing a pmap, then reading it in and comparing it to itself 'pre-writing'. Is that correct? In that case, is there a reason that the pmap is generated here explicitly but in the other tests `KrMC_pmaps_filename` is used?

jwaiton (commented):
I've read through the PR, very interesting work! My comments at the moment are mostly cosmetic/about tests. 

I have read all the new lazy functions, but my knowledge of generators is a bit murky so I need to take more time to understand exactly how they work here in the pmap loader.

invisible_cities/io/pmaps_io_test.py
gonzaponte: Probably a leftover from a previous version. Will fix.

invisible_cities/io/pmaps_io_test.py
gonzaponte: I prefer to test the functions directly. A separate test ensures that `load_pmaps` (meant to be the interface to these functions(*)) works correctly with both values of the `lazy` parameter. But that's another layer. Testing the functions directly is more robust; in theory, something in `load_pmaps` could prevent errors in one of the inner functions from being spotted.

(*) Perhaps, we should "hide" the inner functions by prefixing an underscore to their names. This indicates the user that they should use `load_pmaps` with the corresponding value of `lazy` and not `load_pmaps_whatever`. What do you think?

invisible_cities/io/pmaps_io_test.py
gonzaponte: I'm struggling to find a good name. Do you have any suggestions?

invisible_cities/io/pmaps_io_test.py
gonzaponte: Hmm, I didn't read the test, actually, I just renamed it and refactored it to make it work with the new version. But I don't see a reason why it shouldn't follow the same logic as the other ones, so I will change it.

gonzaponte: I just realized that some of other tests are wrong, as they compare the output of `load_pmaps_as_df_eager` with the output of a fixture, which only reads pmaps using that very function. So it's testing nothing essentially. I will refactor this test to make it simpler (but keep the logic) and fix the other ones.

invisible_cities/io/pmaps_io_test.py
jwaiton: I guess `test_load_pmaps_returns_correct_format` or `...correct_type_and_shape`? The `lazy` and `eager` are both tested so I don't think they need to be specified distinctly.

invisible_cities/io/pmaps_io_test.py
jwaiton: That makes sense. Testing the inner function explicitly is important :+1: 

On the (*), I don't see why not other than convention. Is the use of leading underscores implemented anywhere else in IC? For some users this may be more of a hindrance than a help if they are unfamiliar with its effects.

gonzaponte: This is more of a Python convention than an IC one. Even though we don't follow PEP8 entirely, this specific rule is a common convention in the Python community. It's described [here](https://peps.python.org/pep-0008/#descriptive-naming-styles) (last paragraph).

jwaiton: I think I'm being overly concerned. My worry was that (as is described in the PEP8 doc you linked) adding a leading underscore means indiscriminate importing (`from components import *`) will not import the corresponding inner function. I just assumed that could cause issues with users if it hasn't been used much in IC before and they aren't familiar with PEP8.

I think it's a good change to add the underscore, I'm just concerned in case notebooks start crashing and people aren't sure why.

gonzaponte: The takeaway message here is that using this convention is just a way of expressing something. That something is that users should exclusively use the `load_pmaps` function, and not the other ones. We don't forbid their use, just discourage it. If a user insists on calling the inner functions directly, they will have to import them explicitly. Unfortunately, there is no other solution for the case you mention than learning a bit more about the language.

invisible_cities/io/pmaps_io.py
```diff
@@ -78,8 +80,23 @@ def _make_tables(hdf5_file, compression):
     return pmp_tables
 
 
-def load_pmaps_as_df(filename):
+def check_file_integrity(file):
+    events_run      = file.root.Run  .events.read(field="evt_number")
+    events_pmaps_s1 = file.root.PMAPS.S1    .read(field="event")
+    events_pmaps_s2 = file.root.PMAPS.S2    .read(field="event")
+    if set(events_run) != set(events_pmaps_s1.tolist() + events_pmaps_s2.tolist()):
+        raise InvalidInputFileStructure("Inconsistent data: event number mismatch")
```
jwaiton, line 88: This function isn't tested explicitly, but is in the `eager` and `lazy` df functions below. Should it be tested explicitly? For example, does it behave correctly when there are multiple S2 peaks per event?

gonzaponte: Yes, it should be tested explicitly. I will add some tests.

I had to put this on hold for a bit, but I was trying to rewrite the tests that were "testing nothing" as we discussed in a separate comment.

invisible_cities/io/pmaps_io_test.py
```diff
@@ -29,9 +32,102 @@
...
+    return s1, s2, si, s1pmt, s2pmt
+
+
+@fixture(scope="session")
+def two_pmaps(two_pmaps_evm, two_pmaps_dfs, output_tmpdir):
+    pmap_filename = os.path.join(output_tmpdir, "two_pmaps.h5")
+    run_number    = 0 # irrelevant
+    timestamp     = 0 # irrelevant
+
+    with tb.open_file(pmap_filename, "w") as output_file:
```
jwaiton, line 120: A very nice set of fixtures!

If you run the tests multiple times, this method here (line 121 onwards, not sure why it cropped it like this) will overwrite the existing `"two_pmaps.h5"` file every time right? I assume its a very small gain, but might it be worth checking that the file doesnt exist before writing to it?

invisible_cities/io/pmaps_io_test.py
```diff
@@ -320,6 +320,18 @@ def test_load_pmaps_as_df_lazy(KrMC_pmaps_filename):
         assert_dataframes_equal(df_lazy, df_eager)
 
 
+@mark.parametrize("skip" , (0, 1, 2, 3))
+@mark.parametrize("nread", (0, 1, 2, 3))
+def test_load_pmaps_as_df_lazy_subset(KrMC_pmaps_filename, skip, nread):
+    """Ensure the lazy and non-lazy versions provide the same result"""
```
jwaiton: The docstring here implies that the `eager` version is implemented, which isn't true right? It just checks that the subset output from the lazy method has the same number of unique entries as is expected by `nread`.

invisible_cities/io/pmaps_io_test.py
```diff
@@ -371,6 +383,16 @@ def test_load_pmaps_lazy(KrMC_pmaps_filename):
         assert_PMap_equality(pmap_lazy, pmaps_eager[evt])
 
 
+@mark.parametrize("skip" , (0, 1, 2, 3))
+@mark.parametrize("nread", (0, 1, 2, 3))
+def test_load_pmaps_lazy_subset(KrMC_pmaps_filename, skip, nread):
```
jwaiton, line 388: Same as comment on `test_load_pmaps_as_df_lazy_subset`, except instead it just checks that the lengths are the same.

invisible_cities/cities/components.py
```diff
@@ -567,11 +567,11 @@ def pmap_from_files(paths):
...
 
-            check_lengths(event_info, pmaps)
-
-            for evtinfo in event_info:
+            for evtinfo, (evt, pmap) in zip(event_info, pmaps):
                 event_number, timestamp = evtinfo.fetch_all_fields()
-                yield dict(pmap=pmaps[event_number], run_number=run_number,
+                if event_number != evt:
+                    raise RuntimeError("Inconsistent data: event number mismatch")
+                yield dict(pmap=pmap, run_number=run_number,
```
jwaiton: This check for inconsistent data is implemented, but not tested. Similar functionality is tested for `check_file_integrity`, but that only appears to be implemented for `..._as_df_...` functions. Is it possible for these to be merged in some capacity?

invisible_cities/io/pmaps_io.py
```diff
@@ -141,6 +182,19 @@ def load_pmaps(filename):
     return pmap_dict
 
 
+def load_pmaps_lazy(filename, skip=0, n=None):
```
jwaiton, line 280: I would add in the PR first comment a note of what these two parameters (`skip` and `n`) do for posterity, if these functions aren't going to have docstrings.

jwaiton (commented):
It looks good! I like the new tests and fixtures. Just a few comments.

invisible_cities/io/pmaps_io_test.py
gonzaponte: The file won't exist, because it's being written in a temp folder, which is different every time we run the tests. As you say, it's a very small file, so the time spent here over the time it takes the test suite to run is negligible.

invisible_cities/io/pmaps_io.py
gonzaponte: I'll do both (write it in the PR and add docstrings).

invisible_cities/cities/components.py
gonzaponte: I don't know what you mean by merging here, but I'll add a test for this reader.
Regarding `check_file_integrity`, it's only implemented for the `as_df` readers because the other ones call these, so that check is performed in any case. Is that what you were referring to?

invisible_cities/io/pmaps_io_test.py
gonzaponte: Copy-pasting has its risks... :)

invisible_cities/cities/components.py
jwaiton: Yes that's what I was referring to. Both errors provided the same message with different error types, so I assumed the check for `Inconsistent data` could be merged into a function. On second reading I can see that doesn't make sense, apologies.

gonzaponte: I will also change the Exception to be the same in both cases.

jwaiton (approved):
The lazy `PMap` reader allows for memory to be utilised more efficiently when reading in PMaps, and is fully tested with documentation to explain how the relevant functions work. Good job! Approved :smile:

PR #918, opened 2024-11-07
Mark scaled_pedestal tests as flaky

gonzaponte:
These tests fail fairly frequently because they rely on statistics. We will review them (#919), but for the time being, they are marked as flaky, so they don't bother us when they fail sporadically.

### DIFF
--- a/invisible_cities/calib/spe_response_test.py
+++ b/invisible_cities/calib/spe_response_test.py
@@ -191,6 +191,7 @@ def dark_spectrum_global():
     return parameters, pedestal + signal
 
 
+@flaky(max_runs=2)
 def test_scaled_dark_pedestal_pedestal(dark_spectrum_local):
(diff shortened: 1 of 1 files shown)


jwaiton:
A simple fix to a problem I've had many times.

Are there any more tests that should be marked as flaky? If so, it may be nice to contain all of them in one PR. If not I'll approve the changes post-haste.

gonzaponte:
Hmm, the database tests sometimes fail due to connection errors. I have a branch somewhere where they are marked as `xfail` allowing only those types of errors, but I think this falls under a slightly different category. The db tests probably need to have their own workflow and not be executed together with the rest of the test suite.

Other than that, I am not aware of any other flaky tests that aren't marked as such, but they might pop up in the future.

jwaiton (approved):
A simple fix that improves testing consistency, approved :)

PR #921, opened 2024-11-12
Silence warnings produced in test suite

### LINKED ISSUE #875: Fix test suite warnings
The test suite is producing several warnings. Wouldn't it be nice to get a clean and clear green light when they pass? Let's aim for that!

gonzaponte:
This was creating too much noise in the test suite output. This PR silences all warnings. For this, I introduced a `dataclass` holding all warning filters. This allows for the reuse of filters while giving them a meaningful name.

Closes #875.

### DIFF
--- a/.github/workflows/test_suite.yml
+++ b/.github/workflows/test_suite.yml
@@ -24,12 +24,25 @@ jobs:
     # - name: Fix Conda permissions on macOS
     #   run: sudo chown -R $UID $CONDA
     #   if: runner.os == 'macOS'
+
     - name: Install IC
       run: |
         source $CONDA/etc/profile.d/conda.sh
         source manage.sh work_in_python_version_no_tests ${{ matrix.python-version }}
+
+
     - name: Run tests
       run: |
+        set -o pipefail # necessary for test failure to propagate to `tee`
         source $CONDA/etc/profile.d/conda.sh
         source manage.sh work_in_python_version_no_tests ${{ matrix.python-version }}
-        PYTEST_ADDOPTS=--color=yes HYPOTHESIS_PROFILE=travis-ci bash manage.sh run_tests_par
+        PYTEST_ADDOPTS=--color=yes HYPOTHESIS_PROFILE=travis-ci bash manage.sh run_tests_par | tee pytest_output
+
+    - name: Warning summary
+      run: |
+        ./scripts/check_warnings pytest_output | tee warnings
+        if [ -s warnings ]; then
+            echo "::warning ::There are warnings raised in the test suite"
+        else
+            echo "No warnings were raised."
+        fi
--- a/.github/workflows/test_suite.yml
+++ b/.github/workflows/test_suite.yml
@@ -38,6 +38,50 @@ jobs:
         source manage.sh work_in_python_version_no_tests ${{ matrix.python-version }}
         PYTEST_ADDOPTS=--color=yes HYPOTHESIS_PROFILE=travis-ci bash manage.sh run_tests_par | tee pytest_output
 
+    - name: Test warning-catching script
+      run: |
+        source $CONDA/etc/profile.d/conda.sh
+        source manage.sh work_in_python_version_no_tests ${{ matrix.python-version }}
+
+        # write dummy tests
+        cat > dummy.py <<EOF
+        from warnings import warn
+        def test_warn():
+          warn("bla bla bla", UserWarning)
+        def test_pass():
+          pass
+        EOF
+
+        echo "###################################"
+        cat dummy.py
+        echo "###################################"
+
+        pytest dummy.py -k "warn" > warn
+        pytest dummy.py -k "pass" > pass
+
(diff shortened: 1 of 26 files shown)


gonzaponte:
I've also introduced a GHA step to check for warnings in the test suite. This does not block the merge, it simply produces a warning that's not even too obvious. One has to enter the GHA log to see it, but I haven't found a better way.

.github/workflows/test_suite.yml
```diff
@@ -24,12 +24,25 @@ jobs:
...
+        PYTEST_ADDOPTS=--color=yes HYPOTHESIS_PROFILE=travis-ci bash manage.sh run_tests_par | tee pytest_output.txt
+
+    - name: Warning summary
+      run: |
+        ./scripts/check_warnings pytest_output.txt | tee warnings
+        if [ -s warnings ]; then
+            echo "::warning ::There are warnings raised in the test suite"
+        else
+            echo "No warnings were raised."
+        fi
```
jwaiton: It feels a bit silly and possible convoluted to ask for a test of a warning script for tests. But is there any way we could test that this output is provided if an error does occur in pytest? Possibly by writing a false warning to a file and running `check_warnings` on it to ensure you get the appropriate output?

invisible_cities/cities/beersheba_test.py
```diff
@@ -103,7 +108,9 @@ def test_beersheba_exact_result( deco
                 assert_tables_equality(got, expected, rtol=1e-6)
 
 
-@mark.filterwarnings("ignore:.*not of kdst type.*:UserWarning")
+@ignore_warning.no_config_group
+@ignore_warning.str_length
+@ignore_warning.not_kdst
```
jwaiton, line 113: It feels a bit odd to have separate `ignore_warning`s setup on each test function. 

Would it be possible to setup some sort of child class of the `ignore_warning` class that includes all the standard warnings we want to avoid  (`no_config group`, `str_length`, `not_kdst`, and possibly `no_hits`) and apply it as a general filter to most of these functions? 

This is assuming that applying an `ignore_warning.not_kdst` to a function that doesn't currently have it wont be an issue, which I could be wrong about.

invisible_cities/cities/esmeralda_test.py
```diff
@@ -176,6 +186,7 @@ def test_esmeralda_all_hits_after_drop_voxels(esmeralda_config, Th228_hits, conf
 
 
 #TODO: refactor paolina to include this as a filter
+@ignore_warning.no_config_group
```
jwaiton, line 189: similarly, this ignore_warning for no_config group is (I think) applied to every test in this file. Is it possible to make that a single line that just states "ignore this warning for all tests in `esmeralda_test.py`"?

jwaiton (commented):
Looks good generally! I just have a few comments.

invisible_cities/cities/beersheba_test.py
gonzaponte: That could hide warnings that might be meaningful to the developer. There is already a way to silence warnings on a global level in the `pytest.ini` file, but that would silence them everywhere, even where we don't expect one. This way we document which warnings are expected while showing that we understand them (hopefully).

invisible_cities/cities/esmeralda_test.py
gonzaponte: I don't know, there might be, but I am hesitant to do so for a similar reason. If someone in the future adds a test that is not supposed to raise a warning, but it does, the developer will not see it.

.github/workflows/test_suite.yml
gonzaponte: Hmm, I agree it would be nice, but I'm not sure how to do it without making it silly. I tested it in my fork before opening the PR. See for instance [this](https://github.com/gonzaponte/IC/actions/runs/11820831339) workflow run. Images attached for reference.
image
image

jwaiton: That is quite a large output for a single test case. My guess would be that in the GHA you would run `pytest` in a particular directory with a single (possibly a dummy) test that is guaranteed to produce a warning, and check the corresponding behaviour from `check_warnings` occurs. This does feel a bit overbearing though, so if you think it's too much I'm happy to approve the PR :)

gonzaponte: Just to clarify, this is the output for the full test suite. I've added a test in the workflow. This is to avoid having unnecessary files in the repo, so they are created on the fly and they only exist in the GHA runner.

jwaiton: Yes sorry thats what I meant. The output of the full test suite is quite heavy for a simple warning test. The new commit looks good and keeps the repo clean :)

When it clears the tests, I'll approve the PR

jwaiton (approved):
This PR reduces clutter caused by warnings in the test suite. It also includes new GHAs that look for new warnings not silenced in this PR, and test that the 'warning checker' functions as expected. Good job!

PR #922, opened 2024-11-13
Bump actions/checkout version in GHA workflows

gonzaponte:
The current version issues the following warning:

```
The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2.
For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/
```

Bumpng to the latest version seems to get rid of it.

### DIFF
--- a/.github/workflows/check_db_calibration.yml
+++ b/.github/workflows/check_db_calibration.yml
@@ -21,7 +21,7 @@ jobs:
         python-version: [3.8]
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
     - id: files
       uses: jitterbit/get-changed-files@v1
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


jwaiton (approved):
Good change, we move from node16 to node20 with checkout@v4. From reading the [release](https://github.com/actions/checkout/releases) changelogs from v2 -> v3 -> v4, it seems like there are no backwards compatibility issues and the tests execute as expected.

jwaiton:
There may be more complex behaviour in the differences between node16 & 20 but I can't find anything significant in documentation that warns of this, or complaints online of said incompatibilities between versions. Again, the tests run as expected looking at the output so I'm happy :)

gonzaponte:
As far as I know, that should not affect the user interface, so it shouldn't matter.

PR #927, opened 2024-12-30
Update tests for seed functions 

carhc:
This PR updates some tests in `invisible_cities/reco/krmap_functions_test.py` of PR #872 . There were some missing cases and therefore the implementation was not completely tested.


gonzaponte (approved):
Perfect! Good job.

PR #937, opened 2025-07-17
Add more normalization strategies

gonzaponte:
Update normalization strategies to include more useful options:
- Median: less sensitive to outliers than mean
- Region: use a specific region of the chamber to compute the reference value. It uses the median for the values within the specified region. There are two different ways of specifying such region:
  - xrange + yrange: a rectangle
  - origin + radius: a circle

### DIFF
--- a/invisible_cities/reco/corrections.py
+++ b/invisible_cities/reco/corrections.py
@@ -228,7 +228,7 @@ def get_normalization_factor(map_e0    : ASectorMap,
     if norm_strat is NormStrategy.max:
         norm_value =  map_e0.e0.max().max()
     elif norm_strat is NormStrategy.mean:
-        norm_value = np.mean(np.mean(map_e0.e0))
+        norm_value = np.nanmean(map_e0.e0.values.flatten())
     elif norm_strat is NormStrategy.kr:
         norm_value = 41.5575 * units.keV
     elif norm_strat is NormStrategy.custom:
--- a/invisible_cities/reco/corrections.py
+++ b/invisible_cities/reco/corrections.py
@@ -8,6 +8,7 @@ from   typing      import Callable
 from   typing      import Optional
 
 from .. core.core_functions import in_range
+from .. core.core_functions import shift_to_bin_centers
 from .. core                import system_of_units      as units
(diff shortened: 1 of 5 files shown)


invisible_cities/reco/corrections.py
```diff
@@ -228,17 +233,43 @@ def get_normalization_factor(map_e0    : ASectorMap,
...
+            selx = in_range(binsx, *norm_options["xrange"])
+            sely = in_range(binsy, *norm_options["yrange"])
+            e0s  = map_e0.e0.loc[sely, selx]
+        elif "origin" in norm_options and "radius" in norm_options:
+            x0, y0 = norm_options["origin"]
+            x , y  = np.meshgrid(binsx, binsy)
+            sel    = in_range((x-x0)**2 + (y-y0)**2, 0, norm_options["radius"]**2)
+            e0s    = map_e0.e0.values[sel]
+        else:
+            s  = "Invalid options for noramlization strategy `region`. Valid options:\n"
```
carhc:
```suggestion
            s  = "Invalid options for normalization strategy `region`. Valid options:\n"
```

invisible_cities/reco/corrections.py
```diff
@@ -268,9 +299,8 @@ def apply_all_correction_single_maps(map_e0         : ASectorMap,
...
     norm_strat : NormStrategy
         Provides the desired normalization to be used.
-    norm_value : Float(optional)
-        If norm_strat is selected to be custom, user must provide the
-        desired scale.
+    norm_options : dict
+        Extra normalization options for some strategies.
     krscale_output : Bool
         If true, the returned factor will take into account the scaling
         from pes to the Kr energy scale.
```
carhc, line 306: I would suggest deleting this in the docstring, also for `apply_all_correction`.

```suggestion

```

carhc: Oh, I see that it's already fixed for the later. Sorry!

carhc (approved):
Interesting tools to incorporate! Everything looks good and is well tested. Nice job! 

(my comments do not involve changing the code itself, so I approve it already)

PR #938, opened 2025-07-31
Bump miniconda version and use it in GHA

### LINKED ISSUE #931: Bump miniconda version
The current miniconda version (4.9.2) is over 4 years old and needs to be updated.

gonzaponte:
Our repository comes with a setup script that provides the user with `miniconda` if they don't have one. The version of the `miniconda` package we download has been fixed for far too long. This PR bumps this version to the latest version that still supports python 3.8. This closes #931, but we will also need to update python to a newer version, which should como with an update to the `miniconda` installer as well.


jwaiton (approved):
A much needed update, may the rollout of newer software versions continue 🙏

PR #942, opened 2025-10-02
Handle nested dictionaries in config parameter storage

### LINKED ISSUE #941: Configs stored in h5 files too short, losing parameter information
The ability to store configuration parameters was introduced in #891, with a maximum string length applied as `300` in this line:
[components.py - L224](https://github.com/next-exp/IC/blob/master/invisible_cities/cities/components.py#L224),

This is an issue specifically with Beersheba, where the `deconv_params` dictionary quickly overcomes this 300 character threshold.

A simple solution would be to increase this limit, but perhaps there is a more elegant solution in unpacking the dictionary parameters.

Ian0sborne:
This PR fixes #941 whereby some arguments in `write_city_configuration` were over the 300 character threshold. This issue occurred when passing dictionary arguments without unpacking them. This PR makes the `write_city_configuration` function sensitive to dictionary arguments and unpacks them, assigning sensible names. I have also added a case for this in the test function.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -211,10 +211,37 @@ def index_tables(file_out):
                 table.colinstances[colname].create_index()
 
 
+def dict_to_string(arg : dict,
+                   parent_key : str= ""):
+    '''
+    A function that recusively flattens nested dictionaries and converts the values to strings.
+
+    Each nested key is combined with its parent keys and associated with its respective value 
+    as a string.
+
+    Parameters
+    ----------
+    arg        : the dictionary to flatten which can contain nested dictionaries
+    parent_key : the string containing the prefix to use for keys in the flattened dictionary (default "").
+
+    Returns
+    -------
+    flat_dict : flattened dictionary containing no nested dictionaries and where all values are strings
+    '''
+    flat_dict = {}
+    for k, v in arg.items():
+        new_key = f"{parent_key}.{k}" if parent_key else k
+        if isinstance(v, dict):
+            flat_dict.update(dict_to_string(v, new_key))
+        else:
+            flat_dict[new_key] = str(v)
+    return flat_dict
+
+
 def write_city_configuration( filename : str
                             , city_name: str
                             , args     : dict):
-    args = {k: str(v) for k,v in args.items()}
+    args = dict_to_string(args)
 
     with tb.open_file(filename, "a") as file:
         df = (pd.Series(args)
--- a/invisible_cities/cities/components_test.py
+++ b/invisible_cities/cities/components_test.py
@@ -575,15 +575,10 @@ def test_write_city_configuration(config_tmpdir):
         assert str(value) == df.value.loc[var]
 
     # considering just the nested dictionary
-    var, value = list(args.items())[-1]
-    for var1, value1 in value.items():
-        if not isinstance(value1, dict):
-            assert f"{var}.{var1}" in df.index        
-            assert  str(value1) == df.value.loc[f"{var}.{var1}"]
-        else:
-            for var2, value2 in value1.items():
-                assert f"{var}.{var1}.{var2}" in df.index        
-                assert  str(value2) == df.value.loc[f"{var}.{var1}.{var2}"]
+    assert 'g.a'        in df.index  and str(args['g']['a'])           == df.value.loc['g.a']
+    assert 'g.b'        in df.index  and str(args['g']['b'])           == df.value.loc['g.b']
+    assert 'g.c.alpha'  in df.index  and str(args['g']['c']['alpha'])  == df.value.loc['g.c.alpha']
+    assert 'g.c.beta'   in df.index  and str(args['g']['c']['beta'])   == df.value.loc['g.c.beta']
 
 
 def test_copy_cities_configuration(config_tmpdir):
--- a/invisible_cities/cities/components_test.py
+++ b/invisible_cities/cities/components_test.py
@@ -560,6 +560,7 @@ def test_write_city_configuration(config_tmpdir):
         d = "two strings".split(),
         e = [1,2,3],
         f = np.linspace(0, 1, 5),
+        g = {'a' : 2, 'b' : 3, 'c' : {'alpha' : 5, 'beta' : 10}}
     )
     write_city_configuration(filename, city_name, args)
     with tb.open_file(filename, "r") as file:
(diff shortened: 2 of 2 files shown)


gonzaponte:
This looks pretty good.

I would it to be a bit more general, allowing an arbitrary level of nesting. I would go for a recursive approach.

Ian0sborne:
I have tried to come up with an iterative approach by adding a new function. Not sure if this is ugly though?

```
def flatten_dict(arg : dict,
                parent_key : str= "",
                dev : str= "."):
    flat_dict = {}
    for k, v in arg.items():
        new_key = f"{parent_key}{dev}{k}" if parent_key else k
        if isinstance(v, dict):
            flat_dict.update(flatten_dict(v, new_key))
        else:
            flat_dict[new_key] = str(v)
    return flat_dict
```

This would output something like:
image

Let me know what you think and I will implement it

invisible_cities/cities/components.py
```diff
@@ -211,10 +211,39 @@ def index_tables(file_out):
                 table.colinstances[colname].create_index()
 
 
+def flatten_dict(arg : dict,
```
gonzaponte: This name refers more to the solution you've had to implement for the case of nested dictionaries rather than the "problem" it solves. I suggest something along the lines of "stringify dict". Try to come up with a cool name :).

invisible_cities/cities/components_test.py
```diff
@@ -560,6 +560,7 @@ def test_write_city_configuration(config_tmpdir):
         d = "two strings".split(),
         e = [1,2,3],
         f = np.linspace(0, 1, 5),
+        g = {'a' : 2, 'b' : 3}
```
gonzaponte: Add yet another key `'c'` that holds a dictionary to make the nesting one level deeper.

invisible_cities/cities/components_test.py
```diff
@@ -568,8 +569,13 @@ def test_write_city_configuration(config_tmpdir):
...
     for var, value in args.items():
-        assert var in df.index
-        assert str(value) == df.value.loc[var]
+        if not isinstance(value, dict):
+            assert var in df.index
+            assert str(value) == df.value.loc[var]
+        else:
+            for var1, value1 in value.items():
+                assert f"{var}.{var1}" in df.index        
+                assert  str(value1) == df.value.loc[f"{var}.{var1}"]
```
gonzaponte: I think in this particular case it will be more readable to ignore the entry `"g"` in the loop and run the assertion separately after the loop, especially if you implement another level of nesting as suggested above.

gonzaponte (changes requested):
Apologies for the late reply. I've had a hectic couple of weeks.

The proposed solution is excellent, I think. Just a minor comment: you either remove the `dev` argument (what does it stand for?) or pass it as an argument when you call the function from within the loop. I don't think we want to make it this generic, but both approaches are fine by me.

Ian0sborne:
No need to apologise! 

You're absolutely right Gonzalo: the "dev" was meant to stand for "divider" (but I made a typo). There's no need for it to be an argument so I'll implement that and the rest of the changes now.

invisible_cities/cities/components_test.py
```diff
@@ -560,17 +560,31 @@ def test_write_city_configuration(config_tmpdir):
...
+    var, value = list(args.items())[-1]
+    for var1, value1 in value.items():
+        if not isinstance(value1, dict):
+            assert f"{var}.{var1}" in df.index        
+            assert  str(value1) == df.value.loc[f"{var}.{var1}"]
+        else:
+            for var2, value2 in value1.items():
+                assert f"{var}.{var1}.{var2}" in df.index        
+                assert  str(value2) == df.value.loc[f"{var}.{var1}.{var2}"]
+
```
gonzaponte, line 582: Since there are only 4 cases to test for `"g"`, I think it would be more readable to just lay them out explicitly.

invisible_cities/cities/components.py
```diff
@@ -211,10 +211,37 @@ def index_tables(file_out):
                 table.colinstances[colname].create_index()
 
 
+def dict_to_string(arg : dict,
+                parent_key : str= ""):
```
gonzaponte: alignment

invisible_cities/cities/components_test.py
Ian0sborne: Do you mean something like this?

```
    # considering just the nested dictionary
    assert str(args['g']['a']) == df.value.loc['g.a']
    assert str(args['g']['b']) == df.value.loc['g.b']
    assert str(args['g']['c']['alpha']) == df.value.loc['g.c.alpha']
    assert str(args['g']['c']['beta']) == df.value.loc['g.c.beta']
```

gonzaponte: yes, plus `'g.a' in df.index` and the alike. They can be and-ed in each assertion

Ian0sborne: Implemented!

gonzaponte: Perfect, can you improve the alignment to highlight the symmetries?

Ian0sborne: Something like this? Or would you like the individual statements aligned too (for `in` and `==`)?

image

gonzaponte: I would align all three: `in`, `and`, `==`

gonzaponte:
Sorry, I thought I have already commented here. This is almost ready, let's rewrite the commit history following [these](https://github.com/next-exp/IC/pull/943#issuecomment-3422512247) recommendations. After that, I'll approve it.

Ian0sborne:
Hi Gonzalo! Is the commit history okay now?

gonzaponte:
The second commit says "Alter function & test...", but the changes are more like "Introduce function and alter test ...". I suggest you put the second version of the test (the one prior to "simplification") already in the first commit and then introduce the final version of the function in the second commit, leaving the third one as it is now. Let me know if you need help with this.

Ian0sborne:
I'll try this now, will let you know if I need help. Thanks!

Ian0sborne:
I think this should be better now.

I was not able to put the second version of the testing function in the first commit because I committed it at the same time as the "final version" of the unpacking function. For this reason, I thought they would be better in the second commit together.

Let me know what you think of these changes

gonzaponte:
The history has been rewritten with peras to learn some magit.

gonzaponte (approved):
Handles nested dictionaries elegantly in the config output. And it's tested. Nice job!

gonzaponte:
@Luismiguelvillar please merge!

PR #943, opened 2025-10-07
Alter list type to Tuples for sample_width and bin_size in Beersheba. 

### LINKED ISSUE #908: Make `sample_width`, `bin_size`, `diffusion` & `drop_dist` tuples
All these parameters involved in Beersheba are of fixed size and we don't want them to be changed once they're assigned (immutable). Hence it would be beneficial to convert them to tuples as it would ensure they aren't altered in the running of the cities.

All of these parameters are defined within the [beersheba config](https://github.com/next-exp/IC/blob/master/invisible_cities/config/beersheba.conf), but converting them all will require ensuring that all functions in which each variable is used is altered across the entire repository.

Luismiguelvillar:
This PR fixes #908 by changing the lists parameters in the config file of Beersheba to Tuples. At the same time it corrects functions which take this parameters by changing their expected input from lists to Tuples.

### DIFF
--- a/invisible_cities/cities/beersheba.py
+++ b/invisible_cities/cities/beersheba.py
@@ -86,13 +86,12 @@ from .. types.symbols          import HitEnergy
 from .. types.symbols          import InterpolationMethod
 from .. types.symbols          import CutType
 from .. types.symbols          import DeconvolutionMode
-
+from .. types.ic_types         import Tuple2Dor3D
 
 from typing import Tuple
 from typing import List
 from typing import Optional
 from typing import Union
-
 
 def event_info_adder(timestamp : float, dst : pd.DataFrame):
     return dst.assign(time=timestamp/1e3, nsipm=0, Xrms=0, Yrms=0)
@@ -104,8 +103,8 @@ def deconvolve_signal(det_db           : pd.DataFrame,
                       e_cut            : float,
                       n_iterations     : int,
                       iteration_tol    : float,
-                      sample_width     : List[float],
-                      bin_size         : List[float],
+                      sample_width     : Tuple2Dor3D,
+                      bin_size         : Tuple2Dor3D,
                       satellite_params : Union[dict, NoneType],
                       diffusion        : Optional[Tuple[float, float, float]]=(1., 1., 0.3),
                       energy_type      : Optional[HitEnergy]=HitEnergy.Ec,
--- a/invisible_cities/reco/deconv_functions.py
+++ b/invisible_cities/reco/deconv_functions.py
@@ -25,7 +25,7 @@ from ..core .configure      import check_annotations
 from .. types.ic_types      import NoneType
 from .. types.symbols       import InterpolationMethod
 from .. types.symbols       import CutType
-
+from .. types.ic_types      import Tuple2Dor3D
 
 def collect_component_sizes(im_mask : np.ndarray) -> (np.ndarray, np.ndarray):
     '''
@@ -236,7 +236,7 @@ def drop_isolated_clusters(distance   :  List[float],
 
     return drop_event
 
-def deconvolution_input(sample_width : List[float     ],
+def deconvolution_input(sample_width : Tuple2Dor3D,
                         det_grid     : List[np.ndarray],
                         inter_method : InterpolationMethod = InterpolationMethod.cubic
                        ) -> Callable:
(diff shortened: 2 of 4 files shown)


invisible_cities/reco/deconv_functions.py
```diff
@@ -181,7 +181,7 @@ def drop_isolated_sensors(df : pd.DataFrame) -> pd.DataFrame:
     return drop_isolated_sensors
 
 
-def deconvolution_input(sample_width : List[float     ],
+def deconvolution_input(sample_width : Tuple[float     ],
```
gonzaponte: The correct type here would be a 2- or 3-tuple of floats. The correct way to type-hint it is
`Union[Tuple[float, float], Tuple[float, float, float]]`

I suggest you declare the type beforehand to reduce the noise. The same applies to `deconvolve` in this file and `deconvolve_signal` in `beersheba.py`.

invisible_cities/cities/beersheba.py
```diff
@@ -103,8 +105,8 @@ def deconvolve_signal(det_db           : pd.DataFrame,
                       e_cut            : float,
                       n_iterations     : int,
                       iteration_tol    : float,
-                      sample_width     : List[float],
-                      bin_size         : List[float],
+                      sample_width     : sampleWidth,
+                      bin_size         : Tuple[float],
```
gonzaponte: I believe `bin_size` has the same type hint as `sample_width`, doesn't it? Therefore the type should have a more general name. Also, you can import it from the other file.

Luismiguelvillar: It does make sense to do that (import the data type). But is not there a more "general file" in which I can declare the alias of this "custom" data type and import it from there? I think it is kind of messy to define the alias in, for example, `deconv_functions` and import it from there. But if you think that is fine I can defenitely do that.

gonzaponte: Good point. `types/ic_types.py` is probably the best place for this.

Luismiguelvillar: perfect! Thank you!

gonzaponte:
Perfect. Let's:
- rebase this branch onto master (you will need to resolve a conflict)
- re-write the commit history
  - Capitalize first letter
  - Use an imperative voice
  - Be more specific about the changes. For instance:
    - "alter list to tuples" could benefit from a bit of more information like you are referring to the type hints, and specifically in this corner of the project.
    - "implemented suggested changes" needs a lot of context to know what it is about. Also, there is no need to reference if something was a reviewer suggestion or not. Focus the message on describing the changes implemented in the commit. 
    - Your third commit message is pretty good. The "in ic_types.py" bit is unnecessary (an implementation detail), but that's ok.

After that, I'll approve the PR.

Luismiguelvillar:
> Perfect. Let's:
> 
> * rebase this branch onto master (you will need to resolve a conflict)
> * re-write the commit history
>   
>   * Capitalize first letter
>   * Use an imperative voice
>   * Be more specific about the changes. For instance:
>     
>     * "alter list to tuples" could benefit from a bit of more information like you are referring to the type hints, and specifically in this corner of the project.
>     * "implemented suggested changes" needs a lot of context to know what it is about. Also, there is no need to reference if something was a reviewer suggestion or not. Focus the message on describing the changes implemented in the commit.
>     * Your third commit message is pretty good. The "in ic_types.py" bit is unnecessary (an implementation detail), but that's ok.
> 
> After that, I'll approve the PR.

I have already done it! Do you think this is fine?

gonzaponte:
That's great. Just to make sure I didn't convey the wrong message, I didn't mean to say that you should squash everything into a single commit. In this case it's a good approach, but that is not the case in general :).

gonzaponte (approved):
Better type hints for beersheba-related functions. Good job!

Luismiguelvillar:
> Better type hints for beersheba-related functions. Good job!

Yes! I figured hahaha. Should i continue with "Squash and merge" then?

gonzaponte:
You mean merging the PR? Unless it's urgent we tend to have a merger that has not been involved in the PR. So I think @Ian0sborne can merge your PR and you can merge Ian's.

Luismiguelvillar:
> You mean merging the PR? Unless it's urgent we tend to have a merger that has not been involved in the PR. So I think @Ian0sborne can merge your PR and you can merge Ian's.

Perfect! Sounds good!

Issue #945, opened 2025-10-29
Add IC tag into config


jwaiton:
As raised in today's software meeting, it would be beneficial to have the IC-tag included in the config group of each processed h5.

When a tag isn't available (say, you're on a branch with no specific tag), the fork, branch and commit hash should be returned (perhaps this should be included regardless). 

The tagless option shouldn't be applied if the process is an official production.

PR #947, opened 2025-11-03
Add git information to config file

Ian0sborne:
This PR addresses #945 by including Git information into city config files. A function `add_git_info(conf)` is introduced, which adds information about the IC tag, upstream remote name, branch name and commit hash to the city config file.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -23,6 +23,7 @@ import pandas as pd
 import warnings
 import math
 import os
+import subprocess
 
 from .. dataflow                  import                  dataflow as  fl
 from .. dataflow.dataflow         import                      sink
@@ -137,6 +138,8 @@ def city(city_function):
         conf.files_in = input_files
         conf.file_out = expandvars(conf.file_out)
 
+        conf = add_git_info(conf)
+
         conf.event_range  = event_range(conf)
         # TODO There were deamons! self.daemons = tuple(map(summon_daemon, kwds.get('daemons', [])))
 
@@ -196,6 +199,78 @@ def create_timestamp(rate: float) -> float:
         return timestamp
 
     return create_timestamp_
+
+
+def get_git_tag() -> str:
+    """Returns the current tag if on a tagged commit, or '-' if not."""
+    try:
+        tag = subprocess.check_output(
+            ['git', 'describe', '--tags', '--exact-match'],
+            stderr=subprocess.DEVNULL
+        ).decode('ascii').strip()
+        return tag
+    except subprocess.CalledProcessError:
+        print("Warning: No tag found for current commit.")
+        return "-"
+    except Exception as e:
+        print(traceback.format_exc())
+        return "-"
+
+
+def get_git_upstream_remote() -> str:
+    """Returns the upstream remote name, or '-' if not found."""
+    try:
+        upstream = subprocess.check_output(
+            ["git", "rev-parse", "--abbrev-ref", "@{upstream}"],
+            stderr=subprocess.DEVNULL
+        ).decode("ascii").strip()
+        remote_name = upstream.split('/')[0]  # take the part before '/'
+        return remote_name
+    except subprocess.CalledProcessError:
+        print("Warning: No upstream remote configured for this branch.")
+        return "-"
+    except Exception as e:
+        print(traceback.format_exc())
+        return "-"
+
+
+def get_git_branch() -> str:
+    """Returns the current branch name, or '-' if not found."""
+    try:
+        branch = subprocess.check_output(
+            ['git', 'branch', '--show-current'],
+            stderr=subprocess.DEVNULL
+        ).decode('ascii').strip()
+        return branch
+    except subprocess.CalledProcessError:
+        print("Warning: No branch found.")
+        return "-"
+    except Exception as e:
+        print(traceback.format_exc())
+        return "-"
+
+
+def get_git_commit_hash() -> str:
+    """Returns the current commit hash, or '-' if not found."""
+    try:
+        hash = subprocess.check_output(
+            ['git', 'rev-parse', 'HEAD'],
+            stderr=subprocess.DEVNULL
+        ).decode('ascii').strip()
+        return hash
+    except subprocess.CalledProcessError:
+        print("Warning: No commit hash found.")
+        return "-"
+    except Exception as e:
+        print(traceback.format_exc())
+
+
+def add_git_info(conf):
+    conf.tag = get_git_tag()
+    conf.upstream = get_git_upstream_remote()
+    conf.branch = get_git_branch()
+    conf.hash = get_git_commit_hash()
+    return conf
 
 
 def index_tables(file_out):
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -201,17 +201,27 @@ def create_timestamp(rate: float) -> float:
     return create_timestamp_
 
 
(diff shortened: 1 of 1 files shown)


invisible_cities/cities/components.py
```diff
@@ -198,6 +200,80 @@ def create_timestamp_(event_number: Union[int, float]) -> float:
...
+            ['git', 'describe', '--tags', '--exact-match'],
+            stderr=subprocess.DEVNULL
+        ).decode('ascii').strip()
+        return tag
+    except subprocess.CalledProcessError:
+        print("Warning: No tag found for current commit.")
+        return "-"
+    except Exception as e:
+        print(traceback.format_exc())
+        return "-"
```
gonzaponte: This pattern:
```
try:
   var = subprocess.check_output(some_command, stderr=subprocess.DEVNULL).decode("ascii").strip()
   return var
except subprocess.CalledProcessError:
    print(some_message)
except Exception as e:
    print(traceback.format_exc())
    return "-"
```
repeats multiple times. It would be great to encapsulate the logic into a function to avoid duplication. It will also significantly enhance the signal-to-noise ratio.

I also think it makes sense to return `None` when the value does not exist, since it will be converted to a string anyway.

Ian0sborne: Hey Gonzalo. I think I've implemented this functionality, it's a lot more succinct now:
```python
def fetch_git_info(git_command : str, 
                   git_parameter : str) -> str | None:
    """Returns requested Git information, or None if not found."""
    git_command = git_command.split() # turns the string input into a list of strings
    try:
        var = subprocess.check_output(git_command, stderr=subprocess.DEVNULL).decode('ascii').strip()
        return var
    except subprocess.CalledProcessError:
        print(f"Warning: No {git_parameter} found.")
        return None
    except Exception as e:
        print(traceback.format_exc())


def add_git_info():
    git_info = dict(
        IC_tag = fetch_git_info("git describe --tags --exact-match", "git tag"),
        upstream_name = fetch_git_info("git rev-parse --abbrev-ref @{upstream}", "upstream remote"),
        branch_name = fetch_git_info("git branch --show-current", "branch"),
        commit_hash = fetch_git_info("git rev-parse HEAD", "commit hash")
    )
    return git_info
```

The only issue is the fact that I've forfeited the functionality of the previous `get_git_upstream_remote()` function which would convert the output from {upstream remote}/{branch} to just {upstream remote}. Is this something we care about, should I find a way to implement it back in?

Ian0sborne: To be fair, keeping only the upstream remote name is not that ugly to implement, the simplest way I can think of is this, let me know if you'd like me to include this.
```python
def fetch_git_info(git_command : str, 
                   git_parameter : str):
    """Returns requested Git information, or None if not found."""
    git_command = git_command.split() # turns the string input into a list of strings
    try:
        var = subprocess.check_output(git_command, stderr=subprocess.DEVNULL).decode('ascii').strip()
        if git_parameter == "upstream remote":
            var = var.split('/')[0]
        return var
    except subprocess.CalledProcessError:
        print(f"Warning: No {git_parameter} found.")
        return None
    except Exception as e:
        print(traceback.format_exc())
```

gonzaponte: Loving this. You can also apply the `.split('/')[0]` bit to the return value in `upstream_name = fetch_git_info(...).split('/')[0]`, which is more appropriate.

Ian0sborne: Good catch! If I do that I could probably just remove `git_parameter` entirely and instead print out the `git_command` in the error message. What do you think?

gonzaponte: Sounds good.

invisible_cities/cities/components.py
```diff
@@ -198,6 +201,42 @@ def create_timestamp_(event_number: Union[int, float]) -> float:
...
+    git_command = git_command.split() # turns the string input into a list of strings
+    try:
+        git_output = subprocess.run(git_command, text=True, capture_output=True, check=True).stdout.strip()
+        return git_output
+    except subprocess.CalledProcessError as e:
+        print(f"Following error encountered when running: " + ' '.join(git_command))
+        print(e.stderr)
+        return None
+    except Exception as e:
+        print(traceback.format_exc())
```
gonzaponte: Probably better to do `git_command.split()` inline in the call to `subprocess.run` so you don't have to join it back in the print below.

Ian0sborne: Will do, the problem now is that using `.split('/')[0]` within the `add_git_info()` function is cause issues because it gets called even when there isn't a remote upstream so it cannot `.split` a `None` type. Should I put it back into the `run_git_command()` function?

gonzaponte: right. Then you can return `""` or, even better, something like `"Unknown"` or `"None"`.

Ian0sborne:
I still haven't finished all the comments, but what do you think of the testing function so far? I wanted to hear your opinion before I committed it.
```python
def test_add_git_info():

    try:
        # keep a copy of the current branch
        current_branch = run_git_command("git branch --show-current")

        # create a testing branch
        testing_branch = 'function-testing-branch'
        run_git_command(f"git checkout -b {testing_branch}")

        # create an empty testing commit
        run_git_command("git commit --allow-empty -m 'create_dummy_commit'")
        testing_commit_hash = run_git_command("git log --pretty=format:%H -n 1") # alternative way of extracting hash

        # create a testing remote upstream
        testing_upstream = 'function-testing-upstream'
        run_git_command(f"git remote add {testing_upstream} .") # use current repo as remote
        run_git_command(f"git branch --set-upstream-to {testing_upstream}/{testing_branch}")
        run_git_command(f"git push -u {testing_upstream} {testing_branch}") # add this otherwise remote = None

        # create a temporary tag for testing
        testing_tag = 'v.function.testing.tag'
        run_git_command(f"git tag {testing_tag}")
        
        extracted_git_info = add_git_info()

        assert extracted_git_info['branch_name']   == testing_branch
        assert extracted_git_info['commit_hash']   == testing_commit_hash
        assert extracted_git_info['upstream_name'] == testing_upstream
        assert extracted_git_info['IC_tag']        == testing_tag

    except Exception as e:
        print(f"Something went wrong: {e}")
        raise

    finally:
        # cleanup happens no matter what, each command is within a try/except so that if one fails the rest still run
        # switches back to the original branch
        try:
            run_git_command(f"git checkout {current_branch}")
        except Exception as e:
            print(f"Failed to restore branch {current_branch}: {e}")
        
        try:
            run_git_command(f"git tag -d {testing_tag}")
        except Exception as e:
            print(f"Failed to delete tag {testing_tag}: {e}")
        # delete local testing branch
        try:
            run_git_command(f"git branch -D {testing_branch}")
        except Exception as e:
            print(f"Failed to delete branch {testing_branch}: {e}")
        # delete remote testing branch
        try:
            run_git_command(f"git push {testing_upstream} --delete {testing_branch}")
        except Exception as e:
            print(f"Failed to delete remote branch {testing_branch} on {testing_upstream}: {e}")

        try:
            run_git_command(f"git remote remove {testing_upstream}")
        except Exception as e:
            print(f"Failed to remove remote {testing_upstream}: {e}")
```

gonzaponte:
Please remind me where we left things here. Is this ready for another review?

Ian0sborne:
Hey Gonzalo! We still have to commit the new testing function and clean up some of the weird commits we made when testing stuff together. I will get back on it on Monday.

Issue #861, opened 2024-02-26
ICAROS: City Tools


jahernando:
Repository with Kr-Calibration tools

- [ ] functions to apply the Kr corrections
- [ ] access and display of the calibration data map
- [ ] functions to compare two calibration maps
- [ ] NB step by step ICARO

bpalmeiro:
Not sure if here is the best place to have this (advice otherwise), but for book-keeping when we do the documentation:

https://github.com/next-exp/IC/pull/865#discussion_r1900679623

PR #948, opened 2025-12-19
Add functions to get 3D krypton maps

mcidlaso:
These scripts contain functions to:  
- Make cuts on the outputs from sophronia to get the selected events and the efficiencies of the cuts
- Get the lifetime and drift velocity for each run 
- Compute 3D krypton maps and apply them to correct the data
- Some tests for these functions (still have a lot of work to do)


invisible_cities/icaros/correction_functions.py
```diff
@@ -0,0 +1,51 @@
+import pandas as pd
+import numpy  as np
+from scipy.interpolate import griddata
+
+
+
+def normalization(krmap, method, x_low, x_high, y_low, y_high):
```
gonzaponte: The last 4 parameters need to be optional, since they are only relevant for `method == "region"`. I also think it would be interesting to gather them in a dictionary to simplify the logic.

invisible_cities/icaros/correction_functions.py
```diff
@@ -0,0 +1,51 @@
...
+
+    mu_values = krmap.mu
```
gonzaponte: since you are doing `.dropna()` everywhere, you can do it here already

invisible_cities/icaros/correction_functions_test.py
```diff
@@ -0,0 +1,43 @@
...
+def test_method_norm():
+
+    k_vals = np.arange(10)
+    i_vals = np.arange(12)
+    j_vals = np.arange(12)
+
+    k, i, j = np.meshgrid(k_vals, i_vals, j_vals, indexing='ij')
+    k = k.ravel()
+    i = i.ravel()
+    j = j.ravel()
```
gonzaponte:
```suggestion
    k, i, j = map(np.ravel,
                  np.meshgrid(k_vals, i_vals, j_vals, indexing='ij'))
```

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,149 @@
...
+
+import itertools
+
+"""
+
+This script contains functions to compute NaN_maps, 3D_maps from kdsts and to merge them to create an 'uniform' map.
+
+"""
+
+def create_NaN_map(xy_range, dt_range, xy_nbins, dt_nbins):
```
gonzaponte:
```suggestion
def create_empty_map(xy_range, dt_range, xy_nbins, dt_nbins):
```

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,149 @@
...
+    """
+
+    xy_bins = np.linspace(xy_range[0], xy_range[1], xy_nbins + 1)
+    dt_bins = np.linspace(dt_range[0], dt_range[1], dt_nbins + 1)
+
+    #shift to bin centers invisible_cities.core.core_functions
+
+    i_range = np.arange(0, len(xy_bins)-1)
+    j_range = np.arange(0, len(xy_bins)-1)
+    k_range = np.arange(0, len(dt_bins)-1)
```
gonzaponte: `xy_bins` and `dt_bins` are only used to compute their length, which are given by the variables `xy_nbins` and `dt_nbins`

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,149 @@
...
+        index=range(len(combinations)),
+        columns=['nevents', 'mu', 'sigma', 'mu_error', 'sigma_error']
+    )
+    Nan_map = pd.concat([combinations, Nan_map], axis=1)
+
+    return Nan_map
+
+
+def med_fun(df):
+    sigma = (0.04/2.35)*df.S2e.median()
```
gonzaponte: Add a comment explaining this value

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,149 @@
...
+        np.nan,
+        index=range(len(combinations)),
+        columns=['nevents', 'mu', 'sigma', 'mu_error', 'sigma_error']
+    )
+    Nan_map = pd.concat([combinations, Nan_map], axis=1)
+
+    return Nan_map
+
+
+def med_fun(df):
```
gonzaponte: rename to `median`, `get_median` or similar

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,149 @@
...
+    return pd.DataFrame({
+         'nevents': len(df),
+         'median': df.S2e.median(),
+         'sigma': sigma,
+         'median_error': sigma/np.sqrt(len(df)),
+         'sigma_error': np.nan}, index = [0]) #error de std?
+
+
+
+def fit_function(df, bins, size =10):
```
gonzaponte: rename to `gaussian_fit` or similar
rename `bins` to `ebins`, `energy_bins`, or similar, to clarify that they refer to energy
rename `size` to something that gives a better hint of its meaning (e.g. `min_events`)

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,149 @@
...
+    if len(df)<size:
+        results = med_fun(df)
+
+    bin_centers = shift_to_bin_centers(bin_edges)
```
gonzaponte: if `bin_edges` is only used to compute `bin_centers`, then the argument to this function should be `bin_centers`

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,149 @@
...
+
+        results = pd.DataFrame({'nevents': len(df), 'mu': np.nan, 'sigma': np.nan, 'mu_error': np.nan, 'sigma_error': np.nan}, index = [0])
+
+        return results
+    results = pd.DataFrame({'nevents': len(df), 'mu': f.values[1], 'sigma': f.values[2], 'mu_error': f.errors[1], 'sigma_error': f.errors[2]}, index = [0])
+
+    return results
+
+
+def map_3D_fits(df, xy_range, dt_range, xy_nbins, dt_nbins):
```
gonzaponte: try to come up with a name like `verb_noun`, like `create_something`

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,149 @@
...
+
+    df = df.loc[:,['X', 'Y', 'DT', 'S2e']]
+
+    mask = in_range(df.X, *xy_range) & in_range(df.Y, *xy_range) & in_range(df.DT, *dt_range)
+
+    df = df[mask]
+
+    x = df.X
+    y = df.Y
+    dt = df.DT
```
gonzaponte:
```suggestion
    x, y, dt, _ = df.values.T
```
and then you don't need to do .values on each column

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,149 @@
...
+
+    xy_bins = np.linspace(*xy_range, xy_nbins + 1)
+    dt_bins = np.linspace(*dt_range, dt_nbins + 1)
+
+    map_df = df.assign( i = np.digitize(x.values, bins = xy_bins) - 1,
+                        j = np.digitize(y.values, bins = xy_bins) - 1,
+                        k = np.digitize(dt.values, bins = dt_bins) - 1)
```
gonzaponte: Add a comment explaining the -1 at the end

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+    result = map_df.groupby(['k', 'i', 'j']).apply(fit_function, bins = np.linspace(3000, 9000, 101))
```
gonzaponte: the argument bins should probably be an argument to this function as well

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,149 @@
...
+
+    result = result.reset_index()
+
+    if 'level_3' in result.columns:
+        result = result.drop(columns = 'level_3')
```
gonzaponte: This should be deterministic, either the condition is true or false, but it will not change depending on the data, I don't think

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,149 @@
...
+
+    return result
+
+
+def merge_maps(NaN_map, map_3D):
+
+    full_map = pd.concat([map_3D, NaN_map], ignore_index = True)
+    full_map = full_map.groupby(['k', 'i','j']).first().reset_index()
```
gonzaponte: Explain this bit in a comment

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,149 @@
...
+def include_coordinates(map, xy_range, dt_range, xy_nbins, dt_nbins):
+
+    xy_bins = np.linspace(*xy_range, xy_nbins + 1)
+    dt_bins = np.linspace(*dt_range, dt_nbins + 1)
+
+    x = shift_to_bin_centers(xy_bins)
+    y = shift_to_bin_centers(xy_bins)
+    dt = shift_to_bin_centers(dt_bins)
+
+    combinaciones_bins = pd.DataFrame(
```
gonzaponte: english :)

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,149 @@
...
+    itertools.product(dt, x, y),
+    columns=['dt', 'x', 'y'])
+
+
+    map = pd.concat([combinaciones_bins.reset_index(drop=True), map.reset_index(drop=True)], axis = 1)
```
gonzaponte: why do you need to reset indices here?

invisible_cities/icaros/krmap_functions_test.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def test_empty_medfun():
+    empty_dst = pd.DataFrame(columns = ['DT', 'x', 'y', 'S2e'])
+    result = med_fun(empty_dst)
+
+    assert result.shape == (1, 5)
+    assert result['nevents'].sum() == 0
+    for col in ['median', 'sigma', 'median_error', 'sigma_error']:
+        assert pd.isna(result[col].iloc[0])
```
gonzaponte:
```suggestion
        assert pd.isna(result[col].iloc[0]), f"{col} is not a nan"
```

this adds a bit of information in case of failure

invisible_cities/icaros/krmap_functions_test.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+
+def test_empty_medfun_2():
```
gonzaponte:
```suggestion
def test_medfun_empty_input_does_not_raise():
```

invisible_cities/icaros/krmap_functions_test.py
```diff
@@ -0,0 +1,85 @@
+import numpy as np
+import pandas as pd
+from invisible_cities.core.core_functions import in_range, shift_to_bin_centers
+import itertools
+from new_icaromaps import create_NaN_map, med_fun, fit_function, map_3D_fits, merge_maps, include_coordinates, is_even
+from pytest import raises
+
+
+def test_empty_medfun():
```
gonzaponte: Try to be a bit more descriptive in the test name. Also keep the structure `test_TESTEDFUN_DESCRIPTION`:

```suggestion
def test_medfun_empty_input():
```

invisible_cities/icaros/krmap_functions_test.py
```diff
@@ -0,0 +1,85 @@
...
+
+def test_fit_fun():
+    empty_dst = pd.DataFrame(columns = ['DT', 'x', 'y', 'S2e'])
+    result = fit_function(empty_dst)
+
+    assert result == med_fun(empty_dst)
+
+
+def test_Nevents():
+    d = {'DT': np.empty(6, dtype=int), 'x' : np.empty(6), 'y' : np.empty(6),'S2e' : np.ones(6)}
```
gonzaponte: why `dtype=int` here?

invisible_cities/icaros/krmap_functions_test.py
```diff
@@ -0,0 +1,85 @@
...
+    assert fit_function(df_test, bins = 5)['nevents'].iloc[0] == N
+
+
+def test_merge_maps():
+    Nan_map_test = create_NaN_map(xy_range = (-500, 500), dt_range = (0, 1400), xy_nbins = 100, dt_nbins = 10)
+    d = {'nevents': np.empty(6, dtype = int), 'mu' : np.empty(6), 'sigma' : np.empty(6), 'mu_error' : np.empty(6), 'sigma_error' : np.empty(6)}
+    map_3D_test = pd.DataFrame(data = d, index = range(0, 6))
+    assert merge_maps(Nan_map_test, map_3D_test).shape == Nan_map_test.shape
+
+test_merge_maps()
```
gonzaponte: remove this line

invisible_cities/icaros/lifetime_vdrift_functions.py
```diff
@@ -0,0 +1,137 @@
...
+import matplotlib.pyplot as plt
+from scipy import stats
+from matplotlib.patches import Circle, Rectangle
+from typing import Tuple
+from invisible_cities.core.stat_functions import poisson_sigma
+import invisible_cities.core.fit_functions as fit
+
+
+
+def select_lifetime_region(df, bins, x0, y0, r, half_size, circle=True):
```
gonzaponte: we should review this function interactively

invisible_cities/icaros/lifetime_vdrift_functions.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+    return df_region, mean_values
+
+
+
+def LT_fit(DT, s2e, p0):
+
+    def exponential(t, e, tau):
+        return e*np.exp(-t/tau)
```
gonzaponte: Use `expo` from `fit_functions` (careful with the negative sign)

invisible_cities/icaros/lifetime_vdrift_functions.py
```diff
@@ -0,0 +1,137 @@
...
+
+    popt, pcov = curve_fit(exponential, DT, s2e, p0 = p0)
```
gonzaponte: Use `fit` from `fit_functions`

invisible_cities/icaros/lifetime_vdrift_functions.py
```diff
@@ -0,0 +1,137 @@
...
+        The slope parameter that controls the steepness of the sigmoid curve.
+    offset : float
+        The vertical offset added to the sigmoid function.
+    Returns
+    -------
+    np.array
+        Array of computed sigmoid values for x array.
+    '''
+    sigmoid = scale / (1 + np.exp(-slope * (x - inflection))) + offset
+    return sigmoid
```
gonzaponte: we can move this function to `fit_functions`

invisible_cities/icaros/selection_functions.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+
+def eff_of_sel(sel, name = "", do_print = True):
+    nsel, ntot = np.sum(sel), len(sel)
+    eff = nsel/float(ntot)
+    if do_print:
+        seff = "{:.4f}".format(eff)
+        print(f"{name} efficiency {seff}, {nsel}/{ntot}.")
+    return eff
```
gonzaponte: I think this function is replaced by `eff_of_selection`, isn't it?

invisible_cities/icaros/selection_functions.py
```diff
@@ -0,0 +1,148 @@
...
+
+def load_files(path_data, n):
+    filenames = []
+    for i in range(len(n)):
+        path_run =  f"/trigger1/ldc{i}"
+        path = path_data + path_run
+        file = [os.path.join(path, f) for f in os.listdir(path) if f.endswith(".h5")]
+        filenames.append(file)
+    filenames = list(itertools.chain.from_iterable(filenames))
+    return filenames
```
gonzaponte: This function needs to be more generic, instead of assuming that the folder is structured like
```
top
└── trigger1
    ├── ldc1
    ├── ldc2
    ├── ldc3
    ├── ldc4
    ├── ldc5
    ├── ldc6
    └── ldc7
```

Maybe it can search for all `.h5` files recursively

invisible_cities/icaros/selection_functions.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+
+def eff_of_selection(df_before, df_after, name = ""):
+    events_before = df_before.event.nunique()
+    events_after = df_after.event.nunique()
+
+    eff = events_after/events_before
+
+    print(f"{name} efficiency {eff}, {events_after}/{events_before}.")
```
gonzaponte: Wrap it in `if name:`?

invisible_cities/icaros/selection_functions.py
```diff
@@ -0,0 +1,148 @@
...
+    df_DTband = kdst[sel_DTband]
+
+    eff_DT = eff_of_selection(kdst, df_DTband, 'band selection')
+
+    return df_DTband, eff_DT
+
+
+def select_Xrays(kdst, low_xrays, high_xrays):
+
+    sel_xrays = (kdst.Ec >= low_xrays) & (kdst.Ec <= high_xrays)
```
gonzaponte: use `in_range`

invisible_cities/icaros/selection_functions.py
```diff
@@ -0,0 +1,148 @@
...
+    df_Xrays = kdst[sel_xrays]
+
+    eff_Xrays = eff_of_selection(kdst, df_Xrays, 'remove xrays')
+
+    return df_Xrays, eff_Xrays
+
+
+def select_1S1_1S2(kdst):
+    a = kdst.groupby('event')['time'].count()
+    events_1S1_1S2 = a[a == 1].index.values
```
gonzaponte: explain this and rename `a` to something more self-explanatory

invisible_cities/icaros/selection_functions.py
```diff
@@ -0,0 +1,148 @@
...
+    df_1S1_1S2 = kdst[kdst.event.isin(events_1S1_1S2)]
+
+    eff_1S1_1S2 = eff_of_selection(kdst, df_1S1_1S2, '1S1 & 1S2')
+
+    return df_1S1_1S2, eff_1S1_1S2
+
+
+def select_S2t(kdst, low_S2t, high_S2t):
+    range_S2t = (low_S2t, high_S2t)
+    sel_S2t = in_range(kdst.S2t, *range_S2t)
```
gonzaponte:
```suggestion
    sel_S2t = in_range(kdst.S2t, low_S2t, high_S2t)
```

invisible_cities/icaros/selection_functions.py
```diff
@@ -0,0 +1,148 @@
...
+def select_Rmax(kdst, R_max):
+    df_Rmax = kdst[kdst.R <= R_max]
+
+    eff_Rmax = eff_of_selection(kdst, df_Rmax, f'events with R less than {R_max}')
+
+    return df_Rmax, eff_Rmax
+
+
+def select_DTrange(kdst, low_DT, high_DT):
+    df_DTrange = kdst[(kdst.DT >= low_DT) & (kdst.DT <= high_DT)]
```
gonzaponte: Use `in_range`

invisible_cities/icaros/selection_functions.py
```diff
@@ -0,0 +1,148 @@
...
+
+    eff_DTrange = eff_of_selection(kdst, df_DTrange, f'events in DT range [{low_DT}, {high_DT}]')
+
+    return df_DTrange, eff_DTrange
+
+
+def select_nsipm(kdst, low_nsipm, high_nsipm):
+    sel_nsipm = (kdst.Nsipm >= low_nsipm) & (kdst.Nsipm <= high_nsipm)
```
gonzaponte: Use `in_range`

invisible_cities/icaros/selection_functions.py
```diff
@@ -0,0 +1,148 @@
...
+
+def apply_selections(kdst, run_number,  dtrms2_low, dtrms2_upp, low_xrays, high_xrays, low_S2t, high_S2t, R_max, low_DT, high_DT, low_nsipm, high_nsipm):
+
+    df, eff_DTband = select_diffusion_band(kdst, dtrms2_low, dtrms2_upp)
+
+    df, eff_Xrays = select_Xrays(df, low_xrays, high_xrays)
+
+    df, eff_1S1_1S2 = select_1S1_1S2(df)
+
+    df, eff_S2t = select_S2t(df, low_S2t, high_S2t)
```
gonzaponte: This one should go before the 1S1+1S2 selection, I think

invisible_cities/icaros/selection_functions.py
```diff
@@ -0,0 +1,148 @@
...
+
+    df_final, eff_nsipm = select_nsipm(df, low_nsipm, high_nsipm)
+
+    total_efficiency = eff_of_selection(kdst, df_final, f'total events after all selections')
+
+    d = {'eff diffusion band': [eff_DT], 'eff X rays': [eff_Xrays], 'eff 1S1 & 1S2': [eff_1S1_1S2],
+         'eff S2 trigger time': [eff_S2t], 'eff Rmax': [eff_Rmax], 'eff range DT': [eff_DTrange],
+         'eff number of SiPMS': [eff_nsipm], 'total efficiency': [total_eff]}
+
+    df_efficiencies = pd.DataFrame(data = d)
```
gonzaponte: You can also have numbers in the dictionary (instead of lists of 1 element) and then add `index=[0]` when creating the dataframe

gonzaponte (changes requested):
First round

invisible_cities/icaros/correction_functions.py
```diff
@@ -0,0 +1,53 @@
+import pandas as pd
+import numpy  as np
+from scipy.interpolate import griddata
+
+
+def normalization(krmap, method, xy_params = None):
+
+    mu_values = krmap.mu.dropna()
```
gonzaponte: `dropna` should affect all options of method. Right now, it's only affecting the first two, the other three do not use this variable. You should redefine kr_map so it doesn't contain `nan` values.:
`krmap = krmap.dropna(subset=["mu"])`. Then use `krmap.mu.something` everywhere

invisible_cities/icaros/correction_functions.py
```diff
@@ -0,0 +1,53 @@
+import pandas as pd
+import numpy  as np
+from scipy.interpolate import griddata
+
+
+def normalization(krmap, method, xy_params = None):
```
gonzaponte: We will discuss how to do this with symbols

invisible_cities/icaros/correction_functions.py
```diff
@@ -0,0 +1,53 @@
...
+    if method == 'mean chamber':
+        E_reference_chamber = mu_values.mean()
+        return E_reference_chamber
+
+    if method == 'mean anode':
+        mu_values_anode = krmap[krmap.k == 0].mu
+        E_reference_anode = mu_values_anode.mean()
+        return E_reference_anode
+
+    mask_region = (krmap['x'] <= xy_params['x_high']) & (krmap['x'] >= xy_params['x_low']) & (krmap['y'] <= xy_params['y_high']) & (krmap['y'] >= xy_params['y_low'])
```
gonzaponte: use `in_range`

invisible_cities/icaros/correction_functions_test.py
```diff
@@ -0,0 +1,43 @@
...
+        'y': y,
+        'dt': dt,
+        'mu': mu
+    })
+
+    region_mask = (
+        (map_test.x <= 100) & (map_test.x >= -100) &
+        (map_test.y <= 100) & (map_test.y >= -100)
+    )
+
```
gonzaponte: `in_range`

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,316 @@
...
+         'median': df.S2e.median(),
+         'sigma': sigma,
+         'median_error': sigma/np.sqrt(len(df)),
+         'sigma_error': np.nan}, index = [0]) #error de std?
+
+
+
+def gaussian_fit(df, ebins, min_events = 10):
+    if len(df)< min_events:
+        results = get_median(df)
```
gonzaponte:
```suggestion
        return get_median(df)
```

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,316 @@
...
+    metadata = {
+        'rmax'        : df.R.max(),
+        'zmax'        : df.Z.max(),
+        'bin_size_dt' : (dt_range[1] - dt_range[0]) / dt_nbins,
+        'bin_size_x'  : (xy_range[1] - xy_range[0]) / xy_nbins,
+        'bin_size_y'  : (xy_range[1] - xy_range[0]) / xy_nbins,
+        'method'      : norm_method,
+        'dtbins'      : [np.unique(krmap.k.values).tolist()],
+        'xbins'       : [np.unique(krmap.i.values).tolist()],
+        'ybins'       : [np.unique(krmap.j.values).tolist()],
```
gonzaponte:
```suggestion
        'dtbins'      : [krmap.k.unique().to_list()],
        'xbins'       : [krmap.i.unique().to_list()],
        'ybins'       : [krmap.j.unique().to_list()],
```

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,316 @@
...
+    """
+    y, x  = np.histogram(data, bins)
+    x     = shift_to_bin_centers(x)
+    seed  = gauss_seed(x, y)
+    
+    if sigma:
+        sigma = poisson_sigma(y)
+        f     = fit(gauss, x, y, seed, sigma = sigma)
+    else:
+        f     = fit(gauss, x, y, seed)
```
gonzaponte:
```suggestion
    sigma = poisson_sigma(y) if sigma else None
    f     = fit(gauss, x, y, seed, sigma = sigma)
```

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,316 @@
...
+    assert np.all(f.values != seed)
```
gonzaponte:
```suggestion
    if not np.all(f.values != seed):
        raise RuntimeError("a meaningful error message")
```

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,316 @@
...
+    
+    time_edges = np.arange(t_min, t_max + slice_seconds, slice_seconds)
+    
+    dataframes = []
+
+    for i in range(len(time_edges) - 1):
+        t_start = time_edges[i]
+        t_end   = time_edges[i + 1]
+        
+        mask = (df.time >= t_start) & (df.time < t_end)
```
gonzaponte: `in_range`

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,316 @@
...
+        
+    return dataframes
+  
+
+
+def get_time_evol(df, run_number, bins_lt, x0, y0, shape, shape_size, p0, dtrange, low_DT, high_DT, bins_Ec, sigma = False, seed = None): 
+    
+    df_tevols = []
+    
+    for i in range(len(df)):
```
gonzaponte: rename `df` to `dfs` and then do `for df in dfs:`. With this you don't need to put `dfs[i]` everywhere, but simply `df`

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,316 @@
...
+        
+        kdst_in_region = select_lifetime_region(df[i], bins_lt, x0, y0, shape, shape_size)
+        magnitudes, uncertainties = LT_fit(kdst_in_region.DT, kdst_in_region.S2e, p0)
+        lifetime = -magnitudes[1]
+        ulifetime = uncertainties[1]
+        
+        e0 = magnitudes[0]
+        e0u = uncertainties[0]
+    
+        dv, udv = compute_drift_v(df[i].DT.to_numpy(), df[i].DT.nunique(), dtrange, seed)
```
gonzaponte: why is the number of bins set to the number of unique dt values?

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,316 @@
...
+        
+        f = quick_gauss_fit(df[i].Ec_2.values, bins_Ec, sigma = sigma)
+    
+        mu = f.values[1]
+        u_mu = f.errors[1]
+    
+        sigma = f.values[2]
+        u_sigma = f.errors[2]
```
gonzaponte:
```suggestion
        _,   mu,   sigma = f.values
        _, u_mu, u_sigma = f.errors
```

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,316 @@
...
+    
+        R = np.sqrt(8*np.log(2))*(sigma/mu) #FWHM (krypton paper)
+        u_R = R * (u_mu**2/mu**2 + u_sigma**2/sigma**2)**0.5
+    
+        mask = (df[i].DT.values > low_DT) & (df[i].DT.values < high_DT)
```
gonzaponte: `in_range`

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,316 @@
...
+
+
+def get_time_evol(df, run_number, bins_lt, x0, y0, shape, shape_size, p0, dtrange, low_DT, high_DT, bins_Ec, sigma = False, seed = None): 
+    
+    df_tevols = []
+    
+    for i in range(len(df)):
+        
+        mean = df[i].S2e.mean()
+        std = df[i].S2e.std()/np.sqrt(len(df[i]))
```
gonzaponte: define `sqrtn = np.sqrt(len(df[i]))`  and use this variable for the rest of the function to reduce the clutter

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,316 @@
...
+        
+        mask = (df.time >= t_start) & (df.time < t_end)
+        df_slice = df[mask]
+        dataframes.append(df_slice)
+        
+    return dataframes
+  
+
+
+def get_time_evol(df, run_number, bins_lt, x0, y0, shape, shape_size, p0, dtrange, low_DT, high_DT, bins_Ec, sigma = False, seed = None):
```
gonzaponte: let's talk about how to refactor this using `groupby`

invisible_cities/icaros/control_plots_functions.py
```diff
@@ -0,0 +1,531 @@
...
+freq = lambda : plt.ylabel("frequency")
+
+
+def monitor_S1S2(df, run_number):
+    fig, axs = plt.subplots(2, 2, figsize=(10, 8))
+    
+    nevents = len(df['event'].unique())
+    
+    df2 = df.groupby('event s1_peak'.split()).first().reset_index()
+    df2_ = df2.groupby('event').count()
```
gonzaponte: I think you don't need to define a new variable here and reuse df2. Same for `df1` below

invisible_cities/icaros/control_plots_functions.py
```diff
@@ -0,0 +1,531 @@
...
+                   f'events: {nevents}\n'
+                   f'mean: {df1_.nS2.mean():.2f}\n'
+                   f'std: {df1_.nS2.std():.2f}')
+    axs[0, 1].set_xlabel('Number of S2')
+    axs[0, 1].set_title('nS2 distribution')
+    axs[0, 1].grid(True)
+    axs[0, 1].legend()
+
+
+    axs[1, 0].hist(df.S1t, bins=100, histtype = 'step', density=True, label =
```
gonzaponte: I think here you should use `df2`, no? And `df1` for `S2t`

invisible_cities/icaros/control_plots_functions.py
```diff
@@ -0,0 +1,531 @@
...
+    axs[0, 0].hist(df1_.nS1, ns1bins,
+                   density=True, histtype='step', label=
+                   f'run: {run_number}\n'
+                   f'events: {nevents}\n'
+                   f'mean: {df1_.nS1.mean():.2f}\n'
+                   f'std: {df1_.nS1.std():.2f}')
+    axs[0, 0].set_xlabel('Number of S1')
+    axs[0, 0].set_title('nS1 distribution')
+    axs[0, 0].grid(True)
+    axs[0, 0].legend()
```
gonzaponte, line 92: This plot is already defined in the previous function, so perhaps the previous function is not necessary

invisible_cities/icaros/control_plots_functions.py
```diff
@@ -0,0 +1,531 @@
...
+    
+    
+def monitor_S1(df, ebins, ns1bins, s1hbins, s1wbins):
+    fig, axs = plt.subplots(2, 2, figsize=(10, 8))
+    
+    nevents = len(df['event'].unique())
+    
+    df1 = df.groupby("event s2_peak".split()).first()
+    df1_ = df1.groupby('event').count()
+    df1__ = df1.groupby('event').mean()
```
gonzaponte: Not sure why you use the mean here

invisible_cities/icaros/control_plots_functions.py
```diff
@@ -0,0 +1,531 @@
...
+def monitor_kr_distribution(df, bins, dtr2_bins):
+
+    sel = in_range(df.S2e, 7.5e3, 9.5e3)
+    sel_DT = in_range(df.DT, 20, 1350)
+
+    DT = df.DT[sel_DT]
+    R2 = df.X[sel_DT]**2 + df.Y[sel_DT]**2
+    
+    DT = DT[sel].dropna()
+    R2 = R2[sel].dropna()
```
gonzaponte: I think this won't work, because `sel` has a different length than `DT` and `R2`, but I don't see why not apply both selections at the same time.

invisible_cities/icaros/control_plots_functions.py
```diff
@@ -0,0 +1,531 @@
...
+
+
+    axs[1].hist2d(DT, R2, dtr2_bins);
+    axs[1].set_xlabel("DT ($\mu$s)"); 
+    axs[1].set_ylabel("R$^2$ (mm$^2$)");
+
+    
+        
+    
+def monitor_S1_multipleruns(run_data, ebins, ns1bins, s1hbins, s1wbins):
```
gonzaponte: I don't think we will use multiple runs here

invisible_cities/icaros/selection_functions.py
```diff
@@ -0,0 +1,157 @@
...
+from scipy import stats
+from correction_functions import apply_3Dmap, normalization
+from pathlib import Path
+import itertools
+
+
+dtrms2_low = lambda dt: -0.7 + 0.030 * (dt-20) # Gonzalo's
+dtrms2_upp = lambda dt: 2.6 + 0.036 * (dt-20) # Gonzalo's2
+dtrms2_cen = lambda dt:  1.0 + 0.033 * (dt-20)
+def dist_to_bandcenter(df): return df.Zrms**2 - dtrms2_cen(df.DT)
```
gonzaponte: Is this function used somewhere?

invisible_cities/icaros/selection_functions.py
```diff
@@ -0,0 +1,157 @@
...
+import matplotlib.pyplot as plt
+from invisible_cities.core.core_functions import in_range, shift_to_bin_centers
+from scipy import stats
+from correction_functions import apply_3Dmap, normalization
+from pathlib import Path
+import itertools
+
+
+dtrms2_low = lambda dt: -0.7 + 0.030 * (dt-20) # Gonzalo's
+dtrms2_upp = lambda dt: 2.6 + 0.036 * (dt-20) # Gonzalo's2
```
gonzaponte: For now it's ok, but we should have these as inputs at some point

invisible_cities/icaros/selection_functions.py
```diff
@@ -0,0 +1,157 @@
...
+dtrms2_cen = lambda dt:  1.0 + 0.033 * (dt-20)
+def dist_to_bandcenter(df): return df.Zrms**2 - dtrms2_cen(df.DT)
+
+
+def load_files(path):
+    return [str(p) for p in Path(path).rglob("*.h5")]
```
gonzaponte: Probably not worth making it a function. We will see.

invisible_cities/icaros/selection_functions.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    xy_params must be a dictionary: {'x_high': , 'x_low': , 'y_high': , 'y_low': }
+    """
+    
+    corrected_energy = apply_3Dmap(map3D, norm_method, kdst.DT, kdst.X, kdst.Y, kdst.S2e, xy_params = xy_params, keV = keV)
+    
+    col_name = 'Ec' if 'Ec' not in kdst.columns else 'Ec_2'
+    kdst[col_name] = corrected_energy.values
+
+    return kdst
```
gonzaponte: This function should be in correcion_functions, I think.
Make `col_name` an argument of this function as well.

invisible_cities/icaros/lifetime_vdrift_functions.py
```diff
@@ -0,0 +1,112 @@
...
+from scipy import stats
+from matplotlib.patches import Circle, Rectangle
+from typing import Tuple
+from invisible_cities.core.stat_functions import poisson_sigma
+import invisible_cities.core.fit_functions as fit
+from invisible_cities.core.core_functions import in_range, shift_to_bin_centers
+
+
+
+def select_lifetime_region(df, bins, x0, y0, shape, shape_size):
```
gonzaponte: We will do this with symbols as well

invisible_cities/icaros/lifetime_vdrift_functions.py
```diff
@@ -0,0 +1,112 @@
...
+
+    df = df.dropna(subset=['X','Y','S2e', 'DT'])
+
+    if shape == 'circle':
+
+        x1, y1 = x0-shape_size, y0-shape_size
+
+        mask_df = ((df.X >= x1) & (df.X <= x1+2*shape_size) & (df.Y >= y1) &
+                   (df.Y <= y1+2*shape_size))
```
gonzaponte: `in_range`

invisible_cities/icaros/lifetime_vdrift_functions.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+    return df_region
+
+
+
+def LT_fit(DT, s2e, p0):
+
+    f  = fit.fit(fit.expo, DT, s2e, p0)
+    return f.values, (f[2][0], f[2][1])
```
gonzaponte: Probably not worth making a function just with this

invisible_cities/icaros/krmap_functions_test.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def test_medfun_empty_input_does_not_raise():
+    empty_dst = pd.DataFrame(columns = ['DT', 'x', 'y', 'S2e'])
+
+    #with raises(ValueError):
+    med_fun(empty_dst)
+
+
+def test_gaussianfit_does_fit_right():
```
gonzaponte:
```suggestion
def test_gaussianfit_empty():
```

invisible_cities/icaros/krmap_functions_test.py
```diff
@@ -0,0 +1,94 @@
...
+    
+
+
+def test_fitfun_medfun_Nevents():
+    d = {'DT': np.empty(6), 'x' : np.empty(6), 'y' : np.empty(6),'S2e' : np.ones(6)}
+    df_test = pd.DataFrame(data = d, index = range(0, 6))
+    N = len(df_test)
+    result_med = med_fun(df_test)
+    assert result_med['nevents'].iloc[0] == N
+    assert fit_function(df_test, bins = 6)['nevents'].iloc[0] == N
```
gonzaponte:
```suggestion
    result_med = med_fun(df_test)
    result_fit = fit_function(df_test, bins = 6)
    assert result_med['nevents'].iloc[0] == N
    assert result_fit['nevents'].iloc[0] == N
```

invisible_cities/icaros/krmap_functions_test.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def test_merge_maps():
+    Nan_map_test = create_empty_map(xy_range = (-500, 500), dt_range = (0, 1400), xy_nbins = 100, dt_nbins = 10)
+    d = {'nevents': np.empty(6, dtype = int), 'mu' : np.empty(6), 'sigma' : np.empty(6), 'mu_error' : np.empty(6), 'sigma_error' : np.empty(6)}
+    map_3D_test = pd.DataFrame(data = d, index = range(0, 6))
+    assert merge_maps(Nan_map_test, map_3D_test).shape == Nan_map_test.shape
```
gonzaponte, line 292: `map_3D_test` needs to contain the columns `k`, `i`, and `j`.

gonzaponte (changes requested):
More comments. We will need to build up a bit more testing and put everything together in a city.

gonzaponte:
Closes #860.
Closes #861.
Closes #862.
Closes #863.
Closes #864.

invisible_cities/icaros/selection_functions.py
gonzaponte: Now that these are in the config file, you can remove them.

invisible_cities/icaros/selection_functions_test.py
```diff
@@ -0,0 +1,135 @@
...
+                              'nS1': np.array([1, 2, 2, 2, 3, 1, 1, 1, 4, 1, 1]),
+                              'nS2': np.full_like(time, 1, dtype = int)})
+
+    kdst_selected, eff = select_1S1_1S2(kdst_test)
+
+    kdst_sel = kdst_selected.groupby('event').count()
+
+    assert np.all(kdst_sel.nS1.values == 1)
+    assert np.all(kdst_sel.nS2.values == 1)
+    assert np.all((kdst_sel.nS1.values) & (kdst_sel.nS2.values)) == 1
```
gonzaponte: this line is probably not what you are trying to do. I'm surprised it even works.

invisible_cities/icaros/lifetime_vdrift_functions.py
```diff
@@ -0,0 +1,93 @@
...
+
+        df_region = df[mask_df]
+        return df_region
+
+    if shape is SelRegionMethod.circle:
+
+        mask_df = (df.X - x0)**2 + (df.Y - y0)**2 <= shape_size**2
+        df_region = df[mask_df]
+        return df_region
+
```
gonzaponte, line 49:
```suggestion
    if shape is SelRegionMethod.square:
        x1, y1 = x0-shape_size, y0-shape_size
        mask_df = in_range(df.X, x1, x1+2*shape_size) & in_range(df.Y, y1, y1+2*shape_size)

    if shape is SelRegionMethod.circle:
        mask_df = (df.X - x0)**2 + (df.Y - y0)**2 <= shape_size**2

    df_region = df[mask_df]
    return df_region
```
Take the common part outside of the if statements

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,773 @@
...
+                     dt_nbins : int) -> pd.DataFrame:
+    """
+    - Uses every possible indices combination.
+    - Creates a DataFrame filled with NaNs for each (i, j, k) combination.
+    Parameters
+    ----------
+    xy_range : tuple
+      Range in x and y (mm) inside which the map is being computed.
+    dt_range : tuple
+      Range in drift time (micro seconds) (same as z) inside which the map is being computed.
```
gonzaponte: the units are not implicit, remove them.
also, strictly speaking, it's not the same as z, so that can also be misleading.

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,773 @@
...
+    xy_nbins : int
+      Number of map bins for x and y.
+    dt_nbins : int
+      Number of map bins for drift time.
+    Returns
+    -------
+    Nan_map : pd.DataFrame
+      Dataframe filled with NaNs for each (i,j,k) combination.
```
gonzaponte: i,j,k are not defined in the docstring, so better say something like "for each combination of z, x, y bins

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,773 @@
...
+    return Nan_map
+
+
+def get_median(df        : pd.DataFrame,
+               nbins_S2e : int = None) -> pd.DataFrame:
+    """
+    Computes the median value of a given variable
+    -We consider that for len(df) < 3, it makes no sense to compute the std.
+    -Kr energy resolution is ~4%, so the std for S2e median would be ~0.04/2.35.
+    Parameters
```
gonzaponte, line 98: update docstring

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,773 @@
...
+    except:
+
+        return pd.DataFrame({'nevents': len(df),
+                                 'mu': np.nan,
+                                 'sigma': np.nan,
+                                 'mu_error': np.nan,
+                                 'sigma_error': np.nan}, index = [0])
+
+    if not in_range(f.values[1], 0.5*df.S2e.median(), 1.5*df.S2e.median()):
+        return get_median(df)
```
gonzaponte: document this behaviour in the docstring

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,773 @@
...
+
+def fit_map(df            : pd.DataFrame,
+            xy_range      : tuple,
+            dt_range      : tuple,
+            xy_nbins      : int,
+            dt_nbins      : int,
+            S2e_range     : tuple,
+            fit_function  : Callable,
+            min_events    : int = 50,
+            nbins_S2e     : int = None) -> pd.DataFrame:
```
gonzaponte: let's discuss offline how to remove the optional arguments here.

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,773 @@
...
+      Full regularized map.
+
+    """
+    mu = krmap.mu.values.reshape(mapshape)
+    for k in range(mu.shape[0]):
+        for i in range(1, mu.shape[1] -1):
+            for j in range(1, mu.shape[2] -1):
+                if np.isnan(mu[k,i,j]):
+                    vals = [mu[k,i+ii,j+jj] for ii in (-1, 0, 1) for jj in (-1, 0, 1) if not (ii==jj==0)]
+                    if np.count_nonzero(np.isnan(vals))>3: continue
```
gonzaponte: define a variable `max_nan_neighbours=3` at the top and use it here

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,773 @@
...
+                   dt_range     : tuple,
+                   xy_nbins     : int,
+                   dt_nbins     : int,
+                   S2e_range    : tuple,
+                   fit_function : Callable,
+                   min_events   : int = 50,
+                   nbins_S2e    : int = None) -> pd.DataFrame:
+    """
+    Computes a full 3D map merging (with merge_maps) a map full of nans from create_empty_map
+    with a 3D data map from fit_map and adding spatial coordinates using include_coordinates.
```
gonzaponte: Avoid the implementation details. Give the overall picture. For instance:

Computes a 3D map from the data in `df` given a range in xy and dt and the number of bins for each coordinate. The map values in each voxel are computed using `fit_function`. A data quality filter is applied requiring S2e to be in a given range `S2e_range`.

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,773 @@
...
+      Full map representing all bins (every possible indices combination), filled with
+      data from fit_map,  with NaNs in those bins where there is no data and spatial
+      dt, x, y coordinates.
+    """
+    ts = int(df.time.values.min())
+    tf_slice = int(df.time.values.max())
+
+    t_interval = tf_slice - ts
+
+    t_hours = t_interval/3600 #from seconds to hours
```
gonzaponte: Rewrite this in a single line, and don't round to ints

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,773 @@
...
+    volume = (dt_binsize*xy_binsize*xy_binsize)*0.001 #from mm^3 to cm^3
+
+    NaN_map = create_empty_map(xy_range, dt_range, xy_nbins, dt_nbins)
+    map_3D_fit = fit_map(df, xy_range, dt_range, xy_nbins, dt_nbins, S2e_range, fit_function, min_events, nbins_S2e)
+
+    map = merge_maps(NaN_map, map_3D_fit)
+    coor_map = include_coordinates(map, xy_range, dt_range, xy_nbins, dt_nbins)
+
+    full_map = coor_map.assign(density = coor_map['nevents']/volume/t_hours)
+    #density in events*hour/cm^3
```
gonzaponte: - put the comment above the line it applies to
-  it's events/hour/cm3
- bring the line that defines `volume` next to where it's used
- same for `t_hours`

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,773 @@
...
+
+    mask = in_range(df.DT.values, *s1_DTrange)
+    s1e_cath = df[mask].S1e.values.mean()
+    us1e_cath = df[mask].S1e.values.std()/np.sqrt(len(df))
+
+    ts = int(df.time.values.min())
+    tf_slice = int(df.time.values.max())
+
+    t_interval = tf_slice - ts
+
```
gonzaponte: single line, no int

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,773 @@
...
+    t_evols : pd.DataFrame
+      Dataframe containing time evolution parameters for each time slice.
+    """
+
+    slice_seconds = slice_hours * 3600
+
+    t_min = df.time.min()
+    t_max = df.time.max()
+
+    time_bins = np.arange(t_min, t_max + slice_seconds, slice_seconds)
```
gonzaponte: `t_min` and `t_max` are a bit superfluous, use them directly

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,773 @@
...
+    metadata : pd.DataFrame
+      Metadata dataframe.
+    t_evol : pd.DataFrame
+      Time evolution dataframe.
+    Returns
+    -------
+    Hdf file containing in each node each one of the inputs.
+    """
+
+    metadata.to_hdf(name, key = 'metadata', mode = 'a')
```
gonzaponte: shouldn't it be `mode="w"`?

Explain why this is a special case with a comment

invisible_cities/icaros/correction_functions.py
```diff
@@ -0,0 +1,224 @@
...
+from .. core    .core_functions import in_range
+from .. types   .symbols import NormMethod
+
+
+def normalization(krmap    : pd.DataFrame,
+                  method   : NormMethod,
+                  xy_params : dict = None) -> float:
+    """
+    Given a input krypton map, normalizes the whole map to a certain value
+    as a function of the chosen method.
```
gonzaponte: It doesn't normalize the whole map. It returns the reference value which is later used for normalization

invisible_cities/icaros/correction_functions.py
```diff
@@ -0,0 +1,224 @@
...
+    anode = krmap[krmap.k == 0]
+
+    if method is NormMethod.mean_region_anode:
+        E_reference_slice_anode = anode.mu.mean()
+        return E_reference_slice_anode
+
+    if method is NormMethod.median_region_anode:
+        E_median_region_anode = anode.mu.median()
+        return E_median_region_anode
+
```
gonzaponte, line 80: raise an error here. The only way to get to the end of this function if is `method` is not a `NormMethod`

invisible_cities/icaros/correction_functions.py
```diff
@@ -0,0 +1,224 @@
...
+
+def apply_3Dmap(krmap        : pd.DataFrame,
+                norm_method  : NormMethod,
+                dt           : pd.core.series.Series,
+                x            : pd.core.series.Series,
+                y            : pd.core.series.Series,
+                E            : pd.core.series.Series,
+                xy_params    : dict = None,
+                keV          : bool = False) -> pd.core.series.Series:
```
gonzaponte: I think we discussed using an arbitraty unit instead of `keV`

invisible_cities/icaros/correction_functions.py
```diff
@@ -0,0 +1,224 @@
...
+    Ec = E * correction_factor
+
+    if keV:
+        Ec = Ec * (41.55 / norm)
+
+    return Ec
+
+
+
+def apply_correctionmap_inplace(kdst        : pd.DataFrame,
```
gonzaponte:
```suggestion
def apply_correctionmap_inplace_kdst(kdst        : pd.DataFrame,
```

invisible_cities/cities/zemrude.py
```diff
@@ -0,0 +1,293 @@
...
+from .. types   .symbols          import NormMethod, SelRegionMethod, MapFitFunction
+from .. icaros  .control_plots_functions import make_control_plots, plot_time_evolution_with_errors_and_dates
+
+
+from pathlib import Path
+from typing import Sequence
+from typing import Optional
+from typing import Tuple
+from typing import Union
+from .components import *
```
gonzaponte: Replace `*` with the actual imports you need.
Tip: remove this line and run `pyflakes` on this file

gonzaponte (changes requested):
another round of comments

invisible_cities/cities/zemrude.py
```diff
@@ -0,0 +1,293 @@
...
+from .. icaros  .krmap_functions         import gaussian_fit_ready
+from .. icaros  .krmap_functions         import get_median
+from .. icaros  .krmap_functions         import compute_metadata
+from .. icaros  .krmap_functions         import get_time_evol
+from .. icaros  .krmap_functions         import save_map
+from .. icaros  .control_plots_functions import make_control_plots
+from .. icaros  .control_plots_functions import plot_time_evolution_with_errors_and_dates
+from .. types   .symbols              import NormMethod
+from .. types   .symbols              import SelRegionMethod
+from .. types   .symbols              import MapFitFunction
```
gonzaponte: align and sort IC imports

invisible_cities/cities/zemrude.py
```diff
@@ -0,0 +1,293 @@
...
+from typing import Tuple
+from typing import Union
+from typing import Callable
+from typing import List
+from typing import Iterator
+from typing import Dict
+
+
+def concatenated_dsts_from_files(path: List[str], group: str, node:str)-> Iterator[Dict[str,Union[pd.DataFrame, int, np.ndarray]]]:
+
```
gonzaponte: remove this blank line

invisible_cities/cities/zemrude.py
```diff
@@ -0,0 +1,293 @@
...
+def select_dst(dtrms2_low, dtrms2_upp, low_xrays, high_xrays, low_S2t, high_S2t, R_max, low_DT, high_DT, low_nsipm, high_nsipm):
+    def apply_selections_dst(df):
+        return apply_selections(df, dtrms2_low, dtrms2_upp, low_xrays, high_xrays, low_S2t, high_S2t, R_max, low_DT, high_DT, low_nsipm, high_nsipm)
+    return apply_selections_dst
+
+
+def create_selfmap(xy_range, dt_range, xy_nbins, dt_nbins, S2e_range, fit_function, nbins, min_events):
+    if fit_function == MapFitFunction.gaussian:
+        fit_function = gaussian_fit_ready
```
gonzaponte: This should be
```suggestion
        fit_function = gaussian_fit_ready(nbins, min_events)
```
right?

invisible_cities/cities/zemrude.py
```diff
@@ -0,0 +1,293 @@
...
+                              , args = 'dst'
+                              , out = ('selected_dst', 'efficiencies'))
+
+
+    if fit_function is MapFitFunction.gaussian:
+        fit_function = gaussian_fit_ready(min_events,
+                                          nbins)
+    else:
+        fit_function = get_median
+
```
gonzaponte: This is already done in `create_selfmap`

invisible_cities/cities/zemrude.py
```diff
@@ -0,0 +1,287 @@
...
+                                            dt_range,
+                                            xy_nbins,
+                                            dt_nbins)
+                               , args = ('selected_dst', '3D_krmap')
+                               , out = 'metadata')
+
+    apply_3Dmap_to_data = fl.map( apply_selfmap(norm_method,
+                                                xy_params,
+                                                'Ec_2',
+                                                 unit = keV)
```
gonzaponte: 4th argument is misaligned

invisible_cities/cities/zemrude.py
```diff
@@ -0,0 +1,287 @@
...
+    get_t_evol = fl.map( time_evol(slice_hours,
+                                    x0,
+                                    y0,
+                                    shape,
+                                    shape_size,
+                                    dtbins_dv,
+                                    s1_DTrange,
+                                    bins_Ec,
+                                    error
+                                    )
```
gonzaponte: revisit alignment

invisible_cities/cities/zemrude.py
```diff
@@ -0,0 +1,287 @@
...
+                                                 bins,
+                                                 dtr2_bins,
+                                                 statistic,
+                                                 x0,
+                                                 y0,
+                                                 shape,
+                                                 shape_size,
+                                                 xy_range_plot,
+                                                 names
+                                                 )
```
gonzaponte: revisit alignment

invisible_cities/cities/zemrude.py
```diff
@@ -0,0 +1,287 @@
...
+                    , pipe   = fl.pipe(apply_preliminary_map,
+                                           apply_selections,
+                                           compute_3D_map,
+                                           compute_metadata,
+                                           apply_3Dmap_to_data,
+                                           get_t_evol,
+
+                                           fl.fork(save_everything,
+                                                   make_control_plots,
+                                                   plot_t_evol))
```
gonzaponte: revisit alignment

invisible_cities/cities/zemrude.py
```diff
@@ -0,0 +1,287 @@
...
+
+                                 , args = ('dst', 'selected_dst', 'corrected_dst','efficiencies', 'run_number')
+                                 )
+
+    plot_t_evol = fl.sink(time_evol_plots(plots_out)
+                          , args = ('time_evol'))
+
+
+    with tb.open_file(file_out, "w", filters=tbl.filters(compression)):
+        pass
```
gonzaponte, line 270: is this to overwrite the output file, if it exists and set the compression? If so, explain it in a comment

invisible_cities/core/fit_functions_test.py
```diff
@@ -252,6 +252,34 @@ def test_power_at_zero(a, b):
...
+    assert np.all(np.diff(y) >= 0)
+
+
+def test_sigmoid_shape(): ##input shape = output shape
+    x = np.linspace(-10, 10, 100)
+    y = fitf.sigmoid(x, scale = 1, inflection = 0, slope = 1, offset = 0)
+
+    assert y.shape == x.shape
+
+
```
gonzaponte: Remove comments

invisible_cities/icaros/control_plots_functions.py
```diff
@@ -0,0 +1,782 @@
...
+from scipy import stats
+
+from .. types.symbols import SelRegionMethod
+from .. core.fit_functions import fit
+from .. core.fit_functions import profileX
+from .. core.fit_functions import expo
+from .. core.fit_functions import sigmoid
+from .. core.core_functions import in_range
+from .. core.core_functions import shift_to_bin_centers
+from .. icaros.lifetime_vdrift_functions import select_lifetime_region
```
gonzaponte: sort and align imports

invisible_cities/icaros/correction_functions.py
```diff
@@ -0,0 +1,219 @@
...
+
+    raise ValueError(f'Unsuported NormMethod: {method}')
+
+
+def apply_3Dmap(krmap       : pd.DataFrame,
+                norm_method : NormMethod,
+                dt          : pd.core.series.Series,
+                x           : pd.core.series.Series,
+                y           : pd.core.series.Series,
+                E           : pd.core.series.Series,
```
gonzaponte: change `pd.Series` to `np.ndarray`

invisible_cities/icaros/correction_functions.py
```diff
@@ -0,0 +1,219 @@
...
+                xy_params   : dict  = None,
+                unit        : Union[float, NoneType]  = keV) -> pd.core.series.Series:
```
gonzaponte:
```suggestion
                xy_params   : Union[dict, NoneType]  = None,
                unit        : Union[float, NoneType]  = keV) -> np.ndarray:
```

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,722 @@
...
+from .. icaros.lifetime_vdrift_functions import select_lifetime_region
+from .. icaros.lifetime_vdrift_functions import compute_drift_v
+
+from typing import Callable
+from typing import Union
+
+"""
+This script contains functions to compute empty maps, 3D maps from kdsts and merge them to create an 'uniform' map.
+It also contains functions to get and store map info (metadata) and time evolution.
+"""
```
gonzaponte: This goes at the very beginning of the file, before the imports.

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,722 @@
...
+    Nan_map = pd.DataFrame(np.nan,
+                           index=range(len(combinations)),
+                           columns=['nevents', 'mu', 'sigma', 'mu_error', 'sigma_error']
+                           )
+    Nan_map = pd.concat([combinations, Nan_map], axis=1)
+
+    return Nan_map
+
+
+def get_median(var : pd.core.series.Series) -> pd.DataFrame:
```
gonzaponte:
```suggestion
def get_median(var : np.ndarray) -> pd.DataFrame:
```

invisible_cities/icaros/krmap_functions.py
```diff
@@ -0,0 +1,722 @@
...
+                         'mu'         : mu,
+                         'sigma'      : sigma,
+                         'mu_error'   : sigma/np.sqrt(len(var)),
+                         'sigma_error': np.nan},
+                         index = [0]
+                         )
+
+
+
+def gaussian_fit(var        : pd.core.series.Series,
```
gonzaponte:
```suggestion
def gaussian_fit(var        : np.ndarray,
```

invisible_cities/icaros/lifetime_vdrift_functions.py
```diff
@@ -0,0 +1,91 @@
+import numpy as np
+import pandas as pd
+from typing import Tuple
+from invisible_cities.core.stat_functions import poisson_sigma
+import invisible_cities.core.fit_functions as fit
+from invisible_cities.core.core_functions import in_range, shift_to_bin_centers
+from invisible_cities.types.symbols import SelRegionMethod
+
```
gonzaponte, line 3: organize and align the imports

invisible_cities/icaros/lifetime_vdrift_functions.py
```diff
@@ -0,0 +1,91 @@
...
+    if shape is SelRegionMethod.square:
+
+        x1, y1 = x0-shape_size, y0-shape_size
+        mask_df = in_range(df.X, x1, x1+2*shape_size) & in_range(df.Y, y1, y1+2*shape_size)
+
+
+    if shape is SelRegionMethod.circle:
+
+        mask_df = (df.X - x0)**2 + (df.Y - y0)**2 <= shape_size**2
+
```
gonzaponte, line 47: remove spurious blank lines

invisible_cities/icaros/lifetime_vdrift_functions_test.py
```diff
@@ -0,0 +1,96 @@
...
+import pandas as pd
+import numpy  as np
+
+from pytest import fixture
+
+from .. core   .core_functions import in_range
+from .. core   .core_functions import fix_random_seed
+from .. types  .symbols        import SelRegionMethod
+from .. icaros .lifetime_vdrift_functions import select_lifetime_region
+from .. icaros .lifetime_vdrift_functions import compute_drift_v
```
gonzaponte: align imports

gonzaponte (commented):
A few cosmetic changes and we are done. We will rewrite the history together to reduce the number of commits.

invisible_cities/cities/zemrude.py
```diff
@@ -0,0 +1,275 @@
...
+            ,pipe   = fl.pipe(apply_preliminary_map,
+                               apply_selections,
+                               compute_3D_map,
+                               compute_metadata
+                               apply_3Dmap_to_data,
+                               get_t_evol,
+
+                       fl.fork(save_everything,
+                               make_control_plots,
+                               plot_t_evol))
```
gonzaponte:
```suggestion
            ,pipe   = fl.pipe(apply_preliminary_map,
                               apply_selections,
                               compute_3D_map,
                               compute_metadata
                               apply_3Dmap_to_data,
                               get_t_evol,

                               fl.fork(save_everything,
                                       make_control_plots,
                                       plot_t_evol))
```

invisible_cities/cities/zemrude.py
```diff
@@ -0,0 +1,275 @@
...
+        return apply_selections(df, dtrms2_low, dtrms2_upp, low_xrays, high_xrays, low_S2t, high_S2t, R_max, low_DT, high_DT, low_nsipm, high_nsipm)
+    return apply_selections_dst
+
+
+def create_selfmap(xy_range, dt_range, xy_nbins, dt_nbins, S2e_range, fit_function, nbins, min_events):
+    if fit_function == MapFitFunction.gaussian:
+        fit_function = gaussian_fit_ready(nbins, min_events)
+    else:
+        fit_function = get_median
+    def create_map(df):
```
gonzaponte, line 71: Let's add a protection here:

```suggestion
    if fit_function == MapFitFunction.gaussian:
        fit_function = gaussian_fit_ready(nbins, min_events)
    elif fit_function == MapFitFunction.median:
        fit_function = get_median
    else:
        raise ValueError(f"Invalid fit function {fit_function}")

    def create_map(df):
```

gonzaponte (approved):
The 3D version of our beloved Kr maps. An important addition developed with peras, reasonably tested and validated externally. Good job!

PR #950, opened 2026-01-14
Avoid floating point comparison issues

gonzaponte:
Some tests that check for an "exact result" have been problematic because they rely on floating point comparisons. This is machine-dependent, which means a test may pass locally, but fail on GHA or viceversa. With direct comparisons, this is usually addressed by adding a tolerance, which we already do. However there are other cases where this isn't enough
- cumulative rounding errors: typical of beersheba, where the iterative process seems to accumulate errors beyond what one might expect using doubles (?)
- discretization of continuous values: typical of paolina, where voxelization is sensible to tiny differences between floating point values, despite our attempts to avoid them.

In this PR, we bypass all these issues by rounding values to a safe number while maintaining enough precision.

### DIFF
--- a/invisible_cities/reco/paolina_functions.py
+++ b/invisible_cities/reco/paolina_functions.py
@@ -32,6 +32,14 @@ def bounding_box(seq : BHit) -> Sequence[np.ndarray]:
     return (reduce(np.minimum, posns, MAX3D),
             reduce(np.maximum, posns, MIN3D))
 
+def round_hits_positions_in_place(hits, decimals=5):
+    """
+    Rounds the hits positions to `decimals` decimals to avoid floating point
+    comparison issues. The operation is performed inplace to avoid an
+    unnecessary copy.
+    """
+    for hit in hits:
+        hit.xyz = np.round(hit.xyz, decimals)
 
 def voxelize_hits(hits             : Sequence[BHit],
                   voxel_dimensions : np.ndarray,
--- a/invisible_cities/reco/paolina_functions_test.py
+++ b/invisible_cities/reco/paolina_functions_test.py
@@ -33,6 +33,7 @@ from .. evm.event_model import Cluster
 from .. evm.event_model import Voxel
 
 from . paolina_functions import bounding_box
+from . paolina_functions import round_hits_positions_in_place
 from . paolina_functions import energy_of_voxels_within_radius
 from . paolina_functions import find_extrema
 from . paolina_functions import find_extrema_and_length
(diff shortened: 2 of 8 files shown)


Luismiguelvillar (commented):
It looks good to me. I just have a question regarding the chosen precision. Is there a specific reason for rounding to 6 decimal places? 

`deco_dst.E = np.round(deco_dst.E, 6)`

gonzaponte:
Not a fundamental reason, but a practical one. Rounding to more digits produced discrepancies between my computer and GHA. 6 digits seems enough to maintain precision while avoiding these issues.

invisible_cities/reco/paolina_functions_test.py
```diff
@@ -165,6 +166,24 @@ def sum_energy(seq):
     assert sum_energy(voxels) == approx(sum_energy(hits))
 
 
+@given(bunch_of_hits)
+def test_round_hits_positions_in_place(hits):
```
Luismiguelvillar, line 167: Should we also test how the function behaves with edge cases such as NaNs, infinities, or empty hit lists?

Luismiguelvillar:
> Not a fundamental reason, but a practical one. Rounding to more digits produced discrepancies between my computer and GHA. 6 digits seems enough to maintain precision while avoiding these issues.

Makes sense! But maybe it will be a good idea to let the user change this precision, similar to the xyz in  hits?

invisible_cities/reco/paolina_functions.py
```diff
@@ -32,6 +32,14 @@ def bounding_box(seq : BHit) -> Sequence[np.ndarray]:
     return (reduce(np.minimum, posns, MAX3D),
             reduce(np.maximum, posns, MIN3D))
 
+def round_hits_positions_in_place(hits, decimals=5):
```
Luismiguelvillar, line 35: It looks good, and modifying the hits in place was a good choice.
My only minor concern is that the function is not protected against NaNs, infinities, or other non-finite values. This is not particularly critical, since `np.round()` will simply leave those non-finite values unchanged (furthermore, I do not think hits with non-finite coordinates will make it to this point).

gonzaponte: I guess the expected result in that case (that shouldn't happen, but I've seen worse) is to leave those values unchanged. I'll add dedicated tests as requested above.

Luismiguelvillar (approved):
Everything looks fine!

gonzaponte:
@mcidlaso please merge

PR #952, opened 2026-03-02
Clean up the mess

gonzaponte:
This PR:

- Removes Penthesilea out of the way so we don’t need to maintain it going forward
- Cleans up unused files, imports, variables, and other dead code
- Makes no changes to production logic — this is purely cosmetic
- Adds pyflakes to the dependencies and a GHA step that verifies the source files for errors or unused code.


invisible_cities/database/test_data/dst_NEXT_v1_00_05_Kr_ACTIVE_0_0_7bar_PMP_10evt.h5
jwaiton, line 1: Is it a good idea to remove all these files? I'm assuming it's not an issue as they'll be retrievable historically, but I'd prefer it such that if someone did need to resurrect penthesilea, they'd be able to ensure the tests still work :)

invisible_cities/cities/beersheba.py
```diff
@@ -336,7 +336,7 @@ def drop_isolated( distance   : List[float],
         drop = drop_isolated_sensors(distance, redist_var)
     elif len(distance) == 3:
         if nhits is None:
-            raise TypeError(f"Applying 3-dimensional dropping of isolated hits requires parameter nhits which is missing.")
+            raise TypeError("Applying 3-dimensional dropping of isolated hits requires parameter nhits which is missing.")
```
jwaiton, line 339: For my understanding, is there a particular stylistic reason for the recent purging of f-strings? I understand that they're not doing anything here, but they also do no harm right?

invisible_cities/core/testing_utils_test.py
```diff
@@ -75,7 +75,7 @@ def test_assert_tables_equality_fails_different_names():
         assert_tables_equality(table1, table2)
 
 
-def test_assert_tables_equality_fails_different_values():
+def test_assert_tables_equality_fails_different_values_2():
```
jwaiton: Nitpicky, but I think this test should sit next to its sibling `test_assert_tables_equality_fails_different_values()`, or its name should be more descriptive, perhaps `test_assert_tables_equality_fails_different_columns` based on the descriptive comment.

invisible_cities/reco/corrections_test.py
```diff
@@ -290,7 +290,7 @@ def test_get_normalization_factor_region_norm_raises_exception_when_no_range(map
                   get_normalization_factor,
                   map_e, NormStrategy.region, **options)
 
-def test_get_normalization_factor_custom_norm_raises_exception_when_no_value(map_filename):
+def test_get_normalization_factor_custom_norm_raises_exception_when_no_value_2(map_filename):
```
jwaiton: As far as I can tell `test_get_normalization_factor_custom_norm_raises_exception_when_no_value()` and `test_get_normalization_factor_custom_norm_raises_exception_when_no_value_2()` are identical right? Why have both?

jwaiton (changes requested):
Looks good! Glad to see a museum being opened 🥳 

I have a few relatively mundane comments, otherwise it looks ready to go!

invisible_cities/cities/beersheba.py
gonzaponte: Perhaps the strongest reason is that it is misleading: it signals an intention that isn't there. Something like this depends on something, when it doesn't.

invisible_cities/database/test_data/dst_NEXT_v1_00_05_Kr_ACTIVE_0_0_7bar_PMP_10evt.h5
gonzaponte: Mmm, I agree with your logic, but I am also weary of all the mess we are keeping around. I will move them to the museum as well.

invisible_cities/reco/corrections_test.py
gonzaponte: It seems they are the same indeed. I just changed the name cause pyflakes was complaining and didn't pay attention. Will delete the duplicate.

jwaiton (approved):
This PR implements some 'spring cleaning' of the repository, removing an unused city and all its related components, improving the cosmetics and implementing pyflakes within our GHA testing suite.

Great work :)

PR #953, opened 2026-03-05
Implement Pyrrha functionalities in Irene

Ian0sborne:
This PR addresses the data processing issues cause by the SiPM thresholds in Irene. It retains standard Irene functionalities but also implements the new Pyrrha selection methods for SiPMs, which relies on a spatial selection rather than a charge threshold. 

The user is free to decide how they would rather process the raw waveforms (using the old Irene or Pyrrha), as the config file allows for either method to be selected.

### DIFF
--- a/invisible_cities/calib/calib_sensors_functions.py
+++ b/invisible_cities/calib/calib_sensors_functions.py
@@ -136,15 +136,18 @@ def pmt_subtract_maw(cwfs, n_maw=100):
     return cwfs - maw
 
 
-def calibrate_sipms(sipm_wfs, adc_to_pes, *, bls_mode=BlsMode.mode):
+def calibrate_sipms(sipm_wfs, adc_to_pes, *, thr = None, bls_mode=BlsMode.mode):
     """
     Subtract baseline, and calibrates waveforms to pes.
     """
-    #thr  = to_col_vector(np.full(sipm_wfs.shape[0], thr))
     bls  = subtract_baseline(sipm_wfs, bls_mode=bls_mode)
     cwfs = calibrate_wfs(bls, adc_to_pes)
-    return cwfs
-    #return np.where(cwfs > thr, cwfs, 0)
+    if thr is None:
+        return cwfs
+    else:
+        # if you apply a threshold here, apply as usual
+        thr  = to_col_vector(np.full(sipm_wfs.shape[0], thr))
+        return np.where(cwfs > thr, cwfs, 0)
 
 
 def subtract_mean  (wfs): return subtract_baseline(wfs, bls_mode=BlsMode.mean  )
--- a/invisible_cities/calib/calib_sensors_functions.py
+++ b/invisible_cities/calib/calib_sensors_functions.py
@@ -136,15 +136,15 @@ def pmt_subtract_maw(cwfs, n_maw=100):
     return cwfs - maw
 
 
-def calibrate_sipms(sipm_wfs, adc_to_pes, thr, *, bls_mode=BlsMode.mode):
+def calibrate_sipms(sipm_wfs, adc_to_pes, *, bls_mode=BlsMode.mode):
     """
-    Subtracts the baseline, calibrates waveforms to pes
-    and suppresses values below `thr` (in pes).
+    Subtract baseline, and calibrates waveforms to pes.
     """
-    thr  = to_col_vector(np.full(sipm_wfs.shape[0], thr))
+    #thr  = to_col_vector(np.full(sipm_wfs.shape[0], thr))
     bls  = subtract_baseline(sipm_wfs, bls_mode=bls_mode)
     cwfs = calibrate_wfs(bls, adc_to_pes)
-    return np.where(cwfs > thr, cwfs, 0)
+    return cwfs
+    #return np.where(cwfs > thr, cwfs, 0)
 
 
 def subtract_mean  (wfs): return subtract_baseline(wfs, bls_mode=BlsMode.mean  )
--- a/invisible_cities/calib/calib_sensors_functions.py
+++ b/invisible_cities/calib/calib_sensors_functions.py
@@ -136,18 +136,13 @@ def pmt_subtract_maw(cwfs, n_maw=100):
     return cwfs - maw
 
 
-def calibrate_sipms(sipm_wfs, adc_to_pes, *, thr = None, bls_mode=BlsMode.mode):
+def calibrate_sipms(sipm_wfs, adc_to_pes, *, bls_mode=BlsMode.mode):
     """
     Subtract baseline, and calibrates waveforms to pes.
     """
     bls  = subtract_baseline(sipm_wfs, bls_mode=bls_mode)
     cwfs = calibrate_wfs(bls, adc_to_pes)
-    if thr is None:
-        return cwfs
-    else:
-        # if you apply a threshold here, apply as usual
-        thr  = to_col_vector(np.full(sipm_wfs.shape[0], thr))
-        return np.where(cwfs > thr, cwfs, 0)
+    return cwfs
 
 
 def subtract_mean  (wfs): return subtract_baseline(wfs, bls_mode=BlsMode.mean  )
--- a/invisible_cities/calib/calib_sensors_functions.py
+++ b/invisible_cities/calib/calib_sensors_functions.py
@@ -59,8 +59,9 @@ def modes  (wfs): return to_col_vector(mode  (wfs, axis=1))
 
 def subtract_baseline(wfs, *, bls_mode=BlsMode.mean):
     """
-    Subtract the baseline to all waveforms in the input
-    with a specific algorithm.
+    Subtract the baseline from all waveforms in the input
+    with a specific algorithm, leaving individual samples
+    that are already 0 untouched.
 
     Parameters
     ----------
@@ -75,15 +76,18 @@ def subtract_baseline(wfs, *, bls_mode=BlsMode.mean):
     Returns
     -------
     bls: np.ndarray with shape (n, m)
-        Baseline-subtracted waveforms.
+        Baseline-subtracted waveforms, with originally-zero
+        samples left as 0.
     """
 
-    if   bls_mode is BlsMode.mean     : return wfs - means     (wfs)
-    elif bls_mode is BlsMode.median   : return wfs - medians   (wfs)
-    elif bls_mode is BlsMode.mode     : return wfs - modes     (wfs)
-    elif bls_mode is BlsMode.scipymode: return wfs - scipy_mode(wfs, axis=1)
+    if   bls_mode is BlsMode.mean     : baseline = means     (wfs)
+    elif bls_mode is BlsMode.median   : baseline = medians   (wfs)
+    elif bls_mode is BlsMode.mode     : baseline = modes     (wfs)
+    elif bls_mode is BlsMode.scipymode: baseline = scipy_mode(wfs, axis=1)
     else:
         raise TypeError(f"Unrecognized baseline subtraction option: {bls_mode}")
+
+    return np.where(wfs == 0, 0, wfs - baseline) # subtracts the baseline to non-zero samples only
 
 
 def calibrate_wfs(wfs, adc_to_pes):
--- a/invisible_cities/calib/calib_sensors_functions.py
+++ b/invisible_cities/calib/calib_sensors_functions.py
@@ -157,10 +157,10 @@ def sipm_subtract_median_and_calibrate(sipm_wfs, adc_to_pes): return calibrate_w
 
 # Dict of functions for SiPM processing
 sipm_processing = {
-    SiPMCalibMode.subtract_mode            :      subtract_mode                ,# For gain extraction
-    SiPMCalibMode.subtract_median          :      subtract_median              ,# For gain extraction
-    SiPMCalibMode.subtract_mode_calibrate  : sipm_subtract_mode_and_calibrate  ,# For PDF calculation
-    SiPMCalibMode.subtract_mean_calibrate  : sipm_subtract_mean_and_calibrate  ,# For PDF calculation
-    SiPMCalibMode.subtract_median_calibrate: sipm_subtract_median_and_calibrate,# For PDF calculation
-    SiPMCalibMode.subtract_mode_zs         : calibrate_sipms                    # For data processing
+    SiPMCalibMode.subtract_mode              :      subtract_mode                ,# For gain extraction
+    SiPMCalibMode.subtract_median            :      subtract_median              ,# For gain extraction
+    SiPMCalibMode.subtract_mode_calibrate    : sipm_subtract_mode_and_calibrate  ,# For PDF calculation
+    SiPMCalibMode.subtract_mean_calibrate    : sipm_subtract_mean_and_calibrate  ,# For PDF calculation
+    SiPMCalibMode.subtract_median_calibrate  : sipm_subtract_median_and_calibrate,# For PDF calculation
+    SiPMCalibMode.subtract_baseline_calibrate: calibrate_sipms                    # For data processing
 }
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -836,6 +836,28 @@ def threshold_sipm_selection(thr_sipm_type
     return threshold_sipm_selection
 
 
+def pyrrha_sipm_selection(selection_function : Callable
+                         , selection_kwargs    : dict
+                         , proximity_threshold : float
+                         , padding_radius      : float
+                         , run_number : int
+                         , detector_db : str):
+    '''
+    Function that applies a generic selection function to the sipms, which can be used to 
+    implement a spatial SiPM selection method (called Pyrrha).
+    '''
+    def pyrrha_sipm_selection(wfs, indices):
+        return wfm.spatial_selection_method(wfs, 
+                                            selection_function, 
+                                            selection_kwargs, 
+                                            proximity_threshold,
+                                            padding_radius,
+                                            run_number,
+                                            detector_db)
+
+    return pyrrha_sipm_selection
+
+
 def calibrate_with_mean(dbfile, run_number):
     DataSiPM   = load_db.DataSiPM(dbfile, run_number)
     adc_to_pes = np.abs(DataSiPM.adc_to_pes.values)
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -84,6 +84,7 @@ from .. types  .ic_types          import        types_dict_summary
 from .. types  .ic_types          import         types_dict_tracks
 from .. types  .symbols           import                    WfType
 from .. types  .symbols           import                   CutAlgo
+from .. types  .symbols           import       SiPMSelectionMethod
 from .. types  .symbols           import               RebinMethod
 from .. types  .symbols           import                SiPMCharge
 from .. types  .symbols           import                   BlsMode
@@ -836,12 +837,12 @@ def threshold_sipm_selection(thr_sipm_type
     return threshold_sipm_selection
 
 
-def pyrrha_sipm_selection(selection_function : Callable
+def pyrrha_sipm_selection(selection_method     : SiPMSelectionMethod
                          , selection_kwargs    : dict
                          , proximity_threshold : float
                          , padding_radius      : float
-                         , run_number : int
-                         , detector_db : str):
+                         , run_number          : int
+                         , detector_db         : str):
     '''
     Function that applies a generic selection function to the sipms, which can be used to 
     implement a spatial SiPM selection method (called Pyrrha).
@@ -849,7 +850,7 @@ def pyrrha_sipm_selection(selection_function : Callable
     def pyrrha_sipm_selection(wfs, indices):
         return wfm.spatial_selection_method(wfs, 
                                             indices,
-                                            selection_function, 
+                                            selection_method, 
                                             selection_kwargs, 
                                             proximity_threshold,
                                             padding_radius,
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -797,6 +797,23 @@ def calibrate_sipms(dbfile, run_number):
     return calibrate_sipms
 
 
+
+
+def threshold_sipm_selection(thr_sipm_type, thr_sipm, detector_db, run_number):
+    '''
+    Function that applies thresholding to the sipms in standard irene manner
+    '''
+
+    # extract sipm threshold
+    sipm_thr = get_actual_sipm_thr(thr_sipm_type, thr_sipm, detector_db, run_number)
+
+    def threshold_sipm_selection(wfs):
+        return wfm.charge_threshold_method(wfs, threshold = sipm_thr)
+
+    return threshold_sipm_selection
+
+
+
 def calibrate_with_mean(dbfile, run_number):
     DataSiPM   = load_db.DataSiPM(dbfile, run_number)
     adc_to_pes = np.abs(DataSiPM.adc_to_pes.values)
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -809,8 +809,8 @@ def apply_cutting_function(algo, **cutting_params):
         # temporary solution, think of a nicer method
         func = threshold_sipm_selection(**cutting_params)
 
-    def apply_cutting_function(wfs, indices):
-        return func(wfs, indices)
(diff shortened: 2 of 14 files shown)


invisible_cities/calib/calib_sensors_functions.py
```diff
@@ -136,15 +136,18 @@ def pmt_subtract_maw(cwfs, n_maw=100):
...
-    thr  = to_col_vector(np.full(sipm_wfs.shape[0], thr))
     bls  = subtract_baseline(sipm_wfs, bls_mode=bls_mode)
     cwfs = calibrate_wfs(bls, adc_to_pes)
-    return np.where(cwfs > thr, cwfs, 0)
+    if thr is None:
+        return cwfs
+    else:
+        # if you apply a threshold here, apply as usual
+        thr  = to_col_vector(np.full(sipm_wfs.shape[0], thr))
+        return np.where(cwfs > thr, cwfs, 0)
```
Ian0sborne: Even though this is not used in the code, this is needed for 4 tests in `calib_sensor_functions_test.py`.

gonzaponte: Add another function that does the thresholds suppression, redo those tests doing the threshold suppression outside `calibrate_sipms`, and finally remove the `else` branch here

Ian0sborne: The function `charge_threshold_method()` in `reco/wfm_functions.py` does the zero suppression step. Should I make a function that does just that? If I do this, do I remove the test that rely on zero suppression from `calib_sensor_functions_test.py` and implement them specifically for this new function?

invisible_cities/calib/calib_sensors_functions.py
Ian0sborne: This name may be misleading now after the change, however I'm reluctant to change it in case it breaks other things (maybe configs?)

gonzaponte: Change it, if it affects anyone, that's @RhiannonCoupeNeutrinos, and you are in close proximity to resolve it quickly :)

invisible_cities/cities/components.py
```diff
@@ -1226,7 +1274,7 @@ def integrate_wfs(wfs):
 def compute_and_write_pmaps(detector_db, run_number, pmt_samp_wid, sipm_samp_wid,
                   s1_lmax, s1_lmin, s1_rebin_stride, s1_stride, s1_tmax, s1_tmin,
                   s2_lmax, s2_lmin, s2_rebin_stride, s2_stride, s2_tmax, s2_tmin, thr_sipm_s2,
-                  h5out, sipm_rwf_to_cal=None):
+                  h5out, apply_cut, sipm_rwf_to_cal=None):
```
gonzaponte: I think we need a better name for `apply_cut.` Think of something along the lines of "sipm selection algorithm"

invisible_cities/reco/peak_functions.py
```diff
@@ -76,13 +77,18 @@ def build_pmt_responses(indices, times, widths, ccwf,
...
-     sipm_wfs)   = select_wfs_above_time_integrated_thr(sipm_wfs_,
-                                                        thr_sipm_s2)
+    if apply_cut is not None:
+        # apply cut before slicing and rebinning
+        (sipm_ids,
+         sipm_wfs)   = apply_cut(sipm_wfs)
+    else:
+        # give all sipm ids as index if no cut is applied
+        sipm_ids     = np.arange(sipm_wfs.shape[0])
+
```
gonzaponte: What does `apply_cut = None` represent? I thought there were only 2 options: threshold or pyrrha

invisible_cities/reco/peak_functions_test.py
```diff
@@ -350,10 +353,18 @@ def test_build_sipm_responses(wf_with_indices):
     wfs_slice       = wfs[:, indices]
     peak_integrals  = wfs_slice.sum(axis=1)
     below_thr_index = np.argmin (peak_integrals)
-    # next_float doesn't work here
     thr             = peak_integrals[below_thr_index] * 1.000001
+
+    # next_float doesn't work here
```
gonzaponte, line 356: This comment belongs above the previous line

invisible_cities/reco/wfm_functions.py
```diff
@@ -129,3 +136,211 @@ def compare_cwf_blr(cwf, pmtblr, event_list, window_size=500):
             DIFF.append(diff)
 
     return np.array(DIFF)
+
+
+def zero_wfs_below_threshold(wfs         : np.ndarray,
+                             zeroing_thr : Optional[float] = 2.) -> np.ndarray:
```
gonzaponte: The threshold should not be an optional parameter (update docstring as well)

invisible_cities/reco/wfm_functions.py
```diff
@@ -129,3 +136,211 @@ def compare_cwf_blr(cwf, pmtblr, event_list, window_size=500):
...
+    Boolean numpy array of shape (n_sipms,) where True indicates that the SiPM is selected.
+    """
+    charges = np.sum(wfs, axis=1)
+    threshold = np.median(charges) + nsigma * np.std(charges)
+    return charges >= threshold
+
+
+def charge_threshold_method(wfs             : np.ndarray,
+                            zeroing_thr     : Optional[float] = 2.,
+                            integration_thr : Optional[float] = 5.) -> Tuple[np.ndarray, np.ndarray]:
```
gonzaponte: Remove defaults here as well

invisible_cities/reco/wfm_functions.py
```diff
@@ -129,3 +136,211 @@ def compare_cwf_blr(cwf, pmtblr, event_list, window_size=500):
...
+    """
+    # zero entries below threshold
+    zwfs = zero_wfs_below_threshold(wfs, zeroing_thr)
+
+    # returns selected ids and waveforms above integral
+    return select_wfs_above_time_integrated_thr(zwfs, integration_thr)
+
+
+def top_n_method(wfs : np.ndarray,
+                 n   : Optional[int] = 10) -> np.ndarray:
```
gonzaponte: no default

invisible_cities/reco/wfm_functions.py
```diff
@@ -129,3 +136,211 @@ def compare_cwf_blr(cwf, pmtblr, event_list, window_size=500):
...
+    Returns
+    -------
+    2D array of shape (n_sipms, n_time_bins) containing the input waveforms with entries below threshold set to zero.
+    """
+    thr = to_col_vector(np.full(wfs.shape[0], zeroing_thr))
+    return np.where(wfs > thr, wfs, 0)
+
+
+def median_std_method(wfs    : np.ndarray,
+                      nsigma : Optional[float] = 3.) -> np.ndarray:
```
gonzaponte, line 163: Contrary to the other cases, I think this default is well motivated

invisible_cities/reco/wfm_functions.py
```diff
@@ -129,3 +136,211 @@ def compare_cwf_blr(cwf, pmtblr, event_list, window_size=500):
...
+    selected_ids[idx] = True
+    return selected_ids
+
+
+def kill_isolated_sipms(selected_ids        : np.ndarray,
+                        sipm_x              : np.ndarray,
+                        sipm_y              : np.ndarray,
+                        proximity_threshold : float) -> np.ndarray:
+    """
+    For the SiPMs that have passed the previous selection, scans through the SiPMs to check if they
```
gonzaponte: The initial sentence implies that the user needs to understand the overall flow to understand this function, which should (almost) never be the case. One should be able to understand a function on its own. They might not understand how it fits in the whole picture, but that's a different problem.

invisible_cities/reco/wfm_functions.py
```diff
@@ -129,3 +136,211 @@ def compare_cwf_blr(cwf, pmtblr, event_list, window_size=500):
...
+
+    return selected_ids_no_isolated
+
+
+def apply_circular_padding(selected_ids_no_isolated : np.ndarray,
+                           sipm_x                   : np.ndarray,
+                           sipm_y                   : np.ndarray,
+                           padding_radius           : float) -> np.ndarray:
+    """
+    For the SiPMs that pass the previous selection, creates circular padding of radius padding_radius,
```
gonzaponte: Same here

invisible_cities/reco/wfm_functions.py
```diff
@@ -129,3 +136,211 @@ def compare_cwf_blr(cwf, pmtblr, event_list, window_size=500):
...
+        sipm_y,
+        proximity_threshold
+    )
+
+    sipm_ids_with_signal_ = apply_circular_padding(
+        selected_ids_no_isolated_,
+        sipm_x,
+        sipm_y,
+        padding_radius
+    )
```
gonzaponte, line 344: why the trailing underscores?

gonzaponte (changes requested):
I will take a deeper look, but here is a first go.

invisible_cities/reco/wfm_functions.py
Ian0sborne: The idea is that they shouldn't be used by the user, but we can remove that

invisible_cities/reco/peak_functions.py
Ian0sborne: I think this is a leftover from when we were first making the implementation, it doesn't make sense anymore I agree

Ian0sborne: However in the functions prior to this one, `apply_cut` is set as `None` in the parameters, see:

https://github.com/next-exp/IC/pull/953/changes/b268b328b15ec69715987f1b3c5fe45a9d10f6b2#diff-da42ca1aac462421c8d637d77af246ebd19d1c5c9d1764c251b771b420a1cacdR103

https://github.com/next-exp/IC/pull/953/changes/b268b328b15ec69715987f1b3c5fe45a9d10f6b2#diff-da42ca1aac462421c8d637d77af246ebd19d1c5c9d1764c251b771b420a1cacdR130

https://github.com/next-exp/IC/pull/953/changes/b268b328b15ec69715987f1b3c5fe45a9d10f6b2#diff-da42ca1aac462421c8d637d77af246ebd19d1c5c9d1764c251b771b420a1cacdR154

Should we leave this as is?

Ian0sborne: Okay I removed the `else` clause and some tests for the `peak_functions.py` use to the `None` case for the `apply_cut` parameter, so now without the `else` fallback they fail

```
FAILED invisible_cities/reco/peak_functions_test.py::test_build_peak_development[S2-40-True] - TypeError: 'NoneType' object is not callable
FAILED invisible_cities/reco/peak_functions_test.py::test_find_peaks_s2_style - TypeError: 'NoneType' object is not callable
FAILED invisible_cities/reco/peak_functions_test.py::test_get_pmap - TypeError: 'NoneType' object is not callable
``` 

Should we redo the functions that are using `apply_cut = None` so that there is some parameter passes through?

invisible_cities/cities/components.py
Ian0sborne: I agree, I'll do `sipm_selection_algo`, but I will change it last so there's no back and forth

invisible_cities/reco/peak_functions.py
gonzaponte: I think there are only two or three options, but correct me if I'm wrong:
- `None` meant "no selection", so everything is kept. You would need to add this option to `CutAlgo`
- `None` meant "threshold selection". In this case you need to adapt the tests by explicitly sending the original default threshold parameters.
- `None` meant "threshold selection but with some (non-defaul) specific value of the parameters (e.g. thr=0). You would need to tweak the tests to specify these values as well.

invisible_cities/reco/wfm_functions.py
gonzaponte: Yes, please. These variables only exist within the scope of this function.

invisible_cities/reco/peak_functions.py
Ian0sborne: `None` in this case means no cut is applied at all, I think the best option would be to implement this explicitly as an option.

I noticed that in the old version of the tests that failed above they set `thr_sipm_s2 = -1`, this is essentially equivalent to setting no threshold the way I see it.

If you agree I'll make an option for no threshold and that should fix things.

gonzaponte: Yup. Go ahead.

Ian0sborne: Done. What do you think?

Also should I make these functions have `CutAlgo.no_cut` as the default option?

https://github.com/next-exp/IC/commit/b268b328b15ec69715987f1b3c5fe45a9d10f6b2#diff-da42ca1aac462421c8d637d77af246ebd19d1c5c9d1764c251b771b420a1cacdR103

https://github.com/next-exp/IC/commit/b268b328b15ec69715987f1b3c5fe45a9d10f6b2#diff-da42ca1aac462421c8d637d77af246ebd19d1c5c9d1764c251b771b420a1cacdR130

https://github.com/next-exp/IC/commit/b268b328b15ec69715987f1b3c5fe45a9d10f6b2#diff-da42ca1aac462421c8d637d77af246ebd19d1c5c9d1764c251b771b420a1cacdR154

invisible_cities/cities/components.py
```diff
@@ -797,19 +787,89 @@ def calibrate_pmts(cwf):# -> CCwfs:
...
     def calibrate_sipms(rwf):
         return csf.calibrate_sipms(rwf,
                                    adc_to_pes = adc_to_pes,
-                                   thr        = thr_sipm,
                                    bls_mode   = BlsMode.mode)
 
     return calibrate_sipms
 
 
+def apply_cutting_function(algo, **cutting_params):
```
```suggestion
def select_cutting_function(algo, **cutting_params):
```

or
```suggestion
def select_cutting_algorithm(algo, **cutting_params):
```


would be more appropriate

invisible_cities/cities/components.py
```diff
@@ -797,19 +787,89 @@ def calibrate_pmts(cwf):# -> CCwfs:
...
+        func = pyrrha_sipm_selection(**cutting_params)
+    elif algo is CutAlgo.no_cut:
+        func = no_cut_sipm_selection()
+    else:
+        raise ValueError(f"Unsupported cutting algorithm: {algo!r}. Expected one of {list(CutAlgo)}")
+
+    def apply_cutting_function(wfs):
+        return func(wfs)
+
+    return apply_cutting_function
```
gonzaponte: this is the same as `return func`. you can also use the return statement directly in the if/elif branches.

invisible_cities/cities/components.py
```diff
@@ -797,19 +787,89 @@ def calibrate_pmts(cwf):# -> CCwfs:
...
+    '''
+    Function that applies thresholding to the sipms in standard irene manner,
+    by zeroing all waveform values below a threshold.
+    '''
+    # assume that if the detector_db is None, you return sipm threshold as the number provided
+    if detector_db is None:
+        sipm_thr = thr_sipm
+    else:
+        # extract sipm threshold
+        sipm_thr = get_actual_sipm_thr(thr_sipm_type, thr_sipm, detector_db, run_number)
```
gonzaponte: If you add comments, make them say something that is not obvious from the code.  In this case, since it is more a description of the function overall (rather than an implementation detail or trick), I would use the docstring to explain _what_ the function does.

invisible_cities/reco/wfm_functions.py
```diff
@@ -129,3 +136,217 @@ def compare_cwf_blr(cwf, pmtblr, event_list, window_size=500):
...
+    Parameters
+    ----------
+    wfs         : 2D array of shape (n_sipms, n_time_bins) containing the waveforms of each SiPM.
+    zeroing_thr : Charge threshold for zero suppression in PE.
+
+    Returns
+    -------
+    2D array of shape (n_sipms, n_time_bins) containing the input waveforms with entries below threshold set to zero.
+    """
+    thr = to_col_vector(np.full(wfs.shape[0], zeroing_thr))
```
gonzaponte:
```suggestion
    thr = np.full((wfs.shape[0], 1), zeroing_thr)
```

invisible_cities/reco/wfm_functions.py
```diff
@@ -129,3 +136,217 @@ def compare_cwf_blr(cwf, pmtblr, event_list, window_size=500):
...
+
+    Returns
+    -------
+    Tuple of np arrays including all passing sipm ids and the corresponding waveforms
+    """
+    # zero entries below threshold
+    zwfs = zero_wfs_below_threshold(wfs, zeroing_thr)
+
+    # returns selected ids and waveforms above integral
+    return select_wfs_above_time_integrated_thr(zwfs, integration_thr)
```
gonzaponte: remove comments

invisible_cities/reco/wfm_functions.py
```diff
@@ -129,3 +136,217 @@ def compare_cwf_blr(cwf, pmtblr, event_list, window_size=500):
...
+    ----------
+    selected_ids_no_isolated : Boolean array of shape (n_sipms,) corresponding to the "relevant" SiPMs.
+    sipm_x                   : 1D array of shape (n_sipms,) containing the x positions of the SiPMs.
+    sipm_y                   : 1D array of shape (n_sipms,) containing the y positions of the SiPMs.
+    padding_radius           : Distance threshold in mm used to create circular padding around selected SiPMs.
+
+    Returns
+    -------
+    sipm_ids_with_signal : Boolean array of shape (n_sipms,) where True indicates that the SiPM is selected.
+    """
```
gonzaponte, line 284: Whenever possible, try to frame the function in a more general way. Here, you use variable names and an overall function explanation that is perhaps easy to follow if you read the whole Pyrrha implementation, but not on their own. It should be possible to read and interpret this function without understanding Pyrrha as a whole.

gonzaponte (changes requested):
A few more comments. I still need to check a few things, but at least you can make some progress while I do that.

invisible_cities/reco/wfm_functions.py
Ian0sborne: This caused one of the tests to fail so I reverted it back to the original

```
_______________________________________________ test_calibrate_sipms_individual_thresholds _______________________________________________

toy_sipm_signal = (array([[47, 44,  8, 41, 34,  0, 25, 85, 10, 11],
       [83, 75, 18, 89, 40, 65, 99, 14, 97, 89],
       [75, 63, 84,...6, 0.18372304, 0.40110688, 0.44504872, 0.48178363,
       0.26737329, 0.34721906, 0.39026542, 0.31321725, 0.2410451 ]))

    def test_calibrate_sipms_individual_thresholds(toy_sipm_signal):
        (signal_adc, adc_to_pes,
         _, signal_zs_individual_thresholds,
         _, individual_thresholds) = toy_sipm_signal
    
        ccwf = csf.calibrate_sipms(signal_adc, adc_to_pes, bls_mode=BlsMode.mode)
>       zs_wf = wfm.zero_wfs_below_threshold(ccwf, zeroing_thr=individual_thresholds)

invisible_cities/calib/calib_sensors_functions_test.py:236: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
invisible_cities/reco/wfm_functions.py:155: in zero_wfs_below_threshold
    thr = np.full((wfs.shape[0], 1), zeroing_thr)
/Users/user/miniconda/envs/IC-3.8-2024-06-08/lib/python3.8/site-packages/numpy/core/numeric.py:344: in full
    multiarray.copyto(a, fill_value, casting='unsafe')
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = (array([[ 8.],
       [89.],
       [17.],
       [ 3.],
       [ 6.],
       [59.],
       [ 2.],
       [23.],
     ...6, 0.18372304, 0.40110688, 0.44504872, 0.48178363,
       0.26737329, 0.34721906, 0.39026542, 0.31321725, 0.2410451 ]))
kwargs = {'casting': 'unsafe'}
relevant_args = (array([[ 8.],
       [89.],
       [17.],
       [ 3.],
       [ 6.],
       [59.],
       [ 2.],
       [23.],
     ...8372304, 0.40110688, 0.44504872, 0.48178363,
       0.26737329, 0.34721906, 0.39026542, 0.31321725, 0.2410451 ]), None)

>   ???
E   ValueError: could not broadcast input array from shape (10,) into shape (10,1)

<__array_function__ internals>:180: ValueError
```

gonzaponte (commented):
I'm happy with this. As discussed, let's rebase and fix the problems that may arise there. After that, some cosmetics and we are done!

invisible_cities/cities/components.py
```diff
@@ -802,20 +805,82 @@ def calibrate_pmts(cwf):# -> CCwfs:
...
+    else:
+        raise ValueError(f"Unsupported cutting algorithm: {algo!r}. Expected one of {list(CutAlgo)}")
+
+
+def threshold_sipm_selection(thr_sipm_type
+                            , thr_sipm
+                            , thr_sipm_s2
+                            , run_number
+                            , detector_db = None):
+    '''
```
```suggestion
def threshold_sipm_selection( thr_sipm_type
                            , thr_sipm
                            , thr_sipm_s2
                            , run_number
                            , detector_db = None):
    '''
```

invisible_cities/cities/components.py
```diff
@@ -802,20 +805,82 @@ def calibrate_pmts(cwf):# -> CCwfs:
...
+
+    return threshold_sipm_selection
+
+
+def pyrrha_sipm_selection(selection_method     : SiPMSelectionMethod
+                         , selection_kwargs    : dict
+                         , proximity_threshold : float
+                         , padding_radius      : float
+                         , run_number          : int
+                         , detector_db         : str):
```
```suggestion
def pyrrha_sipm_selection( selection_method     : SiPMSelectionMethod
                         , selection_kwargs    : dict
                         , proximity_threshold : float
                         , padding_radius      : float
                         , run_number          : int
                         , detector_db         : str):
```

invisible_cities/cities/components.py
```diff
@@ -802,20 +805,82 @@ def calibrate_pmts(cwf):# -> CCwfs:
...
+
+
+def threshold_sipm_selection(thr_sipm_type
+                            , thr_sipm
+                            , thr_sipm_s2
+                            , run_number
+                            , detector_db = None):
+    '''
+    Function that applies thresholding to the sipms in standard irene manner,
+    by zeroing all waveform values below a threshold.
```
gonzaponte: Please rephrase this not in terms of history, but in terms of behaviour

invisible_cities/reco/wfm_functions.py
```diff
@@ -3,10 +3,17 @@
 authors: J.J. Gomez-Cadenas, G. Martinez
 """
 import numpy as np
+from typing  import Optional
+from typing  import Callable
+from typing  import Tuple
 
-from .. core.core_functions import define_window
+from .. core.core_functions import define_window, to_col_vector
```
gonzaponte:
```suggestion
from .. core.core_functions import define_window
from .. core.core_functions import to_col_vector
```

invisible_cities/types/symbols.py
```diff
@@ -132,12 +132,12 @@ class SensorType(AutoNameEnumBase):
...
-    subtract_mode_calibrate   = auto()
-    subtract_mean_calibrate   = auto()
-    subtract_median_calibrate = auto()
-    subtract_mode_zs          = auto()
+    subtract_mode                        = auto()
+    subtract_median                      = auto()
+    subtract_mode_calibrate              = auto()
+    subtract_mean_calibrate              = auto()
+    subtract_median_calibrate            = auto()
+    subtract_baseline_calibrate          = auto()
```
gonzaponte:
```suggestion
    subtract_mode               = auto()
    subtract_median             = auto()
    subtract_mode_calibrate     = auto()
    subtract_mean_calibrate     = auto()
    subtract_median_calibrate   = auto()
    subtract_baseline_calibrate = auto()
```

gonzaponte (commented):
This is ready to be merged. Just some cosmetic changes. Please run `pyflakes` on the files you've modified to verify there aren't unused imports or variables.

gonzaponte (approved):
An important addition to the signal reconstruction, allowing for a more efficient and more correct storage of sipm information. Great job!

PR #954, opened 2026-03-09
Fix Z column dtype in hits and deco outputs

gonzaponte:
... which was 32-bit float. To be consistent with the rest of the variables, we force it to be 64-bit.


jwaiton (approved):
This PR updates the dtype for Z in hits and deco outputs by ensuring that it is stored as a pythonic `float` (64 bit by default).

All files that rely on this change have been altered to match the updated data type.

Good work!

Issue #920, opened 2024-11-09
Review tests in `psf_functions_test.py`


gonzaponte:
... specifically

- `test_add_empty_sensors_and_normalize_q`
- `test_add_empty_sensors_and_normalize_q_file`
- `test_hdst_psf_processing`

The input data for these tests is not the expected one. We expect hits that correspond to one z-sample and one sensor. Because these tests only check for continuity, we've never noticed they are incorrect.

We should also add more meaningful tests.

PR #955, opened 2026-03-09
Bump Python version to 3.13

gonzaponte:
Several changes, particularly in pandas, mean that we need to do things slightly differently.

### DIFF
--- a/invisible_cities/cities/beersheba.py
+++ b/invisible_cities/cities/beersheba.py
@@ -307,9 +307,12 @@ def cut_over_Q(q_cut, redist_var):
     cut = cut_and_redistribute_df(f"Q > {q_cut}", redist_var)
 
     def cut_over_Q(df):  # df shall be an event cdst
-        cdst = df.groupby(['event', 'npeak']).apply(cut).reset_index(drop=True)
-
-        return cdst
+        if not len(df): return df
+        df = (df.groupby("event time npeak".split())
+                .apply(cut, include_groups=False)
+                .reset_index(level=3, drop=True) # drop this artifical index level
+                .reset_index())                  # and restore the "real" index
+        return df
 
     return cut_over_Q
 
@@ -344,8 +347,11 @@ def drop_isolated( distance   : List[float],
 
 
     def drop_isolated(df): # df shall be an event cdst
-        df = df.groupby(['event', 'npeak']).apply(drop).reset_index(drop=True)
-
+        if not len(df): return df
+        df = (df.groupby("event time npeak".split())
+                .apply(drop, include_groups=False)
+                .reset_index(level=3, drop=True) # drop this artifical index level
+                .reset_index())                  # and restore the "real" index
         return df
 
     return drop_isolated
--- a/invisible_cities/database/load_db_test.py
+++ b/invisible_cities/database/load_db_test.py
@@ -11,6 +11,8 @@ from pytest  import fixture
 from pytest  import mark
 
 from . import load_db as DB
+from ..core.testing_utils import assert_dataframes_close
+
 
 def test_pmts_pd(db):
     """Check that we retrieve the correct number of PMTs."""
@@ -62,8 +64,8 @@ def test_DetectorGeometry(dbnew):
 
 
 def test_mc_runs_equal_data_runs(db):
-    assert (DB.DataPMT (db.detector, -3550).values == DB.DataPMT (db.detector, 3550).values).all()
-    assert (DB.DataSiPM(db.detector, -3550).values == DB.DataSiPM(db.detector, 3550).values).all()
+    assert_dataframes_close(DB.DataPMT (db.detector, -3550), DB.DataPMT (db.detector, 3550))
+    assert_dataframes_close(DB.DataSiPM(db.detector, -3550), DB.DataSiPM(db.detector, 3550))
 
 
 @fixture(scope='module')
@@ -141,7 +143,7 @@ def test_database_is_being_cached(db_fun, db):
         for item_first, item_second in zip(first_call, second_call):
             assert np.allclose(item_first, item_second)
     else:
-        assert np.all(first_call.values == second_call.values)
+        assert_dataframes_close(first_call, second_call)
     assert time_second_call < 1e6 * time_first_call
 
 
--- a/invisible_cities/detsim/sensor_utils.py
+++ b/invisible_cities/detsim/sensor_utils.py
@@ -66,8 +66,8 @@ def sensor_order(detector_db: str, run_number : int,
     n_pmt, n_sipm = get_n_sensors(detector_db, run_number)
     pmt_shape     = (n_pmt , length_pmt )
     sipm_shape    = (n_sipm, length_sipm)
-    def ordering(sensor_order : pd.Int64Index  ,
-                 sensor_resp  : np.ndarray     ,
+    def ordering(sensor_order : pd.Index  ,
+                 sensor_resp  : np.ndarray,
                  sensor_shape : Tuple[int, int]) -> np.ndarray:
         sensors = np.zeros(sensor_shape, int)
         sensors[sensor_order] = sensor_resp
--- a/invisible_cities/detsim/sensor_utils.py
+++ b/invisible_cities/detsim/sensor_utils.py
@@ -103,4 +103,4 @@ def pmt_and_sipm_bin_width(file_name: str) -> Tuple[float, float]:
         raise tb.NoSuchNodeError('No useful binning info found')
     pmt_wid  = sns_bins.bin_width[sns_bins.index.str.contains( 'Pmt')].iloc[0]
     sipm_wid = sns_bins.bin_width[sns_bins.index.str.contains('SiPM')].iloc[0]
-    return pmt_wid, sipm_wid
+    return float(pmt_wid), float(sipm_wid)
--- a/invisible_cities/detsim/sensor_utils.py
+++ b/invisible_cities/detsim/sensor_utils.py
@@ -103,4 +103,4 @@ def pmt_and_sipm_bin_width(file_name: str) -> Tuple[float, float]:
         raise tb.NoSuchNodeError('No useful binning info found')
     pmt_wid  = sns_bins.bin_width[sns_bins.index.str.contains( 'Pmt')].iloc[0]
     sipm_wid = sns_bins.bin_width[sns_bins.index.str.contains('SiPM')].iloc[0]
-    return float(pmt_wid), float(sipm_wid)
+    return pmt_wid, sipm_wid
--- a/invisible_cities/detsim/sensor_utils.py
+++ b/invisible_cities/detsim/sensor_utils.py
@@ -62,8 +62,8 @@ def sensor_order(detector_db: str, run_number : int,
     n_pmt, n_sipm = get_n_sensors(detector_db, run_number)
     pmt_shape     = (n_pmt , length_pmt )
     sipm_shape    = (n_sipm, length_sipm)
-    def ordering(sensor_order : pd.Int64Index  ,
-                 sensor_resp  : np.ndarray     ,
+    def ordering(sensor_order : pd.Index  ,
+                 sensor_resp  : np.ndarray,
                  sensor_shape : Tuple[int, int]) -> np.ndarray:
         sensors = np.zeros(sensor_shape, int)
         sensors[sensor_order] = sensor_resp
--- a/invisible_cities/detsim/sensor_utils.py
+++ b/invisible_cities/detsim/sensor_utils.py
@@ -99,4 +99,4 @@ def pmt_and_sipm_bin_width(file_name: str) -> Tuple[float, float]:
         raise tb.NoSuchNodeError('No useful binning info found')
     pmt_wid  = sns_bins.bin_width[sns_bins.index.str.contains( 'Pmt')].iloc[0]
     sipm_wid = sns_bins.bin_width[sns_bins.index.str.contains('SiPM')].iloc[0]
-    return pmt_wid, sipm_wid
+    return float(pmt_wid), float(sipm_wid)
--- a/invisible_cities/detsim/sensor_utils.py
+++ b/invisible_cities/detsim/sensor_utils.py
@@ -99,4 +99,4 @@ def pmt_and_sipm_bin_width(file_name: str) -> Tuple[float, float]:
         raise tb.NoSuchNodeError('No useful binning info found')
     pmt_wid  = sns_bins.bin_width[sns_bins.index.str.contains( 'Pmt')].iloc[0]
     sipm_wid = sns_bins.bin_width[sns_bins.index.str.contains('SiPM')].iloc[0]
-    return float(pmt_wid), float(sipm_wid)
+    return pmt_wid, sipm_wid
--- a/invisible_cities/evm/event_model.py
+++ b/invisible_cities/evm/event_model.py
@@ -361,16 +361,17 @@ class KrEvent(Event):
     def store(self, table):
         row = table.row
 
-        s1_peaks = range(int(self.nS1)) if self.nS1 else [-1]
-        s2_peaks = range(int(self.nS2)) if self.nS2 else [-1]
+        dummy    = np.iinfo(np.uint16).max
+        s1_peaks = range(int(self.nS1)) if self.nS1 else [0]
+        s2_peaks = range(int(self.nS2)) if self.nS2 else [0]
         self.fill_defaults()
 
         for i in s1_peaks:
             for j in s2_peaks:
                 row["event"  ] = self.event
                 row["time"   ] = self.time
-                row["s1_peak"] = i if i>=0 else np.iinfo(np.uint16).max + i + 1
-                row["s2_peak"] = j if j>=0 else np.iinfo(np.uint16).max + j + 1
+                row["s1_peak"] = i if self.nS1 else dummy
+                row["s2_peak"] = j if self.nS2 else dummy
                 row["nS1"    ] = self.nS1
                 row["nS2"    ] = self.nS2
 
--- a/invisible_cities/evm/event_model.py
+++ b/invisible_cities/evm/event_model.py
@@ -369,8 +369,8 @@ class KrEvent(Event):
             for j in s2_peaks:
                 row["event"  ] = self.event
                 row["time"   ] = self.time
-                row["s1_peak"] = i
-                row["s2_peak"] = j
+                row["s1_peak"] = i if i>=0 else np.iinfo(np.uint16).max + i + 1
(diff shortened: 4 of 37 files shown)


manage.sh
```diff
@@ -65,7 +65,7 @@ function install_conda {
     fi
 }
 
-CONDA_ENV_TAG=2024-11-07
+CONDA_ENV_TAG=2026-03-09
```
jwaiton: [This commit](https://github.com/next-exp/IC/pull/955/changes/eda48bd08b25ba770346c029c821ea41e8ff63fa#diff-a008ebe7403b9bde1fadc20a7d91e0fe53d87a3f40edaa30c6fd5b7e14a1a6c3R68) also alters the `conda_env_tag`. To be really pernickety, it would make sense to highlight this in the commit message right? Or separate them.

pyproject.toml
```diff
@@ -0,0 +1,17 @@
+[project]
+name         = "invisible_cities"
+version      = "0.1.0"
+description  = "NEXT blah blah"
```
jwaiton: Good description 🐱

pyproject.toml
```diff
@@ -0,0 +1,17 @@
...
+license      = {text = "TODO"}
```
jwaiton: License is already described as the MIT license [here](https://github.com/next-exp/IC/blob/master/doc/ic_authors_and_legal.rst). This should be quick to implement right?

manage.sh
```diff
@@ -217,7 +217,7 @@ function download_test_db {
 }
 
 function compile_cython_components {
-    python setup.py develop
+    pip3 install -e .
```
jwaiton, line 222: I'm not exactly sure how this works. Does `pip3` use `setup.py` by default to apply the cythonisation?

invisible_cities/database/load_db_test.py
```diff
@@ -62,8 +64,8 @@ def test_DetectorGeometry(dbnew):
 
 
 def test_mc_runs_equal_data_runs(db):
-    assert (DB.DataPMT (db.detector, -3550).values == DB.DataPMT (db.detector, 3550).values).all()
-    assert (DB.DataSiPM(db.detector, -3550).values == DB.DataSiPM(db.detector, 3550).values).all()
+    assert_dataframes_close(DB.DataPMT (db.detector, -3550), DB.DataPMT (db.detector, 3550))
+    assert_dataframes_close(DB.DataSiPM(db.detector, -3550), DB.DataSiPM(db.detector, 3550))
```
jwaiton, line 68: Does it matter that these now use tolerances (as it relies on [assert_frame_equal()](https://pandas.pydata.org/docs/reference/api/pandas.testing.assert_frame_equal.html)) rather than requiring the two dataframes to be exactly equal?

invisible_cities/reco/psf_functions_test.py
```diff
@@ -44,6 +44,7 @@ def test_add_empty_sensors_and_normalize_q(ICDATADIR):
     assert np. isclose(hdst_processed.Q.sum(), hdst.Q.sum())
 
 
+@mark.skip(reason="test uses bad input data")
 def test_add_empty_sensors_and_normalize_q_file(ICDATADIR):
     PATH_IN        = os.path.join(ICDATADIR, "exact_Kr_tracks_with_MC.h5")
```
jwaiton: Based on the [commit](https://github.com/next-exp/IC/pull/955/changes/390d1dde732b16443f4de655d2fc665c1ba6afed#r2930706799) message:
```
Skip tests
Unfortunately I don't remember what I meant by bad input data. Is this
real?
```
It looks like you mistrust `exact_Kr_tracks_with_MC.h5`. I'm unsure why that would be, although I haven't looked at the data explicitly.

invisible_cities/io/mcinfo_io_test.py
```diff
@@ -543,9 +543,9 @@ def new_format_mcsensor_data(ICDATADIR):
                                                    'TP_SiPM'     ],
                                                   name='sns_name'))
     evt_no         = 1280
-    sns_list       = [20005  , 20016  , 57  , 51  ]
+    sns_list       = [20016  , 20005  , 57  , 51  ]
     charge_list    = [    1  ,     1  ,  1  ,  1  ]
-    time_list      = [   48.0,    27.0, 90.0, 50.0]
+    time_list      = [   27.0,    48.0, 90.0, 50.0]
```
jwaiton, line 548: I'm assuming you've tested this with both versions of python (3.8 & 3.13) with their respective packages, and see this flip? It would be an interesting test if not. 

Checking my own version of IC (using 3.8), the dataframes show up with the order you've altered it to here:
Image
And then locally (to avoid any funny business with hdf5view):

Image

So perhaps this is moreso that the test passed silently in the past with no complaints, when it shouldn't have?

invisible_cities/reco/psf_functions.py
```diff
@@ -81,32 +81,17 @@ def add_empty_sensors_and_normalize_q(df       : pd.DataFrame,
...
-    df_out['nsipm'] = len(df)
-
-    return df_out
+    all_sens_df = pd.DataFrame(dict(X=sensors.X, Y=sensors.Y, Z=df.Z.min(), Q=0, E=0, Ec=0))
+    combined    = pd.concat([df, all_sens_df]    , ignore_index=True)
+    combined    = combined.groupby("X Y Z".split(), as_index=False).agg("first")
+    combined    = combined.ffill().bfill()
+    combined['NormQ'] = combined.Q / combined.Q.sum()
+    combined['nsipm'] = 1
+    return combined
```
jwaiton, line 93: your new method no longer uses the `var` character, right? should it be removed as an argument?

invisible_cities/evm/event_model.py
```diff
@@ -369,8 +369,8 @@ def store(self, table):
             for j in s2_peaks:
                 row["event"  ] = self.event
                 row["time"   ] = self.time
-                row["s1_peak"] = i
-                row["s2_peak"] = j
+                row["s1_peak"] = i if i>=0 else np.iinfo(np.uint16).max + i + 1
+                row["s2_peak"] = j if j>=0 else np.iinfo(np.uint16).max + i + 1
```
jwaiton: This defaults the value to `65535 + i + 1` if its negative. To avoid any data type issues, `s1_peak` and `s2_peak` are saved with a datatype larger than `np.uint16` right?

invisible_cities/reco/paolina_functions_test.py
```diff
@@ -178,7 +178,7 @@ def test_round_hits_positions_in_place(hits):
     round_hits_positions_in_place(hits)
 
     pos = np.asarray([h.pos for h in hits])
-    assert np.all(np.in1d(pos, [0, 1e-5, -1e-5]))
+    assert np.all(np.isin(pos, [0, 1e-5, -1e-5]))
```
jwaiton: [This commit](https://github.com/next-exp/IC/pull/955/changes/f0acec071e06f5a24946f6ee963951b47417e2a1) should be fixup'd with the [prior commit](https://github.com/next-exp/IC/pull/955/changes/6bb3f7c066f9fda29f77cfd0160db86a0309aebd) right?

invisible_cities/io/mcinfo_io.py
```diff
@@ -63,7 +63,8 @@ def write_mctables(table_dict : Dict):
                     orig_indx = tbl.file_index.unique()
                     new_indx  = np.arange(first_indx                 ,
                                           first_indx + len(orig_indx))
-                    tbl.file_index.replace(orig_indx, new_indx, inplace=True)
+                    for orig, new in zip(orig_indx, new_indx):
+                        tbl.replace(dict(file_index=orig), new, inplace=True)
```
jwaiton, line 67: This is in the [commit](https://github.com/next-exp/IC/pull/955/changes/9b397567e0f0873fd739e7ac8328d43bfb2e7749#diff-66a7d5cce8e82dcd5dca37868a5ba1c9e10b735b0534a7977fd3fdcebf573ca5R66-R67) titled:
```
Avoid using inplace operations on columns
```
Perhaps this is my misunderstanding, but isn't this still an inplace operation? You're just looping over the indices rather than applying it across the whole df

invisible_cities/reco/psf_functions.py
```diff
@@ -113,14 +115,16 @@ def hdst_psf_processing(dsts     : pd.DataFrame,
...
+           .apply( add_empty_sensors_and_normalize_q
+                 , list("XY"), ranges, database
+                 , include_groups=False)
+           .assign( Zpeak = lambda df: df.Z.min()
+                  , RelX  = lambda df: df.X - df.Xpeak
+                  , RelY  = lambda df: df.Y - df.Ypeak
+                  , RelZ  = 0.
+                  ))
+    hdst.reset_index(level=3, drop=True , inplace=True)
+    hdst.reset_index(         drop=False, inplace=True)
```
jwaiton: There are nice comments explaining this in the [next commit](https://github.com/next-exp/IC/pull/955/changes/6c17f471db44c6a1f6ca0b26b5c66f2ce18a6d42). I think it would be nice to add them here also.

invisible_cities/cities/beersheba.py
```diff
@@ -344,8 +347,11 @@ def drop_isolated( distance   : List[float],
 
 
     def drop_isolated(df): # df shall be an event cdst
-        df = df.groupby(['event', 'npeak']).apply(drop).reset_index(drop=True)
-
+        if not len(df): return df
+        df = (df.groupby("event time npeak".split())
+                .apply(drop, include_groups=False)
+                .reset_index(level=3, drop=True) # drop this artifical index level
```
jwaiton, line 353: Perhaps this is me not understanding the indexing well enough, but why do you need to `reset_index()` twice here, compared to once before?

setup.py
```diff
@@ -13,6 +14,15 @@
           .removeprefix("v")
           )
 
+#                    digits dot digits dot digits
```
jwaiton, line 19: Perhaps a better comment here would be nice. I'm assuming its extracting the numbers from a standard semantic versioning pattern (3.13.123).

setup.py
```diff
@@ -13,6 +14,15 @@
...
+version_pattern = "^ (\d+) (\.) (\d+) (\.) (\d+) $".replace(" ", "")
+
+# HACK:
+# in GHA we only pick up the last commit in the repo history and the the command
+# above returns the commit hash. If this happens, use a dummy version number.
+if not re.match(version, version_pattern):
+    version = "0.0.0"
+
```
jwaiton, line 27: I'm not sure I like this, wont this default every PR to version "0.0.0" unless specifically defined under the most recent tag?

invisible_cities/reco/pmaps_functions_test.py
```diff
@@ -30,7 +30,7 @@ def test_rebin_peak_rebin_factor_1(pk):
     assert_Peak_equality(rebinned_pk, expected_pk)
 
 
-@given(peaks())
+@given(peaks().filter(lambda pk: pk[1].pmts.sum_over_sensors.sum()>1e-20))
```
jwaiton, line 35: maybe add a comment explaining what this is for. I can see this filtering of really small pmt sensors signals becoming rather obtuse in the future.

invisible_cities/detsim/sensor_utils.py
```diff
@@ -103,4 +103,4 @@ def pmt_and_sipm_bin_width(file_name: str) -> Tuple[float, float]:
         raise tb.NoSuchNodeError('No useful binning info found')
     pmt_wid  = sns_bins.bin_width[sns_bins.index.str.contains( 'Pmt')].iloc[0]
     sipm_wid = sns_bins.bin_width[sns_bins.index.str.contains('SiPM')].iloc[0]
-    return pmt_wid, sipm_wid
+    return float(pmt_wid), float(sipm_wid)
```
jwaiton: This feels like a bit of a patch. Is there a reason why `sns_bins.bin_width` itself cannot be stored as floats?

invisible_cities/reco/hits_functions.py
```diff
@@ -141,7 +141,7 @@ def merge_NN_hits(hits: pd.DataFrame, same_peak: bool = True) -> pd.DataFrame:
 
 
 def empty_hit( event : int  , timestamp: float, peak_no: int
-             , x_peak: float, y_peak   : float, z      : float
+             , x_peak: float, y_peak   : float
```
jwaiton, line 141: [This commit](https://github.com/next-exp/IC/pull/955/changes/f3164deffb5f433b508820d4bcdf804f6d71dac0) appears to be related to [these lines](https://github.com/next-exp/IC/pull/955/changes/f3164deffb5f433b508820d4bcdf804f6d71dac0#diff-93799367305ef95d290c756b5d2ff019e42be5617bc89a1e3bde337815721ab9R244) based on the description, but it also includes a bunch of removals of `Z` from dataframes. Are these related?

jwaiton (changes requested):
First read over, its good work and very well commented. I just have a few questions.

manage.sh
gonzaponte: It is kind of implicit; changing the list of packages in any way must come with a change in the tag. But I'm not against it. The second commit will go away once #952 is merged.

manage.sh
gonzaponte: Sadly, I don't remember the details. Just that setup.py is obsolete, and this is the procedure now.

invisible_cities/database/load_db_test.py
gonzaponte: Not really. What I consider more important is that this is the correct way of doing the comparison, as these dataframes contain floats.

invisible_cities/reco/psf_functions_test.py
gonzaponte: Ok, I think I found the issue. The input data contains hits with positions that do not correspond to sensors. Not sure why this is the case. If you are happy with it, I can replace these two tests with new ones that are a bit less misterious.

invisible_cities/io/mcinfo_io_test.py
gonzaponte: I don't recall the details (I did this months ago), but essentially, yes, I checked with both versions and the data seemed to be ordered differently at some point in the process. My feeling is that it was wrong before, and now it's ok.

invisible_cities/evm/event_model.py
gonzaponte: No, it doesn't. The maximum number that the else branch can give is 65535 for i/j=-1, more negative values of i/j will only make it smaller, so there is no risk of overflow. In practice I think the only negative value i/j can have is -1, but this is more general. Not sure if it needs to be more general, though. -1 is a placeholder value for when an event doesn't have s1 or s2 peaks. Seeing how confusing this can be, I lean towards simplifying it. What's your take?

Also, the second line uses i instead of j, will fix that.

invisible_cities/io/mcinfo_io.py
gonzaponte: Poor wording. The change does not avoid doing things inplace overall. It avoids using column operations with the inplace argument which is no longer allowed. In the new pandas version, the interface forces you to do one of:
- do the operation on the dataframe, using the arguments to restrict the operation to a subset of columns (this is the bit that you highlighted)
- to the operation on the column, without inplace, and reassign it to the same column (this is the other half of the commit)

I will reword the commit and add this explanation.

invisible_cities/cities/beersheba.py
gonzaponte: This has to do with a change in the groupby-apply behaviour of pandas. In the older version, the groups passed to `apply` contained the grouping columns. The new behaviour promotes removing the grouping columns from the groups. For now it still allows both uses via the `include_groups` argument, but it will become the default (or even fixed) behaviour in the future.

Before the function used in `apply` (`drop`) was returning a dataframe that contained the grouping columns. `apply` would then combine the resulting dataframes using the grouping columns as index. Finally, we were resetting and dropping the _entire index_ because the grouping columns were already in the dataframe.

Now, the grouping columns are not passed to `drop`, so the dataframe it returns does not contain them. When `apply` combines these dataframes using the grouping columns, we must preserve the index in order not to lose this information. However, the dataframe produced in `drop` contains another (meaningless) index. This is just because dataframes must have an index. What we do in this case is to just drop the meaningless index and restore the grouping columns which have been set as the index.

Will add this description to the commit message.

setup.py
gonzaponte: Yes, it is precisely what I intended: set a default dummy value for GHA. But the version number doesn't mean anything for GHA. Do you see an issue here?

invisible_cities/detsim/sensor_utils.py
gonzaponte: They are stored as floats. The thing is that pandas now returns wrapped types for operations like  `iloc`. So instead of `float` you get `pd.Float64` or something like that. Actually, I just checked and this is not needed, it's something that showed up during the debugging phase of something else. I'm happy to remove it since the tests pass without this commit.

invisible_cities/reco/hits_functions.py
gonzaponte: Yes, it's the same `groupby` behaviour I explained in a previous reply.

invisible_cities/reco/psf_functions_test.py
jwaiton: I think that would be optimal :)

invisible_cities/evm/event_model.py
jwaiton: As you say, there is no need for it to be so complex as the `s1_peaks` and `s2_peaks` are defined a few lines above with -1 being the only negative value. I think it would be nice to simplify this :)

setup.py
jwaiton: I think I'm just confusing myself here. I assumed the resulting `setup()` would write the version out to [pyproject.toml](https://github.com/gonzaponte/IC/blob/python3.13/pyproject.toml). My issue was that if you apply a tag to a specific commit, you'd then need to add another commit where the toml file changes, losing the tag.

It appears that this code doesn't alter `pyproject.toml`'s version though! So it's no problem. One issue I have then is: do we alter pyproject.toml manually each time we implement a tag?

gonzaponte: all this happens on the fly during installation, so the version installed will show up as "0.0.0", but it doesn't really matter on GHA.

invisible_cities/reco/psf_functions_test.py
jwaiton: Resolved in [9aadbc](https://github.com/next-exp/IC/commit/9aadbc8)

jwaiton (approved):
This PR bumps the python version from 3.8 to 3.13, requiring some alterations to the code to ensure IC functions as intended.

Very nice work!

gonzaponte:
Closes #920.

PR #957, opened 2026-04-20
Rewrite paolina using DataFrames

gonzaponte:
Paolina functions were written using the `HitCollection` and `Hit` classes, which are cumbersome. Pandas' Dataframes are a much better tool to handle this type of data and, in practice, we always use them. This PR rewrites some of the paolina functionality avoiding these unnecessary intermediate objects.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -608,7 +608,8 @@ def pmap_from_files(paths):
 @check_annotations
 def hits_and_kdst_from_files( paths : List[str]
                             , group : str
-                            , node  : str ) -> Iterator[Dict[str,Union[HitCollection, pd.DataFrame, MCInfo, int, float]]]:
+                            , node  : str
+                            ) -> Iterator[Dict[str,Union[ pd.DataFrame , pd.DataFrame , MCInfo, int, float]]]:
     """
     Reader of hits files. For each event it produces a dictionary containing
     - hits        : a DataFrame
@@ -696,9 +697,8 @@ def MC_hits_from_files(files_in : List[str], rate: float) -> Generator:
 
 
 @check_annotations
-def dhits_from_files(paths: List[str]) -> Iterator[Dict[str,Union[HitCollection, pd.DataFrame, MCInfo, int, float]]]:
-    """Reader of the files, yields HitsCollection, pandas DataFrame with
-    run_number, event_number and timestamp."""
+def dhits_from_files(paths: List[str]) -> Iterator[Dict[str, Union[pd.DataFrame, pd.DataFrame, MCInfo, int, float]]]:
+    """Reader of the files, yields a hits dataframe, a kdst, run_number, event_number and timestamp."""
     for path in paths:
         try:
             dhits_df = load_dst (path, 'DECO', 'Events')
@@ -1372,8 +1372,8 @@ def make_event_summary(event_number  : int         ,
     topology_info : DataFrame
         Dataframe containing track information,
         output of track_blob_info_creator_extractor
-    paolina_hits  : HitCollection
-        Hits table passed through paolina functions,
+    paolina_hits  : pd.DataFrame
+        Hits passed through paolina functions,
         output of track_blob_info_creator_extractor
     kdst          : DataFrame
         Kdst information read from sophronia output
@@ -1446,7 +1446,8 @@ def track_blob_info_creator_extractor(vox_size         : Tuple[float, float, flo
                                       max_num_hits     : int
                                      ) -> Callable:
     """
-    For a given paolina parameters returns a function that extract tracks / blob information from a HitCollection.
+    For a given set of paolina parameters returns a function that extract tracks / blob
+    information from a set of hits.
 
     Parameters
     ----------
@@ -1465,9 +1466,12 @@ def track_blob_info_creator_extractor(vox_size         : Tuple[float, float, flo
 
     Returns
     ----------
-    A function that from a given HitCollection returns a pandas DataFrame with per track information.
+    A function that from a given set of hits returns
+      - general track information dataframe
+      - hits with associated track ids
+      - flag to indicate whether there are hits outside of the correction-map domain
     """
-    def create_extract_track_blob_info(hits):
+    def create_extract_track_blob_info(hits: pd.DataFrame) -> Tuple[pd.DataFrame, pd.DataFrame, bool]:
         df = pd.DataFrame(columns=list(types_dict_tracks.keys()))
         hits = hits.assign(track_id=-1)
         if len(hits) > max_num_hits:
@@ -1650,7 +1654,7 @@ def hits_thresholder(threshold_charge : float, same_peak : bool ) -> Callable:
 
     Returns
     ----------
-    A function that takes HitCollection as input and returns another object with
+    A function that takes a set of hits as input and returns another object with
     only non NN hits of charge above threshold_charge.
     The energy of NN hits is redistributed among neighbors.
     """
@@ -1683,8 +1687,8 @@ def hits_corrector( filename     : str
 
     Returns
     ----------
-    A function that takes a HitCollection as input and returns
-    the same object with modified Ec and Z fields.
+    A function that takes a set of hits as input and returns another one with Ec
+    and Z fields assigned. Input data is not modified.
     """
     maps      = read_maps(os.path.expandvars(filename))
     get_coef  = apply_all_correction( maps
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -1349,19 +1349,9 @@ def calculate_and_save_buffers(buffer_length    : float        ,
 
 @check_annotations
 def Efield_copier(energy_type: HitEnergy):
-    def copy_Efield(hitc : HitCollection) -> HitCollection:
-        mod_hits = []
-        for hit in hitc.hits:
-            hit = Hit(hit.npeak,
-                      Cluster(hit.Q, xy(hit.X, hit.Y), hit.var, hit.nsipm),
-                      hit.Z,
-                      hit.E,
-                      xy(hit.Xpeak, hit.Ypeak),
-                      s2_energy_c=getattr(hit, energy_type.value),
-                      Ep=getattr(hit, energy_type.value))
-            mod_hits.append(hit)
-        mod_hitc = HitCollection(hitc.event, hitc.time, hits=mod_hits)
-        return mod_hitc
+    def copy_Efield(df: pd.DataFrame) -> pd.DataFrame:
+        return df.assign( Ec = df[energy_type.value]
+                        , Ep = df[energy_type.value])
     return copy_Efield
 
 
@@ -1582,7 +1572,7 @@ def hitc_to_df(hitc: HitCollection):
                                      , track_id = hit .track_id
                                      , Ep       = hit .Ep), index=[0]))
     df = pd.concat(hits, ignore_index=True)
-    df = df.astype(dict(event=np.int64, npeak=np.uint16, Ec=np.float64, Ep=np.float64))
+    df = df.astype(dict(event=np.int64, npeak=np.uint16, Ec=np.float64, track_id=np.int64, Ep=np.float64))
     return df
 
 
@@ -1600,7 +1590,7 @@ def compute_and_write_tracks_info(paolina_params, h5out,
                                   hit_type, filter_hits_table_name,
                                   hits_writer):
 
-    to_hitc            = fl.map(hitc_from_df, item="hits")
+    to_hitc            = fl.map(hitc_from_df, item="Ep_hits")
 
     filter_events_nohits = fl.map(lambda x : len(x) > 0,
                                       args = 'hits',
@@ -1652,8 +1642,8 @@ def compute_and_write_tracks_info(paolina_params, h5out,
     return pipe( filter_events_nohits
                , fl.branch(write_no_hits_filter)
                , hits_passed.filter
-               , to_hitc
                , copy_Efield
+               , to_hitc
                , create_extract_track_blob_info
                , sort_hits_
                , filter_events_topology
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -1548,27 +1548,6 @@ def sort_hits(hits):
     return hits.sort_values("Z X Y".split())
 
 
-def hitc_to_df(hitc: HitCollection):
-    hits = []
-    for hit in hitc.hits:
-        hits.append(pd.DataFrame(dict( event    = hitc.event
-                                     , time     = hitc.time
-                                     , npeak    = hit .npeak
-                                     , Xpeak    = hit .Xpeak
-                                     , Ypeak    = hit .Ypeak
-                                     , X        = hit .X
-                                     , Y        = hit .Y
-                                     , Z        = hit .Z
-                                     , Q        = hit .Q
-                                     , E        = hit .E
-                                     , Ec       = hit .Ec
-                                     , track_id = hit .track_id
-                                     , Ep       = hit .Ep), index=[0]))
-    df = pd.concat(hits, ignore_index=True)
-    df = df.astype(dict(event=np.int64, npeak=np.uint16, Ec=np.float64, track_id=np.int64, Ep=np.float64))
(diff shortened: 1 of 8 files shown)


invisible_cities/cities/components.py
jwaiton: This function no longer yields a `HitsCollection` right? Would be a good excuse to come back and rewrite this docstring, and possibly reformat the typehint, since its a bit horrendous legibility-wise

invisible_cities/cities/components.py
```diff
@@ -1582,7 +1572,7 @@ def hitc_to_df(hitc: HitCollection):
                                      , track_id = hit .track_id
                                      , Ep       = hit .Ep), index=[0]))
     df = pd.concat(hits, ignore_index=True)
-    df = df.astype(dict(event=np.int64, npeak=np.uint16, Ec=np.float64, Ep=np.float64))
+    df = df.astype(dict(event=np.int64, npeak=np.uint16, Ec=np.float64, track_id=np.int64, Ep=np.float64))
```
jwaiton: Is there any particular reason why `track_id` is now passed through, and no other components of the hit collection? It seems to me that it would be passed through regardless in the line above.

invisible_cities/cities/components.py
```diff
@@ -1600,7 +1590,7 @@ def compute_and_write_tracks_info(paolina_params, h5out,
                                   hit_type, filter_hits_table_name,
                                   hits_writer):
 
-    to_hitc            = fl.map(hitc_from_df, item="hits")
+    to_hitc            = fl.map(hitc_from_df, item="Ep_hits")
```
jwaiton: Is there a reason for redefining the item here? To avoid overwriting a separate `hits` table? This gets removed entirely as far as I can tell, so will remove this comment at that point.

invisible_cities/reco/paolina_functions.py
```diff
@@ -23,40 +24,39 @@
 from typing import Tuple
 from typing import Dict
 
-MAX3D = np.array([float(' inf')] * 3)
-MIN3D = np.array([float('-inf')] * 3)
-
-def bounding_box(seq : BHit) -> Sequence[np.ndarray]:
+def bounding_box(hits: pd.DataFrame) -> Tuple[np.ndarray, np.ndarray]:
     """Returns two arrays defining the coordinates of a box that bounds the voxels"""
```
jwaiton: Nicer docstrings would be nice, although not a priority

invisible_cities/reco/paolina_functions.py
```diff
@@ -81,31 +80,22 @@ def   sizes(a : np.ndarray) -> np.ndarray:
...
-    h_indices = np.array([(i, j, k) for i, j, k in zip(indx_coordinates[0], indx_coordinates[1], indx_coordinates[2])])
+    # find the bins where hits fall into
+    # numpy.histogramdd() uses [,) intervals...
+    # ...except for the last one, which is [,]: hits on the last edge,
+    # if any, must fall into the last bin
+    # @gonzaponte: This should not happen, as we are increasing the voxel
+    # dimensions a few lines above
+    hits_indices = np.array([ np.clip(np.digitize(coords, bins, right=False) - 1, 0, n-1)
+                              for coords, bins, n in zip(xyz.T, edges, number_of_voxels)
+                            ]).T
```
jwaiton: I've always found this section of code to be a bit cryptic. Perhaps the comment could be reduced to be less so.

invisible_cities/reco/paolina_functions.py
```diff
@@ -81,31 +80,22 @@ def   sizes(a : np.ndarray) -> np.ndarray:
...
     voxels = []
-    for (x,y,z) in np.stack(nz).T:
-
-        indx_comp = (h_indices == (x, y, z))
-        hits_in_bin = list(h for i, h in zip(indx_comp, hits) if all(i))
+    true_dimensions = np.array([size_x[0], size_y[0], size_z[0]])
+    for indices_voxel in np.stack(np.nonzero(E)).T:
+        x, y, z     = indices_voxel
+        indices     = np.where(np.all(hits_indices == indices_voxel, axis=1))[0]
+        hits_in_bin = hits.iloc[indices]
```
jwaiton: Maybe a comment would be nice here as well, or perhaps this whole 'determine which voxel bin a hit lands in' section would be better served as a separate function.

I'm surprised there isn't already some voxelisation tooling that can do this for you, but thats beyond the scope of this PR I'm assuming.

invisible_cities/reco/paolina_functions.py
```diff
@@ -314,48 +300,52 @@ def drop_end_point_voxels(voxels           : Sequence[Voxel],
...
+                             , weights = the_vox.hits[e_type]
+                             , axis    =  0)
+
+        distance_to_bary = lambda hs: np.linalg.norm(hs[xyz] - bary_pos, axis=1)
+        distances        = [(v, distance_to_bary(v.hits)) for v in the_neighbour_voxels]
+        min_dist         = min(ds.min() for (_, ds) in distances)
+
+        hits_at_min_dist = lambda ds: np.argwhere(np.isclose(ds, min_dist)).flatten()
+        min_hits         = [(v, hits_at_min_dist(ds)) for (v, ds) in distances]
+        min_hits         = [pair for pair in min_hits if pair[1].size]
```
jwaiton: This is much easier to follow, but it may be worth adding a comment of the logic as its still a bit obtuse.

Something like:
```
# determine all hits in neighbouring voxels within a minimum distance from the main voxels barycentre position
```

In general, I think it would be a bit nicer if the logic of each step of this function was commented a bit better, as its quite a dense read! It appears to me that the logic matches on both sides, but it's not the simple to follow.

invisible_cities/reco/paolina_functions.py
```diff
@@ -314,48 +300,52 @@ def drop_end_point_voxels(voxels           : Sequence[Voxel],
...
+            v.hits.iloc[row_idx, col_idx] = current_energy * (1 + the_vox.E/total_closest_e)
 
-        ### add dropped voxel energy to closest hits/voxels, proportional to the hits/voxels energy
-        sum_en_h = sum(getattr(h, e_type) for h in min_hs)
-        sum_en_v = sum(v.E                for v in min_vs)
-        for h in min_hs:
-            setattr(h, e_type, getattr(h, e_type) + getattr(h, e_type) * the_vox.E/sum_en_h)
-        for v in min_vs:
-            v.E = sum(getattr(h, e_type) for h in v.hits)
+        # recopmute voxel energy
```
jwaiton: typo :smile_cat:

invisible_cities/reco/paolina_functions_test.py
```diff
@@ -268,81 +254,70 @@ def is_close_to_integer(n):
         assert is_close_to_integer(separation_over_size).all()
 
 
-@given(bunch_of_hits, box_sizes)
+@given(bunch_of_hits(), box_sizes)
 def test_hits_energy_in_voxel_is_equal_to_voxel_energy(hits, requested_voxel_dimensions):
     voxels = voxelize_hits(hits, requested_voxel_dimensions, strict_voxel_size=False)
     for v in voxels:
-        assert sum(h.E for h in v.hits) == v.E
+        assert v.hits.E.sum() == approx(v.E)
```
jwaiton: is there a reason this is now `approx()`? this happens in `test_voxels_with_no_hits()` also

invisible_cities/reco/paolina_functions_test.py
```diff
@@ -268,81 +254,70 @@ def is_close_to_integer(n):
...
 
-    hits_df   = load_mchits_df(hit_file)
-    hits_dict = cast_mchits_to_dict(hits_df)
-    hit_seq   = hits_dict[evt_number]
-    voxels    = voxelize_hits(hit_seq                  ,
-                              vox_size                 ,
-                              strict_voxel_size = False)
+    hits_df = load_mchits_df(hit_file)
+    hits_df = hits_df.loc[evt_number].rename(columns=dict(x="X", y="Y", z="Z", energy="E"))
+    voxels  = voxelize_hits(hits_df, vox_size, strict_voxel_size=True)
```
jwaiton: This may be my ignorance, but the file appears to have hits in it, at what point are you testing voxels that don't contain hits here?

File in question:

Image

invisible_cities/cities/components.py
```diff
@@ -1383,26 +1383,22 @@ def make_event_summary(event_number  : int          ,
...
-    x, y, z = map(np.array, zip(*pos))
-    r = np.sqrt(x**2 + y**2)
-
-    e     = [h.Ec  for h in paolina_hits.hits]
-    ave_pos = np.average(pos, weights=e, axis=0)
-    ave_r   = np.average(r  , weights=e, axis=0)
-
+    r       = np.sqrt(hits.X**2 + hits.Y**2)
+    ave_pos = np.average(hits["X Y Z".split()], weights=hits.Ec, axis=0)
+    ave_r   = np.average(r                    , weights=hits.Ec, axis=0)
```
jwaiton, line 1408: I always think its a bit deceptive that these are labelled `ave_`. While its true in the sense that they are determined with `np.average`, they're weighted by the energies, making them no longer solely geometric in nature. This is just me being nitpicky though.

invisible_cities/cities/components.py
```diff
@@ -1470,89 +1466,81 @@ def track_blob_info_creator_extractor(vox_size         : Tuple[float, float, flo
...
-                        hit.track_id = tID
-                        track_hits.append(hit)
-
-            #change dtype of columns to match type of variables
-            df = df.apply(lambda x : x.astype(types_dict_tracks[x.name]))
-            track_hitc.hits.extend(track_hits)
-        return df, track_hitc, out_of_map
+        hits = hits.assign(track_id=-1)
+        if len(hits) > max_num_hits:
+            return df, hits, True
```
jwaiton, line 1492: I think I've raised this before, but this is in principle a silent 'filter' right? There should be some notification to the user that they're losing event topology information here.

invisible_cities/cities/components.py
```diff
@@ -1470,89 +1466,81 @@ def track_blob_info_creator_extractor(vox_size         : Tuple[float, float, flo
...
+        plf.round_hits_positions_in_place(hits, 5)
```
jwaiton, line 1493: The default decimals is 5, is there a need to explicitly put this?

invisible_cities/cities/components.py
```diff
@@ -1470,89 +1466,81 @@ def track_blob_info_creator_extractor(vox_size         : Tuple[float, float, flo
...
+            energy = plf.get_track_energy(t)
+            numb_of_hits   = sum(len(v.hits) for v in t.nodes())
+            numb_of_voxels = len(t.nodes())
+            numb_of_tracks = len(tracks   )
+            hits_from_track = ( pd.concat([v.hits for v in t.nodes()], ignore_index=True)
+                                  .assign(R = lambda df: np.sqrt(df.X**2 + df.Y**2))
+                              )
+
+            ave_pos = np.average(hits_from_track["X Y Z".split()], weights=hits_from_track.Ep, axis=0)
+            ave_r   = np.average(hits_from_track.R               , weights=hits_from_track.Ep, axis=0)
```
jwaiton, line 1526: Likewise here, these are not inherently geometrical. Just a nitpick.

invisible_cities/cities/components.py
```diff
@@ -721,11 +721,10 @@ def dhits_from_files(paths: List[str]) -> Iterator[Dict[str,Union[HitCollection,
...
                 kdst  =  kdst_df.loc[this_event] if isinstance(kdst_df, pd.DataFrame) \
                                                               else None
-                ### It makes no sense to use the 'io.hits_io.hits_from_df' function here
-                ### (as well as in 'hits_and_kdst_from_files') since the majority of the
-                ### 'evm.Hit' parameters don't appear in the dst (particularly running
-                ### Isaura) and must be set to -1. This proceure should be revisited and
-                ### rethought in the near future, with the aim of changing the event model.
+
+                # this are just patches, the hit architecture should be reviseted
+                dhits = dhits.assign(Q=dhits.E)
```
jwaiton, line 736: Beersheba defaults to converting the deconvoluted image straight to energy, presumably to avoid two conversions (`deconv weight --> Q --> E`), this can be seen [here](https://github.com/next-exp/IC/blob/master/invisible_cities/cities/beersheba.py#L229) and then written [here](https://github.com/next-exp/IC/blob/master/invisible_cities/cities/beersheba.py#L259)

invisible_cities/museum/penthesilea/penthesilea.py
```diff
@@ -51,7 +51,27 @@
...
+                                     , Y        = hit .Y
+                                     , Z        = hit .Z
+                                     , Q        = hit .Q
+                                     , E        = hit .E
+                                     , Ec       = hit .Ec
+                                     , track_id = hit .track_id
+                                     , Ep       = hit .Ep), index=[0]))
+    df = pd.concat(hits, ignore_index=True)
+    df = df.astype(dict(event=np.int64, npeak=np.uint16, Ec=np.float64, track_id=np.int64, Ep=np.float64))
+    return df
```
jwaiton, line 74: 🥳

jwaiton (commented):
Good work! This is a really important change that I'm excited to see implemented.

I've commented commit-by-commit, so some of the commits may no longer be useful (and some outdated ones may still be, for example I still think the `dhits_from_files()` docstring is a bit barebones).

invisible_cities/reco/paolina_functions_test.py
gonzaponte: It should have been `approx` before, as we are comparing floats. I just took the chance to do it properly

invisible_cities/reco/paolina_functions_test.py
gonzaponte: That's a good question. I never understood how we could end up with hitless voxels. In the branch I have for the sequel PR, I removed all checks for hitless voxels and nothing terrible seems to happen...

invisible_cities/cities/components.py
gonzaponte: Isn't this called a _weighted average_? What would be a better terminology?

invisible_cities/cities/components.py
gonzaponte: I was thinking that if we ever decide to change the number of decimal places, we want to change the caller, not the default value. Perhaps it's just better to make it a non-optional argument.

invisible_cities/cities/components.py
gonzaponte: During this PR, `to_hitc` is being pushed down the pipeline. Each commit modifies one of the steps in this function to work with dataframes. Once this is done, the output needs to be converted to hitc to carry on with the rest of the chain. In this commit, the function that transforms the hits does not put the output back into the "hits" label, but on the "Ep_hits" label. Therefore, `to_hitc` has to change its input label when it's put passed this function.

invisible_cities/cities/components.py
gonzaponte: I am sure there is one, but I can't remember. It probably has to do with a specific case where there is a type mismatch in the output. This usually happens with empty dataframes, but I can't remember if this was the case.

invisible_cities/reco/paolina_functions.py
gonzaponte: I understand what it says, but not how that translates into the code below, so I don't know how to improve it.
If you don't mind, I am getting rid of this in the next PR, so we can leave it as is for a few weeks.

invisible_cities/reco/paolina_functions.py
gonzaponte: > I'm surprised there isn't already some voxelisation tooling that can do this for you, but thats beyond the scope of this PR I'm assuming.

That is literally what `digitize` does. It's all just a bit convoluted. As before, if you don't mind, this whole thing goes away in the next PR, so I wouldn't spend too much time on it.

invisible_cities/reco/paolina_functions_test.py
jwaiton: Okay! So this can be left for the sequel PR?

jwaiton (approved):
This PR removes the use of the `HitCollection` class from paolina, retrofitting all relevant functions to instead work with dataframes. General improvements to the docstrings and code clarity were made in parallel.

Good work!

PR #960, opened 2026-06-02
Propagate v2 LSC features

gonzaponte:
Some of the features developed for v2 at LSC where left behind. This PR propagates them to the main development branch. The main changes are

- Update on the handling of masked sensors
- Update database entry ordering
- Store calibration constants in irene

The first two are related to the way we store the data in NEXT-100 that somehow has changed since NEXT-White. The last one simply adds another layer of redundancy for better bookkeeping.

### DIFF
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -1750,3 +1750,9 @@ def sensor_masker(detector_db, run_number):
                , rwf_sipm[active_sipms])
 
     return mask_sensors
+
+
+def copy_db_info(file_in, file_out):
+    with tb.open_file(file_in) as file:
+        if "DB" in file.root:
+            file.copy_node(file.root.DB, file_out.root, recursive=True)
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -790,8 +790,8 @@ def build_pmap(detector_db, run_number, pmt_samp_wid, sipm_samp_wid,
 
 def calibrate_pmts(dbfile, run_number, n_maw, thr_maw):
     DataPMT    = load_db.DataPMT(dbfile, run_number = run_number)
+    DataPMT    = DataPMT.loc[lambda df: df.Active==1]
     adc_to_pes = np.abs(DataPMT.adc_to_pes.values)
-    adc_to_pes = adc_to_pes[adc_to_pes > 0]
 
     def calibrate_pmts(cwf):# -> CCwfs:
         return csf.calibrate_pmts(cwf,
@@ -803,6 +803,7 @@ def calibrate_pmts(dbfile, run_number, n_maw, thr_maw):
 
 def calibrate_sipms(dbfile, run_number, thr_sipm):
     DataSiPM   = load_db.DataSiPM(dbfile, run_number)
+    DataSiPM   = DataSiPM.loc[lambda df: df.Active==1]
     adc_to_pes = np.abs(DataSiPM.adc_to_pes.values)
 
     def calibrate_sipms(rwf):
--- a/invisible_cities/cities/components.py
+++ b/invisible_cities/cities/components.py
@@ -777,12 +777,13 @@ def build_pmap(detector_db, run_number, pmt_samp_wid, sipm_samp_wid,
                     stride       = s2_stride,
                     rebin_stride = s2_rebin_stride)
 
-    datapmt = load_db.DataPMT(detector_db, run_number)
-    pmt_ids = datapmt.SensorID[datapmt.Active.astype(bool)].values
-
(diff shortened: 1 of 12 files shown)


invisible_cities/cities/components.py
```diff
@@ -1736,3 +1736,14 @@ def hits_clusterizer( min_samples : int
 
 def identity(x : Any) -> Any:
     return x
+
+
+def sensor_masker(detector_db, run_number):
+    active_pmts  = load_db.DataPMT (detector_db, run_number).Active.values.astype(bool)
+    active_sipms = load_db.DataSiPM(detector_db, run_number).Active.values.astype(bool)
```
jwaiton, line 1743: We should have better documentation on how to interact/work with the databases. I'll open an issue in sw-docs [here](https://github.com/next-exp/sw-docs/issues/41).

invisible_cities/cities/components.py
```diff
@@ -426,18 +426,19 @@ def find_signal(wfs: pd.Series) -> List[int]:
...
+    coeff_c    = DataPMT.coeff_c  .values.astype(np.double)[pmt_active]
+    coeff_blr  = DataPMT.coeff_blr.values.astype(np.double)[pmt_active]
 
     def deconv_pmt(RWF):
+        assert len(RWF) == len(coeff_c)
         CWF = pedestal_function(RWF[:, :n_baseline]) - RWF
-        return np.array(tuple(map(blr.deconvolve_signal, CWF[pmt_active],
-                                  coeff_c              , coeff_blr      )))
+        CWF = map(blr.deconvolve_signal, CWF, coeff_c, coeff_blr)
+        return np.array(tuple(CWF))
```
jwaiton, line 440: Is the type wrapping needed? Isn't the output of `blr.deconvolve_signal()` already a `np.array`? As seen [here](https://github.com/next-exp/IC/blob/161438d756b19171ca5f31c9f26d9ca85d885c4b/invisible_cities/sierpe/blr.pyx).

invisible_cities/cities/hypathia.py
```diff
@@ -140,10 +145,11 @@ def hypathia( files_in        : OneOrManyFiles
                                     event_count_in.spy,
                                     mcrd_to_rwf,
                                     simulate_pmt,
-                                    pmt_sum,
-                                    zero_suppress,
                                     simulate_sipm_response_,
                                     discretize_signal,
+                                    mask_sensors,
+                                    pmt_sum,
+                                    zero_suppress,
```
jwaiton, line 152: Is there a reason why `pmt_sum` and `zero_suppress` moved below `simulate_pmt` & `simulate_sipm_response_`? This seems to imply (to me) that you may now be simulating the pmt and sipm responses for sipms and pmts that aren't active.

invisible_cities/cities/components.py
```diff
@@ -427,9 +427,9 @@ def find_signal(wfs: pd.Series) -> List[int]:
 
 
 def deconv_pmt(dbfile, run_number, n_baseline,
-               selection=None, pedestal_function=csf.means):
+               mask=True, pedestal_function=csf.means):
```
jwaiton, line 430: Do we want the masking to be user-controllable via city configs?

jwaiton (changes requested):
Good work! Small comments, but mostly in shape for merging.

There are a lack of tests for the new functionality, I believe these will need to be addressed at a later date. In particular:

- `deconv_pmt()` new argument: `mask`,
- `sensor_masker()` as a whole,
- `store_db_info()` as a whole,
- `build_sipm_responses()` now taking the `sipm_ids`, and all related functions (`build_peak()`, `find_peaks()`, `get_pmap()`)

invisible_cities/cities/components.py
gonzaponte: As discussed offline:
`tuple` consumes the map iterator, this results in a tuple of 1d arrays; `np.array` combines the 1d arrays into a single 2d array

invisible_cities/cities/hypathia.py
gonzaponte: I don't remember if there was a better motivation, but I believe it's just the minimal change to make it work. We should revisit this because it seems that we are indeed simulating things that we don't need to.

invisible_cities/cities/components.py
gonzaponte: No, that's not meant to be a user choice. The inner function is parametrized to account for two modes of operation:
- data processing, where we always want to ignore those channels (except for some cases, we can't process the waveforms of the masked sensors in a meaningful way)
- calibration, where we (can or might) do something different

jwaiton (approved):
This PR introduces functionality for sensor masking as well as the storage of calibration constants and a few alterations to the handling of databases.

Good work!

PR #962, opened 2026-06-04
Update size of some persistent data types

gonzaponte:
Due to the large number of reconstructed peaks, sometimes our counters surpass the maximum values for 8-bit unsigned integers. This PR promotes these counters to 16-bit uints, which should be enough in practice. Moreover, we add some overflow protection to avoid related issues that were silently going through before the update to python 3.13 (#955).

### DIFF
--- a/invisible_cities/core/core_functions.py
+++ b/invisible_cities/core/core_functions.py
@@ -4,11 +4,13 @@ This module includes utility functions.
 """
 import time
 from contextlib import contextmanager
+from warnings    import warn
 
 import numpy as np
 
 from typing import Sequence
 from typing import Tuple
+from typing import Union
 
 from ..types.symbols import NormMode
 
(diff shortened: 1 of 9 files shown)


jwaiton:
The overflow protection is nice, but shouldn't there be some warning that the user has reached the overflow limit? Having a bunch of peaks/sensor numbers with the same value may not be clear enough to the end user that they reached the int 16 bit limit.

gonzaponte:
That's reasonable, yes.

jwaiton (approved):
This PR increases the memory allocation for `peak` and `nsipm` from `UInt8Col` to `UInt16Col`, increasing the maximal value from 255 to 65535! 

Hopefully further increases wont be an issue for a while, and if it is, the user will be warned with the overflow checker also implemented in this PR.

Great work :)

